PR #25, opened 2020-10-19
Minor fixes.

inferno-chromium:


cmd/root.go
```diff
@@ -27,9 +27,9 @@ var (
 )
 
 var rootCmd = &cobra.Command{
-	Use:   "scorecard",
-	Short: "Security scorecards!",
-	Long:  `A scorecard program!`,
+	Use:   "./scorecard --repo=<repo_url> [--checks=check1,...]",
+	Short: "Open Source Scorecards",
```
inferno-chromium, line 31: Unsure if we are calling this "Security scorecards or Open Source Scorecards", but we can change this later i guess.

inferno-chromium:
minor fix, TBR

cmd/root.go
dlorenc: Yeah either works

PR #49, opened 2020-11-10
Bug fixes

inferno-chromium:
- Fix typo for default output function.
- For signed tags, get the only required last 5. matches signed
releases.


inferno-chromium:
TBR since typo fix is imp.

PR #56, opened 2020-11-10
Add a script to output in csv that can be run daily.

dlorenc:


cron/cron.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+SOURCE="${BASH_SOURCE[0]}"
+input=$(dirname $SOURCE)/projects.txt
+output=$(date +"%m-%d-%Y").csv
+touch $output
+while read -r line
+do
+    echo $line
+    go run . --repo=$line --format=csv 2>/dev/null >> $output
```
inferno-chromium, line 10: fyi, We need to remove that "[Starting]", "[Finishing]", "OUTPUT CSV" part from output when it is csv, so that output is clean. maybe also some way to not repeat headers.

inferno-chromium (approved):
LGTM with some changes that can done in this cl or next.

cron/cron.sh
dlorenc: That stuff actually goes to stderr, so it won't end up this file. For this script I piped stderr to null so it won't clog things up, but a "quiet" flag or something should work too.

inferno-chromium: Good point!

Issue #39, opened 2020-10-28
Searching for vuln reporting info outside of GH's security.md


annabellebertooch:
For the Security Policy field, some projects that use GitHub do have a documented security policy, but are not using the GitHub Security Policy feature, so scraping just for GH's Security Policy via security.md would inaccurately show a false in this field.

I think this is most commonly seen with projects that were established before the Security Policy feature, or for projects mirror to GitHub.

inferno-chromium:
Please provide an example repo (and file path to policy).

dlorenc:
Yeah, if you could point us at a few examples that would help!

inferno-chromium:
We want to make this discoverable by people, so standardizing on github security policy makes sense here. we also support the default community health files as well. So closing, if you have examples, feel free to reopen.

PR #57, opened 2020-11-10
extend Security-Policy check to check default community health files

chair6:
The current Security-Policy check only looks for policy files within the target repository, which is not the only way of specifying a security policy.

This extends the check to also look for a security policy that is set via GitHub's default community health file functionality (see https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/creating-a-default-community-health-file).

Seems relevant to https://github.com/ossf/scorecard/issues/39.

### DIFF
--- a/checks/security_policy.go
+++ b/checks/security_policy.go
@@ -15,7 +15,10 @@
 package checks
 
 import (
+    "fmt"
+    "net/http"
 	"path"
+    "strings"
 
 	"github.com/google/go-github/v32/github"
 	"github.com/ossf/scorecard/checker"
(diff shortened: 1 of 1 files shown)


checks/security_policy.go
```diff
@@ -41,6 +45,25 @@ func SecurityPolicy(c checker.Checker) checker.CheckResult {
 			}
 		}
 	}
+    // alternatively check repository owner for default community health file
+    r, _, err := c.Client.Repositories.Get(c.Ctx, c.Owner, ".github")
+    if err != nil {
+        return checker.RetryResult(err)
+    }
+    url := fmt.Sprintf("https://raw.githubusercontent.com/%s/.github/{/ref}/SECURITY.md", c.Owner)
```
inferno-chromium, line 53: This can get ratelimited, why not use c.Client.Repositories.DownloadContents api for this ?

inferno-chromium (changes requested):

checks/security_policy.go
chair6: good call.. helps to know the API. :) pushed an update that uses this instead..

inferno-chromium (approved):

PR #59, opened 2020-11-12
Add a Dockerfile and k8s cron job to upload files to GCS each night.

dlorenc:


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang
```
inferno-chromium, line 1: nit: probably need the license header in this and cron.sh files.

cron/cron.sh
```diff
@@ -9,3 +9,5 @@ do
     echo $line
     go run . --repo=$line --format=csv 2>/dev/null >> $output
```
inferno-chromium, line 10: should we build it once, and then just run ./scorecard binary.

cron/cron.sh
```diff
@@ -9,3 +9,5 @@ do
...
 done < "$input"
+
+gsutil cp $output gs://ossf-scorecards/
```
inferno-chromium, line 13: maybe gs://ossf-scorecards/ should come from ENV var, that ENV var can be set in Dockerfile,

ENV SCORECARS_BUCKET gs://ossf-scorecards

inferno-chromium (approved):

cron/cron.sh
dlorenc: Fixed up the Dockerfile into a proper multi-stage one.

cron/cron.sh
dlorenc: handled this in the k8s yaml instead, it's basically the same thing

inferno-chromium (approved):

PR #65, opened 2020-11-12
Add a JSON format mode.

dlorenc:
This is usable as is, but is also desigend to be easy to import into a database.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -17,12 +17,14 @@ package cmd
 import (
 	"context"
 	"encoding/csv"
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	goflag "flag"
 
@@ -44,6 +46,7 @@ var (
 
 const (
 	formatCSV     = "csv"
+	formatJSON    = "json"
 	formatDefault = "default"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -104,6 +109,38 @@ var rootCmd = &cobra.Command{
...
+		Date: d.Format("2006-01-01"),
+	}
+	for _, r := range results {
+		or.Checks = append(or.Checks, checkResult{
+			CheckName:  r.Name,
+			Pass:       r.Cr.Pass,
+			Confidence: r.Cr.Confidence,
+		})
+	}
+	b, err := json.Marshal(or)
```
inferno-chromium, line 137: nit: s/b/output

cmd/root.go
```diff
@@ -151,7 +188,7 @@ func init() {
 	// Add the zap flag manually
 	rootCmd.PersistentFlags().AddGoFlagSet(goflag.CommandLine)
 	rootCmd.Flags().Var(&repo, "repo", "repository to check")
-	rootCmd.Flags().StringVar(&format, "format", formatDefault, "output format. allowed values are [default, csv]")
+	rootCmd.Flags().StringVar(&format, "format", formatDefault, "output format. allowed values are [default, csv, json]")
```
inferno-chromium, line 191: Can you update README.md as well.

inferno-chromium (approved):

cmd/root.go
dlorenc: Good catch.

PR #69, opened 2020-11-13
Add support for and hookup app based authentication for higher rate l…

dlorenc:
…imiting.

This also configures it in our nightly cron cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,9 @@ gs://ossf-scorecards/11-11-2020.json
 ...
 ```
 
+The latest results are also always available at
+`https://storage.googleapis.com/ossf-scorecards/latest.json`.
+
 The list of projects that are checked each night is available in the
 [`cron/projects.txt`](https://github.com/ossf/scorecard/blob/main/cron/projects.txt)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -26,10 +26,17 @@ gs://ossf-scorecards/11-11-2020.json
 ...
 ```
 
+<<<<<<< Updated upstream
```
inferno-chromium, line 29: Need to fix this conflict, i linkified some files last night.

README.md
```diff
@@ -94,6 +103,18 @@ export GITHUB_AUTH_TOKEN=<your access token>
 set GITHUB_AUTH_TOKEN=<your access token>
 ```
 
+We also support GitHub App Installation based authentication for higher rate-limit quotas.
+If you have an installed GitHub App and key file, you can use these three environment variables, following the commands
```
inferno-chromium, line 107: "you can instead use" ?

i think it is either GITHUB_AUTH_TOKEN or these three right ?

inferno-chromium (approved):

PR #70, opened 2020-11-13
Tweak the Dockerfile to make it more generally useful.

dlorenc:


kimsterv:
great shipment of value

PR #71, opened 2020-11-14
Add support for gitlab repositories

MheniMerz:


inferno-chromium (changes requested):
This is not as simple as this.

You would need to add go-gitlab support in each check
See https://github.com/xanzy/go-gitlab

PR #79, opened 2020-11-24
Add log message when regular release is found.

inferno-chromium:


inferno-chromium:
requested by @htuch.

See
go run . --repo=https://github.com/kubernetes/kubernetes --checks=Signed-Releases --show-details

PR #81, opened 2020-11-26
Sort/uniquify the projects list before running.

dlorenc:
This way we can add things in multiple sections and not worry about deduping.


cron/projects.txt
```diff
@@ -1,3 +1,4 @@
+# Top Projects from Criticality
```
inferno-chromium, line 1: nit: "Criticality Score"

inferno-chromium (approved):
Thanks!

Issue #82, opened 2020-11-27
security policy check


vosytniazhskyi:
You check only for "security.md" and "SECURITY.md" but Some project has "SECURITY.**MD**".

inferno-chromium:
Can you give some examples ?
Ideally it should be just SECURITY.md as per github docs - https://docs.github.com/en/free-pro-team@latest/github/managing-security-vulnerabilities/adding-a-security-policy-to-your-repository

vosytniazhskyi:
Hello.
Last project that I've tested:
https://github.com/IdentityServer/IdentityServer4/security.md
https://api.github.com/repos/IdentityServer/IdentityServer4/contents/

 "name": "SECURITY.MD",
    "path": "SECURITY.MD"

If I change the list of strings for securityFile all works as expected.

for _, securityFile := range []string{"SECURITY.md", "security.md",
"SECURITY.MD"}


with best regards

Vasyl

PR #83, opened 2020-11-27
Allow other accepted variants of security.md case styles.

inferno-chromium:
Fixes https://github.com/ossf/scorecard/issues/82


inferno-chromium:
minor fix, TBR.

PR #85, opened 2020-11-30
Add SonarCloud to the SAST check

tom-vanbraband-sonarsource:


README.md
```diff
@@ -128,7 +128,7 @@ The following checks are all run against the target project:
 | CII-Best-Practices | Does the project have a [CII Best Practices Badge](https://bestpractices.coreinfrastructure.org/en)? |
 | Pull-Requests | Does the project use [Pull Requests](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-requests) for all code changes? |
 | Fuzzing | Does the project use [OSS-Fuzz](https://github.com/google/oss-fuzz)? |
-| SAST | Does the project use static code analysis tools, e.g. [CodeQL](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository#enabling-code-scanning-using-actions)? |
+| SAST | Does the project use static code analysis tools, e.g. [CodeQL](https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository#enabling-code-scanning-using-actions), [SonarCloud](https://sonarcloud.io/github? |
```
inferno-chromium, line 131: nit: close the bracket ')'

checks/sast.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/ossf/scorecard/checker"
 )
 
+var sast_tools map[string]bool = map[string]bool{"github-code-scanning": true, "sonarcloud": true}
```
inferno-chromium, line 22: nit: camelCase for variable sastTools

inferno-chromium: Can you think of any others to add here ?

inferno-chromium (changes requested):
LGTM with minor changes needed.

checks/sast.go
tom-vanbraband-sonarsource: Some actions/apps that can be found [here](https://github.com/marketplace/category/security) could make sense to check for, but it would be hard to create an exhaustive list. 

Instead of checking for specific apps another approach could be to check if the app that created the check is categorized under security in the marketplace. However this might then be a bit too inclusive, since not every app listed is a SAST tool

dlorenc (approved):

dlorenc:
Nice!

Issue #95, opened 2020-12-19
Show negative results in details


moorereason:
Howdy!  New here.  :wave: 🤠 

While using scorecard to evaluate projects I contribute to, I found myself having to read (and patch) the checks code to understand why I was seeing failing or marginal results.  It would be very helpful to see negative results in the `--show-details` output instead of only positive results.  It makes it easier for newcomers like me to learn what is considered bad practice, and I assume it would be help everyone to investigate any negative findings.

For example, I've patched scorecard locally to give negative results (starting with `!!`) for some checks:

```text
$ ./scorecard --repo=github.com/gohugoio/hugo --show-details --checks=Code-Review,CI-Tests,Pull-Requests,Signed-Releases,Signed-Tags
RESULTS
-------
CI-Tests: Fail 4
    !! found committed PR without CI test: 8075
    !! found committed PR without CI test: 8070
    CI test found: context: continuous-integration/travis-ci/pr, url: https://api.github.com/repos/gohugoio/hugo/statuses/1056701da088d5e87e6e31cdc6e0c455862697cc
    !! found committed PR without CI test: 8059
    CI test found: context: continuous-integration/travis-ci/pr, url: https://api.github.com/repos/gohugoio/hugo/statuses/34ecd28779d4836c74d4e71a5227b196d5d2cbec
    CI test found: context: continuous-integration/travis-ci/pr, url: https://api.github.com/repos/gohugoio/hugo/statuses/713792077b06504e9fdd0c8abdc2aebace5dcf2a
    CI test found: context: continuous-integration/travis-ci/pr, url: https://api.github.com/repos/gohugoio/hugo/statuses/3a38df2c4f96e560ddedc5fd5266fb54972a7fb6
    !! found committed PR without CI test: 8020
    CI test found: context: continuous-integration/travis-ci/pr, url: https://api.github.com/repos/gohugoio/hugo/statuses/7475bb324f208c8acd5fb3b6aa0f1b07c8d26d94
    CI test found: context: continuous-integration/travis-ci/push, url: https://api.github.com/repos/gohugoio/hugo/statuses/cd5a53bb240ef23977c03d2a47cd76f8ee915899
    !! found committed PR without CI test: 7999
    CI test found: context: continuous-integration/travis-ci/pr, url: https://api.github.com/repos/gohugoio/hugo/statuses/15ee0346fc60265c36530dcc3b70be3955f006ef
[...]

PR #98, opened 2020-12-20
Show negative results in some check details

moorereason:
I have one commit per check in this PR.  See commit messages for more details.

I'd like to get more checks updated, but these are the simple ones.

Updates #95

### DIFF
--- a/checks/pull_requests.go
+++ b/checks/pull_requests.go
@@ -53,6 +53,7 @@ func PullRequests(c checker.Checker) checker.CheckResult {
 		commitMessage := commit.GetCommit().GetMessage()
 		if strings.Contains(commitMessage, "\nReviewed-on: ") {
 			totalWithPrs++
+			c.Logf("found gerrit reviewed commit: %s", commit.GetSHA())
 			continue
 		}
 
@@ -62,6 +63,7 @@ func PullRequests(c checker.Checker) checker.CheckResult {
 		}
 		if len(prs) > 0 {
 			totalWithPrs++
+			c.Logf("found commit with PR: %s", commit.GetSHA())
 		} else {
 			c.Logf("!! found commit without PR: %s, committer: %s", commit.GetSHA(), commit.GetCommitter().GetLogin())
 		}
--- a/checks/pull_requests.go
+++ b/checks/pull_requests.go
@@ -62,6 +62,8 @@ func PullRequests(c checker.Checker) checker.CheckResult {
 		}
 		if len(prs) > 0 {
 			totalWithPrs++
+		} else {
+			c.Logf("!! found commit without PR: %s, committer: %s", commit.GetSHA(), commit.GetCommitter().GetLogin())
 		}
 	}
 	c.Logf("found PRs for %d out of %d commits", totalWithPrs, total)
--- a/checks/signed_releases.go
+++ b/checks/signed_releases.go
@@ -33,6 +33,8 @@ func SignedReleases(c checker.Checker) checker.CheckResult {
 		return checker.RetryResult(err)
 	}
 
+	artifactExtensions := []string{".asc", ".minisig", ".sig"}
+
 	totalReleases := 0
 	totalSigned := 0
 	for _, r := range releases {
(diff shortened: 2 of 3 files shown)


checks/signed_releases.go
```diff
@@ -59,13 +61,19 @@ func SignedReleases(c checker.Checker) checker.CheckResult {
...
 			break
 		}
 	}
 
 	if totalReleases == 0 {
+		c.Logf("no releases found")
 		return checker.InconclusiveResult
 	}
+
+	c.Logf("found signed artifacts for %d of %d releases", totalSigned, totalReleases)
```
inferno-chromium, line 77: for %d out of

checks/pull_requests.go
```diff
@@ -62,6 +62,8 @@ func PullRequests(c checker.Checker) checker.CheckResult {
 		}
 		if len(prs) > 0 {
 			totalWithPrs++
```
inferno-chromium, line 64: i think we don't do Logf for two places where it is incremented, can you add a regular message for those too.

checks/signed_tags.go
```diff
@@ -61,10 +60,18 @@ func SignedTags(c checker.Checker) checker.CheckResult {
 			return checker.RetryResult(err)
 		}
 		if gt.GetVerification().GetVerified() {
-			c.Logf("signed tag found: %s, commit: %s", t.Name, sha)
+			c.Logf("verified tag found: %s, commit: %s", t.Name, sha)
```
inferno-chromium, line 63: verified is slightly confusing word, that is why we keep check name as signed. i prefer signed for all of these, thoughts ?

checks/signed_tags.go
```diff
@@ -61,10 +60,18 @@ func SignedTags(c checker.Checker) checker.CheckResult {
...
 			totalSigned++
+		} else {
+			c.Logf("!! unverified tag found: %s, commit: %s, reason: %s", t.Name, sha, gt.GetVerification().GetReason())
 		}
 	}
 
+	if totalReleases == 0 {
+		c.Logf("no releases found")
```
inferno-chromium, line 71: i think we need to change totalReleases->totalTags and also update this message to "no tags found"

inferno-chromium (changes requested):
Thanks a lot, minor changes needed.

checks/signed_releases.go
moorereason: Done

checks/pull_requests.go
moorereason: Added a log here and a few lines up for gerrit logs.  PTAL

checks/signed_tags.go
moorereason: Done

inferno-chromium (approved):

checks/signed_tags.go
moorereason: I changed the terminology for a repo such as git, as shown below.  They're signed but GH can't verify the key, so the check fails.  Logging "unsigned tag found" in this scenario is confusing.

I'd like to keep the new "reason" entry in the output either way, but I can change back to "signed" instead of "verified" if you wish.
```
go run . --repo=github.com/git/git --show-details --checks=Signed-Tags
Starting [Signed-Tags]
Finished [Signed-Tags]

RESULTS
-------
Signed-Tags: Fail 10
    !! unverified tag found: v2.29.0, commit: 0c6ff04c974f0aab239e9ba46307e1d1a4904d66, reason: unknown_key
    !! unverified tag found: v2.29.1, commit: 6c5034eca4973ae22eee0436e34f9f010895bea3, reason: unknown_key
    !! unverified tag found: v2.29.2, commit: 1ec19b7757a1acb11332f06e8e812b505490afc6, reason: unknown_key
    !! unverified tag found: v2.30.0-rc0, commit: a8eaf9de52c2d49799d7dc724e688ccbfa74390c, reason: unknown_key
    !! unverified tag found: v2.30.0-rc1, commit: cb70effebd91d9e0d4ce81650785e973b5d16ad1, reason: unknown_key
    found 0 out of 5 verified tags
```

Issue #99, opened 2020-12-21
Fails when trailing slash is passed in the URL for the GitHub repository


naveensrinivasan:
The code fails to fetch the repository information when there is a trailing slash in the URL. 

```
./scorecard --repo=https://github.com/volcano-sh/volcano/ --show-details
Starting [Active]
Starting [CI-Tests]
Starting [CII-Best-Practices]
Starting [Code-Review]
Starting [Contributors]
Starting [Frozen-Deps]
Starting [Fuzzing]
Starting [Pull-Requests]
Starting [SAST]
Starting [Security-Policy]
Starting [Signed-Releases]
Starting [Signed-Tags]
Finished [Fuzzing]
Finished [CII-Best-Practices]
Finished [SAST]
Finished [Pull-Requests]
Finished [Frozen-Deps]
Finished [Contributors]
Finished [Signed-Releases]
Finished [Signed-Tags]
Finished [CI-Tests]
Finished [Active]
Finished [Code-Review]
Finished [Security-Policy]

RESULTS
-------
Active: Fail 0
    error, retrying: GET https://api.github.com/repos/volcano-sh/volcano//releases: 404 Not Found []
CI-Tests: Fail 0
    error, retrying: GET https://api.github.com/repos/volcano-sh/volcano//pulls?state=closed: 404 Not Found []
CII-Best-Practices: Pass 10
    badge level: passing
Code-Review: Fail 0
    error, retrying: GET https://api.github.com/repos/volcano-sh/volcano//commits: 404 Not Found []
Contributors: Fail 0
    error, retrying: GET https://api.github.com/repos/volcano-sh/volcano//contributors: 404 Not Found []
Frozen-Deps: Fail 0
    error, retrying: GET https://api.github.com/repos/volcano-sh/volcano/: 404 Not Found []
Fuzzing: Fail 10
Pull-Requests: Fail 0
    error, retrying: GET https://api.github.com/repos/volcano-sh/volcano//commits: 404 Not Found []
SAST: Fail 10
Security-Policy: Fail 10
Signed-Releases: Fail 0
    error, retrying: GET https://api.github.com/repos/volcano-sh/volcano//releases: 404 Not Found []
Signed-Tags: Fail 0
    error, retrying: Could not resolve to a Repository with the name 'volcano-sh/volcano/'.
    ```

naveensrinivasan:
Fixed by #100.

PR #100, opened 2020-12-21
fix - URL with trailing slash

naveensrinivasan:
- Fixes the URL with a trailing slash #99 
- Changed the URL parsing to net package implementation.
- Included tests for URL parsing.

### DIFF
--- a/pkg/scorecard.go
+++ b/pkg/scorecard.go
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"regexp"
+	"net/url"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


pkg/scorecard_test.go
```diff
@@ -0,0 +1,79 @@
...
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "Valid http address",
+			fields: fields{
+				Host:  "github.com",
+				Owner: "kubeflow",
```
inferno-chromium, line 40: nit: pick a different owner and repo to make sure we parse them right.

pkg/scorecard.go
```diff
@@ -48,13 +48,21 @@ func (r *RepoURL) Type() string {
...
+	}
+
+	split := strings.SplitN(strings.Trim(u.Path, "/"), "/", 2)
+	if len(split) != 2 {
 		log.Fatalf("invalid repo flag: [%s], pass the full repository URL", s)
 	}
-	r.Host, r.Owner, r.Repo = split[0], split[1], split[2]
+
+	if len(strings.TrimSpace(split[0])) == 0 || len(strings.TrimSpace(split[1])) == 0 {
+		log.Fatalf("invalid repo flag: [%s] pass the full repository URL", s)
```
inferno-chromium, line 62: nit: comma after [%s]

inferno-chromium (approved):
LGTM with minor nit changes needed.

pkg/scorecard_test.go
naveensrinivasan: Do you want to check if the parsed value is the expected value?

inferno-chromium: yes right, just want to make sure owner and repo part are parsed correctly too.

pkg/scorecard.go
naveensrinivasan: Took care of it.

pkg/scorecard_test.go
naveensrinivasan: Took care of it.

inferno-chromium (approved):

PR #104, opened 2020-12-22
Update README.md

chrisjsimpson:
The getting started readme example was:

> ./scorecard --repo=github.com/kubernetes/kubernetes

Which results in:
`Error: invalid argument "github.com/kubernetes/kubernetes" for "--repo" flag: unsupported host:`

So it has been corrected to include the scheme:
```
./scorecard --repo=https://github.com/kubernetes/kubernetes
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ The program only requires one argument to run, the name of the repo:
 
 ```shell
 $ go build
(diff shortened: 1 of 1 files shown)


inferno-chromium:
This was not intended and is a recent regression, fixing in https://github.com/ossf/scorecard/pull/106, so closing here as doc change is not needed.

PR #106, opened 2020-12-22
Allow skipping scheme, fix regression.

inferno-chromium:
Fix regression from
https://github.com/ossf/scorecard/commit/f7cb4d70e67f7551aa0caf84eca6a0c10919b3c7


inferno-chromium:
TBR

naveensrinivasan:
There should probably a test to catch this.

inferno-chromium:
> There should probably a test to catch this.

Can you add one, or i can do after i am back from holidays.

PR #108, opened 2020-12-23
Refactor CI-Tests to show negative results

### LINKED ISSUE #96: Erroneous inconclusive CI-Tests result
Given:

https://github.com/ossf/scorecard/blob/199161753d0a5c2b2832ea3126165261ca4ac86c/checks/ci_tests.go#L58-L59

Shouldn't that be `if totalMerged == 0`?

If we have 10 merged PRs and zero tests found for those PRs, that should not be an inconclusive result.  It should `Fail 10` not `Fail 0`.  Or am I missing something?

moorereason:
Add negative check results to the CI-Tests output.

Assuming that a repo will only support one CI system, GithubStatuses and
GithubCheckRuns are merged into a single CITests function.  Since both
GithubStatuses and GithubCheckRuns were essentially validating the same
PRs, it makes more sense to keep all of that state together in a single
check.

Additionaly, a single check can reduce the number of API queries once we
detect the CI system in use.

Fixes #96
Updates #95

### DIFF
--- a/checks/ci_tests.go
+++ b/checks/ci_tests.go
@@ -22,10 +22,10 @@ import (
 )
 
 func init() {
-	registerCheck("CI-Tests", checker.MultiCheck(GithubStatuses, GithubCheckRuns))
+	registerCheck("CI-Tests", CITests)
 }
 
-func GithubStatuses(c checker.Checker) checker.CheckResult {
+func CITests(c checker.Checker) checker.CheckResult {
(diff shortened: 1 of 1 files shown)


inferno-chromium (approved):
One function has its pros and cons, but i kind of agree that saving the api use count is more important, also, less code duplication, just that dont like code complication, but dont have other ideas.

inferno-chromium:
@moorereason - can you please help to fix the other checks as well to log negative results. Right now, it is a mix and slightly confusing. Your help is appreciated!

PR #113, opened 2020-12-28
Add --show-details to the cron job.

### LINKED ISSUE #112: Add details to the public data
It would be great if the public data included the details (`--show-details`) as well as the pass/fail and confidence indicator. This would give consumers of those data files a bit more information when parsing them.

scovetta:
This change will add a `details` section to the cron job output.

Fixes #112


dlorenc (approved):
Thanks!

PR #114, opened 2020-12-31
feat - nonroot docker container

naveensrinivasan:
- Changed the docker container to non-root
- Fixed the hadolint dockerfile warnings.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,12 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang
+FROM golang:1.15 as builder
 COPY . /go/src/github.com/ossf/scorecard
 WORKDIR /go/src/github.com/ossf/scorecard
 RUN [ "go", "build", "."]
 
-FROM gcr.io/cloud-builders/gsutil
+FROM gcr.io/distroless/base:nonroot
 WORKDIR /go/src/github.com/ossf/scorecard
-COPY --from=0 /go/src/github.com/ossf/scorecard /go/src/github.com/ossf/scorecard
+COPY --from=builder /go/src/github.com/ossf/scorecard /go/src/github.com/ossf/scorecard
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -11,12 +11,12 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang
+FROM golang:1.15 as builder
 COPY . /go/src/github.com/ossf/scorecard
 WORKDIR /go/src/github.com/ossf/scorecard
 RUN [ "go", "build", "."]
 
-FROM gcr.io/cloud-builders/gsutil
+FROM gcr.io/distroless/base:nonroot
```
dlorenc, line 19: Ah, I don't think this part is safe. We actually use gsutil in the nightly cron job from inside the container to upload results to GCS here: https://github.com/ossf/scorecard/blob/main/cron/cron.sh

We could either reimplement that part in Go or try to use a separate container.

Any other ideas?

naveensrinivasan: IMO we could have a separate Dockerfile just for the scorecard without using gsutil and publish that image for pulling down the scorecard. How does that sound?

dlorenc: SGTM! We could move this one into the cron directory to make it more clear as well.

naveensrinivasan: Sounds good. Thanks

README.md
```diff
@@ -170,6 +170,9 @@ These may be specified with the `--format` flag.
 * Any criteria in the scorecard must be actionable. It should be possible, with help, for any project to "check all the boxes".
 * Any solution to compile a scorecard should be usable by the greater open source community to monitor upstream security.
 
+## Docker
+* The Dockerfile in the root directory utilizes [experminetal features](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md)
```
dlorenc, line 174: nit: typo, experimental.

Do you need a specific version of docker to use this or any specific flags?

naveensrinivasan: Took care of it. Thanks

dlorenc (approved):

PR #115, opened 2020-12-31
e2e tests for signed tags and signed releases

naveensrinivasan:
This is the start of the `e2e` testing implementation.

- Implemented e2e tests using ginkgo for validating signed tags and signed releases.
- [ginkgo](https://github.com/onsi/ginkgo) is utilized as a standard BDD testing framework in other projects like [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder).


dlorenc (approved):
Wow, thanks!

PR #117, opened 2021-01-03
Feat - implemented goreleaser for releases

naveensrinivasan:
Implemented [goreleaser](https://github.com/goreleaser/goreleaser-action) for releasing the code to Github. The release would happen only for `tag` pushes

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

The release would be automated and would be something like this. 
image

https://github.com/naveensrinivasan/scorecard/releases/tag/v1.4

### DIFF
--- a/.github/workflows/goreleaser.yaml
+++ b/.github/workflows/goreleaser.yaml
@@ -1,0 +1,29 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - "*" # triggers only if push new tag version, like `0.8.4` or else
+
+jobs:
+  goreleaser:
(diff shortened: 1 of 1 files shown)


dlorenc:
Nice! Thanks!

dlorenc:
Do I need to do anything to configure the secret, or is that automatic?

naveensrinivasan:
> Do I need to do anything to configure the secret, or is that automatic?

It is automatic https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#using-the-github_token-in-a-workflow

Issue #120, opened 2021-01-05
Active != 90-day releases


david-a-wheeler:
The current definition of "active" is:

* Did the project get any commits and releases in last 90 days?

I think this is a *terrible* definition. Many projects intentionally do not release every 90 days, nor do their users want that kind of churn.

Can we please change "and" to "or"? Or alternatively, remove "and releases" entirely?

inferno-chromium:
I think we mean EITHER commits OR releases. Should i fix that in docs or you want to get rid of releases completely. I think just commits should be sufficient, i am trying to remember why i added releases as well.

inferno-chromium:
removing releases.

PR #122, opened 2021-01-05
Remove releases from active check.

inferno-chromium:
Fixes https://github.com/ossf/scorecard/issues/120


inferno-chromium:
TBR minor fix.

PR #153, opened 2021-02-08
Docker releases to the GitHub Docker registry

naveensrinivasan:
N/A

This will release the docker container to the GitHub docker registry.
An example https://github.com/naveensrinivasan/scorecard/packages/613485

The user would have to compile the docker image 

The docker images will be available to pull

None

* **Other information**:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,18 +14,17 @@
 
 # syntax = docker/dockerfile:1-experimental
 
-FROM --platform=${BUILDPLATFORM} golang:1.15 as base
+FROM  golang:1.15 as base
 WORKDIR /src
 ENV CGO_ENABLED=0
-COPY go.* .
+COPY go.* ./
 RUN go mod download
-COPY . .
+COPY . ./
 
 FROM base AS build
 ARG TARGETOS
 ARG TARGETARCH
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,18 +14,17 @@
...
 RUN go mod download
-COPY . .
+COPY . ./
 
 FROM base AS build
 ARG TARGETOS
 ARG TARGETARCH
-RUN --mount=type=cache,target=/root/.cache/go-build \
-  GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /out/scorecard .
+RUN --mount=type=cache,target=/root/.cache/go-build  go build -o /out/scorecard .
```
inferno-chromium, line 27: I don't know the impact of these changes. @dlorenc can you review this PR.

dlorenc: Just to make sure I understand, the build args aren't needed here because they're auto set as env vars? Or because we're just only building one platform?

naveensrinivasan: Because we are building for only one platform. That is the reason.

inferno-chromium (approved):

PR #170, opened 2021-02-13
Fix - Organization checks for members

naveensrinivasan:

Bugfix for Org checks for members

The current check ignores the org the members are in

None

* **Other information**:

### DIFF
--- a/checks/contributors.go
+++ b/checks/contributors.go
@@ -15,6 +15,8 @@
 package checks
 
 import (
+	"encoding/json"
+	"fmt"
 	"strings"
 
 	"github.com/google/go-github/v32/github"
@@ -26,6 +28,9 @@ func init() {
 }
 
 func Contributors(c checker.Checker) checker.CheckResult {
+	type orgnaizations []struct {
+		Login string `json:"login"`
+	}
 	contribs, _, err := c.Client.Repositories.ListContributors(c.Ctx, c.Owner, c.Repo, &github.ListContributorsOptions{})
 	if err != nil {
 		return checker.RetryResult(err)
@@ -38,6 +43,28 @@ func Contributors(c checker.Checker) checker.CheckResult {
 			if err != nil {
 				return checker.RetryResult(err)
 			}
+			// TODO - Figure out the real API for getting user orgs
+			url := fmt.Sprintf("https://api.github.com/users/%s/orgs", contrib.GetLogin())
+			resp, e := c.HttpClient.Get(url)
+
+			if e != nil {
+				c.Logf("unable to get org members for %s", contrib.GetLogin())
+			} else {
+				var orgs orgnaizations
+
+				err = json.NewDecoder(resp.Body).Decode(&orgs)
+
+				if err != nil {
+					c.Logf("unable to get  decode json for for %s org", contrib.GetLogin())
+				} else {
+					if len(orgs) > 0 {
+						companies[orgs[0].Login] = struct{}{}
+						continue
+					}
+				}
+			}
+
+			defer resp.Body.Close()
 			company := u.GetCompany()
 			if company != "" {
 				company = strings.ToLower(company)
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,10 @@
+checksum:
+
+  # Algorithm to be used.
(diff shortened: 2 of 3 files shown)


checks/contributors.go
```diff
@@ -38,6 +43,28 @@ func Contributors(c checker.Checker) checker.CheckResult {
...
+			} else {
+				var orgs orgnaizations
+
+				err = json.NewDecoder(resp.Body).Decode(&orgs)
+
+				if err != nil {
+					c.Logf("unable to get  decode json for for %s org", contrib.GetLogin())
+				} else {
+					if len(orgs) > 0 {
+						companies[orgs[0].Login] = struct{}{}
```
inferno-chromium, line 61: So, this does not do the filtering we do in the lines below [e.g. strip inc., llc, etc]. Probably need to move filtering to a helper function and then call in these 2 places.

inferno-chromium (approved):
LGTm with one small change needed.

checks/contributors.go
```diff
@@ -38,6 +43,28 @@ func Contributors(c checker.Checker) checker.CheckResult {
...
+
+			if e != nil {
+				c.Logf("unable to get org members for %s", contrib.GetLogin())
+			} else {
+				var orgs orgnaizations
+
+				err = json.NewDecoder(resp.Body).Decode(&orgs)
+
+				if err != nil {
+					c.Logf("unable to get  decode json for for %s org", contrib.GetLogin())
```
inferno-chromium, line 58: nit: 2 spaces after get. two repeated fors, also it is user for %s user or just %s.

inferno-chromium (approved):
LGTm with one small change needed.

checks/contributors.go
```diff
@@ -38,6 +43,28 @@ func Contributors(c checker.Checker) checker.CheckResult {
...
+				} else {
+					if len(orgs) > 0 {
+						companies[orgs[0].Login] = struct{}{}
+						continue
```
inferno-chromium, line 62: Also, why continue, i think we need to check company as well.

inferno-chromium (approved):
One more change needed, see comment.

checks/contributors.go
naveensrinivasan: Your org's is not shown, because you don't have any orgs that are public.

naveensrinivasan: We shouldn't count for the same contributor in `org` and `company` and that is the reason for continue. Here is my id https://api.github.com/users/naveensrinivasan/orgs , I had to enable the public settings in every org I am in.

checks/contributors.go
naveensrinivasan: Took care of it. Thanks

checks/contributors.go
naveensrinivasan: Took care of it.

inferno-chromium:
E.g. check https://api.github.com/users/dlorenc/orgs vs https://github.com/dlorenc 
https://github.com/dlorenc shows openssf org, whereas https://api.github.com/users/dlorenc/orgs does not. So, might need to consider both??

Also, i dont see filtering logic moved to helper function. maybe github org name does not need filtering (now that i think about it, it can't have llc, inc, etc).

naveensrinivasan:
🤦 My commit got messed up while rebasing on the refactoring.

naveensrinivasan:
> E.g. check https://api.github.com/users/dlorenc/orgs vs https://github.com/dlorenc
> https://github.com/dlorenc shows openssf org, whereas https://api.github.com/users/dlorenc/orgs does not. So, might need to consider both??
> 
You are able to see `dlorenc` in `openssf` because you are also in that org. Or else you will not be able to see the org private status

 
image

inferno-chromium:
> 🤦 My commit got messed up while rebasing on the refactoring.

No worries Sir!, can fix this sometime next week, enjoy long weekend.

> > E.g. check https://api.github.com/users/dlorenc/orgs vs https://github.com/dlorenc
> > https://github.com/dlorenc shows openssf org, whereas https://api.github.com/users/dlorenc/orgs does not. So, might need to consider both??
> 
> You are able to see `dlorenc` in `openssf` because you are also in that org. Or else you will not be able to see the org private status

You are so right! cant see this openssf from incognito, so no change needed.

Issue #173, opened 2021-02-14
Fix golang-lint ci issues


naveensrinivasan:
Enable golang-lint ci issues in the code based on this config.

```
---
run:
  concurrency: 6
  deadline: 5m
linters:
  disable-all: true
  enable:
    - asciicheck
    - bodyclose
    - deadcode
    - depguard
    - dogsled
    - dupl
    - errcheck
    - errorlint
    - exhaustive
    - exportloopref
    - gci
    - gochecknoinits
    - gocognit
    - goconst
    - gocritic
    - gocyclo
    - godot
    - godox
    - goerr113
    - gofmt
    - gofumpt
    - goheader
    - goimports
    - golint
    - gomnd
    - gomodguard
    - goprintffuncname
    - gosec
    - gosimple
    - govet
    - ineffassign
    - interfacer
    - lll
    - makezero
    - maligned
    - misspell
    - nakedret
    - nestif
    - noctx
    - nolintlint
    - paralleltest
    - prealloc
    - predeclared
    - rowserrcheck
    - scopelint
    - sqlclosecheck
    - staticcheck
    - structcheck
    - stylecheck
    - thelper
    - tparallel
    - typecheck
    - unconvert
    - unparam
    - unused
    - varcheck
    - whitespace
    - wrapcheck
    # - exhaustivestruct
    # - forbidigo
    # - funlen
    # - gochecknoglobals
    # - nlreturn
    # - testpackage
    # - wsl
linters-settings:
  gci:
    local-prefixes: sigs.k8s.io/security-profiles-operator
  errcheck:
    check-type-assertions: true
    check-blank: true
  godox:
    keywords:
      - BUG
      - FIXME
      - HACK
  gocritic:
    enabled-checks:
      # Diagnostic
      - appendAssign
      - argOrder
      - badCond
      - caseOrder
      - codegenComment
      - commentedOutCode
      - deprecatedComment
      - dupArg
      - dupBranchBody
      - dupCase
      - dupSubExpr
      - exitAfterDefer
      - flagDeref
      - flagName
      - nilValReturn
      - offBy1
      - sloppyReassign
      - weakCond
      - octalLiteral

      # Performance
      - appendCombine
      - equalFold
      - hugeParam
      - indexAlloc
      - rangeExprCopy
      - rangeValCopy

      # Style
      - assignOp
      - boolExprSimplify
      - captLocal
      - commentFormatting
      - commentedOutImport
      - defaultCaseOrder
      - docStub
[...]

PR #184, opened 2021-02-15
Feature - Include additional linters for golangci

naveensrinivasan:
N/A

Linting #173 

It does not have additional lighters enabled.

Feature - Include additional linters for golangci. This is settings is based on https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/.golangci.yml

None

* **Other information**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,13 +15,12 @@ tidy:
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint:
-	@[ -f $(GOLANGCI_LINT) ] || { \
+	rm -f $(GOLANGCI_LINT) || :
 	set -e ;\
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) v1.29.0 ;\
-	}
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,13 +15,14 @@ tidy:
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint:
+	rm -rf bin || :
 	@[ -f $(GOLANGCI_LINT) ] || { \
```
inferno-chromium, line 19: This file exists check can be removed then if we are removing the dir/file anyway at start.

Makefile
```diff
@@ -15,13 +15,14 @@ tidy:
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint:
+	rm -rf bin || :
```
inferno-chromium, line 18: Why remove entire dir, why not just remove the file e.g. rm -f $(GOLANGCI_LINT) ?

inferno-chromium (approved):
LGTM with a change needed.

Makefile
naveensrinivasan: I agree we can just remove the `rm -f $(GOLANGCI_LINT)`

naveensrinivasan: Took care of it!

Makefile
naveensrinivasan: Took care of it! Thanks

PR #186, opened 2021-02-16
Slow down to every few days.

dlorenc:

The cron runs every day right now

The cron will only run twice a week. We added more projects than we can scan in one day.

* **Other information**:


naveensrinivasan (approved):

inferno-chromium (approved):

naveensrinivasan:
Looks like there is a bug in the action. I will address that as part of circleci  move of the e2e tests.

PR #199, opened 2021-02-19
Feature - Vendor dependencies

naveensrinivasan:

Feature vendor go packagees

Download external dependecies

  Vendoring gives your application projects the most durability since the project owns every line of
  source code it needs to build the applications. If you want a **_reproducible build_** without needing
  to rely on external services (like module mirrors) and being connected to the network,
  vendoring is the solution.
  
  These are other benefits of vendoring:
  
  If dependencies are removed from the VCS or somehow proxy servers lose modules, you are covered.
  Upgrading dependencies can be seen by running diffs and you maintain a history.
  You will have the ability to trace and debug your dependencies and test changes if necessary.
  Once you run a go mod tidy and go mod vendor your changes will be replaced.
  
  source for this comments https://www.ardanlabs.com/blog/2020/04/modules-06-vendoring.html

No

* **Other information**:


inferno-chromium (commented):
Sorry i don't like this approach of putting vendor deps in source code and checking it in. this may be important for super critical software, but i dont think scorecards.

naveensrinivasan:
Can you please explain the reasons? 

> Sorry i don't like this approach of putting vendor deps in source code and checking it in. this may be important for super critical software, but i dont think scorecards.

naveensrinivasan:
What happens when one of the dependencies decides to delete the repo? How are we going to deal with that?

inferno-chromium:
> Can you please explain the reasons?
> 
> > Sorry i don't like this approach of putting vendor deps in source code and checking it in. this may be important for super critical software, but i dont think scorecards.

Atleast for go, they can be all freezed nicely in mod files (with shas), why bloat the repo and check in deps source explicitly. it becomes more cumbersome to do updates, too many file changes, and history grows significantly.

if deps decides to delete repo, we should move to alternate one. all these should be pretty common.

naveensrinivasan:
> > Can you please explain the reasons?
> > > Sorry i don't like this approach of putting vendor deps in source code and checking it in. this may be important for super critical software, but i dont think scorecards.
> 

> Atleast for go, they can be all freezed nicely in mod files (with shas), why bloat the repo and check in deps source explicitly. it becomes more cumbersome to do updates, too many file changes, and history grows significantly.

I agree with SHA's

> if deps decides to delete repo, we should move to alternate one. all these should be pretty common.

Now that is a risk. I am good for now. We can take this discussion later :)

Issue #217, opened 2021-02-24
Enable docker build check as part of CI


naveensrinivasan:
The docker build and tests are not part of CI. Enable the check for docker build and tests.

naveensrinivasan:
Closing it https://github.com/ossf/scorecard/pull/219

PR #219, opened 2021-02-25
Feature - Include e2e tests for docker

naveensrinivasan:
N/A

Feature

Docker does not have e2e tests, #217 

1. Included e2e tests for docker.
2. Included`.Dockerignore` to ignore files.
3. Included Docker builds in the Makefile.

None

* **Other information**:

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -5,7 +5,8 @@ on:
     types: [ok-to-test-command]
 
 name: Integration tests
-
+env:
+  IMAGE_NAME: scorecard
 jobs:
   integration-trusted:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -92,6 +93,27 @@ jobs:
...
+          if [[ ! -v GITHUB_AUTH_TOKEN ]]; then
+              echo "GITHUB_AUTH_TOKEN is not set, skip test"
+              exit 1
+          elif [[ -z "$GITHUB_AUTH_TOKEN" ]]; then
+              echo "GITHUB_AUTH_TOKEN is set to the empty string, skip test"
+              exit 1
+          else
+            go get github.com/onsi/ginkgo/ginkgo@v1.14.2
+            go mod download
+           	mkdir -p bin
```
inferno-chromium, line 112: maybe we should put results in output directory instead of bin. we do this in one more place.
also indent seems off by 2 spaces.

inferno-chromium (approved):

.github/workflows/integration.yml
naveensrinivasan: Agreed, we need to put the results in the output directory. I will fix that part of another PR for the entire repo.

PR #222, opened 2021-02-26
Reduced GitHub API calls for security check

### LINKED ISSUE #221: Reducing GitHub API calls in security policy check
The check for security policy code  https://github.com/ossf/scorecard/blob/main/checks/security_policy.go#L31-L47 could potentially use `16` API calls.

This can be reduced to `1` call by using downloading the archive of the repository and checking for the files locally, similar to this https://github.com/ossf/scorecard/blob/main/checks/frozen_deps.go#L37-L46

naveensrinivasan:
No, the `e2e` test cover for these

Reduce the number of API calls to GitHub. Closes #221 

The current behavior uses up to `16` calls to GitHub API for checking the existence of Security.md 

Reduced the number of calls to GitHub API from 16 to a max of 2 calls.
Utilized tarball to download and check for the contents of those files.

No

* **Other information**:

### DIFF
--- a/checks/checkforfile.go
+++ b/checks/checkforfile.go
@@ -1,0 +1,76 @@
+// Copyright 2020 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


checks/checkforfile.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/ossf/scorecard/checker"
+)
+
+// CheckIfFileExists downloads the tar of the repository and calls the predicate to check
+// for the occurence.
+func CheckIfFileExists(c checker.Checker, predicate func(name string,
+	Logf func(s string, f ...interface{})) bool) checker.CheckResult {
+	const pass int = 10
+	passResult := checker.CheckResult{
```
inferno-chromium, line 31: can we move this and const pass int=10 in checker/check.go, next to InconclusiveResult. PassResult would be a common use result.

inferno-chromium (approved):
So awesome!

PR #224, opened 2021-02-26
feat - Included tests for disk cache

naveensrinivasan:

Tests for disk cache.

Not many tests for disk cache.

Included tests for disk cache.
Cleaned up tests.

None

* **Other information**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,8 @@ GOBIN ?= $(GOPATH)/bin
 GINKGO ?= $(GOBIN)/ginkgo
 IMAGE_NAME = scorecard
 OUTPUT = output
-
+FOCUS_DISK_TEST="E2E TEST:Disk Cache|E2E TEST:executable"
+IGNORED_CI_TEST="E2E TEST:blob|E2E TEST:Disk Cache|E2E TEST:executable"
 .PHONY: help
 help:  ## Display this help
 	@awk \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -57,16 +58,35 @@ e2e: build check-env ginkgo
...
+	rm -rf cache
+	mkdir $(OUTPUT)
+	mkdir cache
+	@echo Focusing on these tests $(FOCUS_DISK_TEST)
+	USE_DISK_CACHE=1 DISK_CACHE_PATH="./cache" \
+				   ./scorecard --repo=https://github.com/ossf/scorecard \
+				   --show-details --metadata=openssf  --format json > ./$(OUTPUT)/results.json
+	USE_DISK_CACHE=1 DISK_CACHE_PATH="./cache" ginkgo -p  -v -cover --focus=$(FOCUS_DISK_TEST)  ./e2e/...
+	# Rerun the test with the disk cache
+	USE_DISK_CACHE=1 DISK_CACHE_PATH="./cache" ./scorecard \
```
inferno-chromium, line 85: I don't understand the difference from these two commands next from the previous ones. Are previous ones supposed to be run with USE_DISK_CACHE=0 ?

inferno-chromium: Can you also move "./scorecard" to next line, to match other command.

inferno-chromium (approved):

Makefile
naveensrinivasan: The whole idea is to start without a cache in the first run and in the second run it should fetch from the cache. 

The goal of this test is to make sure the cache didn't have any issues.

inferno-chromium: Makes sense, thanks for clarifying this.

Issue #33, opened 2020-10-26
Feature: Add a mode to point at a Package URL instead of a github URL


dlorenc:
This would allow scanning language-package-manager dependencies directly, rather than forcing a user to first resolve the package name to a github URL themselves.

Something like:

```shell
scorecard --npm-package=electorn
```

to catch https://portswigger.net/daily-swig/open-source-security-malicious-npm-packages-broadcast-sensitive-user-data-online

PR #226, opened 2021-03-01
Add pypi and ruby gems package support.

inferno-chromium:
Adds some more package managers to
https://github.com/ossf/scorecard/issues/33

No

Add pypi and ruby gems package support.

Only npm package manager supported

npm, pypi and rubygems supported.

No

* **Other information**:
NA


naveensrinivasan (approved):
Great! I think we need some `e2e` for these, to make sure we don't break these features.

Issue #202, opened 2021-02-21
Reducing GitHub API calls to scale scanning repositories


naveensrinivasan:
The GitHub API calls are throttled which makes it hard to scale the number of repositories to scan and provide results.

The code would have to wait for tens of minutes before continuing 
`{"level":"warn","ts":1613869247.8747272,"caller":"roundtripper/roundtripper.go:139","msg":"Rate limit exceeded. Waiting 44m34.125286853s to retry..."}`

### Scorecard checks for these don't need GitHub API, it requires a Git API

1. Active 
2. Frozen-Deps 
3. CodeQLInCheckDefinitions 
4. Security-Policy
5. Packaging

### Potential solution
1. Clone the Git Repo
2. Git pull on these repo's on a `cron` - to get the updates
3. Use an API to query these repositories directly instead of the GitHub

The https://github.com/go-git/go-git project provides an API on Git which could be used for avoiding the GitHub API limitations.

With `httpcache` https://github.com/ossf/scorecard/issues/80#issuecomment-782723182 and reducing the number of GitHub API calls, we should be able to scale the scanning number of repositoreis.

related to #80

naveensrinivasan:
Also, this approach could be used to Scan not GitHub repositories like `Gitlab` which helps one API irrespective of the provider.

inferno-chromium:
Cloning might be too heavy weight for some big repos, and slow too.

Maybe lets start with httpcache first. We can also scale the number of github tokens. right now, we have 2, we can easily go into 4-5 (seperated with comma).

naveensrinivasan:
> Cloning might be too heavy weight for some big repos, and slow too.
> 
> Maybe lets start with httpcache first. We can also scale the number of github tokens. right now, we have 2, we can easily go into 4-5 (seperated with comma).

Cloning can be async as another cron job and it is a one-time effort. Cloning should not be run as part of scorecard , probably give an additional option to look at a location for cached git repo's if not fetch them from GitHub.com.

inferno-chromium:
> > Cloning might be too heavy weight for some big repos, and slow too.
> > Maybe lets start with httpcache first. We can also scale the number of github tokens. right now, we have 2, we can easily go into 4-5 (seperated with comma).
> 
> Cloning can be async as another cron job and it is a one-time effort. Cloning should not be run as part of scorecard , probably give an additional option to look at a location for cached git repo's if not fetch them from GitHub.com.

Makes sense.

[...]

PR #227, opened 2021-03-01
gitcache - Scaling the scorecard scans

naveensrinivasan:
Not yet

Feature

Scaling issues for scorecard

When scorecard uses gitcache it can scale scanning repositories

None

* **Other information**:

#

### gitcache - Reducing the GitHub API usage for scaling scorecard

To scale 1000's repositories it is important that the codebase is cognizant on the usage of the GitHub API's.

One of the strategies to avoid using the API directly cloning the repository.

1. Clone the repository anonymously (not using GitHub API using token).
2. Tarball the default branch and compress it.
3. Store the compressed file into a blob store GCS.
4. Store the last commit date within the blob.

#### How is the above going to scale scorecard scans?

Scorecard checks for these don't need GitHub API, it requires a Git API.
1. Active - Checks for the last commit in 90 days. This is can be fetched from the above blob storage.
1. Frozen-Deps - Checks for a file within the tarball.
1. CodeQLInCheckDefinitions - Checks for file content within the tarball.
1. Security-Policy - Checks for a file name within the tarball.
1. Packaging - Checks for file content within the tarball.

The number of checks within scorecard https://github.com/ossf/scorecard#checks is `14` out of which12` use GitHub API's. The two that do not use the GitHub API's are `Fuzzing` and `CII`.

With the above strategy, we have reduced around **41%** of the GitHub API calls.

https://github.com/ossf/scorecard/issues/202

#### Is this an optional feature? Can the scorecard run without this option?

This is a separate package that has to run to populate the blob storage. The scorecard will have a flag to either load from the `gitcache` which is the blob store or use the standard API

#### What does the gitcache do? 

image

- It fetches the git repository only with the **lastcommit**
- Extracts the `lastcommit` time.
- Removes the `.git` folder.
- `tar` and `gzip` the folder.
- Updates the `blob` with 
    - `lastcommit` - time of the last commit.
    - `lastsynctime` - The time that this git repository was synced with the blob cache.
    - `tar` - The tarball of the git repository.

#### Can this be used for Non-GitHub repositories like Gitlab?

Yes, this can be used for GitLab or other Non-GitHub repositories.

#### Can gitcache be scaled?

Yes, gitcache can be scaled because it does not hold any state or need to be throttled by an exteranl API. gitcache would be exposed as an HTTP API which can be deployed in `k8s` with autoscaling capabilities.

#### How can scorecard determine when was the last sync?

gitcache stores the last sync for each repository within its folder.

#### Can gitcache be extended to expose other properties of the Git Repo?

Yes, it can be extended for the other properties of the git repo, because it stored as key/value pairs.

#### Has this been tested with real data?

Yes, this has been tested with real data. It is syncing `2000` git repositories from _projects.txt_ which is within the `cron` folder.

The bottom of the picture shows `588` folders.

#### Can this be used for non-scorecard repositories?

Yes, this data can be used by other projects like https://github.com/ossf/criticality_score


gitcache/main.go
```diff
@@ -0,0 +1,149 @@
...
+	"os"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/mholt/archiver"
+)
+
+func main() {
```
naveensrinivasan, line 17: This is an initial implementation. This will be an `HTTP` server.

gitcache/main.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// storing the archive to the blob
+	err = bucket.Set(fmt.Sprintf("gitcache/%s/%s/tar", repo.Owner, repo.Repo), data)
+	if err != nil {
+		log.Panic(err)
+	}
+}
+
+type RepoURL struct {
```
naveensrinivasan, line 114: Not reusing the `RepoURL` struct . https://go-proverbs.github.io/ `A little copying is better than a little dependency.`

We could probably pull out `gitcache` into a separate `repo`

naveensrinivasan:
The integration on the scorecard is not yet done. After this PR gets merged it will be implemented.

dlorenc:
Nice! What do you think about keeping the *full* git repos instead of zipping one branch?

That way the update flow would look like:

* pull gzip from GCS
* unzip git repo
* git pull origin <default-branch>
* update metadata (last sync, etc.)
* gzip, reupload to GCS

It's more data getting stored in GCS, but the git protocol is optimized to only fetch changes from the server (GitHub). So by storing the full history, we'd only be doing an incremental update each time from the remote.

naveensrinivasan:
> Nice! What do you think about keeping the _full_ git repos instead of zipping one branch?
> 
> That way the update flow would look like:
> 
> * pull gzip from GCS
> * unzip git repo
> * git pull origin
> * update metadata (last sync, etc.)
> * gzip, reupload to GCS
> 
> It's more data getting stored in GCS, but the git protocol is optimized to only fetch changes from the server (GitHub). So by storing the full history, we'd only be doing an incremental update each time from the remote.

I agree, that will be better approach in solving this problem. 

I can do that as a following PR.

inferno-chromium:
Why do the back-and-forth from GCS ? Why not cache the repo in local FileStore - https://cloud.google.com/filestore. If anything corrupts or git pull fails, then delete repo and redo?

naveensrinivasan:
> Why do the back-and-forth from GCS ? Why not cache the repo in local FileStore - https://cloud.google.com/filestore. If anything corrupts or git pull fails, then delete repo and redo?

I wasn't aware of that. Does `filestore` support Multiple RW? 

Also that means this solution is exclusive for GCP , unlike blob which is available with other cloud providers.

inferno-chromium:
> https://cloud.google.com/filestore

Yes it support multiple RW, it is like a NFS in project's network. Might need do a lock file before a bot starts updating, to avoid concurrent updates / repo corruption.

i dont think exclusive to gcp matters here, since this is just to cache repos. People can do their own NFS anywhere in their own cloud. From a setup perspective, this just shows as a dir mount on a bot's disk. if someone is just running one bot, they can just pick a bot with a very large disk.

naveensrinivasan:
> > https://cloud.google.com/filestore
> 
> Yes it support multiple RW, it is like a NFS in project's network. Might need do a lock file before a bot starts updating, to avoid concurrent updates / repo corruption.

That is good!

> i dont think exclusive to gcp matters here, since this is just to cache repos. People can do their own NFS anywhere in their own cloud. From a setup perspective, this just shows as a dir mount on a bot's disk. if someone is just running one bot, they can just pick a bot with a very large disk.

I agree with that approach.

Now that the data is cached in a local disk. The consumers of this data still need to get this data. File locking in NFS is hard. My suggestion would be each time we fetch the update, tarball and push it to the GCS bucket so that the consumers (scorecard) can directly hit the bucket and get the data. 

This avoids locking the reader/writer and extending an API on top of this cache. 

The max capacity is 320 TB, won't become an issue as we scale?

naveensrinivasan:
The `k8s` driver for gcp filter store is still in beta also it has 

>Non-root access: By default, GCFS instances are only writable by the root user and readable by all users. Provide a CreateVolume parameter to set non-root owners.

https://github.com/kubernetes-sigs/gcp-filestore-csi-driver

naveensrinivasan:
> > https://cloud.google.com/filestore
> 
> Yes it support multiple RW, it is like a NFS in project's network. Might need do a lock file before a bot starts updating, to avoid concurrent updates / repo corruption.
> 
> i dont think exclusive to gcp matters here, since this is just to cache repos. People can do their own NFS anywhere in their own cloud. From a setup perspective, this just shows as a dir mount on a bot's disk. if someone is just running one bot, they can just pick a bot with a very large disk.

I think for us sticking with GCS buckets would make it stable as the filestore for k8s is not fully mature. There are few gotchas.

As we scale more we thank rethink moving to Filestore. Thoughts?

inferno-chromium:
> > > https://cloud.google.com/filestore
> > 
> > 
> > Yes it support multiple RW, it is like a NFS in project's network. Might need do a lock file before a bot starts updating, to avoid concurrent updates / repo corruption.
> > i dont think exclusive to gcp matters here, since this is just to cache repos. People can do their own NFS anywhere in their own cloud. From a setup perspective, this just shows as a dir mount on a bot's disk. if someone is just running one bot, they can just pick a bot with a very large disk.
> 
> I think for us sticking with GCS buckets would make it stable as the filestore for k8s is not fully mature. There are few gotchas.
> 
> As we scale more we thank rethink moving to Filestore. Thoughts?

Ok, lets stick to gcs then.

naveensrinivasan:
> > > > https://cloud.google.com/filestore
> > > 
> > > 
> > > Yes it support multiple RW, it is like a NFS in project's network. Might need do a lock file before a bot starts updating, to avoid concurrent updates / repo corruption.
> > > i dont think exclusive to gcp matters here, since this is just to cache repos. People can do their own NFS anywhere in their own cloud. From a setup perspective, this just shows as a dir mount on a bot's disk. if someone is just running one bot, they can just pick a bot with a very large disk.
> > 
> > 
> > I think for us sticking with GCS buckets would make it stable as the filestore for k8s is not fully mature. There are few gotchas.
> > As we scale more we thank rethink moving to Filestore. Thoughts?
> 
> Ok, lets stick to gcs then.

Ok, Cool! Can we merge this PR and I can get the next PR with those changes?

gitcache/main.go
```diff
@@ -0,0 +1,149 @@
...
+	// opening the blob
+	bucket, e := New(blob)
+
+	if e != nil {
+		log.Panic(e)
+	}
+
+	tpath := path.Join(tarDir, fmt.Sprintf("%s.tar.gz", repo.Repo))
+	// storing the last commit to the blob
+	err = bucket.Set(fmt.Sprintf("gitcache/%s/%s/lastcommit", repo.Owner, repo.Repo), []byte(fmt.Sprintf("%64b\n", commit.Author.When.Unix())))
```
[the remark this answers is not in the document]
naveensrinivasan: I am going to refactor this to clean up in my next PR.

PR #231, opened 2021-03-02
Implemented docker for gitcache

naveensrinivasan:
N/A

feature

 gitcache does not have a docker image

 Implemented docker for gitcache

none

* **Other information**:


gitcache/main.go
```diff
@@ -4,139 +4,191 @@ import (
...
+	err = os.RemoveAll(path.Join(storage.GitDir, ".git"))
 	if err != nil {
 		log.Panic(err)
 	}
 
-	// creating an archive
-	archive := archiver.NewTarGz()
-
-	err = archive.Archive([]string{dir}, archivePath)
+	data, err = archiveFolder(storage.GitDir, storage.BlobArchiveFile)
```
oliverchang, line 126: If this errors out for some reason, the bucket will be left in a bad state, as we already stored the last commit information etc prior to this. 

Let's write the metadata *after* we archive and upload successfully? 

Have you also considered using GCS metadata instead of individual files as well, which would also prevent these kind of de-sync issues?

gitcache/main.go
```diff
@@ -4,139 +4,191 @@ import (
...
+	tarformat := archiver.DefaultTarGz
+	err := tarformat.Archive([]string{folderToArchive}, archivePath)
+	if err != nil {
+		return nil,
+			errors.Wrapf(err, "unable to read the archive the folder %s to the archive path %s", folderToArchive, archivePath)
+	}
 
-func (r *RepoURL) Type() string {
-	return "repo"
+	data, err := ioutil.ReadFile(archivePath)
```
oliverchang, line 162: This doesn't seem like it scales if we are dealing with very large repos. What if we try to clone e.g. a browser like chromium (which is several GBs)? 

Can we just write them to disk somewhere and upload to GCS that way?

oliverchang (commented):
Nice! 

I'm new to this project, so please bear with me as I attempt o review this :)

gitcache/main.go
```diff
@@ -4,139 +4,191 @@ import (
...
 
-	cwd, err := os.Getwd()
-	if err != nil {
-		log.Panic(err)
+	if blob == "" {
+		sugar.Panic("BLOB_URL env is not set.")
+	}
+
+	// opening the blob
+	bucket, e := New(blob)
```
oliverchang, line 40: This New() call here is also a bit confusing, being in the main package. Let's rename this to NewBucket or something similar.

gitcache/storagepath.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Returns path for blob, archiving and also creates temp directories for archiving.
+func getStoragePath(repo RepoURL) (StoragePath, error) {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return StoragePath{}, errors.Wrap(err, "unable to the current working dir")
+	}
+
+	bucketName := fmt.Sprintf("gitcache/%s/%s/%s", repo.Host, repo.Owner, repo.Repo)
```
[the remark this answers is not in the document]
oliverchang: (Or perhaps this variable should be renamed bucketPath?)

gitcache/main.go
naveensrinivasan: > If this errors out for some reason, the bucket will be left in a bad state, as we already stored the last commit information etc prior to this.

> Let's write the metadata after we archive and upload successfully?

Agreed, moved it to the end of the upload. 

> Have you also considered using GCS metadata instead of individual files as well, which would also prevent these kind of de-sync issues?

I wasn't aware of that. I agree it is a valid point. In the upcoming PR's we can fix this.

gitcache/storagepath.go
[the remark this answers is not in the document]
naveensrinivasan: We need slashes to have hierarchal storage. It is working in GCS.

gitcache/storagepath.go
```diff
@@ -0,0 +1,58 @@
...
+	gitDir := fmt.Sprintf(repo.Host + repo.Owner + repo.Repo)
```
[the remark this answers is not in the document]
naveensrinivasan: Got it! Thanks

gitcache/storagepath.go
```diff
@@ -0,0 +1,58 @@
...
+	GitTarFile         string // the tar file name for the git folder
+	BlobArchiveDir     string // directory which is going archive the git folder without .git
+	BlobArchiveFile    string // the tar file for the git folder without  .git
+	BlobLastCommitPath string // blob path for storing the last commit
+	BlobLastSyncPath   string // blob path for storing the lasy sync time
+	BlobGitFolderPath  string // blob path for storing the GitTarFile
+	BlobArchivePath    string // blob path for storing the archive file BlobArchiveFile
+}
+
+// Returns path for blob, archiving and also creates temp directories for archiving.
```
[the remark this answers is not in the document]
naveensrinivasan: Yes, Took care of it.

gitcache/repourl.go
```diff
@@ -0,0 +1,44 @@
...
+type RepoURL struct {
+	Host, Owner, Repo string
+}
+
+func (r *RepoURL) String() string {
+	return fmt.Sprintf("%s/%s/%s", r.Host, r.Owner, r.Repo)
+}
+
+func (r *RepoURL) Type() string {
+	return "repo"
```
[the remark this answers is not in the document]
naveensrinivasan: The whole RepoURL was copied from the other package. I am still accepting the responsibility. This is not required.

gitcache/repourl.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"log"
+	"net/url"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+type RepoURL struct {
+	Host, Owner, Repo string
```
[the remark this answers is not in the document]
naveensrinivasan: I think the field names are descriptive enough IMO. But I can add comments if that would help.

gitcache/repourl.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	const splitLen = 2
+	split := strings.SplitN(strings.Trim(parsedURL.Path, "/"), "/", splitLen)
+	if len(split) != splitLen {
+		log.Fatalf("invalid repo flag: [%s], pass the full repository URL", s)
+	}
+
+	r.Host, r.Owner, r.Repo = parsedURL.Host, split[0], split[1]
+
```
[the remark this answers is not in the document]
naveensrinivasan: Got it. Thanks

gitcache/main.go
```diff
@@ -4,139 +4,191 @@ import (
...
-	// Clone the given repository to the given directory
-	r, err := git.PlainClone(dir, false, &git.CloneOptions{
-		URL:   fmt.Sprintf("http://%s/%s/%s", repo.Host, repo.Owner, repo.Repo),
-		Depth: 1, // Just fetch the last commit
-	})
+	// checks if there is an existing git repo in the bucket
+	data, exists := bucket.Get(storage.BlobGitFolderPath)
+	alreadyUptoDate := false
+
+	if exists {
```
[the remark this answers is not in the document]
naveensrinivasan: Agreed. Thanks

gitcache/storagepath.go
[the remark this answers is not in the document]
oliverchang: Got it, so this isn't referring to the actual bucket name itself, but rather the name/path of the object. So, let's rename this to bucketPath instead to be clear?

gitcache/main.go
naveensrinivasan: Agree, it might not be able to scale very well.  What is your suggestion with the go-cloud package?

gitcache/main.go
naveensrinivasan: Agreed. Thanks

gitcache/main.go
oliverchang: Looks like the gcsblob package that you are using does not support this, so I would recommend using the main gcs package, which provides methods using: https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.NewWriter and io.Copy or something similar.

naveensrinivasan: I did my research and couldn't find it in the gocloud package. 

We are trying to keep this open across cloud vendors and that is the reason for choosing that package.  

So this would make it only for GCP which is something we should avoid.

naveensrinivasan: Looks that is specific to GCP. We would like to keep it open across cloud vendors.

oliverchang: That's a good point. Looking more closely at go-cloud, looks like they offer this equivalent feature instead: https://pkg.go.dev/github.com/google/go-cloud@v0.7.0/blob#Bucket.NewWriter. Can we use this?

naveensrinivasan: Cool, Yes I think it is possible. Let me give it a try. Thanks

naveensrinivasan: The underlying library that the code is using for `tar` and `untar` doesn't support IO writer. The standard `tar` and `untar` funcs have a problem in handling symlinks. 

I agree this is important. But let's take this as an action item and address it.

oliverchang: OK, let's address this in a future PR. I guess we should be able to use that library to write to a file on disk first, and then uploading that way right?

naveensrinivasan: Yes, It should.

gitcache/main.go
```diff
@@ -4,139 +4,190 @@ import (
...
+	// gets all the path configuration.
+	storage, err := getStoragePath(repo)
 	if err != nil {
-		log.Panic(err)
+		sugar.Panic(err)
+	}
+
+	defer os.RemoveAll(storage.GitDir)
+	defer os.RemoveAll(storage.GitTarDir)
+	defer os.RemoveAll(storage.BlobArchiveDir)
```
oliverchang, line 63: It would be better to make a method so that we can do `defer storage.Cleanup()` rather than expecting the caller to know which ones to cleanup.

oliverchang (approved):
thanks a lot for your change and addressing my comments!

LGTM with just one last suggestion.

gitcache/main.go
naveensrinivasan: Good suggestion! Thanks, implemented it.

PR #244, opened 2021-03-05
docker builds for scorecard cron

naveensrinivasan:
N/A

Bug fix

Docker builds are broken for scorecard cron.

Fixed the docker build for scorecard cron and as well as updated the
integration to test for the docker builds.

None

* **Other information**:

### DIFF
--- a/Dockerfile.gsutil
+++ b/Dockerfile.gsutil
@@ -11,12 +11,21 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.16 as app
-COPY . /go/src/github.com/ossf/scorecard
-WORKDIR /go/src/github.com/ossf/scorecard
-RUN [ "go", "build", "."]
+
+# syntax = docker/dockerfile:1-experimental
+
+FROM  golang:1.16 as base
+WORKDIR /src
+ENV CGO_ENABLED=0
+COPY go.* ./
+RUN go mod download
+COPY . ./
+
+FROM base AS build
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 go build -a -tags netgo -ldflags '-w -extldflags "-static"' -o /out/scorecard .
 
 FROM gcr.io/cloud-builders/gsutil
-WORKDIR /go/src/github.com/ossf/scorecard
-COPY --from=app /go/src/github.com/ossf/scorecard /go/src/github.com/ossf/scorecard
-ENTRYPOINT [ "./scorecard" ]
+COPY --from=build /out/scorecard /
+ENTRYPOINT [ "/scorecard" ] 
--- a/Makefile
+++ b/Makefile
@@ -105,4 +105,5 @@ verify-go-mod: ## Verify the go modules
 dockerbuild: ## Runs docker build
 	$(call ndef, GITHUB_AUTH_TOKEN)
 	docker build . --file Dockerfile --tag $(IMAGE_NAME) 
+	docker build . --file Dockerfile.gsutil --tag $(IMAGE_NAME)-gsutil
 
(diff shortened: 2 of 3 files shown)


Dockerfile.gsutil
```diff
@@ -11,12 +11,21 @@
...
-COPY . /go/src/github.com/ossf/scorecard
-WORKDIR /go/src/github.com/ossf/scorecard
-RUN [ "go", "build", "."]
+
+# syntax = docker/dockerfile:1-experimental
+
+FROM  golang:1.16 as base
+WORKDIR /src
+ENV CGO_ENABLED=0
+COPY go.* ./
```
inferno-chromium, line 20: Is this needed since line 22 copies . completely ?

Makefile
```diff
@@ -105,4 +105,5 @@ verify-go-mod: ## Verify the go modules
 dockerbuild: ## Runs docker build
 	$(call ndef, GITHUB_AUTH_TOKEN)
 	docker build . --file Dockerfile --tag $(IMAGE_NAME) 
+	docker build . --file Dockerfile.gsutil --tag $(IMAGE_NAME)-gsutil
```
inferno-chromium, line 108: Why do we need both Dockerfile and Dockerfile.gsutil commands, ideally just Dockerfile should be sufficient ?

Dockerfile.gsutil
```diff
@@ -11,12 +11,21 @@
...
+RUN go mod download
+COPY . ./
+
+FROM base AS build
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 go build -a -tags netgo -ldflags '-w -extldflags "-static"' -o /out/scorecard .
```
inferno-chromium, line 27: isnt there a make target we can use here ?

Dockerfile.gsutil
```diff
@@ -11,12 +11,21 @@
...
 
 FROM gcr.io/cloud-builders/gsutil
-WORKDIR /go/src/github.com/ossf/scorecard
-COPY --from=app /go/src/github.com/ossf/scorecard /go/src/github.com/ossf/scorecard
-ENTRYPOINT [ "./scorecard" ]
+COPY --from=build /out/scorecard /
+ENTRYPOINT [ "/scorecard" ] ~
```
inferno-chromium, line 31: Why ~ at end ?

inferno-chromium (commented):
Some questions to understand this better.

Makefile
naveensrinivasan: The `gsutil` is required by the `cron` to run and copy the images to `GCS`. 

We don't want the base image to contain `gstuil`  which is unnecessary, increases the chances of vulnerabilities.

Dockerfile.gsutil
naveensrinivasan: Agree, we can use `make build`, I will fix it. It was copied from the previous docker file.

Dockerfile.gsutil
naveensrinivasan: 🤦  My vim key bindings, I will fix it.

Dockerfile.gsutil
naveensrinivasan: I tried that it fails with `go mod download: no modules specified (see 'go help mod download')`, so it looks like it is required.

Dockerfile.gsutil
naveensrinivasan: I am not going to use the `make build` target now. That would require lots of changes to test and other places. So that will be another PR. So let's use this for now.

naveensrinivasan:
Enabled the cloudbuild for cron now. It should be building the latest image as we merge to master.
image

Makefile
inferno-chromium: another image seems redundant just for gsutil, it is pretty trusted code, can keep in same. upto you.

Dockerfile.gsutil
inferno-chromium: SG.

Dockerfile.gsutil
inferno-chromium: Thanks for explaining.

inferno-chromium (approved):

PR #248, opened 2021-03-05
Feat- Parmeterize cache folder in gitcache

naveensrinivasan:
Included some tests.

Feature

The cache temp folder was hardcoded to using the current working
directory.

The /tmp filesystem is often backed by tmpfs, which stores files in memory rather than on disk.

With this, it will be using the directory that is configured in `TEMP_DIR`.The `TEMP_DIR` would be an `emptydir` in k8s.

Included basic e2e tests for gitcache.

None

* **Other information**:
None


gitcache/e2e/e2e_http_test.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"bytes"
+	"net/http"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("E2E TEST:HTTP endpoint-gitcache", func() {
+	url := "http://localhost:8080/"
```
naveensrinivasan, line 25: For now, it is hardcoded. We should make this an ENV variable in the subsequent iterations.

gitcache/pkg/repourl.go
```diff
@@ -32,6 +32,10 @@ func (r *RepoURL) String() string {
 	return fmt.Sprintf("%s/%s/%s", r.Host, r.Owner, r.Repo)
 }
 
+func (r *RepoURL) NonURLString() string {
+	return r.Host + r.Owner + r.Repo
```
inferno-chromium, line 36: do we want any seperator here, like underscore or hyphen? good in case we have overlaps.

gitcache/pkg/storagepath.go
```diff
@@ -36,24 +35,20 @@ type StoragePath struct {
...
-	err = os.Mkdir(gitDir, 0755)
+	err := os.Mkdir(gitDir, 0755)
 	if err != nil {
-		return StoragePath{}, errors.Wrapf(err, "unable to create temp directory %s", gitDir)
+		return StoragePath{}, errors.Wrapf(err, "unable to temp directory %s", gitDir)
 	}
 	gitTarPath := path.Join(gitDir, "gitfolder.tar.gz")
 
-	blobArchiveDir, err := ioutil.TempDir(cwd, gitDir+"tar")
+	blobArchiveDir := path.Join(gitDir + "tar")
```
inferno-chromium, line 48: is path.join even needed?

gitcache/pkg/storagepath.go
```diff
@@ -36,24 +35,20 @@ type StoragePath struct {
...
+	err = os.Mkdir(blobArchiveDir, 0755)
 	if err != nil {
-		return StoragePath{}, errors.Wrapf(err, "unable to create temp directory %s", gitDir+"tar")
+		return StoragePath{}, errors.Wrapf(err, "unable to create temp directory for blob archive %s", gitDir+"tar")
```
inferno-chromium, line 51: use blobArchiveDir instead of gitDir+"tar"

gitcache/e2e/e2e_http_test.go
```diff
@@ -0,0 +1,54 @@
...
+			client := &http.Client{}
+			resp, err := client.Do(req)
+			if err != nil {
+				panic(err)
+			}
+			defer resp.Body.Close()
+			Expect(resp.StatusCode).Should(BeEquivalentTo(200))
+		})
+		It("Should fail when an invalid git repo is passed", func() {
+			jsonStr := []byte(`{"url":"http://foo.com/bar/scorecard"}`)
```
inferno-chromium, line 41: foo.com is a valid domain, can we do something that does not connect, invalid one.

inferno-chromium (approved):

gitcache/pkg/storagepath.go
naveensrinivasan: It should have been,` instead of `+`. Good catch! I will fix it. Thanks

gitcache/pkg/repourl.go
naveensrinivasan: I can agree. I can add a separator.

gitcache/e2e/e2e_http_test.go
naveensrinivasan: I can do that 👍

gitcache/pkg/storagepath.go
naveensrinivasan: 👍

PR #259, opened 2021-03-10
Fix - failed cron job docker

naveensrinivasan:
Tested the changes in k8s cluster. Need a way to figure how to automate it.

Bug fix.

CronJob isn't running.

The cron job was failing because of  the .dockerignore ignoring all the other files.

Removed the dockerignore file. 

Also update the cronjob to pull the latest projects.txt from the GitHub.

No

* **Other information**:
None


cron/cron.sh
```diff
@@ -17,6 +17,7 @@ SOURCE="${BASH_SOURCE[0]}"
 input=$(dirname "$SOURCE")/projects.txt
 output=$(date +"%m-%d-%Y").json
 touch "$output"
+curl https://raw.githubusercontent.com/ossf/scorecard/main/cron/projects.txt >projects.txt
```
inferno-chromium, line 20: >$input might be better ?

cron/cron.sh
```diff
@@ -26,7 +27,7 @@ while read -r proj; do
         continue
     fi
     echo "$proj"
-    ./scorecard --repo="$proj" --show-details --format=json >> "$output"
+    ../scorecard --repo="$proj" --show-details --format=json >> "$output"
```
inferno-chromium, line 30: maybe /scorecard instead of ../scorecard might be better ?

inferno-chromium (approved):

cron/cron.sh
naveensrinivasan: I am going to skip these changes for this PR because I would have to test it again which will kill the existing cron job which is running. I would like it to finish to check its output.

PR #270, opened 2021-03-11
Fix - Made the results.json wellformed from cron

### LINKED ISSUE #269: BUG - cron job latest.json is not wellformed
**Describe the bug**
The `cron` job produces https://storage.googleapis.com/ossf-scorecards/latest.json which is not wellformed `json`

**Expected behavior**
The json should be wellformed.

naveensrinivasan:

Bugfix

fix #269

  Fixed the results.json to be wellformed from the cron job.

 Changed the docker image from gsutil to cloudsdk:slim to `sed` binary which is being used with the cron.sh

 Yes, the latest.json will be wellformed.

* **Other information**:

### DIFF
--- a/Dockerfile.gsutil
+++ b/Dockerfile.gsutil
@@ -26,7 +26,7 @@ ARG TARGETOS
 ARG TARGETARCH
 RUN CGO_ENABLED=0 go build -a -tags netgo -ldflags '-w -extldflags "-static"' -o /out/scorecard .
 
-FROM gcr.io/cloud-builders/gsutil
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:slim 
 COPY --from=build /out/scorecard /
 COPY ./cron/ /cron/
 ENTRYPOINT [ "/scorecard" ] 
--- a/cron/cron.sh
+++ b/cron/cron.sh
@@ -17,7 +17,7 @@ SOURCE="${BASH_SOURCE[0]}"
 input=$(dirname "$SOURCE")/projects.txt
 output=$(date +"%m-%d-%Y").json
 touch "$output"
-curl https://raw.githubusercontent.com/ossf/scorecard/main/cron/projects.txt >projects.txt
+echo "{ \"results\": [" >> "$output"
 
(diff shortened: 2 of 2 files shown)


cron/cron.sh
```diff
@@ -17,7 +17,7 @@ SOURCE="${BASH_SOURCE[0]}"
 input=$(dirname "$SOURCE")/projects.txt
 output=$(date +"%m-%d-%Y").json
 touch "$output"
-curl https://raw.githubusercontent.com/ossf/scorecard/main/cron/projects.txt >projects.txt
+echo "{ \"results\": [" >> "$output"
```
inferno-chromium, line 20: probably want to bring back the curl line ?

Dockerfile.gsutil
```diff
@@ -26,7 +26,7 @@ ARG TARGETOS
 ARG TARGETARCH
 RUN CGO_ENABLED=0 go build -a -tags netgo -ldflags '-w -extldflags "-static"' -o /out/scorecard .
 
-FROM gcr.io/cloud-builders/gsutil
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:slim #using cloud-sdk docker image for `sed` utility
```
inferno-chromium, line 29: that comment is not needed.

inferno-chromium (approved):

cron/cron.sh
naveensrinivasan: I realized the `main` branch will always have the `projects.txt` and it was redundant.

Dockerfile.gsutil
naveensrinivasan: Ok. Thanks

cron/cron.sh
inferno-chromium: latest version right?, makes sense

naveensrinivasan: The cloudbuild runs on merge to main.

 The cron uses the latest image from the cloudbuild.  So it will have the latest projects.txt. So the latest project.txt will be the same in the repo and also be the same inside the container.

PR #291, opened 2021-03-19
:seedling: Changed cron to run everyday

naveensrinivasan:
  No

  Feature

  Cron does not run everyday

  With the latest fix for roundrobin token usage, the cron can run everyday.

None

* **Other information**:
None

### DIFF
--- a/k8s/cron.yaml
+++ b/k8s/cron.yaml
@@ -3,7 +3,7 @@ kind: CronJob
 metadata:
   name: daily-score
 spec:
-  schedule: "0 0 * * 0,3"
+  schedule: "0 2 * * *"
   jobTemplate:
     spec:
       template:


inferno-chromium (commented):
Should we scale the system instead ? like from 2K to like 10K repos from gs://ossf-criticality-score/all.csv ?

naveensrinivasan:
We still need more tweaks like gitcache before we take 10,000 repositories.

naveensrinivasan:
> Should we scale the system instead ? like from 2K to like 10K repos from gs://ossf-criticality-score/all.csv ?

This should help people to get daily scorecard results. I understand there might not be many teams using this. But that should be our goal IMO.

inferno-chromium (approved):

Issue #289, opened 2021-03-18
Feature - provide a way for other OSS projects to include their dependencies to be tracked in scorecard


naveensrinivasan:
**Is your feature request related to a problem? Please describe.**
Similar to #287, there should a standard way to parse the dependencies for critical projects add them to `dailyscore` cron job along with metadata.

For example - scorecard cronjob should help track dependencies for critical projects like kubernetes https://twitter.com/spiffxp/status/1371240897531236354

Instead of scorecard doing the work of parsing each repo and adding it to cronjob, the project should provide an option for other automation that other OSS teams can utilize.

**Describe the solution you'd like**
For example -

1. Parse the `go.mod` file for k8s
2. Find the dependencies 
3. Check if the dependencies are in the scorecard cronjob list `projects.txt`. If not add those
4. If there are already there and include additional metadata

naveensrinivasan:
@inferno-chromium Thoughts?

inferno-chromium:
@naveensrinivasan - this is very go centric, what about other languages like python, etc

There is some work at Google to better evaluate dependency tree structure (@robpike). I recommend waiting on that instead of this.

naveensrinivasan:
@inferno-chromium the initial was only for go. But later expand it to other languages

naveensrinivasan:
> @naveensrinivasan - this is very go centric, what about other languages like python, etc
> 
> There is some work at Google to better evaluate dependency tree structure (@robpike). I recommend waiting on that instead of this.

When does Google likely to have the solution for dependency tree structure?  Based on the timeline we could decide to go one way or the other?

inferno-chromium:
No clue, but i think we have lot of other important things to work on (e.g. scaling, more checks, etc).

naveensrinivasan:
Ok. Let’s keep in the backlog.

azeemshaikh38:
IMO, the focus of Scorecards should be repo-specific automated analysis. Tracking dependencies of a repo is a whole different problem/project of its own. If we plan to tackle it, we should do it orthogonal to Scorecards (maybe in a new repo).

I propose closing this issue in lieu of reducing number of open issues we currently have. If we plan to track dependencies in Scorecard at a future date, we can revisit this.

Issue #287, opened 2021-03-18
Feature - Add metadata for Envoy dependencies to scorecard


naveensrinivasan:
**Is your feature request related to a problem? Please describe.**
Now that scorecard is able to scan 2000+ repositories https://storage.googleapis.com/ossf-scorecards/latest.json it would help to add metadata to these scans easily identify the envoy's dependencies using the `--metadata` feature in the scorecard command line. 

**Describe the solution you'd like**
- Identify the dependencies for the envoy.
-  Probably write a tool to update the `./cron/projects.txt` with some like `project=envoy,dependency=true` to the envoy's dependencies
- The `latest.json` will contain this additional metadata which can help identify scorecard for all of `envoy` dependencies.

naveensrinivasan:
@htuch We are trying to scan all of Envoy's dependencies and add metadata to the results using the `metadata` which can be passed in the command line.

`./scorecard  --repo=https://github.com/ossf/scorecard --metadata=project=openssf,critical=true --format json`

``` json
{
  "Repo": "github.com/ossf/scorecard",
  "Date": "2021-03-18",
  "Checks": [
    {
      "CheckName": "Active",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "Branch-Protection",
      "Pass": false,
      "Confidence": 0,
      "Details": null
    },
    {
      "CheckName": "CI-Tests",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "CII-Best-Practices",
      "Pass": false,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "Code-Review",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "Contributors",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "Frozen-Deps",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "Fuzzing",
      "Pass": false,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "Packaging",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "Pull-Requests",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
      "CheckName": "SAST",
      "Pass": true,
      "Confidence": 10,
      "Details": null
    },
    {
[...]

PR #293, opened 2021-03-19
:sparkles: Refactor cron job

naveensrinivasan:
Tested manually

Feature

The projects.txt does not have an option to provide metadata to the workloads.

  * Refactored cron job from shell script to go. It is easier to maintain than shell script.
  * Included metadata to the projects.txt for envoy
  * Provided an option to have metadata to the projects.txt
  * This will help projects filter their dependency by adding metadata to the output. 
  For example when the latest.json is exported to Bigquery , teams can filter the status of their dependencies.

None

* **Other information**:
None


cron/projects.txt
```diff
@@ -1,2314 +1,2314 @@
...
+github.com/zlib-ng/zlib-ng,
+github.com/zloirock/core-js,
+github.com/ZoneMinder/zoneminder,
+github.com/zsh-users/zsh,
+github.com/zsh-users/zsh-completions,
+github.com/zsh-users/zsh-syntax-highlighting,
+github.com/z-song/laravel-admin,
+github.com/zulip/zulip,
+github.com/mwiede/jsch,
+github.com/grailbio/bazel-compilation-database,"project=envoy,dependency=true"
```
naveensrinivasan, line 2293: Included the metadata which will be published in the results.

naveensrinivasan: @asraa FYI..., made these changes to easily track and include results with these additional metadata. I know you are working on helping with this. Just wanted to give you a heads up.

cron/main.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		panic(err)
+	}
+	defer f.Close()
+	if _, err := f.WriteString("\n]}\n"); err != nil {
+		panic(err)
+	}
+
+	// copying the file to the GCS bucket
+	if err := exec.Command("gsutil", "cp", fileName, fmt.Sprintf("gs://%s", os.Getenv("GCS_BUCKET"))).Run(); err != nil {
```
naveensrinivasan, line 91: Later move it to using the actual API instead of the GCS util

cron/projects.txt
```diff
@@ -1,2314 +1,2314 @@
...
+github.com/LCTT/TranslateProject,
+github.com/LMMS/lmms,
+github.com/Laravel-Backpack/CRUD,
+github.com/Laravel-Lang/lang,
+github.com/LemmyNet/lemmy,
+github.com/LessWrong2/Lesswrong2,
+github.com/Lienol/openwrt,
+github.com/LimeSurvey/LimeSurvey,
+github.com/LonamiWebs/Telethon,
+github.com/LuaJIT/LuaJIT,"project=envoy,dependency=true"
```
naveensrinivasan, line 183: @asraa FYI... The projects.txt is now sorted and also provides metadata in the output to track.

cron/main.go
```diff
@@ -0,0 +1,102 @@
...
+	//nolint
+	const checks string = "--checks=Active,CI-Tests,CII-Best-Practices,Code-Review,Contributors,Frozen-Deps,Fuzzing,Packaging,Pull-Requests,SAST,Security-Policy,Signed-Releases,Signed-Tags"
+
+	if _, err = result.Write([]byte(`{"results":[`)); err != nil {
+		panic(err)
+	}
+	for _, r := range repos {
+		fmt.Println(r.Repo)
+		//nolint
+		cmd := exec.Command("../scorecard", fmt.Sprintf("--repo=%s", r.Repo), fmt.Sprintf("--metadata=%s", r.Metadata), checks, "--show-details", "--format=json")
```
oliverchang, line 55: Can we call some function (or perform some refactoring) within the scorecard module, rather than invoke the binary?

cron/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+		_, err = result.WriteString(fmt.Sprintf("\n%s,", string(data)))
+		if err != nil {
+			panic(err)
+		}
+	}
+	result.Close()
+
+	// sed -i '$d' "$output" # removing the trailing comma which will be last line.
+	if err := exec.Command("sed", "-i", "$d", fileName).Run(); err != nil {
```
oliverchang, line 76: This can be all much simplified by handling this in the loop above rather than trying to emulate the bash script. We can also avoid re-opening the file.

E.g. something like

```go
for i, r := range repos {
  ...

  if i < len(repos) - 1 {
    result.WriteString(",")
  }
  ...
}


result.WriteString("\n]}\n")
```

scripts/main.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	err = csvutil.Unmarshal(data, &repos)
+	if err != nil {
+		panic(err)
+	}
+	sort.Slice(repos, func(i, j int) bool {
+		return repos[i].Repo < repos[j].Repo
+	})
+
+	m := make(map[string]string)
```
oliverchang, line 39: nit: the value can just be a bool.

scripts/main.go
```diff
@@ -0,0 +1,46 @@
...
+	repos := []Repository{}
+	data, err := ioutil.ReadAll(projects)
+	if err != nil {
+		panic(err)
+	}
+	err = csvutil.Unmarshal(data, &repos)
+	if err != nil {
+		panic(err)
+	}
+	sort.Slice(repos, func(i, j int) bool {
```
oliverchang, line 35: We don't need to sort since we are using the map to determine duplicates. 

Alternatively, sort and then detect duplicates if there are 2 of the same next to each other (without using the map). 

Using the map alone seems much simpler to me though.

cron/main.go
```diff
@@ -0,0 +1,102 @@
...
+	if _, err = result.Write([]byte(`{"results":[`)); err != nil {
+		panic(err)
+	}
+	for _, r := range repos {
+		fmt.Println(r.Repo)
+		//nolint
+		cmd := exec.Command("../scorecard", fmt.Sprintf("--repo=%s", r.Repo), fmt.Sprintf("--metadata=%s", r.Metadata), checks, "--show-details", "--format=json")
+		// passing the external github token the cmd
+		cmd.Env = append(cmd.Env, fmt.Sprintf("GITHUB_AUTH_TOKEN=%s", os.Getenv("GITHUB_AUTH_TOKEN")))
+		cmd.Stderr = io.Writer(os.Stdout)
```
oliverchang, line 58: Why not just have stderr pipe to actual stderr rather than stdout?

Makefile
```diff
@@ -107,3 +107,6 @@ dockerbuild: ## Runs docker build
 	docker build . --file Dockerfile --tag $(IMAGE_NAME) 
 	docker build . --file Dockerfile.gsutil --tag $(IMAGE_NAME)-gsutil
 
+checkprojects: ## Validates ./cron/projects.txt
+	cd ./scripts && go build  -o validate
```
oliverchang, line 111: nit: we can just call `go run` rather than `go build` and leave this binary around?

cron/projects.txt
```diff
@@ -1,2314 +1,2314 @@
...
+github.com/Unidata/netcdf-c,
+github.com/UnigramDev/Unigram,
+github.com/Unity-Technologies/Graphics,
+github.com/Unity-Technologies/ml-agents,
+github.com/VSCodeVim/Vim,
+github.com/VSCodium/vscodium,
+github.com/Varying-Vagrant-Vagrants/VVV,
+github.com/VictoriaMetrics/VictoriaMetrics,
+github.com/VirtoCommerce/vc-platform,
+github.com/WAVM/WAVM,"project=envoy,dependency=true"
```
oliverchang, line 347: Just a question since I'm unfamiliar with the metadata here. What do these mean exactly?

oliverchang (commented):
Nice!

cron/main.go
naveensrinivasan: Yes, I agree. But it is a larger refactor. Lets do that later.

oliverchang: SGTM.

cron/projects.txt
naveensrinivasan: The metadata is added to the results. https://github.com/ossf/scorecard/blob/6f2a0f43f48079ce5f2ad0db9cb2fd1e0c7aa221/e2e/executable_test.go#L35

Makefile
naveensrinivasan: go run from the root folder with the subdirectory wasn't working

scripts/main.go
naveensrinivasan: Agreed

cron/main.go
naveensrinivasan: Nice! Agreed.

scripts/main.go
naveensrinivasan: Agreed. Thanks

cron/main.go
naveensrinivasan: Agreed

cron/projects.txt
oliverchang: Thanks, I guess my question is more what `project=envoy,dependency=true` means exactly here. Does this mean envoy uses this as a dependency and relies on this for querying? 

Assuming that's what that means, this doesn't seem to scale, if every project needs to add themselves to their own dependencies here.

naveensrinivasan: It is anything that you can add to the results that the project would like to track. Here is an example where `envoy` would like to track its dependencies. 

The goal is to scorecard 10,000 + repo's also want to track critical repo's.

oliverchang: Sure, this is fine for now but this usecase (envoy tracking dependencies this way) in particular doesn't seem particularly reliable or scalable so we'll likely need to change it.

Dependencies change frequently and we can't have every critical project manually add these labels here. It seems more reliable for envoy to compute its own list of dependencies and then checking scorecards from that list.

naveensrinivasan: https://github.com/ossf/scorecard/issues/289
https://github.com/ossf/scorecard/issues/287

naveensrinivasan: There tracking issues and discussions going with respect to this. This refactor is just so that we can get some data and clean up.

oliverchang: Thanks for providing the context!

cron/main.go
```diff
@@ -0,0 +1,97 @@
...
+			if err != nil {
+				panic(err)
+			}
+		}
+	}
+	if _, err := result.WriteString("\n]}\n"); err != nil {
+		panic(err)
+	}
+
+	result.Close()
```
oliverchang, line 83: last nit: we don't need to explicitly close here anymore since we have the defer above.

oliverchang (approved):
LGTM with one last nit.

cron/main.go
naveensrinivasan: 🤦

naveensrinivasan: Thanks!

inferno-chromium:
Yes is the metadata of envoy project that useful, how do we scale this as @oliverchang mentioned. Like there can be many projects that can have that same dependency.

naveensrinivasan:
> Yes is the metadata of envoy project that useful, how do we scale this as @oliverchang mentioned. Like there can be many projects that can have that same dependency.

Agreed, I didn't want `#` comment on the `projects.txt` to track the `envoy` dependency and that was the reason to this metadata field as a hack.

PR #301, opened 2021-03-22
:sparkles: scorecard badges

### LINKED ISSUE #271: Feature Scorecard badges
Scorecard should provide a badge for repositories to include in their README to display their compliance.

# Scorecard badges

The scorecard should provide badges similar to other https://github.com/badges/shields OSS badges for compliance.

## Goals

* Scorecard should provide badge results based on scorecard runs on the (cron)server to ensure compliance and validation. Similar to Codecov. 
* Scorecard should provide a predictable API to fetch the badges. An example could be https://somefqdn/github/ossf/scorecard/badge
* Scorecard results calculation - _TBD_ - The discussion of calculation should be a separate issue.

## Implementaion

- Scorecard would create a separate HTTP application (within the scorecard repository) which would generate the badge.
- Scorecard would use the Results from the scheduled cron run to generate the badge. The results of the cron are stored within the GCS bucket as `latest.json`
- The HTTP application would be stateless.
- The application would be hosted on Google Cloud Run which will scalability with less maintenance.

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG5rOHMgUkVBRE1FLm1kLT4-c2NvcmVjYXJkIGJhZGdlIHNlcnZpY2U6IEdldCBzY29yZWNhcmQgYmFkZ2VcbnNjb3JlY2FyZCBiYWRnZSBzZXJ2aWNlLT4-R0NTOiBGZXRjaCBsYXRlc3QuanNvblxuc2NvcmVjYXJkIGJhZGdlIHNlcnZpY2UtPj5zY29yZWNhcmQgYmFkZ2Ugc2VydmljZTogY2FsY3VsYXRlIHNjb3JlY2FyZCBzY29yZVxuc2NvcmVjYXJkIGJhZGdlIHNlcnZpY2UtPj5rOHMgUkVBRE1FLm1kOiByZXR1cm4gdGhlIGJhZGdlLnN2Z1xuIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

In this example

- The k8s repository README.md would request the scorecard badge service for the badge
- The badge service would fetch the results from the GC bucket which has the `latest.json` results from the cron job
- The badge service calculates the score and returns the SVG.

naveensrinivasan:
  Manually tested. Need to do automate it.

  Feature

 #271

    * Initial implementation for generating scorecard badges.
    * The calculation of the score is right now simple logic.
    *  The request for the badge would be something like `http://somefqdn/host/owner/repo` example `http://localhost:8000/github.com/18F/identity-idp`
    * The badge service uses the cron results to calculate the score.
    * The service is an HTTP endpoint that is stateless.
   

None

* **Other information**:
Fixes #271
* Example of a generated badge that generated after running `curl -o badge.svg  http://localhost:8000/github.com/18F/identity-idp`
image


naveensrinivasan:
ping @inferno-chromium

oliverchang (changes requested):
Thanks for working on this change! 

However I'm not sure this approach will scale, and it adds significant maintenance cost since this renders the badge on demand and we need to maintain this service for serving the badges.

A much more scalable approach here would be to publish these rendered badges in a public GCS bucket instead so that we don't have to worry about the actual serving.

GCS buckets also allow access over HTTP. This is also what we do for OSS-Fuzz. See e.g. https://oss-fuzz-build-logs.storage.googleapis.com/badges/envoy.svg

naveensrinivasan:
> we

How do you think this won’t scale could you please explain further?

dlorenc:
I'd say it's probably less about scalability, a stateless web server would likely be fine :)

Using GCS would simplify things though, and likely be more reliable/easier to maintain since we just don't have to worry about keeping things running.

naveensrinivasan:
> I'd say it's probably less about scalability, a stateless web server would likely be fine :)
> 
> Using GCS would simplify things though, and likely be more reliable/easier to maintain since we just don't have to worry about keeping things running.

I agree on that. It makes sense.

inferno-chromium:
> > I'd say it's probably less about scalability, a stateless web server would likely be fine :)
> > Using GCS would simplify things though, and likely be more reliable/easier to maintain since we just don't have to worry about keeping things running.
> 
> I agree on that. It makes sense.

it should be very simple to do this in cron code here - https://github.com/ossf/scorecard/blob/main/cron/main.go#L68
add that overall % calc logic in main scorecard code, so in json it gives that number too, and then parse that output in json in cron/main.go and create the badges in a bucket, like ossf-scorecard-badges/repo/url/badge.svg or something. this web service needs some thought and we are still thinking on dash and stuff. so, doing this unblocks and start working right away.

naveensrinivasan:
> > > I'd say it's probably less about scalability, a stateless web server would likely be fine :)
> > > Using GCS would simplify things though, and likely be more reliable/easier to maintain since we just don't have to worry about keeping things running.
> > 
> > 
> > I agree on that. It makes sense.
> 
> it should be very simple to do this in cron code here - https://github.com/ossf/scorecard/blob/main/cron/main.go#L68
> add that overall % calc logic in main scorecard code, so in json it gives that number too, and then parse that output in json in cron/main.go and create the badges in a bucket, like ossf-scorecard-badges/repo/url/badge.svg or something. this web service needs some thought and we are still thinking on dash and stuff. so, doing this unblocks and start working right away.

I agree we could move the score calculation to the main scorecard binary. 

I think it should not be cronjob responsibility to generate the badge. There needs to be separation of concerns. The cronjob responsibility needs to be as is. 

The reason I am calling for that is because others in the community could use the same cronjob to run scorecard on bunch of repos , but they might not want badge generation.

inferno-chromium:
> > > > I'd say it's probably less about scalability, a stateless web server would likely be fine :)
> > > > Using GCS would simplify things though, and likely be more reliable/easier to maintain since we just don't have to worry about keeping things running.
> > > 
> > > 
> > > I agree on that. It makes sense.
> > 
> > 
> > it should be very simple to do this in cron code here - https://github.com/ossf/scorecard/blob/main/cron/main.go#L68
> > add that overall % calc logic in main scorecard code, so in json it gives that number too, and then parse that output in json in cron/main.go and create the badges in a bucket, like ossf-scorecard-badges/repo/url/badge.svg or something. this web service needs some thought and we are still thinking on dash and stuff. so, doing this unblocks and start working right away.
> 
> I agree we could move the score calculation to the main scorecard binary.
> 
> I think it should not be cronjob responsibility to generate the badge. There needs to be separation of concerns. The cronjob responsibility needs to be as is.
> 
> The reason I am calling for that is because others in the community could use the same cronjob to run scorecard on bunch of repos , but they might not want badge generation.

Nobody else is running cron. We can make this configurable which is just like we have GCS_BUCKET, we can do GCS_BADGES_BUCKET, which if set, then we write the badges as well. you can create a seperate cron for badges too, but then there will be race conditions on when to generate etc. Best to keep it simple.

naveensrinivasan:
> > > > > I'd say it's probably less about scalability, a stateless web server would likely be fine :)
> > > > > Using GCS would simplify things though, and likely be more reliable/easier to maintain since we just don't have to worry about keeping things running.
> > > > 
> > > > 
> > > > I agree on that. It makes sense.
> > > 
> > > 
> > > it should be very simple to do this in cron code here - https://github.com/ossf/scorecard/blob/main/cron/main.go#L68
> > > add that overall % calc logic in main scorecard code, so in json it gives that number too, and then parse that output in json in cron/main.go and create the badges in a bucket, like ossf-scorecard-badges/repo/url/badge.svg or something. this web service needs some thought and we are still thinking on dash and stuff. so, doing this unblocks and start working right away.
> > 
> > 
> > I agree we could move the score calculation to the main scorecard binary.
> > I think it should not be cronjob responsibility to generate the badge. There needs to be separation of concerns. The cronjob responsibility needs to be as is.
> > The reason I am calling for that is because others in the community could use the same cronjob to run scorecard on bunch of repos , but they might not want badge generation.
> 
> Nobody else is running cron. We can make this configurable which is just like we have GCS_BUCKET, we can do GCS_BADGES_BUCKET, which if set, then we write the badges as well. you can create a seperate cron for badges too, but then there will be race conditions on when to generate etc. Best to keep it simple.

Sounds good. I will change the implementation. Thanks

oliverchang:
> > > > > > I'd say it's probably less about scalability, a stateless web server would likely be fine :)
> > > > > > Using GCS would simplify things though, and likely be more reliable/easier to maintain since we just don't have to worry about keeping things running.
> > > > > 
> > > > > 
> > > > > I agree on that. It makes sense.
> > > > 
> > > > 
> > > > it should be very simple to do this in cron code here - https://github.com/ossf/scorecard/blob/main/cron/main.go#L68
> > > > add that overall % calc logic in main scorecard code, so in json it gives that number too, and then parse that output in json in cron/main.go and create the badges in a bucket, like ossf-scorecard-badges/repo/url/badge.svg or something. this web service needs some thought and we are still thinking on dash and stuff. so, doing this unblocks and start working right away.
> > > 
> > > 
> > > I agree we could move the score calculation to the main scorecard binary.
> > > I think it should not be cronjob responsibility to generate the badge. There needs to be separation of concerns. The cronjob responsibility needs to be as is.
> > > The reason I am calling for that is because others in the community could use the same cronjob to run scorecard on bunch of repos , but they might not want badge generation.
> > 
> > 
> > Nobody else is running cron. We can make this configurable which is just like we have GCS_BUCKET, we can do GCS_BADGES_BUCKET, which if set, then we write the badges as well. you can create a seperate cron for badges too, but then there will be race conditions on when to generate etc. Best to keep it simple.
> 
> Sounds good. I will change the implementation. Thanks

Thanks Naveen!

naveensrinivasan:
Closing it to open a new one with the latest refacoring

Issue #295, opened 2021-03-20
Cron stopped importing into BQ


dlorenc:
The cron stopped importing in biquery:
![image](https://user-images.githubusercontent.com/1714486/111888839-3582a000-89ae-11eb-9cd2-436a3415a0aa.png)

There's a way to debut this but I forget how right now, you need to use gcloud with the job ids to search for the errors. Opening this bug to track that so I don't forget.

naveensrinivasan:
Wow! Is the Bug query open for everyone? We should fix this and add it to the readme.

dlorenc:
> Wow! Is the Bug query open for everyone? We should fix this and add it to the readme.

Yup! It should be publicly readable. It's just an automated import from the GCS bucket.

naveensrinivasan:
@dlorenc Need your help with this. I don't have permission. I am trying to update the Big Query with the latest json structure.

![image](https://user-images.githubusercontent.com/172697/113446331-9609d800-93c5-11eb-909e-1ba7bf17afaa.png)

inferno-chromium:
@naveensrinivasan - made you bigquery admin, go ahead!

naveensrinivasan:
> @naveensrinivasan - made you bigquery admin, go ahead!

Thanks.

oliverchang:
This probably stopped working due to https://github.com/ossf/scorecard/commit/0eaa4ff3d0dd36cd1d200662e5fc9803cc2fd7be. The BQ import expects newline delimited JSON objects, rather than a well formed JSON file.

dlorenc:
This is still failing. Anything I can do to help fix it?

naveensrinivasan:
AFAIK the json dump has to be changed.

azeemshaikh38:
Dan, see https://github.com/ossf/scorecard/issues/336. I'm working on it right now, but if the failures are blocking work in some way, I can submit a hacky quick-fix solution to unblock this. Let me know.

oliverchang:
Unfortunately this is still failing because the Date format in the latest.json is wrong:

`"Date":"21046-04-21"`

I have a fix in #353

oliverchang:
The last two runs have been green.

naveensrinivasan:
The cron still isn't importing correctly for example the data for this isn't populated in this table`SELECT * FROM `openssf.scorecardcron.scorecard_latest` LIMIT 1000`

[...]

PR #303, opened 2021-03-23
:sparkles: Add script that adds dependencies to project.txt from .bzl files

asraa:
Script to update bazel dependencies by grepping for github.com/{}/{} strings in the *.bzl files.
The script will merge metadata if a repository is already in the list.

Feature to update projects.txt programmatically.

Script programmatically gets Envoy deps by parsing a file on their repo.


naveensrinivasan:
@asraa  Thanks, we would want to add only if it is not there in the projects.txt. If not this will fail https://github.com/ossf/scorecard/blob/0e5b8e63f24511613204c28bdb2bbc6b042fb05c/scripts/main.go#L38

Can we do that check prior to adding it to the projects.txt?

asraa:
> Can we do that check prior to adding it to the projects.txt?

Sure, I'm assuming I'll append metadata even if it is there?

BTW -- is it OK if it's not alphabetical?

naveensrinivasan:
> > Can we do that check prior to adding it to the projects.txt?
> 
> Sure, I'm assuming I'll append metadata even if it is there?
Yes
> 
> BTW -- is it OK if it's not alphabetical?

It is OK.

scripts/update/main.go
```diff
@@ -0,0 +1,57 @@
...
+type Repository struct {
+	Repo     string `csv:"repo"`
+	Metadata string `csv:"metadata,omitempty"`
+}
+
+// Programmatically gets Envoy's dependencies and add to projects.
+func EnvoyDeps() string {
+	client := github.NewClient(nil)
+	ctx := context.Background()
+	fo, _, _, err := client.Repositories.GetContents(ctx, "envoyproxy", "envoy", "bazel/repository_locations.bzl", nil)
```
naveensrinivasan, line 21: Can this be parameterized? I am not aware of the Bazel format being consistent across projects? If it then this can be requested for other bazel projects?

asraa: The dictionary with project_url key structure is Envoy-specific.

Bazel projects will pull in deps in specific files, and you could grep through them for things matching the github.com/{}/{} pattern.

We could do that here and remove the dupes we have through the project URL and dep URL.

asraa: e.g. a typical bazel dep file https://github.com/grpc/grpc/blob/e9de13e6ad98327aa6fcb017ebf375f870a3134b/bazel/grpc_deps.bzl#L157

naveensrinivasan: > The dictionary with project_url key structure is Envoy-specific so we can parse it as a python dict and do analysis/create tables with scripts.
> 
> Bazel projects will pull in deps in specific files, and you could grep through them for things matching the github.com/{}/{} pattern.
> 
> We could do that here and remove the dupes we have through the project URL and dep URL.

Sounds good.

asraa: @naveensrinivasan sorry one more question :) 
any other bazel project that might be interesting to add right now (for testing)? 
i was pointed to https://github.com/google/starlark-go which can interpret *.bzl files and using that API would make this parametrizeable on filename by checking for url values.

naveensrinivasan: > @naveensrinivasan sorry one more question :)

Not an issue 👍  

> any other bazel project that might be interesting to add right now (for testing)?

How about the grpc that you had referred? Could we try that?

> i was pointed to https://github.com/google/starlark-go which can interpret *.bzl files and using that API would make this parametrizeable on filename by checking for url values.

asraa: All done! I added a TODO to change over to starlark, but for now I'm searching the .bzl files for the github string. using startlark-go was a little complicated because navigating the modules was different.

BTW. I joined metadata like
`project=grpc,project=envoy,dependecy=true`. Not sure if that was ideal.

naveensrinivasan: > All done! I added a TODO to change over to starlark, but for now I'm searching the .bzl files for the github string. using startlark-go was a little complicated because navigating the modules was different.
> 

Sounds good!

> BTW. I joined metadata like
> `project=grpc,project=envoy,dependecy=true`. Not sure if that was ideal.

This is great! Thanks!

naveensrinivasan:
Can you please change the title to follow this process https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process?

scripts/update/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/ossf/scorecard/scripts/update
```
naveensrinivasan, line 1: `go mod tidy` please

naveensrinivasan (commented):
Thank you! for the contribution.  A few issues in the linting

Also when you run `make lint` it should fetch the `golangci-lint`

From the `scripts/update` if we run this command it comes with all these warnings.

``` shell
 ../../bin/golangci-lint run
main.go:69:6: Error return value of `regexp.Compile` is not checked (errcheck)
	re, _ := regexp.Compile(`github.com/[^\/]*/[^\/"]*`)
	    ^
main.go:149:15: Error return value of `projects.Seek` is not checked (errcheck)
	projects.Seek(0, 0)
	             ^
main.go:152:22: Error return value of `projects.WriteString` is not checked (errcheck)
	projects.WriteString(string(b))
	                    ^
main.go:25:35: builtinShadow: shadowing of predeclared identifier: new (gocritic)
func JoinMetadata(current string, new string) string {
                                  ^
main.go:26:5: emptyStringTest: replace `len(current) == 0` with `current == ""` (gocritic)
	if len(current) == 0 || len(new) == 0 {
	   ^
main.go:45: File is not `gofmt`-ed with `-s` (gofmt)
	for key, _ := range m {
main.go:72: File is not `gofumpt`-ed (gofumpt)
		repos = append(repos, Repository{match,
			strings.Join([]string{"project=", repo.Repo, ",dependency=true"}, "")})

main.go:85: File is not `gofumpt`-ed (gofumpt)
	var bazel_repos = []RepositoryDepsURL{
main.go:135: File is not `gofumpt`-ed (gofumpt)

main.go:30:2: don't use underscores in Go names; var `cur_entries` should be `curEntries` (golint)
	cur_entries := strings.Split(current, ",")
	^
main.go:31:2: don't use underscores in Go names; var `new_entries` should be `newEntries` (golint)
	new_entries := strings.Split(new, ",")
	^
main.go:151:5: ineffectual assignment to `err` (ineffassign)
	b, err := csvutil.Marshal(repos)
	   ^
main.go:74: unnecessary trailing newline (whitespace)

	}
```
Could these be addressed?

naveensrinivasan:
Also when I ran  `go run main.go ../../cron/projects.txt` the command rewrote the entire file, which could be an issue in git diffs.

If I think about it, go maps don't maintain consistency in ordering. Probably to avoid huge differences we would need it to be sorted. I am sorry I said it need not be [sorted](https://yourbasic.org/golang/sort-map-keys-values/). I wasn't thinking of how this could be an issue. 

If this is a lot of work, don't worry we address it next PR, which I can take up. 

Thanks

oliverchang:
Thanks for contributing this change! 

I might be lacking some context but I think we need to rethink how dependencies are handled. It's not very scalable to have to update this txt file every time a dependency changes, especially if many projects other than Envoy start wanting this. Dependency tracking should be decoupled from the actual scorecards project. 

How is Envoy using this today? Is there any reason why this data needs to be recorded here in projects.txt?

asraa:
Will work on those lints. Actually, I could change it so that it only adds deps at the end and updates those inline so that the diff is minimal. But re-thinking the need:

> How is Envoy using this today? Is there any reason why this data needs to be recorded here in projects.txt?

Actually right now the way we use it doesn't assume that a dep is in projects.txt. We don't use the latest.json (yet) and just use the scorecard binary against our deps using a script. The only thing we want is (1) not to be ratelimited by GitHub API (which is probably doable now that most deps are in projects.txt) and (2) have minimal envoy scripts.
https://github.com/envoyproxy/envoy/blob/25d50667f96ee98d7f6cad8e190a3798ea5cd66f/tools/dependency/ossf_scorecard.py#L3

On the other hand I think https://github.com/ossf/scorecard/issues/287 was spun so that you could filter scorecard by a project (e.g. run --metadata=project=envoy to get a scorecard of all of envoy deps). Another scal-able way would instead of adding to proejcts.txt we run a dependency Checker that adds a Dependency field to each JSON line so people can filter on that. I don't know how easy it is to scan dependencies for all types of projects.

naveensrinivasan:
After some thought, I agree this might not be a scalable solution within a text file. 

But having the dependency information tracked will help projects utilize the `latest.json`/BigQuery https://github.com/ossf/scorecard/issues/295. 

One of the goals of Scorecard is to run 10k Projects https://github.com/ossf/scorecard/pull/291#pullrequestreview-616532657 and if we could tag the dependency information in the results which users can run a query will improve the security posture without having to run and maintain Scorecard.

And this PR is providing building blocks of the above. 



#### Histroy
Some history behind adding envoy dependencies https://github.com/ossf/scorecard/pull/84
https://docs.google.com/document/d/1HbREo7pv7rgeIIjQn6mNpySzQE5rx2Yv9dXm5NqR2N8/edit#heading=h.sb9sheknnavz

#### Related 
https://github.com/ossf/scorecard/issues/289

oliverchang:
Thanks for providing the context. Adding the projects themselves for scorecard analysis is fine while our projects.txt list is small.

My concern is more around the metadata and using that to encode dependency trees for every interested project. This approach isn't going to scale. If this unblocks something on Envoy's side, we can keep this temporarily, but this shouldn't be relied on and users of it should not expect this to always be the way to get their own list of dependencies.

@naveensrinivasan @asraa WDYT?

asraa:
>  but this shouldn't be relied on and users of it should not expect this to always be the way to get their own list of dependencies.

In the long term, WDYT about auto-detecting deps? This function can be more and more generalized to autodetect *.bzl files with dependency info and try to auto-tag (for Bazel projects)

naveensrinivasan:
> > but this shouldn't be relied on and users of it should not expect this to always be the way to get their own list of dependencies.
> 
> In the long term, WDYT about auto-detecting deps? This function can be more and more generalized to autodetect *.bzl files with dependency info and try to auto-tag (for Bazel projects)

I like that idea, and we can extend it to other types of dependencies.

asraa:
Awesome, cleaning this commit up, will add explanation and TODO that this is a stepping stone to auto-detect deps.

naveensrinivasan:
> Awesome, cleaning this commit up, will add explanation and TODO that this is a stepping stone to auto-detect deps.

@asraa I am planning to add something similar to go. I was thinking of extending the work you have done. 

When do you think this will likely be ready?

Thanks

asraa:
> @asraa I am planning to add something similar to go. I was thinking of extending the work you have done.

Can do today -- I'm still a little unclear about whether it's necessary to have a job updating projects.txt and whether deps.dev is better to  provide dependency analysis, but I can clean up today to at least close htis out.

naveensrinivasan:
> > @asraa I am planning to add something similar to go. I was thinking of extending the work you have done.
> 
> Can do today -- I'm still a little unclear about whether it's necessary to have a job updating projects.txt and whether deps.dev is better to provide dependency analysis, but I can clean up today to at least close htis out.

I think we need not add metadata to the project.txt as it makes it harder to maintain.

asraa:
Sounds good, I'll remove the metadata string and just append new deps.

naveensrinivasan (approved):
Thanks, @asraa. Can you please squash the commits?

asraa:
All set!

Issue #304, opened 2021-03-24
Feature - Provide Remediation in results 


naveensrinivasan:
**Is your feature request related to a problem? Please describe.**
Provide description and remediation in the results. Related to #63 
This is also required for https://github.com/ossf/scorecard/issues/193

- [ ] Active
- [ ] Branch-Protection
- [ ] CI-Tests,CII-Best-Practices
- [ ] Code-Review,Contributors
- [ ] Frozen-Deps,Fuzzing
- [ ] Packaging 
- [ ] Pull-Requests
- [ ] SAST
- [ ] Security-Policy
- [ ] Signed-Releases
- [ ] Signed-Tags

PR #305, opened 2021-03-24
:sparkles: Documentation for checks active, frozen-deps, SAST and security policy

naveensrinivasan:
N/A

 #304

Does not provide docs

 Included docs in the results for active, deps,sast and security_policy

None

* **Other information**:
None

### DIFF
--- a/checks/active.go
+++ b/checks/active.go
@@ -51,5 +51,8 @@ func IsActive(c checker.Checker) checker.CheckResult {
 	return checker.CheckResult{
 		Pass:       totalCommits >= numCommits,
 		Confidence: confidence,
+		Description: `This check tries to determine if the project is still "actively maintained". 
+		It currently works by looking for commits within the last 90 days.`,
+		HelpURL: "https://github.com/ossf/scorecard/blob/main/checks.md#active",
 	}
 }
--- a/checks/sast.go
+++ b/checks/sast.go
@@ -21,6 +21,12 @@ import (
 
 var sastTools map[string]bool = map[string]bool{"github-code-scanning": true, "sonarcloud": true}
 
+const (
+	//nolint
+	description string = `This check tries to determine if the project uses static code analysis systems. It currently works by looking for well-known results (CodeQL, etc.) in GitHub pull requests.`
+	helpURL     string = "https://github.com/ossf/scorecard/blob/main/checks.md#sast"
+)
+
 func init() {
 	registerCheck("SAST", SAST)
 }
@@ -37,7 +43,10 @@ func SASTToolInCheckRuns(c checker.Checker) checker.CheckResult {
 		State: "closed",
 	})
 	if err != nil {
-		return checker.RetryResult(err)
+		r := checker.RetryResult(err)
+		r.Description = description
+		r.HelpURL = helpURL
+		return r
 	}
 
 	totalMerged := 0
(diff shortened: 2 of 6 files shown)


checks/active.go
```diff
@@ -51,5 +51,8 @@ func IsActive(c checker.Checker) checker.CheckResult {
 	return checker.CheckResult{
 		Pass:       totalCommits >= numCommits,
 		Confidence: confidence,
+		Description: `This check tries to determine if the project is still "actively maintained".
```
inferno-chromium, line 54: i dont like the idea of adding description here, help url should be enough and we shouldnt duplicate the doc text which can get updated with time (and also will be detailed).

checks/active.go
```diff
@@ -51,5 +51,8 @@ func IsActive(c checker.Checker) checker.CheckResult {
 	return checker.CheckResult{
 		Pass:       totalCommits >= numCommits,
 		Confidence: confidence,
+		Description: `This check tries to determine if the project is still "actively maintained". 
+		It currently works by looking for commits within the last 90 days.`,
+		HelpURL: "https://github.com/ossf/scorecard/blob/main/checks.md#active",
```
inferno-chromium, line 56: we should move https://github.com/ossf/scorecard/blob/main/checks.md into a constant in check.go, and this file can just pass the #<marker-name>

checks/sast.go
```diff
@@ -47,12 +56,19 @@ func SASTToolInCheckRuns(c checker.Checker) checker.CheckResult {
 			continue
 		}
 		totalMerged++
-		crs, _, err := c.Client.Checks.ListCheckRunsForRef(c.Ctx, c.Owner, c.Repo, pr.GetHead().GetSHA(), &github.ListCheckRunsOptions{})
+		crs, _, err := c.Client.Checks.
+			ListCheckRunsForRef(c.Ctx, c.Owner, c.Repo, pr.GetHead().GetSHA(), &github.ListCheckRunsOptions{})
 		if err != nil {
-			return checker.RetryResult(err)
+			r := checker.RetryResult(err)
+			r.Description = description
```
inferno-chromium, line 63: helper function, rather than repeating this.

inferno-chromium (changes requested):

checks/active.go
naveensrinivasan: For `sarif` description is required.SARIF is a standard format and is being used at the GitHub Security center. How do we deal with that?

inferno-chromium: Then need to move all strings data in xml file and then fetch description from there. we can have two fields, summary and detailed description. summary shows by default. we shouldn't put these long strings in code.

naveensrinivasan: image

Which is the summary? The above section?

inferno-chromium: Basically i want https://github.com/ossf/scorecard/blob/main/checks.md to be converted into a machine readable format, with summary, description and remediation fields. And then we read them in code and show it (normal view-summary only or detailed view which shows description and remediation as well). i want to reduce deduplication of text in checks.md and also dont want to hardcode strings in source code.

naveensrinivasan: OK. I understand. But the issue is that if we were covert it to machine-readable then how do we provide the same information as markdown when someone visits the GitHub Repo. Ideally, we have a `web` frontend that can serve this data and can be the single source of truth.

inferno-chromium: I am fine to remove checks.md for now. We can later add it, can you create an issue.

naveensrinivasan: OK. I am fine with that. Just an FYI... this is a chore and would take some time.

naveensrinivasan: This is required to get the SARIF results and only after that, we can publish the scorecard as a GitHub action.

naveensrinivasan: We need more contributors.

inferno-chromium: We have a new member starting on Scorecard project, so we should get help soon. Can table this one for now.

naveensrinivasan: Sounds good. Thanks

oliverchang: big +1 to machine readable format for checker descriptions. We need this info for consumers of scorecard data to show in their own UI.

naveensrinivasan:
Closing this PR to implement it based on the latest docs #345

Issue #179, opened 2021-02-14
New check - Use of tools that update deps like dependabot


naveensrinivasan:
Check if the dependabot is turned on the repository. This helps to identify any vulnerabilities on the project's dependencies.

vin01:
Perhaps this should be a general category for automatic dependency updates, also supporting popular tools like Renovate for dependency updates: https://github.com/renovatebot/renovate

naveensrinivasan:
closing it #322

PR #322, opened 2021-04-02
:sparkles: Automatic dependency update checks

naveensrinivasan:

  Feature

   #179

  * Checks if the dependencies are automatically updated.

None

* **Other information**:
None

### DIFF
--- a/checks/automatic_dependency_update.go
+++ b/checks/automatic_dependency_update.go
@@ -1,0 +1,53 @@
+// Copyright 2020 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checks
+
+import (
+	"strings"
+
+	"github.com/ossf/scorecard/checker"
+)
+
+const check = "Automatic-Dependency-Update"
+
+//nolint
+func init() {
+	registerCheck(check, AutomaticDependencyUpdate)
+}
+
+// AutomaticDependencyUpdate will check the repository if it contains Automatic dependency update.
+func AutomaticDependencyUpdate(c *checker.CheckRequest) checker.CheckResult {
+	result := CheckIfFileExists(check, c, fileExists)
+	if !result.Pass {
+		result.Confidence = 3
+	}
+	return result
+}
+
+// fileExists will validate the if frozen dependencies file name exists.
+func fileExists(name string, logf func(s string, f ...interface{})) (bool, error) {
+	switch strings.ToLower(name) {
+	case ".github/dependabot.yml":
+		logf("dependabot config found: %s", name)
+		return true, nil
+		// https://docs.renovatebot.com/configuration-options/
+	case ".github/renovate.json", ".github/renovate.json5", ".renovaterc.json", "renovate.json",
+		"renovate.json5", ".renovaterc":
+		logf("renovate config found: %s", name)
+		return true, nil
+	default:
+		return false, nil
+	}
+}
--- a/checks/checks.md
+++ b/checks/checks.md
@@ -16,6 +16,13 @@ A project which is not active may not be patched, may not have its dependencies 
 **Remediation steps**
 - There is *NO* remediation work needed here. This is just to indicate your project activity and maintenance commitment.
 
+## Automatic-Dependency-Update 
+
+This check tries to determine if a project has dependencies automatically updated. The checks looks for [dependabot](https://dependabot.com/docs/config-file/) or [renovatebot](https://docs.renovatebot.com/configuration-options/).This check only looks if  it is enabled and does not ensure that it is run and PullRequests are merged. 
+
+**Remediation steps**
+- Signup for automatic dependency updates with dependabot or renovatebot and place the config file in the locations that are recommended by these tools.
+
 ## Branch-Protection 
 
 Branch protection allows defining rules to enforce certain workflows for branches, such as requiring a review or passing certain status checks. This check would work only when the token has [Admin access](https://github.community/t/enable-branch-protection-get-api-without-admin/14197) to the repository. This check determines if the default branch is protected. More specifically, the checks for AllowForcePushes (disabled), AllowDeletions (disabled),  EnforceAdmins (enabled), RequireLinearHistory (enabled), RequiredStatusChecks (enabled and must have non-empty context enabled), RequiredPullRequestReviews (>=1), DismissStaleReviews (enabled), RequireCodeOwnerReviews (enabled). 
--- a/checks/checks.yaml
+++ b/checks/checks.yaml
@@ -54,6 +54,16 @@ checks:
         The file should contain information on what constitutes a vulnerability
         and a way to report it securely (e.g. issue tracker with private issue
         support, encrypted email with a published public key).
+  Automatic-Dependency-Update:
+    description: >-
+      This check tries to determine if a project has dependencies automatically updated.
+      The checks looks for [dependabot](https://dependabot.com/docs/config-file/) or
+      [renovatebot](https://docs.renovatebot.com/configuration-options/). This check only looks if 
+      it is enabled and does not ensure that it is run and pull requests are merged.
+    remediation:
+      - >-
+        Signup for automatic dependency updates with dependabot or renovatebot and place the config
+        file in the locations that are recommended by these tools.
   Contributors:
     description: >-
       This check tries to determine if a project has a set of contributors from
--- a/e2e/automatic_deps_test.go
+++ b/e2e/automatic_deps_test.go
@@ -1,0 +1,58 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/ossf/scorecard/checker"
+	"github.com/ossf/scorecard/checks"
+)
+
+var _ = Describe("E2E TEST:Automatic-Dependency-Update", func() {
+	Context("E2E TEST:Validating dependencies are automatically updated", func() {
+		It("Should return deps are automatically updated for dependabot", func() {
+			l := log{}
+			checker := checker.CheckRequest{
+				Ctx:         context.Background(),
+				Client:      ghClient,
+				Owner:       "ossf",
+				Repo:        "scorecard",
+				GraphClient: graphClient,
+				Logf:        l.Logf,
+			}
+			result := checks.AutomaticDependencyUpdate(&checker)
+			Expect(result.Error).Should(BeNil())
+			Expect(result.Pass).Should(BeTrue())
+		})
+		It("Should return deps are automatically updated for renovatebot", func() {
+			l := log{}
+			checker := checker.CheckRequest{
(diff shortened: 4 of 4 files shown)


checks.md
```diff
@@ -161,3 +161,12 @@ This check tries to determine if the project has branch protection enabled.
...
 - For GitHub, check out the steps
   [here](https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule).
+
+
+## Automatic-Dependency-Update
+
+This check tries to determine if the project has dependencies automatically updated.
+
+**Remediation steps**:
+- Enable dependabot for the project and have a dependabot.yml file in the repository.
```
inferno-chromium, line 171: s/and have a dependabot.yml file in the repository/by placing a .github/dependabot.yml in the root of your repository. See https://docs.github.com/en/code-security/supply-chain-security/enabling-and-disabling-version-updates for more information

checks/automatic_dependency_update.go
```diff
@@ -0,0 +1,37 @@
...
+
+//nolint
+// AutomaticDependencyUpdate will check the repository if it contains Automatic dependecy update.
+func AutomaticDependencyUpdate(c checker.Checker) checker.CheckResult {
+	return CheckIfFileExists(c, fileExists)
+}
+
+// fileExists will validate the if frozen dependecies file name exists.
+func fileExists(name string, logf func(s string, f ...interface{})) bool {
+	return strings.Contains(name, strings.ToLower("dependabot.yml"))
```
inferno-chromium, line 36: this should be exact path, in <root>/.github/dependabot.yml

checks/automatic_dependency_update.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/ossf/scorecard/checker"
+)
+
+//nolint
+func init() {
+	registerCheck("Automatic-Dependency-Update", AutomaticDependencyUpdate)
+}
+
+//nolint
+// AutomaticDependencyUpdate will check the repository if it contains Automatic dependecy update.
```
inferno-chromium, line 29: typo dependency

checks.md
```diff
@@ -161,3 +161,12 @@ This check tries to determine if the project has branch protection enabled.
...
   [here](https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule).
+
+
+## Automatic-Dependency-Update
+
+This check tries to determine if the project has dependencies automatically updated.
+
+**Remediation steps**:
+- Enable dependabot for the project and have a dependabot.yml file in the repository. 
+
```
inferno-chromium, line 172: This is an amazing check, but there are two problems:
1. Dependabot can be enabled on org level
2. Dependabot can be enabled via UI
and this is substantial portion of things, and we would show incorrect result for these. maybe we should bring some of these problems with github api in next openssf sync, e.g. branch protection, this one with knowing security settings like dependabot, etc. Landing this as-is will be problematic since we cannot get right status in many cases and scorecard needs to be accurate. thoughts?

naveensrinivasan: > This is an amazing check, but there are two problems:
> 
> 1. Dependabot can be enabled on org level
> 2. Dependabot can be enabled via UI
>    and this is substantial portion of things, and we would show incorrect result for these. maybe we should bring some of these problems with github api in next openssf sync, e.g. branch protection, this one with knowing security settings like dependabot, etc. Landing this as-is will be problematic since we cannot get right status in many cases and scorecard needs to be accurate. thoughts?

I agree this might be an issue. We can have in-depth discussion. 

Also

- But why not provide an results with lesser confidence, we can never be 100% in some of the cases.

- Another way to solve this issues is to look for commits and check for `Author: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>` and check for signed commits

inferno-chromium: yes can be then

- look for that file .github/dependabot.yml, if exists, pass with high confidence.
- if failed, check for commits [we have to see how far we can go without burning quota, or instead see if we can get user dependabot events by api]. if that exists, pass with medium-high confidence, depending on last update?
- otherwise fail with low confidence (like 3?)

thoughts, can also brainstorm in next meeting with @azeemshaikh38 . but even this is a good start i think.

naveensrinivasan: Sounds good.

asraa: Some projects may have dependabot for tooling (let's say a primary C++ project with no dependabot, but has python deps that utilize dependabot). 

I wonder if we should lower confidence for projects with unsupported languages

laurentsimon: re: dependabot commits. If people have dependabot installed, we might still want to verify that they accept dependabot's PRs once in a while.

inferno-chromium: yes right, this might be a better or definitely needed part here, which is did any dependendabot cls land in a while.

naveensrinivasan:
Going to start working on this.

laurentsimon:
re: @asraa I agree confidence should be lowered for other languages in the repo that are not defined in the dependabot config file. We could list files and their extension: if any of these files are not tracked by the dependabot config, they should reduce the score. It'd be great to weight the scoring based on how much code is tracked or not - maybe by looking at number of lines for each lanaguages.

There's the [language API](https://docs.github.com/en/rest/reference/repos#list-repository-languages) that I initially thought could help, but it unfortunately only works at the repo level, ie we cannot specify folders of interest.

laurentsimon:
@nathannaveen what's left for this PR to land?

naveensrinivasan:
> re: @asraa I agree confidence should be lowered for other languages in the repo that are not defined in the dependabot config file. We could list files and their extension: if any of these files are not tracked by the dependabot config, they should reduce the score. It'd be great to weight the scoring based on how much code is tracked or not - maybe by looking at number of lines for each lanaguages.
> 
> There's the [language API](https://docs.github.com/en/rest/reference/repos#list-repository-languages) that I initially thought could help, but it unfortunately only works at the repo level, ie we cannot specify folders of interest.

I think for the initial release I am not going to target that

naveensrinivasan:
> @nathannaveen what's left for this PR to land?

> 1. if failed, check for commits [we have to see how far we can go without burning quota, or instead see if we can get user dependabot events by api]. if that exists, pass with medium-high confidence, depending on last update?

☝️ This will burn the API quota's

> 2. otherwise fail with low confidence (like 3?) 

For now, we can stay with fail with low confidence. 

Thoughts?

inferno-chromium:
> > @nathannaveen what's left for this PR to land?
> 
> > 1. if failed, check for commits [we have to see how far we can go without burning quota, or instead see if we can get user dependabot events by api]. if that exists, pass with medium-high confidence, depending on last update?
> 
> ☝️ This will burn the API quota's . AFAIK there aren't dependabot events by API.
> 
> > 1. otherwise fail with low confidence (like 3?)
> 
> For now, we can stay with fail with low confidence.
> 
> Thoughts?

Ok this is fine, but please add renovatebot file check in this as well. renovatebot is extremely common, and either one of the two is fine. in documentation, we should say this check only looks if it is enabled and not ensures that it is run and prs merge (maybe tbd for future ?) ?? @azeemshaikh38 @laurentsimon thoughts?

laurentsimon:
> > @nathannaveen what's left for this PR to land?
> 
> > 1. if failed, check for commits [we have to see how far we can go without burning quota, or instead see if we can get user dependabot events by api]. if that exists, pass with medium-high confidence, depending on last update?
> 
> ☝️ This will burn the API quota's . AFAIK there aren't dependabot events by API.
> 
> > 1. otherwise fail with low confidence (like 3?)
> 
> For now, we can stay with fail with low confidence.
> 
> Thoughts?



> > @nathannaveen what's left for this PR to land?
> 
> > 1. if failed, check for commits [we have to see how far we can go without burning quota, or instead see if we can get user dependabot events by api]. if that exists, pass with medium-high confidence, depending on last update?
> 
> ☝️ This will burn the API quota's . AFAIK there aren't dependabot events by API.

Please add a TODO in the code that we should also list commits even when we find the dependabot's config file (to ensure users actually update their dependencies).

re: quota: An alternative approach could be to run the cron jobs with some checks turned off. Within a few days we could run different checks so that, every week, we have all the checks. Do we actually need the cron job to run every single day? (It's ideal, but is it really necessary?)

> 
> > 1. otherwise fail with low confidence (like 3?)
> 
> For now, we can stay with fail with low confidence.
> 
> Thoughts?

I suppose that's the best we can do now. Note that "fail with low confidence" achieves little from a user's point of view. If I use scorecard and I get this result, it's almost like a noop: I have no idea how to interpret it. @asraa who tests scorecard for envoy, so she may comment.

naveensrinivasan:
> > > @nathannaveen what's left for this PR to land?
> > 
> > 
> > > 1. if failed, check for commits [we have to see how far we can go without burning quota, or instead see if we can get user dependabot events by api]. if that exists, pass with medium-high confidence, depending on last update?
> > 
> > 
> > ☝️ This will burn the API quota's . AFAIK there aren't dependabot events by API.
> > > 1. otherwise fail with low confidence (like 3?)
> > 
> > 
> > For now, we can stay with fail with low confidence.
> > Thoughts?
> 
> > > @nathannaveen what's left for this PR to land?
> > 
> > 
> > > 1. if failed, check for commits [we have to see how far we can go without burning quota, or instead see if we can get user dependabot events by api]. if that exists, pass with medium-high confidence, depending on last update?
> > 
> > 
> > ☝️ This will burn the API quota's . AFAIK there aren't dependabot events by API.
> 
> Please add a TODO in the code that we should also list commits even when we find the dependabot's config file (to ensure users actually update their dependencies).
> 
> re: quota: An alternative approach could be to run the cron jobs with some checks turned off. Within a few days we could run different checks so that, every week, we have all the checks. Do we actually need the cron job to run every single day? (It's ideal, but is it really necessary?)

☝️  This is would be a separate issue  

> 
> > > 1. otherwise fail with low confidence (like 3?)
> > 
> > 
> > For now, we can stay with fail with low confidence.
> > Thoughts?
> 
> I suppose that's the best we can do now. Note that "fail with low confidence" achieves little from a user's point of view. If I use scorecard and I get this result, it's almost like a noop: I have no idea how to interpret it. @asraa who tests scorecard for envoy, so she may comment.

checks/automatic_dependency_update.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// fileExists will validate the if frozen dependencies file name exists.
+func fileExists(name string, logf func(s string, f ...interface{})) (bool, error) {
+	switch strings.ToLower(name) {
+	case ".github/dependabot.yml":
+		logf("dependabot config found: %s", name)
+		return true, nil
+		// https://docs.renovatebot.com/configuration-options/
+	case ".github/renovate.json", ".github/renovate.json5", ".renovaterc.json", "renovate.json", "renovate.json5":
```
inferno-chromium, line 46: can you add ".renovaterc" as well
also please add a TODO for "package.json (within a "renovate" section)"

checks/automatic_dependency_update.go
```diff
@@ -0,0 +1,52 @@
...
+const check = "Automatic-Dependency-Update"
+
+//nolint
+func init() {
+	registerCheck(check, AutomaticDependencyUpdate)
+}
+
+// AutomaticDependencyUpdate will check the repository if it contains Automatic dependency update.
+func AutomaticDependencyUpdate(c *checker.CheckRequest) checker.CheckResult {
+	result := CheckIfFileExists(check, c, fileExists)
```
inferno-chromium, line 32: We need to later improve CheckIfFileExists so that it caches content and does not download again if it already exists. We have too many checks now that depend on files. @azeemshaikh38 fyi

checks/checks.md
```diff
@@ -16,6 +16,13 @@ A project which is not active may not be patched, may not have its dependencies
 **Remediation steps**
 - There is *NO* remediation work needed here. This is just to indicate your project activity and maintenance commitment.
 
+## Automatic-Dependency-Update 
+
+This check tries to determine if a project has dependencies automatically updated. The checks looks for [dependabot](https://dependabot.com/docs/config-file/) or [renovatebot](https://docs.renovatebot.com/configuration-options/).This check only looks if  it is enabled and does not ensure that it is run and PullRequests are merged.
```
inferno-chromium, line 21: nite spaces messed up in ".This check only looks if  it"
should be ". This check only looks if it"

inferno-chromium: s/PullRequests/pull requests
no need to capitalize

inferno-chromium (approved):
Thanks a lot!

checks/automatic_dependency_update.go
naveensrinivasan: https://github.com/ossf/scorecard/issues/542

naveensrinivasan: Took care of the other item. Thanks

checks/checks.md
naveensrinivasan: Thanks, took care of it.

checks/automatic_dependency_update.go
naveensrinivasan: https://github.com/ossf/scorecard/issues/545

azeemshaikh38: Commented on the issue. This is definitely an improvement we should work on, but a low priority IMO.

PR #330, opened 2021-04-07
:seedling: Included dependencies for k8s

naveensrinivasan:

  * Included the k8s dependencies.

None

* **Other information**:

None


cron/projects.txt
```diff
@@ -2329,3 +2329,73 @@ github.com/yaml/pyyaml,
 github.com/twisted/incremental,
 github.com/zopefoundation/zope.interface,
 github.com/twisted/constantly,
+github.com/alibaba/hessian2-codec.git,
+github.com/alibaba/hessian2-codec,
```
inferno-chromium, line 2333: this and previous line seems same, maybe we should strip .git from all urls ?
also, can we keep this file sorted with unique lines, it is easy to maintain a sorted list?

naveensrinivasan: The hessian-codec was from the changes for `bazel deps`. I will take care of that. I will also sort it

naveensrinivasan: @inferno-chromium  I took care of the `.git` one. 

The sorting requires a lot more changes to the  `./scripts/update` codebase. That will be in subsequent PR's

inferno-chromium (approved):

PR #337, opened 2021-04-08
✨  Included dependency parsing for go 

naveensrinivasan:

 feature

  * Included dependency parsing of go.mod files.
  * Parse vanity URL in go.mod to add dependencies
  * Updated k8s dependencies based on the vanity URL's.

None

* **Other information**:
None


scripts/update/main.go
```diff
@@ -61,12 +74,94 @@ func GetBazelDeps(repo RepositoryDepsURL) []Repository {
...
+	mods := &gomod{}
+	if err := json.Unmarshal(out.Bytes(), mods); err != nil {
+		log.Panic(err)
+	}
+	// Close the file
+	if err := tmpFile.Close(); err != nil {
+		log.Panic(err)
+	}
+
+	for _, match := range mods.Require {
```
dlorenc, line 135: I think this is actually a common misconception with go.mod files - they do not contain the full set of dependencies required to build an application. For that, you need to use `go list -m all`.

naveensrinivasan: I am guessing `go list -m all` gives transitive dependency. Do we need the transitive dependencies?

dlorenc: I don't think it's that simple but can't explain it perfectly myself. My understanding is that go.mod only contains dependencies that are not the "natural" versions that would be selected by MVS.

naveensrinivasan: So you think we should change the implementation to use the go list -m all instead of go.mod?

dlorenc: Yeah, I understand "go list -m all", but I don't understand go.mod :)

It's possible that go.mod is all we need if we only care about direct dependencies, but from a security standpoint transitive dependencies are just as important.

naveensrinivasan: I understand. I am going to try and see if we can go list -m all with just go.mod and go.sum.

naveensrinivasan: OK, my doubt was correct, can't get the `go list -m all` with just `go.sum` and `go.mod`. We would have to download the entire repository.

naveensrinivasan: I am going to change the implementation to download the repository and then run `go list -m all`

naveensrinivasan: Took care of it. Thanks

dlorenc (approved):

Issue #336, opened 2021-04-08
Feature - refactor cron formatting


azeemshaikh38:
**Is your feature request related to a problem? Please describe.**
The cron job appends the output of  scorecard commandline to an output. This is a problem since, the cron job deals with string output from the scorecard command and tries to get the formatting right. 

**Describe the solution you'd like**
Instead, the cron job should be dealing with in-memory data structure of check results which it can use in any way it likes, including formatting it using pre-existing formatting libraries (like JSON). To do this, we need to refactor such that we have a common library of functions/types which will be used by both the commandline tool as well as the cron job.

@inferno-chromium - this is a fair amount of refactor work, but IMO worth the time/effort. Let me know if this LG to you and I can get started on it.

naveensrinivasan:
> Instead, the cron job should be dealing with in-memory data structure of check results which it can use in any way it likes, including formatting it using pre-existing formatting libraries (like JSON). To do this, we need to refactor such that we have a common library of functions/types which will be used by both the commandline tool as well as the cron job.

A good idea IMO. But with the in-memory you could run into memory bottlenecks as we scale to 10,000 -100,000 repository. 

Will the solution be an issue for such scales?

azeemshaikh38:
Depends on how we handle the results. We could keep writing to file/DB every time a result comes back or we can batch it to reduce IO. But yes, we mostly will not hold all results in memory at the same time.

inferno-chromium:
SGTM.

oliverchang:
Big +1 for this :) This would clean up a lot of how the current outputting is done.

PR #338, opened 2021-04-09
🌱 : Reduce code duplication for follow-up cron refactoring

azeemshaikh38:

Refactors code to avoid code duplication. Follow up CL will replace cron job to use in-memory data structure.
Context: https://github.com/ossf/scorecard/issues/336

No behaviour change.

No

* **Other information**:


lib/check_request.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2020 Security Scorecard Authors
+//
```
naveensrinivasan, line 2: file names with `_` are primarily only for `test` files and for assign platform/arch-only code,
```
➜ cd $GOROOT/src/pkg/math/
➜ ls sqrt*s
sqrt_386.s  sqrt_amd64p32.s  sqrt_amd64.s  sqrt_arm.s
```

lib/check_result.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2020 Security Scorecard Authors
```
naveensrinivasan, line 1: It is a standard approach to keep all the related structs in a single file instead of in separate files

https://golang.org/src/fmt/format.go

>All the files are based on features, and it is best to use a proximity principle approach, where you can find in the same file the definition of what you are using.
Generally, those features are grouped in one file per package, except for large ones, where one package is composed of many files (net, net/http)

https://stackoverflow.com/a/26229922/19407

pkg/scorecard.go
```diff
@@ -80,11 +74,11 @@ func (r *RepoURL) Set(s string) error {
 }
 
 func RunScorecards(ctx context.Context, logger *zap.SugaredLogger,
-	repo RepoURL, checksToRun []checker.NamedCheck) <-chan Result {
-	resultsCh := make(chan Result)
+	repo RepoURL, checksToRun lib.CheckNameToFnMap) <-chan lib.CheckResult {
```
naveensrinivasan, line 77: Is there a specific reason to change slice into a custom type? `lib.CheckNameToFnMap`  `[]checker.NamedCheck`

go is about keeping it simple cognitive load. This would mean someone fixing/reading the code would have to go check type definition.

checks/security_policy.go
```diff
@@ -17,16 +17,18 @@ package checks
 import (
 	"strings"
 
-	"github.com/ossf/scorecard/checker"
+	"github.com/ossf/scorecard/lib"
 )
 
+const securityPolicyStr = "Security-Policy"
```
naveensrinivasan, line 23: In `go` `securityPolicyStr` suffix aren't recommended in the names.

checks/sast.go
```diff
@@ -16,28 +16,30 @@ package checks
 
 import (
 	"github.com/google/go-github/v32/github"
-	"github.com/ossf/scorecard/checker"
+	"github.com/ossf/scorecard/lib"
 )
 
+const sastStr = "SAST"
```
naveensrinivasan, line 22: In go `sastStr` suffix aren't standard. 

The same comment is applicable across other changes

lib/check_result_utils.go
```diff
@@ -12,43 +12,61 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package checker
+package lib
```
naveensrinivasan, line 15: `utils` are recommended in go

https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

pkg/scorecard.go
```diff
@@ -24,18 +24,12 @@ import (
...
 
 	"github.com/google/go-github/v32/github"
-	"github.com/ossf/scorecard/checker"
+	"github.com/ossf/scorecard/lib"
 	"github.com/ossf/scorecard/roundtripper"
 	"github.com/shurcooL/githubv4"
 	"go.uber.org/zap"
 )
 
-type Result struct {
```
naveensrinivasan, line 33: Noticed this was changed from `Result` to `CheckResult`.

The `Result` implies what it is. 

What was the idea of having the `Check` prefix to the `Result`?

naveensrinivasan (changes requested):
Thanks!

It is a large PR, I am going to have to review it once more. I wanted to provide initial feedback.

checks/security_policy.go
inferno-chromium: Without suffix, there is less clarity. I think this confusion comes from check names like SecurityPolicy. So, i think keeping this as-is is fine for clarity, also we should think of renaming check function from like SecurityPolicy -> SecurityPolicyCheck to be more clear [https://github.com/ossf/scorecard/blob/main/checks/security_policy.go#L27]

checks/sast.go
inferno-chromium: See comment above.

lib/check_request.go
inferno-chromium: i dont think that is correct, see e.g. 
https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/dockershim
docker_logs.go
docker_sandbox.go
docker_sandbox_test.go

_test.go is a suffix for tests, but there is no rule against using _ for file names.

naveensrinivasan: Ok. But I am basing my comments based standard library. 

My recommendation was to avoid that as much as possible.

checks/contributors.go
```diff
@@ -66,15 +68,16 @@ func Contributors(c checker.Checker) checker.CheckResult {
 	}
 	c.Logf("companies found: %v", strings.Join(names, ","))
 	const numContributors = 2
```
inferno-chromium, line 70: move this to global too.

checks/signed_releases.go
```diff
@@ -18,19 +18,22 @@ import (
...
 
 	"github.com/google/go-github/v32/github"
-	"github.com/ossf/scorecard/checker"
+	"github.com/ossf/scorecard/lib"
 )
 
-var releaseLookBack int = 5
+const (
+	signedReleasesStr = "Signed-Releases"
+	releaseLookBack   = 5
```
inferno-chromium, line 26: Since you are refactoring things, how about doing s/releaseLookBack/releaseLoockBackDays. some of these vars will be more clear with what they represent.

cmd/root.go
```diff
@@ -195,16 +184,15 @@ func outputJSON(results []pkg.Result) {
 	}
 
 	for _, r := range results {
-		var details []string
+		tmp_result := lib.CheckResult{
```
inferno-chromium, line 187: tmpResult for var names in go (camelCase)

e2e/contributors_test.go
```diff
@@ -36,13 +36,13 @@ var _ = Describe("E2E TEST:CodeReview", func() {
...
 			}
-			result := checks.Contributors(checker)
+			result := checks.Contributors(checkRequest)
 			Expect(result.Error).Should(BeNil())
 			Expect(result.Pass).Should(BeTrue())
 		})
 		It("Should return valid project contributors", func() {
 			l := log{}
-			checker := checker.Checker{
+			lib := lib.CheckRequest{
```
inferno-chromium, line 45: s/lib/checkRequest for var name

lib/check_result.go
inferno-chromium: check_request is fine to be seperate, but i think check_result_utils can just go in this file.

lib/check_runner.go
```diff
@@ -12,27 +12,23 @@
...
 )
 
-type Checker struct {
-	Ctx         context.Context
-	Client      *github.Client
-	GraphClient *githubv4.Client
-	HttpClient  *http.Client
-	Owner, Repo string
-	Logf        func(s string, f ...interface{})
+const numRetries = 3
```
inferno-chromium, line 22: s/numRetries/checkRetries ?

inferno-chromium (approved):
LGTM with lib directory change needed, some other minor comments.

Mostly looks good, left few comments. I don't like the use of lib directory since most of the code in there is checker or check related, so using something like that is a more meaningful package name/directory as per go standards - https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

lib/check_request.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2020 Security Scorecard Authors
```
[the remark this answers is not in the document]
inferno-chromium: pkg is usually for the root package, not for helper stuff. in this case, i dont like the use of lib either as a directory name, what was wrong with using checker itself ? in go, they don't recommend using generic names like utils, base, etc

lib/check_request.go
inferno-chromium: yes i dont disagree, avoid when possible, but fine to use it otherwise.

lib/check_request.go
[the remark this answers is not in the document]
azeemshaikh38: My plan was to have a single package with util fns (I plan to add more in upcoming PRs). But if this is not usually done in  Golang, I'm ok sticking with checker/ itself. Reverted it back to that. PTAL.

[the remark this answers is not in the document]
inferno-chromium: yes more of a golang thingy - https://blog.golang.org/package-names

Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.

Break up generic packages. To fix such packages, look for types and functions with common name elements and pull them into their own package.

lib/check_runner.go
azeemshaikh38: Done.

e2e/contributors_test.go
azeemshaikh38: Done.

cmd/root.go
azeemshaikh38: Done.

checks/signed_releases.go
azeemshaikh38: Done.

checks/contributors.go
azeemshaikh38: Done. Also renamed them, PTAL.

pkg/scorecard.go
azeemshaikh38: The main reason is that we can have a single type which can be used across packages. So I have tried to delete as much new type definitions as possible across the code and have it defined in lib/ (now checker/).

Also, Result was simply CheckResult with check name (metadata, which is likely for the Repo, is never used). Now that checkName is part of CheckResult struct, I decided to remove it.

lib/check_result_utils.go
azeemshaikh38: There was another comment from @inferno-chromium mentioning generic names are to be avoided in Go. So going back to checker/ for now.

checks/sast.go
azeemshaikh38: Ack.

checks/security_policy.go
azeemshaikh38: Ack. Leaving the suffix as is for now.

pkg/scorecard.go
azeemshaikh38: The functionality we want is a map from string -> CheckFn. I don't agree that for the sake of readability we should choose a different data structure which increases code complexity (e.g we loop through all check names to find the fn instead of lookup).

lib/check_result.go
azeemshaikh38: Done.

lib/check_request.go
azeemshaikh38: Will continue to keep it as _ filenames for now, since they help with readability. Can change it in future if needed.

lib/check_request.go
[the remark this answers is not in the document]
azeemshaikh38: Makes sense.

inferno-chromium (approved):

naveensrinivasan (approved):

checks/security_policy.go
oliverchang: How about renaming this to "checkName" everywhere? That communicates the purpose of this variable a lot better IMO, rather than just a variable name which mirrors the actual value.

PR #339, opened 2021-04-10
:bug: Handle vendored repos dependency

naveensrinivasan:

Bug

The dependency parsing does not handle vendored go repositories. It was failing to handle vendored repositories.

  * Handle vendored repos for go dependency

  * Add additional repositories for projects.txt

NA

* **Other information**:
NA


scripts/update/main.go
```diff
@@ -129,14 +136,18 @@ func GetGoDeps(repo RepositoryDepsURL) []Repository {
 		//nolint
 		if strings.HasPrefix(dependency, "github.com") {
 			repourl := RepoURL{}
-			if err := repourl.Set(dependency); err == nil {
+			url := strings.Split(dependency, "/")
```
inferno-chromium, line 139: can you create a helper function for this (with example comment on what this is doing). and use in both places.

inferno-chromium (approved):

azeemshaikh38 (approved):

scripts/update/main.go
azeemshaikh38: +1

naveensrinivasan: Agreed, I will do that. Thanks

PR #343, opened 2021-04-14
:seedling: Fix concurrent cronjob execution

naveensrinivasan:

 Bug

  Multiple cronjobs werer running and slowing down because of use of the same GitHUb Tokens

  * With the increased scans the cronjob is running longer than expected  which was causing the multiple jobs to be executing concurrently.

  * Changed the concurrent policy to "Forbid" to avoid it.

None

* **Other information**:

None

### DIFF
--- a/k8s/cron.yaml
+++ b/k8s/cron.yaml
@@ -18,6 +18,7 @@ metadata:
   name: daily-score
 spec:
   schedule: "0 2 * * *"
+  concurrencyPolicy: "Forbid"
   jobTemplate:
     spec:
       template:


inferno-chromium (approved):

azeemshaikh38:
Doesn't the concurrency come from here: https://github.com/ossf/scorecard/blob/main/pkg/scorecard.go#L83? I wonder if changing Kubernetes concurrency really helps.

naveensrinivasan:
> Doesn't the concurrency come from here: https://github.com/ossf/scorecard/blob/main/pkg/scorecard.go#L83? I wonder if changing Kubernetes concurrency really helps.

This is for the whole batch job. One batch job is running it will not kick in the next batch job unless the first one is complete.

For example

There won't be 2 cron jobs pods running. Prior to this, there were 2 cron jobs that were running, because the k8s scheduler scheduled these jobs even though the first 1 wasn't complete.

HTH

PR #345, opened 2021-04-15
:sparkles: Make checks documentation machine readable.

### LINKED ISSUE #311: Provide canonical descriptions for each checker
Currently in the latest.json, we provide details of each check but there is no canonical description string for each checker.

A canonical description can be useful for consumers to show in their own UI or other tooling.

Perhaps we can add this as a top level "checkers" field to the existing JSON which maps a Check name to a description?

oliverchang:
Make checks.yaml a machine and human readable source of truth for
checks documentation.

A tiny Python script is also added to generate a checks.json
(more universally machine readable) and checks.md (for humans)
from this file.

The resulting checks.md has no visual differences (apart from fixing a small typo).

Fixes #311


docs/checks.json
```diff
@@ -0,0 +1 @@
...ect with OSS-Fuzz by following the instructions [here](https://google.github.io/oss-fuzz/)."]}, "SAST": {"description": "This check tries to determine if the project uses static code analysis systems. It currently works by looking for well-known results ([CodeQL](https://securitylab.github.com/tools/codeql), etc.) in GitHub pull requests.", "remediation": ["Run CodeQL checks in your CI/CD by following the instructions [here](https://github.com/github/codeql-action#usage)."]}, "Active": {"description": "This check tries to determine if the project is still \"actively maintained\". It currently works by looking for commits within the last 90 days.", "remediation": ["There is *NO* remediation work needed here. This is just to indicate your project activity and maintenance commitment."]}, "Branch-Protection": {"description": "This check tries to determine if the project has branch protection enabled.", "remediation": ["Enable branch protection settings in your source hosting provider to avoid force pushes or deletion of your important branches.", "For GitHub, check out the steps [here](https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule)."]}}}
```
naveensrinivasan, line 1: What is the difference between checks.json and checks.yaml?

naveensrinivasan (commented):
Thanks! This is clean.

Why not use `go` instead of python? Because we have linters, codeql everything already set up Also to contribute to this project now we are adding another dependency.

docs/generate_checks.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+import json
+
+import yaml
+
+with open('checks.yaml') as f:
+    checks = yaml.safe_load(f)
+
+with open('checks.json', 'w') as f:
```
inferno-chromium, line 22: is this even needed. ideal scenario would be scorecard tool, read from checks.yaml and add json as part of the latest.json ? if we use checks.json, it can be stale.

inferno-chromium (approved):
LGTM with one comment on json stuff. @azeemshaikh38 should look at this too.

docs/checks.yaml
```diff
@@ -0,0 +1,168 @@
+# Copyright 2021 Security Scorecard Authors
```
azeemshaikh38, line 1: Nit: could we move checks.yaml and the script inside checks/ folder?

docs/generate_checks.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2021 Security Scorecard Authors
```
azeemshaikh38, line 1: Nit: maybe change name to generate_checks_readme/generate_checks_md or something along those lines to clarify what this script does.

docs/generate_checks.py
```diff
@@ -0,0 +1,47 @@
...
+#
```
azeemshaikh38, line 2: Same question as another comment on the PR - any reason to use Python instead of Go? No need to update the language, just curious if there was any reasoning behind the choice.

docs/generate_checks.py
azeemshaikh38: +1. Since we are not using checks.json in this CL, I think we should just not generate it.

azeemshaikh38 (approved):

naveensrinivasan:
After thinking about it, I don't see a reason why we should introduce a new language into the codebase this adds additional maintenance and makes the contributions a little harder. My vote is NOT to introduce python unless python brings the unique feature that `go` does not have.  

This is a fairly simple code that can be written in `go`

oliverchang (commented):
Re choice of Python, my reasoning is that it's a tiny developer script and not part of the core codebase. It's expected to be minimal and not changed frequently (if at all).

The equivalent Go program will be much longer and did not seem like the right tool the for task. 

If others have strong opinions too I can rewrite this.

inferno-chromium:
overall, i like the change, and now understand the need for json file (since latest.json will be line delimited and has data), so automated systems like https://metrics.openssf.org/grafana/d/default/metric-dashboard?orgId=1 can use checks.json to display checks info.
this python vs go thing i would let @azeemshaikh38 @naveensrinivasan to decide. i am fine to merge now and then change to go, should be a simple change, but this moving to checks.yaml is much more important part of change, so that others like @laurentsimon, etc can make doc changes easily.

naveensrinivasan:
> i am fine to merge now and then change to go, should be a simple change

 I am also OK with merging now and then change to go.

azeemshaikh38:
+1 - no need to change language in this PR.

Issue #193, opened 2021-02-17
Integration with CI/CD and GitHub Code Scanning Results


inferno-chromium:
https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/uploading-a-sarif-file-to-github

naveensrinivasan:
That is a cool idea. Probably use Scorecard all the dependencies of the codebase and upload it.

inferno-chromium:
Can even just start with showing it for the repo itself. it wont block the CI, but just run on it. this github token limit might be pain.

naveensrinivasan:
For this, we need to export the results from the scorecard to `sarif` format and after that, it coming up with a `Github action`, I think.

azeemshaikh38:
@laurentsimon I think we can close this now?

laurentsimon:
Yes!

Issue #63, opened 2020-11-12
RFE: Tool / Check should provide suggested remediation


JasonKeirstead:
Currently, a check indicates a simple pass/fail with confidence. For a developer, this does not actually help them know how they should improve the situation. Checks should provide a suggested remediation on how to fix or improve.

inferno-chromium:
You want to propose some recommendation in the doc, maybe https://github.com/ossf/scorecard/blob/main/checks.md ?

JasonKeirstead:
@inferno-chromium I would argue it should output in the tool itself. If there is consensus on that I will try to work on a small PR to get it started.

inferno-chromium:
Yes i meant show the link in tool and improve documentation.

dlorenc:
Yes, makes sense to me. I think having the remediation content itself live on some website with a short summary and links in the tool makes sense.

Some of the guidance could be multiple paragraphs or longer, which could be hard to read in a terminal.

naveensrinivasan:
closing this to track it via #304

PR #347, opened 2021-04-17
:sparkles: Included documentation for Active check

naveensrinivasan:

  Feature

The check results didn't provide a description and remediation.

   - Included the documentation for Active check
   - The building block for SARIF https://github.com/ossf/scorecard/issues/193

 None

* **Other information**:
 None

### DIFF
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -23,6 +23,8 @@ type CheckResult struct {
 	Details     []string
 	ShouldRetry bool  `json:"-"`
 	Error       error `json:"-"`
+	Description string
+	HelpURL     string
 }
 
 func MakeInconclusiveResult(name string) CheckResult {
--- a/checks/documentation.go
+++ b/checks/documentation.go
@@ -1,0 +1,90 @@
+// Copyright 2020 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package checks
+
+import (
+	//nolint
+	_ "embed"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/pkg/errors"
+)
+
+//go:embed checks.yaml
+var checks string
+
+type documentation struct {
+	Checks struct {
+		SecurityPolicy struct {
+			Description string `yaml:"description"`
+		} `yaml:"Security-Policy"`
+		Contributors struct {
+			Description string `yaml:"description"`
+		} `yaml:"Contributors"`
+		FrozenDeps struct {
+			Description string `yaml:"description"`
+		} `yaml:"Frozen-Deps"`
+		SignedTags struct {
+			Description string `yaml:"description"`
+		} `yaml:"Signed-Tags"`
+		SignedReleases struct {
+			Description string `yaml:"description"`
+		} `yaml:"Signed-Releases"`
+		CiTests struct {
+			Description string `yaml:"description"`
+		} `yaml:"CI-Tests"`
+		CodeReview struct {
+			Description string `yaml:"description"`
+		} `yaml:"Code-Review"`
+		CiiBestPractices struct {
+			Description string `yaml:"description"`
+		} `yaml:"CII-Best-Practices"`
+		PullRequests struct {
+			Description string `yaml:"description"`
+		} `yaml:"Pull-Requests"`
+		Fuzzing struct {
+			Description string `yaml:"description"`
+		} `yaml:"Fuzzing"`
+		Sast struct {
+			Description string `yaml:"description"`
+		} `yaml:"SAST"`
+		Active struct {
+			Description string `yaml:"description"`
+		} `yaml:"Active"`
+		BranchProtection struct {
+			Description string `yaml:"description"`
+		} `yaml:"Branch-Protection"`
+	} `yaml:"checks"`
+}
+
+// HelpDocumentation provides the documentation for the checks.
+type HelpDocumentation struct {
+	Description string
+	HelpURL     string
+}
+
+// ActiveHelpDocumentation provides the help documentation for ActiveHelpDocumentation check.
+func ActiveHelpDocumentation() (HelpDocumentation, error) {
+	var helpdocs documentation
+
+	if err := yaml.Unmarshal([]byte(checks), &helpdocs); err != nil {
+		return HelpDocumentation{}, errors.Wrap(err, "unable to unmarshal the json")
+	}
+
+	return HelpDocumentation{
+		Description: helpdocs.Checks.Active.Description,
+		HelpURL:     "https://github.com/ossf/scorecard/blob/main/checks/checks.md#active",
+	}, nil
+}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -185,9 +185,11 @@ func outputJSON(results []checker.CheckResult) {
 
 	for _, r := range results {
 		tmpResult := checker.CheckResult{
-			Name:       r.Name,
-			Pass:       r.Pass,
-			Confidence: r.Confidence,
+			Name:        r.Name,
+			Pass:        r.Pass,
+			Confidence:  r.Confidence,
+			Description: r.Description,
+			HelpURL:     r.HelpURL,
 		}
 		if showDetails {
 			tmpResult.Details = r.Details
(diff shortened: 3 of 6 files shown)


inferno-chromium:
Some things i noticed
- json file is not the source of truth, both checks.json and checks.md are generated files from checks.yaml, so i think program should use checks.yaml directly and we shouldn't modify json file here (like indent can be fixed in auto-generation logic).
- i dont think we want to add long description and remediation in output, since it will inflate latest.json significantly. i think the point was for checks.json would be read by some utility directly alongwith latest.json. however, I do see the point of showing this info when tool is run in standalone fashion like one repo only, so need to see how to show it only for that usecase.

i would let @azeemshaikh38 @oliverchang and @naveensrinivasan to sync here and discuss a solution here, just added my thoughts above.

naveensrinivasan:
> Some things i noticed
> 
> * json file is not the source of truth, both checks.json and checks.md are generated files from checks.yaml, so i think program should use checks.yaml directly and we shouldn't modify json file here (like indent can be fixed in auto-generation logic).
> * i dont think we want to add long description and remediation in output, since it will inflate latest.json significantly. i think the point was for checks.json would be read by some utility directly alongwith latest.json. however, I do see the point of showing this info when tool is run in standalone fashion like one repo only, so need to see how to show it only for that usecase.
> 
> i would let @azeemshaikh38 @oliverchang and @naveensrinivasan to sync here and discuss a solution here, just added my thoughts above.

I agree , I will change the implementation to use the YAML and will add HelpURL which points to the GitHub  instead of remediation.

naveensrinivasan:
> > Some things i noticed
> > 
> > * json file is not the source of truth, both checks.json and checks.md are generated files from checks.yaml, so i think program should use checks.yaml directly and we shouldn't modify json file here (like indent can be fixed in auto-generation logic).
> > * i dont think we want to add long description and remediation in output, since it will inflate latest.json significantly. i think the point was for checks.json would be read by some utility directly alongwith latest.json. however, I do see the point of showing this info when tool is run in standalone fashion like one repo only, so need to see how to show it only for that usecase.
> > 
> > i would let @azeemshaikh38 @oliverchang and @naveensrinivasan to sync here and discuss a solution here, just added my thoughts above.
> 
> I agree , I will change the implementation to use the YAML and will add HelpURL which points to the GitHub instead of remediation.

Updated the implementation to utilize the YAML instead of the `json`

checker/check_result.go
```diff
@@ -23,6 +23,8 @@ type CheckResult struct {
 	Details     []string
 	ShouldRetry bool  `json:"-"`
 	Error       error `json:"-"`
+	Description string
```
azeemshaikh38, line 26: Let's not bloat CheckResult with Description and HelpURL. We should have a separate struct for these which other packages can access as/when needed.

checks/documentation.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2020 Security Scorecard Authors
```
azeemshaikh38, line 1: Change filename -> check_documentation.go

checks/documentation.go
```diff
@@ -0,0 +1,90 @@
...
+			Description string `yaml:"description"`
+		} `yaml:"Active"`
+		BranchProtection struct {
+			Description string `yaml:"description"`
+		} `yaml:"Branch-Protection"`
+	} `yaml:"checks"`
+}
+
+// HelpDocumentation provides the documentation for the checks.
+type HelpDocumentation struct {
```
azeemshaikh38, line 73: We can expose this struct to other packages to get Description/HelpURL.

checks/documentation.go
```diff
@@ -0,0 +1,90 @@
...
+	Description string
+	HelpURL     string
+}
+
+// ActiveHelpDocumentation provides the help documentation for ActiveHelpDocumentation check.
+func ActiveHelpDocumentation() (HelpDocumentation, error) {
```
azeemshaikh38, line 79: Everytime a *HelpDocumentation is called, we will read the YAML file, unmarshal it and then send the result back. Instead, could we only unmarshal the first time, and have a var which saves this in memory so that the next time these fn calls happen, we only read from the in-memory var?

azeemshaikh38 (commented):
If this change is attempting to solve the "upload a SARIF file to GitHub" problem, should we first figure out the design? IMO, we should hold off on this PR before we have a consensus on how the design for the bigger problem would look like.

naveensrinivasan:
> If this change is attempting to solve the "upload a SARIF file to GitHub" problem, should we first figure out the design? IMO, we should hold off on this PR before we have a consensus on how the design for the bigger problem would look like.

It is not just for SARIF, anyone who uses the tool would like to know details and steps to address these issues https://github.com/ossf/scorecard/issues/63

checks/documentation.go
naveensrinivasan: It is already within the `checks` package.  Adding prefix doesn't add any value and also `_` aren't common within standard `go` naming though there are few exceptions.

checker/check_result.go
naveensrinivasan: The result is where the Help documentation needs to have the necessary stuff and this is what is being used in output. 

Where else do you think this could be used? 

How do you see this as bloat?

Also, what would be your suggestion?

checks/documentation.go
naveensrinivasan: The idea behind this is to avoid a global state.  This performance is negligible. IMO avoiding state is cleaner design.

checks/documentation.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/pkg/errors"
+)
+
+//go:embed checks.yaml
+var checks string
+
+type documentation struct {
+	Checks struct {
+		SecurityPolicy struct {
```
oliverchang, line 30: This is a lot of repetition and makes maintenance difficult. Can we marshal these into a `map[string]CheckDescription` or something similar?

cmd/root.go
```diff
@@ -185,9 +185,11 @@ func outputJSON(results []checker.CheckResult) {
 
 	for _, r := range results {
 		tmpResult := checker.CheckResult{
-			Name:       r.Name,
-			Pass:       r.Pass,
-			Confidence: r.Confidence,
+			Name:        r.Name,
+			Pass:        r.Pass,
+			Confidence:  r.Confidence,
+			Description: r.Description,
```
oliverchang, line 191: I also don't think we should include these long strings in these results by default. It will blow up the latest.json and cause a lot of duplication. 

If we need this for another use case, we should figure out another way to do this.

naveensrinivasan: I can think of a flag that either dumps the help information in the json or not. 

Do you have any other suggestions?

checks/documentation.go
naveensrinivasan: There is also a con with that approach, we lose the type-safety and strings are all over the codebase. 

We would have to define a constant for the string and use it which beats the purpose.

cmd/root.go
azeemshaikh38: Me and Oliver synced over this today. This is what we thought - https://github.com/ossf/scorecard/issues/63 wants to address the problem that if Scorecard tool fails, the user should know "why the check failed", "how confident are we about this result" and "how to fix it". This PR won't solve that issue. A simple example - if the check failed due to too many retries, we'd simply be outputting a generic description and helpURL. This does not help our case in making the tool more user friendly.

So here's my suggestion - 

1. Let's update all of our checks to return error_codes on failure. Each unique case of failure should have a unique error_code. These can simply be appended to []details of CheckResult for now.
2. We should then update the YAML to add details/remediation corresponding to these unique failures.
3. Update cmd/root.go to include the remediation string from YAML if any of the checks fail.

What do you think? @oliverchang @inferno-chromium for feedback/opinions.

inferno-chromium: Sounds like a plan, yes giving actual remediation instead of everything is a nice idea. and in yes in tool run, it does not make sense to give everything, (although it can say at end, for more information on check, please see doc at <link>).

oliverchang: Sounds like a good plan to me too. Raising more relevant errors messages seems way more useful. @naveensrinivasan WDYT?

laurentsimon: SG.

Shall we also provide additional information about the checks we make in the command line output (when `--show-details` is provided)? I personally find it useful - even if somehow redundant with the YAML document file. Here's an example output (YES implicitly means it's a positive check outcome):
```
$ scorecard <> --show-details
RESULTS
-------
Repo: some-repo
Active: Fail 
	Confidence: 100%
	Commits in last 90 days greater than 2: NO (1)
Branch-Protection: Fail 
	Confidence: 60%
	AllowForcePushes disabled: YES
	AllowDeletions disabled: YES
	EnforceAdmins enabled: YES
	RequireLinearHistory enabled: YES
	RequiredStatusChecks enabled: YES but has no contexts
	RequiredPullRequestReviews enabled: YES but with only 1 reviews
	DismissStaleReviews enabled: YES
	DismissalRestrictions enabled: NO
	RequireCodeOwnerReviews enabled: YES

```

This is just a first attempt and be changed, of course.

PR #353, opened 2021-04-22
:bug: Fix typo in scorecard date format.

oliverchang:
This is causing the Date to be incorrectly expanded as:

`<Day><Month>6-<Month>-<Day>` rather than `<Year>-<Month>-<Day>`

And causing the BQ import to still fail (#295)


azeemshaikh38 (approved):
Oops! Thanks for finding and fixing this :)

We should have a test to make sure the output of cron doesn't cause these issues. Finding it out after running an entire cron job is terrible :(

oliverchang:
> Oops! Thanks for finding and fixing this :)
> 
> We should have a test to make sure the output of cron doesn't cause these issues. Finding it out after running an entire cron job is terrible :(

Yep we need better tests here :) I'll deploy and re-run the cron.

PR #371, opened 2021-04-29
:seedling: Clean and Validation of cron

naveensrinivasan:

Bug fix

  * Included validation for the cron
  * Removed branch protection check from the cron

 None

* **Other information**:
None

### DIFF
--- a/cron/main.go
+++ b/cron/main.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"os/exec"
 	"time"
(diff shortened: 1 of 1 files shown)


cron/main.go
```diff
@@ -73,6 +79,11 @@ func main() {
 			panic(err)
 		}
 		sugar := logger.Sugar()
+		// deleting branch-protection
```
inferno-chromium, line 82: maybe prefix this with FIXME:, we want to work in future with github to reconsider this decision.

inferno-chromium (approved):

cron/main.go
naveensrinivasan: Agreed, Thanks

PR #348, opened 2021-04-19
:bug: Fix linting issues (1 of n)

chrismcgehee:
* **Please check if the PR fulfills these requirements**
- [ n/a] Tests for the changes have been added (for bug fixes / features)
- [x] PR title follows the guidelines defined in  https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Code quality improvement

* **What is the current behavior?** (You can also link to an open issue here)
When running `bin/golangci-lint run`, there are quite a few linting errors. I figured it would be easier for everyone if I created several smaller PRs instead of one humongous PR. This PR fixes issues from 5 linters:\
gocritic\
paralleltest\
errcheck\
wrapcheck\
whitespace\
Future PRs will fix the rest of the linters as well as introduce the new linters Naveen requested in #173.

* **What is the new behavior (if this is a feature change)?**
n/a

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
No

* **Other information**:
Each commit tackles the issues from a single linter. I would recommend reviewing this PR commit by commit.

PR #392, opened 2021-05-02
🌱 Fix intermittent test failure

chrismcgehee:

Test fix

I believe I introduced a bug in one of the tests in  PR [#348](https://github.com/ossf/scorecard/pull/348). In `roundtripper_test.go`, I added `t.Parallel()` inside the `for` loops. Because these tests are calling `os.Setenv` and `os.Getenv` for the same environment variables, it is possible for one test case to interfere with another. This appears to have happened in https://github.com/ossf/scorecard/runs/2487189008.

These tests should not intermittently fail.

No

* **Other information**:


naveensrinivasan (approved):
:shipit:  Thank you!

PR #393, opened 2021-05-02
:sparkles: Generate docs using go instead of python

naveensrinivasan:

  Docs

  https://github.com/ossf/scorecard/pull/345#issuecomment-821273514

 * Implemented the doc generation from python to go
 * Removed the need for json
 * Sorted the output of the generated markdown

None

* **Other information**:

None


checks/main/main.go
```diff
@@ -0,0 +1,95 @@
...
+	m := make(map[string]map[string]doc)
+	err = yaml.Unmarshal(byteValue, &m)
+	if err != nil {
+		panic(err)
+	}
+	keys := make([]string, 0, len(m["checks"]))
+	for k := range m["checks"] {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
oliverchang, line 49: why sort? we might want to preserve the logical order specified in the yaml

oliverchang (commented):
thanks! Just a comment.

checks/main/main.go
naveensrinivasan: If I am looking for docs I would expect things to be sorted so that I can go look for specific items. 

Most of the docs are sorted and that was the reason.

oliverchang (approved):

PR #395, opened 2021-05-03
🌱 Fix lint issues: govet linter

chrismcgehee:

Code quality improvement

This PR fixes issues from the [govet](https://golang.org/cmd/vet/) linter. With current settings, the only checks enabled are for field alignment and invalid struct field tags. The `fieldalignment` check will look for structs that could take up less memory if the fields are re-ordered. It can also give the optimal ordering which is what I used for this commit. \
Part of issue [#173](https://github.com/ossf/scorecard/issues/173).

n/a

No. With my latest commit, there is no breaking change in this PR.


naveensrinivasan:
Thank you! for taking all the warnings. 

IMO For now let’s not have any breaking changes on the JSON. 

I will let @azeemshaikh38 @inferno-chromium and @oliverchang  also chime in.

azeemshaikh38 (changes requested):
There is a chance this will break our BQ data transfer service, since AFAICT, jagged rows are only allowed in CSV format (
image.

I would err on the side of caution and not risk breaking it for now. We can revisit this later.

checker/check_result.go
```diff
@@ -17,12 +17,12 @@ package checker
 const MaxResultConfidence = 10
 
 type CheckResult struct {
+	Error       error `json:"-"`
 	Name        string
-	Pass        bool
+	Details     []string `json:",omitempty"`
```
azeemshaikh38, line 22: Let's remove the "omitempty" option here. That should allow us to submit the PR without any breaking change?

chrismcgehee: Yep, that will make the PR not have any breaking changes. I have removed that option.

azeemshaikh38 (approved):

chrismcgehee:
This one's ready to merge.

PR #398, opened 2021-05-04
🌱 Fix lint issues: goerr113 linter

chrismcgehee:

Code quality improvement

This PR fixes issues from the [goerr113](https://github.com/Djarvur/go-err113/) linter (checks error handling expressions). One thing this linter checks for is that any custom error wraps an underlying static error. The benefit to this is that calling code can check for errors of a certain type. For example instead of checking if an error message contains the string `"unsupported host"`, the code can use `errors.Is(err, ErrorUnsupportedHost)` to perform a strongly-typed comparison. \
Part of issue [#173](https://github.com/ossf/scorecard/issues/173).

n/a

No


repos/repo_url.go
```diff
@@ -15,12 +15,18 @@
...
+	"errors"
 	"fmt"
 	"log"
 	"net/url"
 	"strings"
 )
 
+var (
+	ErrorUnsupportedHost  = errors.New("unsupported host")
+	ErrorInvalidGithubURL = errors.New("invalid GitHub repo url")
```
naveensrinivasan, line 27: `url` should be `URL`

naveensrinivasan (approved):
Thanks

azeemshaikh38 (approved):

repos/repo_url.go
chrismcgehee: Made the change

Issue #412, opened 2021-05-07
Frozen deps does not check for workflow pinning


laurentsimon:
In our own repo:
```
$grep uses .github/workflows/* | grep master
.github/workflows/codescan.yml:        uses: aquasecurity/trivy-action@master
.github/workflows/integration.yml:        uses: google-github-actions/setup-gcloud@master
.github/workflows/synk-cron.yaml:        uses: snyk/actions/iac@master
.github/workflows/synk.yaml:        uses: snyk/actions/iac@master
```

Shall we pin these? If yes, we should also add this to the Frozen-Dep check.

laurentsimon:
@naveensrinivasan to add the check in scorecard, my current plan is to download the tarball and use a regex for files under `.github/workflow/` to detect `uses: somename@` followed by `vX.Y.Z` or `X.Y.Z` where Y and Z are optional. Basically `@v?\d(\.\d){1,2}$`. I don't like that we have to download the tarball again (it's already done in another check). 

Long-term, we can probably download the tarball once and pass it to other checks that need it.

 wdut?

naveensrinivasan:
@laurentsimon  Tags can be moved. That is the reason teams prefer to pin the SHA instead of tags. So doing checks for tags is not recommended IMO.

Scorecard avoids maintaining the state within the CLI. Passing the downloaded tarball would do that. We would like to keep it stateless. 

I agree to download the tarball again is not utilizing the resources well. But there is a bigger discussion on scaling these API calls and we will be addressing that soon.

laurentsimon:
> @laurentsimon Tags can be moved. That is the reason teams prefer to pin the SHA instead of tags. So doing checks for tags is not recommended IMO.

yes, but do you know if hash pins are supported in github workflows?
We have a dep-pinning check for package management (go, python, etc). Is the reasoning different there? 

> 
> Scorecard avoids maintaining the state within the CLI. Passing the downloaded tarball would do that. We would like to keep it stateless.
> 
> I agree to download the tarball again is not utilizing the resources well. But there is a bigger discussion on scaling these API calls and we will be addressing that soon.

ok.

naveensrinivasan:
> > @laurentsimon Tags can be moved. That is the reason teams prefer to pin the SHA instead of tags. So doing checks for tags is not recommended IMO.
> 
> yes, but do you know if hash pins are supported in github workflows?
> We have a version-pinning check with no hash for package management (go, python, etc). Is the reasoning different there?
> 

[...]

PR #417, opened 2021-05-08
🐛  Pin workflow dependencies

laurentsimon:

This is a patch fix.

We currently don't pin dependencies #412 

We pin dependencies in the workflow.

no

* **Other information**:

### DIFF
--- a/.github/workflows/codescan.yml
+++ b/.github/workflows/codescan.yml
@@ -33,7 +33,7 @@ jobs:
           DOCKER_BUILDKIT=1 docker build -t docker.pkg.github.com/ossf/scorecard:${{ github.sha }} .
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@master
+        uses: aquasecurity/trivy-action@v0.0.15
         with:
           image-ref: 'docker.pkg.github.com/ossf/scorecard:${{ github.sha }}'
           format: 'template'
--- a/.github/workflows/codescan.yml
+++ b/.github/workflows/codescan.yml
@@ -33,7 +33,8 @@ jobs:
           DOCKER_BUILDKIT=1 docker build -t docker.pkg.github.com/ossf/scorecard:${{ github.sha }} .
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@v0.0.15
+        # https://github.com/aquasecurity/trivy-action/releases/tag/0.0.15
+        uses: aquasecurity/trivy-action@341f810bd602419f966a081da3f4debedc3e5c8e
         with:
           image-ref: 'docker.pkg.github.com/ossf/scorecard:${{ github.sha }}'
           format: 'template'
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -64,7 +64,8 @@ jobs:
           go-version: '1.16'
 
       - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@v0.2.0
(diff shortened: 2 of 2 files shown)


naveensrinivasan (changes requested):
Thanks, could you please use the PR template and fill in the details?

.github/workflows/codescan.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           DOCKER_BUILDKIT=1 docker build -t docker.pkg.github.com/ossf/scorecard:${{ github.sha }} .
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@master
+        uses: aquasecurity/trivy-action@v0.0.15
```
naveensrinivasan, line 36: Instead of tag , can you please pin it to a sha? Tags can be moved

.github/workflows/integration.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           go-version: '1.16'
 
       - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@master
+        uses: google-github-actions/setup-gcloud@v0.2.0
```
naveensrinivasan, line 67: Instead of tag , can you please pin it to a sha? Tags can be moved

naveensrinivasan (changes requested):

.github/workflows/integration.yml
laurentsimon: same as above

.github/workflows/codescan.yml
laurentsimon: Would love to but I'm note sure it's supported. 
In addition, the rest of the code uses versioning and not hash pin, so this is consistent and should be good enough for this PR. We can create a new PR to move all the code to using hash at once, if it's doable/supported.

Thoughts?

naveensrinivasan: > Would love to but I'm note sure it's supported.

Have you tried it? It is supported, I have used it in the past 

https://julienrenaux.fr/2019/12/20/github-actions-security-risk/#solution-use-commit-hash-as-version

> In addition, the rest of the code uses versioning and not hash pin, so this is consistent and should be good enough for this PR. We can create a new PR to move all the code to using hash at once, if it's doable/supported.
> 
> Thoughts?

I agree the rest of the actions code isn't using SHA, if we are fixing it then let's do it right.  Unless there are some issues in using the SHA.

inferno-chromium: Yes, i didnt know either, this is definitely something we should do, scorcard should set the benchmark for secure configuration.
E.g. "actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4"
from https://dev.to/mheap/ensure-your-github-actions-are-pinned-to-a-sha-2d4k

inferno-chromium (changes requested):
lets check and fix all action *.yml

inferno-chromium (approved):

.github/workflows/codescan.yml
```diff
@@ -33,7 +33,8 @@ jobs:
           DOCKER_BUILDKIT=1 docker build -t docker.pkg.github.com/ossf/scorecard:${{ github.sha }} .
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@v0.0.15
+        # https://github.com/aquasecurity/trivy-action/releases/tag/0.0.15
```
naveensrinivasan, line 36: I like that you had included the tag which is superhelpful. Thanks 👍

naveensrinivasan (approved):
:shipit:

PR #421, opened 2021-05-08
🌱 Return error instead of log.fatal.

azeemshaikh38:

Removes log.fatal to instead return error.

Fatal failure on callee side instead letting caller decide behaviour.

Caller decides bahviour on failure.

No.


naveensrinivasan (changes requested):
The golint is failing.

naveensrinivasan (approved):

Issue #318, opened 2021-03-30
Scale scorecard from 2K to 100K to a million repos


inferno-chromium:

naveensrinivasan:
I agree with the goal. But I think it should be initially 10k repositories.

azeemshaikh38:
Outcome of initial internal discussion - we'll start with focusing on horizontal scalability of the cron job itself. That is, instead of a single worker machine processing all repositories, we will re-design the architecture such that more workers can be added to improve the performance of the job. To that end, the proposed solution is to use a PubSub architecture - all the repositories to be processed will be pushed to a Topic and multiple subscribers/workers read from this and write to GCS in JSON formatted output.

The JSON output will be sharded, and BQ will import this sharded JSON using BQ data transfer service. 

BQ table will be date partitioned for better scalability. Detailed design discussions about how this sharded JSON output will interact with the BQ table is here - https://github.com/ossf/scorecard/issues/366

azeemshaikh38:
@inferno-chromium @oliverchang @naveensrinivasan 

Update on the GitHub token issue: my thoughts on it so far are that we should consider using [GitHub conditional requests](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests) as a starter. Basically, we store the entire HTTP response from GitHub (along with ETag) indexed by the requestURL. Our subsequent HTTP calls will be backed by this "ETag Cache".

A simple implementation of the "ETag Cache" could be backed by Blob store. Filenames will be the requestURLs and file content will be the HTTP response. This is probably terrible in terms of IO performance, but since our concern here is reducing GitHub token usage, might be a good starting point and we can look into improving the implementation of the ETag cache in the future. What do you guys think? Thoughts/feedback?

jeffmendoza:
I've been doing some conditional request testing. Unfortunately the etag changes for each process invocation (each new auth token), even if the resource didn't change. I was using a GitHup App (https://docs.github.com/en/developers/apps/authenticating-with-github-apps#authenticating-as-a-github-app) though, not sure how it works with a PAT or OAuth app.

azeemshaikh38:
Oh interesting. Did you also try the "Last-Modified" and "If-Modified-Since" options? I would expect that they shouldn't be process dependent.

jeffmendoza:
I only tried `If-None-Match`

azeemshaikh38:
[...]

PR #203, opened 2021-02-21
Feat-Implement httpcache middleware for GitHub API

naveensrinivasan:
* **Please check if the PR fulfills these requirements**
- [x] Tests for the changes have been added (for bug fixes/features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Feature - Caching

* **What is the current behavior?** (You can also link to an open issue here)
#80 #202 

* **What is the new behavior (if this is a feature change)?**
The GitHub API supports conditional requests
https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests

https://github.com/google/go-github supports Conditional requests
https://github.com/google/go-github#conditional-requests

As we are scaling more and more projects this would add a lot of value.

Initial run fetches information using `httpcache` as a middleware,
which caches the HTTP response initially in a large disk (PVC),
probably move to Redis later as a cache instead of disk.

Subsequent `cron runs` will utilize the `httpcache` for checking content modification and
load it from the cache if it isn't modified, which reduces the hitting the
Rate Limit of the GitHub API.

Also fixed the golang-ci warnings.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
None

* **Other information**:

Subsequent cache runs on 50 repositories takes about 18 minutes with `3` GitHub tokens

```
❯ time ./cron.sh
github.com/18F/identity-idp
github.com/1N3/Sn1per
github.com/24pullrequests/24pullrequests
github.com/AElfProject/AElf
github.com/Activiti/Activiti
github.com/AdguardTeam/AdguardFilters
github.com/AlchemyCMS/alchemy_cms
github.com/AliasIO/wappalyzer
github.com/Amanieu/parking_lot
github.com/abonas/kubeclient
github.com/abpframework/abp
github.com/acassen/keepalived
github.com/acmesh-official/acme.sh
github.com/actions/runner
github.com/activeadmin/activeadmin
github.com/activemerchant/active_merchant
github.com/activerecord-hackery/ransack
github.com/activescaffold/active_scaffold
github.com/actix/actix
github.com/actix/actix-extras
github.com/actix/actix-net
github.com/actix/actix-web
github.com/adafruit/Adafruit_CircuitPython_Bundle
github.com/adempiere/adempiere
github.com/afragen/github-updater
github.com/aframevr/aframe
github.com/ai/autoprefixer-rails
github.com/aio-libs/aiohttp
github.com/airbnb/javascript
github.com/airbrake/airbrake
[...]

Issue #80, opened 2020-11-25
Scorecard scalability limitation: Reduce GitHub API calls


htuch:
I've been running scorecard over ~40 Envoy dependencies a few times in the past hour and even using a personal access token for GitHub, I've hit the rate limit. I'm wondering what can be done to make this work better in scorecard. Suggestions:
* Profile GitHub API calls, determine the most costly, optimize to API alternatives. Share details on which are most expensive in public documentation.
* Add support for local caching of results for some time bound period.
* Provide a scorecard network service that collects results for projects periodically.
* Have scorecard JSON examples generalized to some contributor editable file that will collect projects nightly.

dlorenc:
> * Provide a scorecard network service that collects results for projects periodically.

This is the approach we've taken so far. There's a cron that runs every day on ~100 projects and publishes the results in GCS/BigQuery.

I'd be happy to add the envoy deps into that list, it's here: https://github.com/ossf/scorecard/blob/main/cron/projects.txt

dlorenc:
Adding the envoy deps here: #84

naveensrinivasan:
GitHub API supports conditional requests https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests
>Most responses return an ETag header. Many responses also return a Last-Modified header. You can use the values of these headers to make subsequent requests to those resources using the If-None-Match and If-Modified-Since headers, respectively. If the resource has not changed, the server will return a 304 Not Modified.
Making a conditional request and receiving a 304 response does not count against your Rate Limit, so we encourage you to use it whenever possible.

>Making a conditional request and receiving a 304 response does not count against your Rate Limit, so we encourage you to use it whenever possible.

https://github.com/google/go-github supports Conditional requests https://github.com/google/go-github#conditional-requests

As we are scaling more and more projects this would add a lot of value.

cc @inferno-chromium

naveensrinivasan:
A visual representation of the proposed solution.



![image](https://user-images.githubusercontent.com/172697/108604202-6d300500-737a-11eb-86e5-b7ecfc08e85d.png)

1. `k8s cron job` runs on a schedule.
2. Initial run fetches information using `httpcache` as a middleware, which caches the HTTP response initially in a large disk (PVC), probably move to Redis later as a cache instead of disk.
3. Subsequent `cron runs` will utilize the `httpcache` for checking content modification and load it from the cache if it isn't modified, which reduces the hitting the Rate Limit of the GitHub API.

[...]

PR #428, opened 2021-05-10
🌱  Setup PubSub framework code

azeemshaikh38:

Adds framework/skeleton code for setting up PubSub batch job as discussed in https://github.com/ossf/scorecard/issues/318#issuecomment-832348904. This PR also includes code for Master server which publishes data to the repo_request topic.

N/A

No new behaviour added yet in this PR.

No.


naveensrinivasan:
@azeemshaikh38  Some documentation on a large PR would help with the review.  The approach and your thought process. 

For example here is HTTP caching PR https://github.com/ossf/scorecard/pull/203 and some related docs https://github.com/ossf/scorecard#caching
https://github.com/ossf/scorecard/issues/80#issuecomment-782723182

I would suggest that some docs be included for review as well as others to contribute in the future.

azeemshaikh38:
Re: large PR comment. Each individual piece didn't make sense on its own, so had to wait to send this out only after all parts were in order. 

Re: documentation. The PR is based on this comment - https://github.com/ossf/scorecard/issues/318#issuecomment-832348904. What this is PR trying to achieve is - given a list of repos, "master/main.go" will send out requests to a PubSub topic. To do this, it follows these steps:

1. Uses "data/reader.go" to read from the input file. "data/reader.go" abstracts away the format of the file itself and provides a simple HasNext(), GetNext() type interface.
2. Uses "data/request.proto" to create a request proto. Each request corresponds to a single JSON shard file in  https://github.com/ossf/scorecard/issues/318#issuecomment-832348904.
3. Uses "data/publisher.go" to publish to a given topic.
4. Uses "data/blob.go" to write some metadata once all requests are successfully sent. 

I have tried to be as generic as possible to avoid GoogleCloud specific logic. So as much as possible, logic is encapsulated inside gocloud.dev packages. The config/config.go serves a single place to update and use different cloud services if need be.

naveensrinivasan:
> Re: large PR comment. Each individual piece didn't make sense on its own, so had to wait to send this out only after all parts were in order.
> 
> Re: documentation. The PR is based on this comment - [#318 (comment)](https://github.com/ossf/scorecard/issues/318#issuecomment-832348904). What this is PR trying to achieve is - given a list of repos, "master/main.go" will send out requests to a PubSub topic. To do this, it follows these steps:
> 
> 1. Uses "data/reader.go" to read from the input file. "data/reader.go" abstracts away the format of the file itself and provides a simple HasNext(), GetNext() type interface.
> 2. Uses "data/request.proto" to create a request proto. Each request corresponds to a single JSON shard file in  [#318 (comment)](https://github.com/ossf/scorecard/issues/318#issuecomment-832348904).
> 3. Uses "data/publisher.go" to publish to a given topic.
> 4. Uses "data/blob.go" to write some metadata once all requests are successfully sent.
> 
> I have tried to be as generic as possible to avoid GoogleCloud specific logic. So as much as possible, logic is encapsulated inside gocloud.dev packages. The config/config.go serves a single place to update and use different cloud services if need be.

Thanks this helps!

cron/data/blob.go
```diff
@@ -0,0 +1,63 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/ossf/scorecard/cron/config"
+	"gocloud.dev/blob"
+)
+
+const (
+	filePrefixFormat = "2006.01.02/150405/"
```
naveensrinivasan, line 28: Some documentation on the filePrefixformat will help.

naveensrinivasan: What is `150405` signify? It is not clear.

cron/data/blob.go
```diff
@@ -0,0 +1,63 @@
...
+		return fmt.Errorf("%w: %v", errorWriteClose, err)
+	}
+	return nil
+}
+
+func GetBlobFilename(filename string, datetime time.Time) string {
+	return datetime.Format(filePrefixFormat) + filename
+}
+
+func GetShardNumFilename(datetime time.Time) string {
```
naveensrinivasan, line 61: Docs on this function would be helpful. Probably how and the data is going to be sharded.

cron/data/publisher.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return &ret, fmt.Errorf("%w: %v", errorOpenTopic, err)
+	}
+	return &publisherImpl{
+		ctx:   ctx,
+		topic: topic,
+	}, nil
+}
+
+type topicInterface interface {
```
naveensrinivasan, line 53: Instead of `topicInterface` could be just `publisher`? And the method be `publish`.
 https://golang.org/doc/effective_go#interface-names
 
In golang By convention, one-method interface names are nouns denoting the doer of an action.

cron/data/reader.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"io"
+
+	"github.com/jszwec/csvutil"
+	"github.com/ossf/scorecard/repos"
+)
+
+var errorCSVDecoder = errors.New("could not create CSV decoder")
+
+type Reader interface {
```
naveensrinivasan, line 29: > By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc.

https://golang.org/doc/effective_go#interface-names

Why not use the standard `io.Reader` interface? It is used across standard lib and lots of things are built on this convention. https://golang.org/pkg/io/


The iterator pattern used in other languages is not idiomatic go because other things aren't built around this way.

naveensrinivasan (changes requested):
I like the basic structure. Expect few things that aren't idiomatic go and few nits.

cron/data/blob.go
```diff
@@ -0,0 +1,63 @@
...
+var (
+	errorWrite      = errors.New("blobWriter.Write")
+	errorWriteClose = errors.New("blobWriter.Close")
+	errorOpenBucket = errors.New("blob.OpenBucket")
+)
+
+func WriteToBlobStore(ctx context.Context, bucketURL, filename string, data []byte) error {
+	bucket, err := blob.OpenBucket(ctx, bucketURL)
+	if err != nil {
+		return fmt.Errorf("%w: %v", errorOpenBucket, err)
```
oliverchang, line 40: nit: I don't think these fmt.Errorf calls are correct. See examples in https://blog.golang.org/go1.13-errors#TOC_3.3.

You should only have 1 error wrapped by %w, and this should be the underlying `err` that you got back from the previous call.


i.e.

`fmt.Errorf("blob.OpenBucket failed: %w", err)`

cron/master/main.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2021 Security Scorecard Authors
```
oliverchang, line 1: nit: consider a name other than "master" for this. Maybe "controller" ?

cron/master/main.go
```diff
@@ -0,0 +1,84 @@
...
+			continue
+		}
+		request.Repos = append(request.GetRepos(), repoURL.URL())
+		if len(request.GetRepos()) < config.ShardSize {
+			continue
+		}
+		if err := topicPublisher.Publish(&request); err != nil {
+			return shardNum, fmt.Errorf("%w: %v", errorPublishing, err)
+		}
+		_ = request.Repos[:0]
```
oliverchang, line 59: what's this for?

oliverchang, line 59: do you mean to increment shardnum here?

cron/master/main.go
```diff
@@ -0,0 +1,84 @@
...
+	t := time.Now()
+	reposFile, err := os.OpenFile(config.InputReposFile, os.O_RDONLY, 0o644)
+	if err != nil {
+		panic(err)
+	}
+	if reader, err := data.MakeReader(reposFile); err != nil {
+		panic(err)
+	} else if shardNum, err := PublishToRepoRequestTopic(ctx, reader, t); err != nil {
+		panic(err)
+	} else if err := data.WriteToBlobStore(ctx,
```
oliverchang, line 78: the logic is actually kind of hard to follow here when written this way. this is probably more readable without the "else ifs".

oliverchang (commented):
nice! some cursory comments. will take a closer look tomorrow.

cron/data/reader.go
azeemshaikh38: I don't agree with iterator pattern not being a Go way. I see popular Go packages use this pattern, including Google APIs - https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines

However, changing the interface name to `Iterator` to better reflect its purpose better.

cron/data/blob.go
azeemshaikh38: I want to avoid obvious comments in code. The fn does precisely as the name suggests, i.e - given a `time.Time` value, returns a filename `string` which in this case, is a pre-defined metadata file called `.shard_num`. It does not impact data sharding.

cron/config/config.go
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permisshandlerns and
+// limitathandlerns under the License.
+
+package config
+
+const (
+	ResultDataBucketURL string = "gs://ossf-scorecard-data"
+	RequestTopicURL            = "gcppubsub://projects/openssf/topics/scorecard-batch"
```
inferno-chromium, line 19: openssf should be moved to its own var and then RequestTopicURL should embed it.

cron/data/publisher.go
```diff
@@ -0,0 +1,88 @@
...
+	"log"
+	"sync"
+	"sync/atomic"
+
+	"gocloud.dev/pubsub"
+	_ "gocloud.dev/pubsub/gcppubsub" // Needed to link in GCP drivers.
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+var errorPublish = errors.New("total errors when publishing")
```
inferno-chromium, line 30: seems used at only one site, why not just put it there.

cron/data/iterator_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}{
+		{
+			name:     "Basic",
+			filename: "testdata/basic.csv",
+			outcomes: []outcome{
+				{
+					hasError: false,
+					repo: repos.RepoURL{
+						Host:  "github.com",
+						Owner: "18F",
```
inferno-chromium, line 44: nit: should just use fake repo names instead of these weird repo names 18F/1N3, etc

cron/data/request.pb.go
```diff
@@ -0,0 +1,180 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
inferno-chromium, line 1: Add a README.md file on how to generate this (command line used to generate this).

cron/data/publisher.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2021 Security Scorecard Authors
```
inferno-chromium, line 1: publisher*.go seems weird to go in data/ dir, maybe something else?

inferno-chromium (dismissed):

cron/master/main.go
oliverchang: yeah, that's just how Go is :)

cron/data/iterator.go
```diff
@@ -0,0 +1,67 @@
...
+	return reader.err != io.EOF
+}
+
+func (reader *csvReader) Next() (repos.RepoURL, error) {
+	if reader.err != nil {
+		return repos.RepoURL{}, reader.err
+	}
+	ret := repos.RepoURL{}
+	var err error
+	err = ret.Set(reader.next.Repo)
```
oliverchang, line 62: nit: just `err := ` ?

oliverchang (approved):
nice!

naveensrinivasan (dismissed):

cron/config/config.go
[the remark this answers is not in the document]
inferno-chromium: Ideally, project name and gcs bucket should go in a config.yaml and read from there. If some env var is set (e.g. someone's dockerfile who wants to run scorecard service on their project (although rare)), it can use it as an override, otherwise use the default from config.yaml. for tests, you can just set those env vars. this should avoid dependency injection. wdyt @naveensrinivasan @azeemshaikh38 @oliverchang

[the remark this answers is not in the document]
naveensrinivasan: I agree with that https://12factor.net/config. Thanks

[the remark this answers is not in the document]
azeemshaikh38: Without the `scorecard` word, the URL does not indicate that it is related to Scorecard project. How about `scorecard-requests` or `scorecard-batch-requests`?

cron/config/config.go
```diff
@@ -0,0 +1,23 @@
...
+	InputReposFile             = "../projects.txt"
```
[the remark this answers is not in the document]
azeemshaikh38: Done.

cron/controller/main.go
```diff
@@ -0,0 +1,82 @@
...
+	request := data.ScorecardBatchRequest{
+		JobTime:  timestamppb.New(datetime),
+		ShardNum: &shardNum,
+	}
+	topicPublisher, err := data.CreatePublisher(ctx, config.RequestTopicURL)
+	if err != nil {
+		return shardNum, fmt.Errorf("error running CreatePublisher: %w", err)
+	}
+
+	for iter.HasNext() {
```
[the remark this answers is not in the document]
azeemshaikh38: Done.

cron/data/iterator.go
azeemshaikh38: Go complains with `no new variables on left side of :=` if I use `err := `. Maybe I'm missing some Go syntax here?

cron/data/iterator_test.go
azeemshaikh38: Done.

cron/data/publisher.go
azeemshaikh38: Done. Moved to pubsub/

cron/data/publisher.go
azeemshaikh38: This is to avoid the linter from complaining - dynamic errors are not allowed.

cron/config/config.go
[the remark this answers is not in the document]
inferno-chromium: sure, scorecard-batch-requests seems fine.

naveensrinivasan:
There is a binary file in this commit https://github.com/ossf/scorecard/blob/0f01b1501a5f4dad1ef50a5f79edba08e0a82753/cron/cron which could be an accidental commit.

inferno-chromium (approved):
putting approval back as a setting change had reset it.

azeemshaikh38:
> There is a binary file in this commit https://github.com/ossf/scorecard/blob/0f01b1501a5f4dad1ef50a5f79edba08e0a82753/cron/cron which could be an accidental commit.

Deleted.

cron/config/config.go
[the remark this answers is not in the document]
azeemshaikh38: I have added a TODO for this. Will have a follow up PR addressing this to avoid adding more logic in this PR.

azeemshaikh38: Moving these vars to config.yaml in a follow up PR. Will address this there.

cron/data/request.pb.go
azeemshaikh38: Done.

Issue #459, opened 2021-05-16
Feature - manage tool dependencies  via go.mod


naveensrinivasan:
**Is your feature request related to a problem? Please describe.**
Manage tool dependencies via go.mod

**Describe the solution you'd like**
Manage tool dependencies like golint , protoc-gen-go using go.mod https://marcofranssen.nl/manage-go-tools-via-go-modules

azeemshaikh38:
Is this issue same as (or can be merged with): https://github.com/ossf/scorecard/issues/429?

naveensrinivasan:
> Is this issue same as (or can be merged with): #429?

Yes! Closing this in favor of #429

PR #460, opened 2021-05-16
:seedling: Move tool dependencies into go.mod

naveensrinivasan:

 Feature 

  #459

    Moved the tool dependencies into go.mod

 None

* **Other information**:
None


Makefile
```diff
@@ -31,16 +31,9 @@ help:  ## Display this help
...
-$(GINKGO):
-	go get -u github.com/onsi/ginkgo/ginkgo@v1.16.2
-
-$(GOLANGGCI_LINT):
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.40.0
-
-$(PROTOC_GEN_GO):
-	go install google.golang.org/protobuf/cmd/protoc-gen-go
+install: | $(PROTOC) 
+	@echo Installing tools from tools.go
```
azeemshaikh38, line 35: How about this?

```
install:
    go get ./...
```

Makefile
```diff
@@ -133,14 +128,14 @@ unit-test: ## Runs unit test without e2e
 e2e: ## Runs e2e tests. Requires GITHUB_AUTH_TOKEN env var to be set to GitHub personal access token
 e2e: build-scorecard check-env | $(GINKGO)
 	# Run e2e tests. GITHUB_AUTH_TOKEN with personal access token must be exported to run this
-	ginkgo --skip="E2E TEST:executable" -p -v -cover  ./...
+	$(GINKGO) --skip="E2E TEST:executable" -p -v -cover  ./...
 
 ci-e2e: ## Runs CI e2e tests. Requires GITHUB_AUTH_TOKEN env var to be set to GitHub personal access token
-ci-e2e: build-scorecard check-env e2e-cron | $(GINKGO)
+ci-e2e: build-scorecard check-env e2e-cron install| $(GINKGO)
```
azeemshaikh38, line 134: Specifying `install` as dependency will always install all dependencies when we run `make ci-e2e`. Instead, we can have `| $(GINKGO)` and provide a rule for `$(GINKGO)` as:

```
$(GINKGO): install
```

That way, it'll only run the install command if `$(GINKGO)` file doesn't exist. Wdyt?

tools.go
```diff
@@ -0,0 +1,23 @@
+// +build tools
```
azeemshaikh38, line 1: `dependencies.go` or `install.go`?

naveensrinivasan: `dependencies` are ambiguous because of the repo with other dependencies. 
`install` is very generic and confusing with installing what?

tools.go is standard across the `go` community https://github.com/golang/go/issues/25922

Makefile
naveensrinivasan: >go install, with or without a version suffix (as described above), is now the recommended way to build and install packages in module mode. go get should be used with the -d flag to adjust the current module's dependencies without building packages, and use of go get to build and install packages is deprecated. In a future release, the -d flag will always be enabled.

https://tip.golang.org/doc/go1.16#go-command

`go install` is the recommended way. That is the reason. Let me know.

Makefile
naveensrinivasan: Yes agreed. Thanks!

Makefile
naveensrinivasan: Thanks

tools.go
naveensrinivasan: Let me know.

azeemshaikh38: Sounds good.

Makefile
```diff
@@ -62,8 +55,10 @@ update-dependencies: ## Update go dependencies for all modules
 	# Update ./scripts/update go modules
 	cd ./scripts/update && go mod tidy && go mod verify
 
+$(GOLANGGCI_LINT): install
+
 check-linter: ## Install and run golang linter
-check-linter: | $(GOLANGGCI_LINT)
+check-linter: $(GOLANGGCI_LINT)
```
azeemshaikh38, line 61: Shouldn't this be `| $(GOLANGCI_LINT)`? We only care about if this file exists or not.

azeemshaikh38 (approved):

PR #461, opened 2021-05-17
✨  Update RunScorecards API

azeemshaikh38:

Updates the RunScorecard function API.

Currently, RunScorecards library:
 i) forces gitcache to be used for every call.
 ii) instantiates a new HTTP and related clients per-repo-per-check.

i)Allow clients to pass in their custom HTTP and GitHub clients.
ii) More efficient use of these clients due to single instantiation.

No (hopefully). Checked cron/main.go locally, but can only be fully confident after a complete cron job run of this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,15 +153,15 @@ test-disk-cache: build-scorecard | $(GINKGO)
 	mkdir cache
 	@echo Focusing on these tests $(FOCUS_DISK_TEST)
 	USE_DISK_CACHE=1 DISK_CACHE_PATH="./cache" \
-				   ./scorecard \
-				   --repo=https://github.com/ossf/scorecard \
-				   --show-details --metadata=openssf  --format json > ./$(OUTPUT)/results.json
+				./scorecard \
+				--repo=https://github.com/ossf/scorecard \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -142,28 +142,6 @@ ci-e2e: build-scorecard check-env e2e-cron | $(GINKGO)
 	@echo Ignoring these test for ci-e2e $(IGNORED_CI_TEST)
 	ginkgo -p  -v -cover --skip=$(IGNORED_CI_TEST)  ./e2e/...
 
-test-disk-cache: ## Runs disk cache tests
```
naveensrinivasan, line 145: So will this mean customers cannot use the DISK cache? Does this use cache?

azeemshaikh38: My main aim in this PR is to simply update the API such that we can pass in a different HTTP client if/when required (I don't have a concrete solution in mind as of yet). Only reason to change the command line tool was because I thought having a cache here might be an overkill. But there are too many failing tests and I do not want to spend time fixing them.

So, I'll keep the cache everywhere for now, but change the API.

oliverchang (approved):

PR #463, opened 2021-05-17
✨  Add worker to the PubSub framework.

azeemshaikh38:

Adds the subscriber/worker logic to the Pubsub cron job setup.

No.

### DIFF
--- a/cron/pubsub/subscriber.go
+++ b/cron/pubsub/subscriber.go
@@ -1,0 +1,84 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Vershandlern 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permisshandlerns and
+// limitathandlerns under the License.
+
+package pubsub
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/ossf/scorecard/cron/data"
+	"gocloud.dev/pubsub"
+	_ "gocloud.dev/pubsub/gcppubsub" // Needed to link in GCP drivers.
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+var ErrorInParse = errors.New("error during protojson.Unmarshal")
+
+type Subscriber interface {
+	SynchronousPull() (*data.ScorecardBatchRequest, error)
+	Ack()
+	Close() error
+}
+
+func CreateSubscriber(ctx context.Context, subscriptionURL string) (Subscriber, error) {
+	subscription, err := pubsub.OpenSubscription(ctx, subscriptionURL)
+	if err != nil {
+		return nil, fmt.Errorf("error during pubsub.OpenSubscription: %w", err)
+	}
+	ret := subscriberImpl{
+		ctx:          ctx,
+		subscription: subscription,
+	}
+	return &ret, nil
+}
+
+type receiver interface {
+	Receive(ctx context.Context) (*pubsub.Message, error)
+	Shutdown(ctx context.Context) error
+}
+
+type subscriberImpl struct {
+	ctx          context.Context
+	subscription receiver
+	msg          *pubsub.Message
+}
+
+func (subscriber *subscriberImpl) SynchronousPull() (*data.ScorecardBatchRequest, error) {
+	msg, err := subscriber.subscription.Receive(subscriber.ctx)
+	if err != nil {
+		fmt.Printf("error during Receive: %v", err)
+		return nil, nil
+	}
+	subscriber.msg = msg
+
+	ret := &data.ScorecardBatchRequest{}
+	if err := protojson.Unmarshal(msg.Body, ret); err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrorInParse, err)
+	}
+	return ret, nil
+}
+
+func (subscriber *subscriberImpl) Ack() {
+	subscriber.msg.Ack()
+}
+
+func (subscriber *subscriberImpl) Close() error {
+	err := subscriber.subscription.Shutdown(subscriber.ctx)
+	if err != nil {
+		return fmt.Errorf("error during subscription.Shutdown: %w", err)
+	}
+	return nil
+}
--- a/cron/worker/main.go
+++ b/cron/worker/main.go
@@ -1,0 +1,155 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Vershandlern 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permisshandlerns and
+// limitathandlerns under the License.
+
+package main
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"sync"
+
+	"github.com/google/go-github/v32/github"
+	"github.com/ossf/scorecard/checks"
+	"github.com/ossf/scorecard/cron/config"
+	"github.com/ossf/scorecard/cron/data"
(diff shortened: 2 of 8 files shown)


cron/pubsub/subscriber.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type subscriberImpl struct {
+	ctx          context.Context
+	subscription receiver
+	msg          *pubsub.Message
+}
+
+func (subscriber *subscriberImpl) SynchronousPull() (*data.ScorecardBatchRequest, error) {
+	msg, err := subscriber.subscription.Receive(subscriber.ctx)
```
oliverchang, line 64: I recently found out that go-cloud does not allow you to extend the acknowledgement deadline (i.e. the "lease period"), and it does not do it automatically in the background AFAICT. 

This means that the maximum deadline is the 10 minutes allowed by Pub/Sub before the message gets redelivered, which is probably OK if each message is for a single repo, but each message here can potentially be for a large batch of repos which may easily exceed it.

Could we split up the request messages into one-per-repo? There are other benefits too, like better granularity when it comes to retrying requests in case of unexpected transient failures. What are the benefits of batching/sharding requests ourselves?

cron/worker/main.go
```diff
@@ -0,0 +1,136 @@
...
+
+	httpClient, githubClient, graphClient := createNetClients(ctx)
+
+	for {
+		req, err := subscriber.SynchronousPull()
+		if errors.Is(err, pubsub.ErrorInReceive) {
+			fmt.Printf("a non-retryable error has occurred during Receive, exiting: %v", err)
+			break
+		} else if err != nil {
+			panic(err)
```
oliverchang, line 125: Why do we need to distinguish between this case and line 122? They ultimately both cause us to exit right?

oliverchang (commented):
nice! Just a high level question for now. will take a closer look tmrw.

cron/worker/main.go
```diff
@@ -0,0 +1,136 @@
...
+	"go.uber.org/zap"
+)
+
+func processRequest(ctx context.Context,
+	batchRequest *data.ScorecardBatchRequest, bucketURL string,
+	httpClient *http.Client, githubClient *github.Client, graphClient *githubv4.Client) error {
+	repoURLs := make([]repos.RepoURL, len(batchRequest.GetRepos()))
+	var resultCh chan repos.RepoResult
+	var buffer bytes.Buffer
+	checksToRun := checks.AllChecks
```
naveensrinivasan, line 43: We would have to ignore checks for branch-protection.

https://github.com/ossf/scorecard/blob/main/cron/main.go#L100

cron/worker/main.go
```diff
@@ -0,0 +1,136 @@
...
+	githubClient *github.Client, graphClient *githubv4.Client) {
+	cfg := zap.NewProductionConfig()
+	cfg.Level.SetLevel(zap.InfoLevel)
+	logger, err := cfg.Build()
+	if err != nil {
+		panic(err)
+	}
+	sugar := logger.Sugar()
+	// Use our custom roundtripper
+	rt := roundtripper.NewTransport(ctx, sugar)
```
naveensrinivasan, line 91: I am guessing the `ENV` variables would be set in the deployment to the use the `HTTPCache`
```
 BLOB_URL:           gs://ossf-scorecards-cache
 USE_BLOB_CACHE:     true
 ```

**Review by naveensrinivasan — APPROVED** [review]

File: cron/worker/main.go (hunk 4 above)
Comment (azeemshaikh38) [reply cron/worker/main.go:91]: Yes, the deployment of worker should be same/similar to cron job today. Let me know if anything is needed in code to ensure these vars are set.

File: cron/worker/main.go (hunk 3 above)
Comment (azeemshaikh38) [reply cron/worker/main.go:43]: Done, thanks.

File: cron/pubsub/subscriber.go (hunk 1 above)
Comment (azeemshaikh38) [reply cron/pubsub/subscriber.go:64]: > I recently found out that go-cloud does not allow you to extend the acknowledgement deadline (i.e. the "lease period"), and it does not do it automatically in the background AFAICT.
> 
> This means that the maximum deadline is the 10 minutes allowed by Pub/Sub before the message gets redelivered, which is probably OK if each message is for a single repo, but each message here can potentially be for a large batch of repos which may easily exceed it.
> 

Ah, that's not good. We should have the automatic lease management irrespective of which solution we choose. I'm considering re-writing `subscriber.go` using GCS libraries instead of go-cloud. Wdyt?

> Could we split up the request messages into one-per-repo? There are other benefits too, like better granularity when it comes to retrying requests in case of unexpected transient failures. What are the benefits of batching/sharding requests ourselves?

The benefit is determinism. e.g - (i) `controller` knows `num_shards` beforehand which is useful in figuring out job completion status, (ii) the generated files can have filenames which read `shard-xxx` making them easily listable/queryable. 
The other thing to note here is that, with the current implementation, we can set `shard_size = 1` in the config to achieve what you are looking for without any change to code/implementation.

File: cron/worker/main.go (hunk 2 above)
Comment (azeemshaikh38) [reply cron/worker/main.go:125]: This line is a error in our message handling (e.g: JSON parsing error). It means there is something wrong in our code, so we explicitly `panic`.

An error during Receive (line 122) is from the PubSub driver itself. It could be a harmless error (e.g: no messages to poll) and will exit without `panic`.

File: cron/worker/main.go (hunk 4 above)
Comment (naveensrinivasan) [reply cron/worker/main.go:98]: How about validating if the ENV var is set and fail fast if it is not?

Comment (azeemshaikh38) [reply cron/worker/main.go:98]: Isn't that validation already happening during the `roundtripper.NewTransport()` call?

Comment (naveensrinivasan) [reply cron/worker/main.go:98]: No, It is optional. But in the pubsub worker, we need it.

File: cron/pubsub/subscriber.go [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (subscriber *subscriberImpl) Ack() {
+	subscriber.msg.Ack()
+}
+
+func (subscriber *subscriberImpl) Close() error {
+	err := subscriber.subscription.Shutdown(subscriber.ctx)
+	if err != nil {
+		return fmt.Errorf("%w: %v", ErrorInShutdown, err)
```
oliverchang, line 84: Nit: this seems like a misuse of `Errorf`. The idea behind error wrapping is that you should be able to access a chain of underlying errors. This short circuits it to the one you've defined here (ErrorInShutdown).

To do what you want here, you probably instead want something like:

```go
type Error struct {
  Type string
  Err error
}

func (e *Error) Error() string{
  return fmt.Sprintf("%s: %v", e.Type, e.Err)
}

func (e *Error) Unwrap() error {
  return e.Err
}
```

Then construct/use them by doing:

```go
return &Error{type: "Shutdown", Err: error}

var pubsubErr pubsub.Error
if errors.As(err, &pubsubErr) && pubsubErr..... 
...
```

cron/worker/main.go
azeemshaikh38: Done. Added it in the `main` fn.

cron/pubsub/subscriber.go
azeemshaikh38: I see. I was trying to test this in unit_test. I don't feel compelled complicating this code for the sake of a test. I will stick with the the simpler solution here.

cron/pubsub/subscriber.go
azeemshaikh38: Agreed. `shard_size = 250` is just something I chose for now. We probably need to do some tuning to figure out what works best for us.

cron/pubsub/subscriber.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/ossf/scorecard/cron/data"
+	"gocloud.dev/pubsub"
+	_ "gocloud.dev/pubsub/gcppubsub" // Needed to link in GCP drivers.
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+var (
+	ErrorInReceive  = errors.New("error in Receive")
+	ErrorInParse    = errors.New("error during protojson.Unmarshal")
+	ErrorInShutdown = errors.New("error during shitdown")
```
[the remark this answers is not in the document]
azeemshaikh38: Lol. Fixed.

cron/worker/main.go
azeemshaikh38: > I see, thanks. I think this does make things a bit complicated though, because it puts the burden on the caller to know about this detail (and it's not really an error).
> 
> Can we just return `req == nil` and `err == nil` in this case and handle it here accordingly? 

Done.

>Also should we exit here or just sleep for a bit and retry later?

Quitting the program here is what gocloud recommends. But considering we'll be updating from go-cloud to GCS, this may not hold true anymore. 
 
image

oliverchang: Right, we could have also recreated the subscription instead of exiting, but yes no longer necessary for the google cloud libraries.

oliverchang (approved):
LGTM

We migrate the Pub/Sub client library in another PR, if you wish.

azeemshaikh38:
> LGTM
> 
> We migrate the Pub/Sub client library in another PR, if you wish.

Thanks. I think I will attempt the library change in another PR. Would be good to have this PR submitted for now.

PR #469, opened 2021-05-18
:sparkles: Add check for Docker dependency pinning by hash

laurentsimon:

feature

We have no checks for docker dependency hash-pinning

We have a check for docker dependency hash-pinning

no

* **Other information**:


checks/checkforcontent.go
```diff
@@ -78,10 +87,18 @@ func CheckFilesContent(checkName, shellPathFnPattern string, c *checker.CheckReq
...
-		name := names[1]
-		// Filter out files based on path.
-		if match, _ := path.Match(shellPathFnPattern, name); !match {
-			continue
+		fullpath := names[1]
+		filename := path.Base(fullpath)
+
+		if !caseSensitive {
+			fullpath = strings.ToLower(fullpath)
+			filename = strings.ToLower(path.Base(fullpath))
```
inferno-chromium, line 95: maybe move this line before fullpath is lowered in case, before line 94.

filename = strings.ToLower(path.Base(fullpath))
fullpath = strings.ToLower(fullpath)

checks/checkforcontent.go
```diff
@@ -28,7 +28,12 @@ import (
 
 // CheckFilesContent downloads the tar of the repository and calls the onFileContent() function
 // shellPathFnPattern is used for https://golang.org/pkg/path/#Match
-func CheckFilesContent(checkName, shellPathFnPattern string, c *checker.CheckRequest,
+// Warning: the pattern is used to match (1) the entire path AND (2) the filename alone. This means:
+// 	- To scope the search to a directory, use "./dirname/*". Example, for the root directory,
+// 		use "./*".
+//	- A pattern such as "*mypatern*" will match files containing mypattern in *any* directory.
+// The check is case-insensitive.
```
inferno-chromium, line 35: I think it is dependent on caseSensitive, not always case-insensitive.

checks/testdata/Dockerfile-pinned-as
```diff
@@ -0,0 +1,26 @@
...
+
+FrOM base As build
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build-scorecard
+
+from build
+RUN /hello-world
+
+FROM python@sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2
```
inferno-chromium, line 26: missing AS

checks/testdata/Dockerfile-not-pinned-as
```diff
@@ -0,0 +1,26 @@
...
+
+FROM python:3.7 as build
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build-scorecard
+
+FROM build
+RUN /hello-world
+
+FROM python@sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2
```
inferno-chromium, line 26: remove sha, test should be failing.

checks/frozen_deps.go
```diff
@@ -35,19 +36,115 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+	asRegex := regexp.MustCompile(`^(?i)FROM\s+(.*)\s+AS\s+(.*)`)
+	regex := regexp.MustCompile(`^(?i)FROM\s+(.*)`)
+	hashAsRegex := regexp.MustCompile(`^(?i)FROM\s+.*@sha256:[a-f\d]{64}\s+AS\s+(.*)`)
+	hashRegex := regexp.MustCompile(`^(?i)FROM\s+.*@sha256:[a-f\d]{64}`)
+
+	// Read the file line by line.
+	scanner.Split(bufio.ScanLines)
+	r := true
+	nl := 0
+	var al []string
```
inferno-chromium, line 79: nit: instead of al, a more meaningful array name like pinnedImages ?

checks/frozen_deps.go
```diff
@@ -35,19 +36,115 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+	scanner := bufio.NewScanner(strings.NewReader(string(content)))
+	asRegex := regexp.MustCompile(`^(?i)FROM\s+(.*)\s+AS\s+(.*)`)
+	regex := regexp.MustCompile(`^(?i)FROM\s+(.*)`)
+	hashAsRegex := regexp.MustCompile(`^(?i)FROM\s+.*@sha256:[a-f\d]{64}\s+AS\s+(.*)`)
+	hashRegex := regexp.MustCompile(`^(?i)FROM\s+.*@sha256:[a-f\d]{64}`)
+
+	// Read the file line by line.
+	scanner.Split(bufio.ScanLines)
+	r := true
+	nl := 0
```
inferno-chromium, line 78: nit: a more meaningful name, e.g. fromLines or importLines

inferno-chromium (approved):
run `make all` to see lint failures

checks/checkforcontent.go
laurentsimon: I can, but what's the reasoning?
I think we could change it to:
```
fullpath = strings.ToLower(fullpath)			
filename = path.Base(fullpath)
```
to avoid having ToLower iterate over the string again.

checks/testdata/Dockerfile-not-pinned-as
laurentsimon: right. My thinking was there is at least one line in the file that should fail, not necessarily all of them. Shall I change them all or keep it as is?

inferno-chromium: My bad, didnt see the other lines.

checks/checkforcontent.go
inferno-chromium: yes skip, i thought path.Base was dynamic, but it just gets a section of path.

checks/frozen_deps_test.go
```diff
@@ -91,3 +91,87 @@ func TestGithubWorkflowPinning(t *testing.T) {
...
+			var err error
+
+			content, err = ioutil.ReadFile(tt.args.Filename)
+			if err != nil {
+				panic(fmt.Errorf("cannot read file: %w", err))
+			}
+
+			r, err := validateDockerfile(tt.args.Filename, content, tt.args.Logf)
+
+			if (err != nil && tt.want.Error == nil) ||
```
azeemshaikh38, line 169: You could simplify error checking logic by using `errors.Is`, i.e:

```
if !errors.Is(err, tt.want.Error) && r != tt.want.Result {
  t.Errorf()
}
```

checks/frozen_deps.go
```diff
@@ -35,19 +36,115 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
 	return checker.MultiCheckAnd(
 		isPackageManagerLockFilePresent,
 		isGitHubActionsWorkflowPinned,
+		isDockerfilePinned,
```
azeemshaikh38, line 39: Should we consider a using an existing Dockerfile parser like this one - github.com/moby/buildkit/frontend/dockerfile/parser?

checks/frozen_deps.go
```diff
@@ -35,19 +36,115 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+		// where name refers to a pinned image
+		re = regex.FindStringSubmatch(line)
+		if len(re) == 2 && isPresent(pinnedAsNames, re[1]) {
+			continue
+		}
+
+		// FROM name@sha256:hash
+		if !hashRegex.Match([]byte(line)) {
+			r = false
+			logf("!! frozen-deps - %v has non-pinned dependency '%v'", path, line)
```
azeemshaikh38, line 128: Here and other place, can we do early error return instead of logging the failures?

checks/frozen_deps.go
```diff
@@ -35,19 +36,115 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+func isPresent(slice []string, val string) bool {
+	for _, item := range slice {
+		if item == val {
+			return true
+		}
+	}
+	return false
+}
+
+func validateDockerfile(path string, content []byte,
```
naveensrinivasan, line 61: We should be using the docker parser.

checks/frozen_deps.go
naveensrinivasan: +1 We should be using docker parser!

naveensrinivasan (changes requested):
The reason not approving is only because we Dockerfile parser could be used. Writing a custom dockerfile parser is error-prone.

@azeemshaikh38 @oliverchang Thoughts?

checks/checkforcontent.go
```diff
@@ -78,10 +86,18 @@ func CheckFilesContent(checkName, shellPathFnPattern string, c *checker.CheckReq
...
-			continue
+		fullpath := names[1]
+		filename := path.Base(fullpath)
+
+		if !caseSensitive {
+			fullpath = strings.ToLower(fullpath)
+			filename = strings.ToLower(path.Base(fullpath))
+		}
+		// Filter out files based on path/names.
+		if match, _ := path.Match(shellPathFnPattern, fullpath); !match {
```
azeemshaikh38, line 97: Since, we only need the caseSensitive bool here for matching, let's do:

```
match := IsMatchingPath(pattern, path, filename, caseSensitive)
```

That way, the code is not littered with `if caseSensitive` and also the variable values for `shellPathFnPattern`, `fullpath` and `filename` remain the same inside this fn, irrespective of `caseSensitive`.

laurentsimon:
note: `make all` succeeds locally. The failing runs indicates ` The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive.` I've installed `revive` and disabled gofmt for now.

checks/frozen_deps_test.go
laurentsimon: I initially did not do that because the error was created on the fly `return errors.New("file has no FROM keyword")` in `frozen_deps.go`. My intention was to clean this up once we handle errors properly.  
For now, I've created a new error `ErrInvalidDockerfile` at the top of `frozen_deps.go` and changed the unit test as you suggested. 
Long-term, are we going to have all errors in one file? Or at the top of individual files?

checks/frozen_deps.go
laurentsimon: probably a good idea. We still need to use the regex though, so the docker parsing lib will only give us line arguments. Probably will give us peace of mind :-)
I'll make the changes

checks/frozen_deps.go
laurentsimon: I did it this way so we can log all failures. wdut?

checks/checkforcontent.go
laurentsimon: you mean you want a new function called `IsMatchingPath`. Is that correct?

azeemshaikh38: Correct. Basically, that should (hopefully) be the only function whose logic depends on caseSensitive.

checks/frozen_deps.go
azeemshaikh38: I do think we should get details of all failures, but it probably requires a decent amount code re-write. Ideally, we should have a way to return a list of errors and we can pile all of these errors onto it. 

But that might be too much to attempt in this PR, so suggested simply early returning for now.

checks/frozen_deps_test.go
azeemshaikh38: > I initially did not do that because the error was created on the fly `return errors.New("file has no FROM keyword")` in `frozen_deps.go`. My intention was to clean this up once we handle errors properly.
> For now, I will create a new error `ErrInvalidDockerfile` at the top of `frozen_deps.go` and changed the unit test as you suggested.

You didn't need a new error here since you are not comparing error types using `errors.As` but instead the errors themselves using `errors.Is`. But if you already have defined a new error, thats great too.

> Long-term, are we going to have all errors in one file? Or at the top of individual files?

I think we should have them in a single file and not at top of individual files. If you are asking for this PR, its ok to keep it in the same file. We can move all errors out to a single file in a later PR.

laurentsimon: > > I initially did not do that because the error was created on the fly `return errors.New("file has no FROM keyword")` in `frozen_deps.go`. My intention was to clean this up once we handle errors properly.
> > For now, I will create a new error `ErrInvalidDockerfile` at the top of `frozen_deps.go` and changed the unit test as you suggested.
> 
> You didn't need a new error here since you are not comparing error types using `errors.As` but instead the errors themselves using `errors.Is`. 
Actually I tried that and it did not work when the error was created on the fly using `return errors.New("...")`

But if you already have defined a new error, thats great too.
> 
> > Long-term, are we going to have all errors in one file? Or at the top of individual files?
> 
> I think we should have them in a single file and not at top of individual files. If you are asking for this PR, its ok to keep it in the same file. We can move all errors out to a single file in a later PR.
sg

checks/frozen_deps.go
```diff
@@ -35,19 +42,113 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+	// Users may use various names, e.g.,
+	// Dockerfile.aarch64, Dockerfile.template, Dockerfile_template, dockerfile, Dockerfile-name.template
+	// Templates may trigger false positives, e.g. FROM { NAME }.
+
+	// We have what looks like a docker file.
+	// Let's interpret the content as utf8-encoded strings.
+	contentReader := strings.NewReader(string(content))
+	regex := regexp.MustCompile(`.*@sha256:[a-f\d]{64}`)
+
+	r := true
```
oliverchang, line 78: nit: `r` is used quite far away from where it is defined, so it's more helpful to rename this something more meaningful, like `pinned`.

checks/frozen_deps.go
```diff
@@ -20,12 +20,19 @@ import (
...
 
+	"github.com/moby/buildkit/frontend/dockerfile/parser"
 	"github.com/ossf/scorecard/checker"
 	"gopkg.in/yaml.v2"
 )
 
 const frozenDepsStr = "Frozen-Deps"
 
+// ErrFrozenDepsInvalidDockerfile : Invalid docker file.
+var ErrFrozenDepsInvalidDockerfile = errors.New("invalid docker file")
```
oliverchang, line 31: nit: Avoid "stutter" by just naming these `ErrInvalidDockerfile` etc

checks/frozen_deps.go
```diff
@@ -35,19 +42,113 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+	// Templates may trigger false positives, e.g. FROM { NAME }.
+
+	// We have what looks like a docker file.
+	// Let's interpret the content as utf8-encoded strings.
+	contentReader := strings.NewReader(string(content))
+	regex := regexp.MustCompile(`.*@sha256:[a-f\d]{64}`)
+
+	r := true
+	fromFound := false
+	var pinnedAsNames []string
```
oliverchang, line 80: Just use a map[string]bool here to represent a set of pinned names. We can then simplify `isPresent` and not needing to call `isPresent` before we append etc

checks/frozen_deps.go
```diff
@@ -35,19 +42,113 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+		} else {
+			// That should not happen.
+			return false, ErrFrozenDepsInvalidDockerfile
+		}
+
+	}
+
+	// The file should have at least one FROM statement.
+	if !fromFound {
+		logf("end")
```
oliverchang: What is this needed for?

oliverchang (approved):
Very nice! LGTM with some minor suggestions

laurentsimon (commented):
> lint

I have and there's no error locally. Please advise.

azeemshaikh38 (approved):

oliverchang:
> > lint
> 
> I have and there's no error locally. Please advise.

Something about merging your PR locally seems to be confusing `golangci-lint run -n`.

I can instead get the same output as the CI if I run:

`golangci-lint run --new-from-rev 8c2432bd6252bba8f357bb130d546380ec5f6957`

naveensrinivasan (approved):

naveensrinivasan:
> > > lint
> > 
> > 
> > I have and there's no error locally. Please advise.
> 
> Something about merging your PR locally seems to be confusing `golangci-lint run -n`.
> 
> I can instead get the same output as the CI if I run:
> 
> `golangci-lint run --new-from-rev 8c2432bd6252bba8f357bb130d546380ec5f6957`

I use vim as my editor with https://github.com/dense-analysis/ale and golang-lint is automatically detected and informs about the new issues in the code. I don't run anything manually for checking. 

image

laurentsimon:
thanks. I've installed it in vscode as well. Maybe something we could indicate in the contributing doc.

naveensrinivasan:
> thanks. I've installed it in vscode as well. Maybe something we could indicate in the contributing doc.

Absolutely, I can do that for vim. If you can do that for vscode it would help.

PR #472, opened 2021-05-19
🌱  Add Dockerfiles for PubSub batch job

azeemshaikh38:

Adds Dockerfiles for PubSub controller/worker. No behaviour change.

No.

### DIFF
--- a/cron/controller/Dockerfile
+++ b/cron/controller/Dockerfile
@@ -1,0 +1,32 @@
+# syntax = docker/dockerfile:1-experimental
+# Copyright 2020 Security Scorecard Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+FROM  golang:1.16.4@sha256:6f0b0a314b158ff6caf8f12d7f6f3a966500ec6afb533e986eca7375e2f7560f AS base
+WORKDIR /src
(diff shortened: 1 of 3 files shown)


laurentsimon (approved):

cron/controller/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+COPY go.* ./
+RUN go mod download
+COPY . ./
+
+FROM base AS pubsub
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build-pubsub
+
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:slim@sha256:acb34ca33bc83255110d16a749b0da618a96fcc481348ecfe614b3edcabdeaa6
```
naveensrinivasan, line 29: Why do we need a `cloud-sdk:slim`? These are `go binaries` which does not use any of the cloudsdk tools like the command line. The previous cron used `gcloud command to upload to gcs`.

cron/worker/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY go.* ./
+RUN go mod download
+COPY . ./
+
+FROM base AS pubsub
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build-pubsub
+
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:slim@sha256:acb34ca33bc83255110d16a749b0da618a96fcc481348ecfe614b3edcabdeaa6
```
naveensrinivasan, line 29: The same comment applies as above.

cron/controller/Dockerfile
azeemshaikh38: Ah. I just copied this as is from `cron/Dockerfile`. So does this only need a single `FROM` which would be the `golang:1.16.4@sha` import?

naveensrinivasan: It has to be similar to this one https://github.com/ossf/scorecard/blob/main/Dockerfile#L28. It is from distroless container.

azeemshaikh38: Thanks! Working on a PR, will send it out shortly.

PR #474, opened 2021-05-19
🌱 Disabling failing linters

chrismcgehee:

CI infrastructure improvement

New PRs are introducing lint errors even though `make all` runs `golangci-lint run -n`. I believe this is because `golangci-lint run -n` only will lint run the linter on the most recent commit. If a PR has multiple commits, the older ones will not get linted. Also if `go mod tidy` or another command makes changes to the source code, `golangci-lint` will run against those changes and not the ones in the PR.
Issue #173 

This PR disables failing linters and causes `make all` to fail on any error, not just newly introduced ones. This will prevent new lint errors from entering in the majority of the lint checks. When I fix all the errors from a single linter, I will re-enable that linter in `.golangci.yml`.

No

* **Other information**:


Makefile
```diff
@@ -60,7 +60,7 @@ $(GOLANGGCI_LINT): install
 check-linter: ## Install and run golang linter
 check-linter: $(GOLANGGCI_LINT)
 	# Run golangci-lint linter
-	golangci-lint run -n
+	golangci-lint run
```
azeemshaikh38, line 63: Explicitly specify `-c .golangci.yml` here?

azeemshaikh38 (approved):

Makefile
chrismcgehee: Sure, that makes it obvious to the person reading this code that we're using that config file.

naveensrinivasan (approved):
Thanks!

PR #475, opened 2021-05-20
🐛  Embed data files into the go lib files

azeemshaikh38:

Embeds the data files (config.yaml and projects.csv) directly into the libraries which need and use it. 

The PubSub data files were specified with relative path, implying that they would be read relative to the dir where `go run` was called. This is very brittle and can break easily.

The data files are embedded directly into the libraries that use them. There is also the option exposed to users to inject their own files if need be.

No.


cron/config/config_test.go
```diff
@@ -29,6 +31,23 @@ const (
 	prodShardSize    int    = 250
 )
 
+func getByteValueFromFile(filename string) ([]byte, error) {
```
oliverchang, line 34: Why not `ioutil.ReadFile()` ?

oliverchang (approved):

cron/config/config_test.go
azeemshaikh38: Nice, thanks. Still kept this fn though to check for the case where filename is empty.

naveensrinivasan:
The idea of tests files is to make the data swappable at runtime. 

What was the reason for embedding?

azeemshaikh38:
> The idea of tests files is to make the data swappable at runtime.
> 
> What was the reason for embedding?

Reason for embedding is so that the data filepaths are not dependent on where the `go run` command is executed from.

I have refactored such that data is still swappable during tests to as much extent as possible.

Issue #366, opened 2021-04-27
Make BQ import more reliable


oliverchang:
BQ imports should load as WRITE_TRUNCATE into a date partition, to prevent duplicates in case a cron runs twice in the same day. 

This cannot be easily done today with the single latest.json file, as entries can span multiple days (since the run can overlap between days).

For consistency of results within a partition, the "Date" should be consistent for all repo results for jobs started on the same day.

oliverchang:
@azeemshaikh38 might something to consider in the design for the split workers / GCS output.

azeemshaikh38:
The date consistency problem should be easy to handle since the master (which publishes to PubSub topic) can provide the date to be used for all results.

However, the duplicate entries due to the BQ data transfer running twice might be tricky. BQ should delete source files after transfer, however it does not provide strong guarantees here. If the deletion does not happen and BQ runs again before the data is deleted, there is risk of duplication.

oliverchang:
I think we can handle duplicate entries well if we load results by day with WRITE_TRUNCATE into a bigquery partition per https://cloud.google.com/bigquery/docs/creating-column-partitions. 

If all input files have the same date, we can load and replace existing data into table$YYYYMMDD. The input files on GCS would just need to be structured in such a way to support this.

azeemshaikh38:
I'm skeptical of WRITE_TRUNCATE. (1) it is wasteful (2) only some GCS files maybe deleted and looks like WRITE_TRUNCATE overwrites the entire partition (3) if it only overwrites selected rows, that would be best-effort de-duping which only mitigates the problem but doesn't solve it.

One way would be - if BQ sends out notifications on delete failure, we can have a garbage collection job running which will process these notifications and retry deletes.

oliverchang:
I think any additional cloud cost that incurs will be minimal, especially if it's avoiding maintenance costs. We can also just rely on GCS lifecycle policies (rather than a job we have to maintain). 

One potential way to do this might look something like:
```
gs://bucket/YYYYMMDD/unique_run_id_0/*.json
gs://bucket/YYYYMMDD/unique_run_id_1/*.json
...
```

and the load job can load WRITE_TRUNCATE into YYYYMMDD partitions from a single run. This way every load is idempotent for the same unique_run_id, which further helps if a load job fails or needs to be retried, in addition to being resilient to handling multiple runs for the same day.

oliverchang:
[...]

PR #476, opened 2021-05-20
🌱  Use TRUNCATE to load data into BigQuery

azeemshaikh38:

Uses TRUNCATE instead of APPEND to load GCS data into BQ.

Currently, the BQ scheduled data transfer service uses APPEND to load data into BQ table. But, we want the TRUNCATE behaviour for reasons discussed here: https://github.com/ossf/scorecard/issues/366. However, this can't be done for 2 reasons:
(i) the scheduled service does not allow for this option
(ii) TRUNCATE requires that all input data has the same `date` field

This PR implements that behaviour.

No.

### DIFF
--- a/cron/bq/transfer.go
+++ b/cron/bq/transfer.go
@@ -1,0 +1,57 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bq
+
+import (
+	"context"
+	"fmt"
+	"log"
+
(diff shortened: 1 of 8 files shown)


cron/bq/transfer.go
```diff
@@ -0,0 +1,57 @@
...
+	bq, err := bigquery.NewClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("failed to create bigquery client: %w", err)
+	}
+
+	datasetName, err := config.GetBigQueryDataset()
+	if err != nil {
+		return fmt.Errorf("error getting BigQuery dataset: %w", err)
+	}
+	tableName, err := config.GetBigQueryTable()
```
oliverchang, line 40: I think you might need a partition decorator as well: https://cloud.google.com/bigquery/docs/managing-partitioned-table-data#using_a_load_job ?

oliverchang (commented):
nice!

cron/bq/transfer.go
azeemshaikh38: Hmm, I thought it should pick the partition up using `date` column. I'm flag guarding the load job for now - it'll help with the failing tests and also allow me to manually test `bq load` before enabling it in prod.

oliverchang: Yes, please test it. I'm not sure the load job is smart enough to do this without specifying the date partition (and the documentation doesn't seem to suggest this either).

oliverchang (approved):

PR #480, opened 2021-05-20
✨  Add monitoring to checks

azeemshaikh38:

Adds monitoring support to daily cron job.

No monitoring.

Logs per-repo per-check runtime statistics to Google Cloud StackDriver.

No.

### DIFF
--- a/cron/main.go
+++ b/cron/main.go
@@ -25,20 +25,38 @@ import (
 	"strconv"
 	"time"
 
+	"contrib.go.opencensus.io/exporter/stackdriver"
 	"github.com/google/go-github/v32/github"
 	"github.com/jszwec/csvutil"
 	"github.com/ossf/scorecard/checks"
 	"github.com/ossf/scorecard/cron/bq"
+	"github.com/ossf/scorecard/cron/monitoring"
 	"github.com/ossf/scorecard/pkg"
 	"github.com/ossf/scorecard/repos"
 	"github.com/ossf/scorecard/roundtripper"
+	"github.com/ossf/scorecard/stats"
 	"github.com/shurcooL/githubv4"
+	"go.opencensus.io/stats/view"
 	"go.uber.org/zap"
 )
 
 type Repository struct {
 	Repo     string `csv:"repo"`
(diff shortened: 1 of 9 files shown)


cron/main.go
```diff
@@ -84,6 +87,18 @@ func main() {
 	githubClient := github.NewClient(httpClient)
 	graphClient := githubv4.NewClient(httpClient)
 
+	exporter, err := monitoring.NewStackDriverExporter()
```
oliverchang, line 90: rather than inlining this, would be good to have a separate function for this. Something like `startMetricsExporter()` which creates, starts, registers all views, and returns the exporter. Since we plan to add more views it'll get a bit messy here.

oliverchang (approved):
nice! I assume that without any exporters set up, these tag.New/opencensusstats.Record are no-ops?

azeemshaikh38:
> nice! I assume that without any exporters set up, these tag.New/opencensusstats.Record are no-ops?

Hmm, how do mean no exporters are setup? I'm hoping with this PR we start logging these metrics. Am I missing something?

oliverchang:
> > nice! I assume that without any exporters set up, these tag.New/opencensusstats.Record are no-ops?
> 
> Hmm, how do mean no exporters are not setup? I'm hoping with this PR we start logging these metrics. Am I missing something?

If a CLI or library user uses this, where no exporters are set up, all these metrics logging calls are no-ops and won't cause errors right?

cron/main.go
azeemshaikh38: Done.

azeemshaikh38:
> > > nice! I assume that without any exporters set up, these tag.New/opencensusstats.Record are no-ops?
> > 
> > 
> > Hmm, how do mean no exporters are not setup? I'm hoping with this PR we start logging these metrics. Am I missing something?
> 
> If a CLI or library user uses this, where no exporters are set up, all these metrics logging calls are no-ops and won't cause errors right?

Ah I see what you mean. Yeah, these shouldn't cause errors if there is no exporter registered. Tested it locally.

PR #481, opened 2021-05-21
✨  [Frozen-deps]: Ignore testdata/ files

laurentsimon:

bug fix

we don't ignore test files.

we ignore test files.

no

* **Other information**:

### DIFF
--- a/checks/frozen_deps.go
+++ b/checks/frozen_deps.go
@@ -17,6 +17,7 @@ package checks
 import (
 	"errors"
 	"fmt"
+	"path"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


naveensrinivasan (approved):

checks/frozen_deps.go
```diff
@@ -55,12 +56,36 @@ func isDockerfilePinned(c *checker.CheckRequest) checker.CheckResult {
...
 }
 
+func isTestFile(pathfn string) bool {
+	d := path.Dir(pathfn)
+	if d == "testdata" {
+		return true
+	}
+
+	for d != "." {
+		r, err := path.Match("*/testdata", d)
```
azeemshaikh38, line 66: Should we be more generic here and say `path.Match("*test*")`?

checks/frozen_deps.go
```diff
@@ -55,12 +56,36 @@ func isDockerfilePinned(c *checker.CheckRequest) checker.CheckResult {
 	return CheckFilesContent(frozenDepsStr, "*Dockerfile*", false, c, validateDockerfile)
 }
 
+func isTestFile(pathfn string) bool {
```
azeemshaikh38, line 59: Maybe simplify fn?

```
func isTestFile(pathfn string) bool {
	for d := path.Dir(pathfn); d != "."; d = path.Dir(d) {
		match, err := path.Match("*test*", d)
		if err != nil {
			return false
		}
		if match {
			return true
		}
	}
	return false
}
```

azeemshaikh38 (approved):
LGTM. Need to update all instances of `path` variable with something else. Otherwise import `pathlib "path"`.

checks/frozen_deps.go
laurentsimon: I'm worried this is too broad and we'll have false negative. I was not convinced we should do this in the first place to be honest, but I wrote this as a quick fix to the failing `e2e` tests to unblock us. However: what if a repo has genuine docker files under a test folder, and the file is actually used for testing during CI/CD? We really don't want to miss these.
So I decided to remove the check here, and move it to `checkforcontent.go`:

1. Check the repo is scorecard repo
2. Check a folder is `testdata`

This way we only filter out our own repo's test files.

checks/frozen_deps.go
laurentsimon: see new code.

checks/frozen_deps.go
azeemshaikh38: Sounds good. This is a good example of what we were discussing yesterday - we need to find a way where clients can run Scorecards using a config. That allows clients to specify repo-specific files that Scorecard should ignore. Just something to keep in mind for future design discussions.

PR #485, opened 2021-05-22
🌱Fix lint issues: gochecknoinits linter

chrismcgehee:

Code quality improvement

This PR fixes issues from the [gochecknoinits](https://github.com/leighmcculloch/gochecknoinits/) linter (checks that init functions are not used). As discussed in #173, we're going to leave the `init` calls as opposed to refactoring how checks are registered.

n/a

No

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -267,7 +267,7 @@ func fetchGitRepositoryFromRubyGems(packageName string) (string, error) {
 	return v.SourceCodeURI, nil
 }
 
-func init() {
+func InitRootCmd() {
 	// Add the zap flag manually
 	rootCmd.PersistentFlags().AddGoFlagSet(goflag.CommandLine)
 	rootCmd.Flags().Var(&repo, "repo", "repository to check")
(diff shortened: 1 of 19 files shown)


cmd/root.go
```diff
@@ -267,7 +267,7 @@ func fetchGitRepositoryFromRubyGems(packageName string) (string, error) {
 	return v.SourceCodeURI, nil
 }
 
-func init() {
+func InitRootCmd() {
```
azeemshaikh38, line 270: Let's add `//nolint` here too and not change code logic.

chrismcgehee: In that case, should we just remove this linter from .golangci.yml?

naveensrinivasan: Going forward we want to avoid using `init` and that is the reason to have this linter. So we should not be removing from the .golangci.yml

chrismcgehee: Changed these to `//nolint`.

naveensrinivasan (approved):

cmd/root.go
naveensrinivasan: Thank you!

PR #490, opened 2021-05-22
🐛  Fix bug in GitHub token access

azeemshaikh38:

Remove the oauth library usage which was causing issues in token access. 
The `oauth.Token()` fn relies on [`Token.Expiry` field](https://github.com/golang/oauth2/blob/f6687ab2804c/token.go#L50). Without this, it always [returns the same token](https://github.com/golang/oauth2/blob/f6687ab2804c/oauth2.go#L301). This was unintentionally working fine before https://github.com/ossf/scorecard/pull/461 since we always created a new instance of the oauth client. 
For our usecase, it is enough if we set the correct `Authorization` header. So updating to use that instead of the oauth library.

FYI, this issue is causing the cron job to be stuck forever.

No.

### DIFF
--- a/roundtripper/github.go
+++ b/roundtripper/github.go
@@ -15,35 +15,36 @@
 package roundtripper
 
 import (
-	"context"
 	"fmt"
 	"net/http"
+	"sync"
 	"sync/atomic"
-
-	"golang.org/x/oauth2"
 )
 
-func MakeOAuthTransport(ctx context.Context, accessTokens []string) http.RoundTripper {
-	tokenSource := &tokenPool{
-		tokens: makeTokenAccessor(accessTokens),
+func MakeGitHubTransport(innerTransport http.RoundTripper, accessTokens []string) http.RoundTripper {
+	return &githubTransport{
+		innerTransport: innerTransport,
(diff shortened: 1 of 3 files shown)


roundtripper/github.go
```diff
@@ -55,14 +56,15 @@ func makeTokenAccessor(accessTokens []string) tokenAccessor {
 type roundRobinAccessor struct {
 	accessTokens []string
 	counter      int64
+	mu           sync.Mutex
 }
 
-func (roundRobin *roundRobinAccessor) next() (*oauth2.Token, error) {
+func (roundRobin *roundRobinAccessor) next() string {
+	roundRobin.mu.Lock()
+	defer roundRobin.mu.Unlock()
```
naveensrinivasan, line 64: Why do we need a lock?

naveensrinivasan (approved):

roundtripper/github.go
azeemshaikh38: Without lock, the fn is not doing what we expect. E.g: thread1 increments `counter`, then context switches and thread2 also increments `counter`. Then both the threads access the same index through `index := c % int64(l)`.

Although not having a lock won't cause any issues, its not technically round robin.

PR #494, opened 2021-05-23
✨ Add k8s file for PubSub jobs

azeemshaikh38:

Adds k8s file for PubSub jobs - controller & worker.

No.

### DIFF
--- a/cron/k8s/worker.yaml
+++ b/cron/k8s/worker.yaml
@@ -1,0 +1,49 @@
+# Copyright 2021 Security Scorecard Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: scorecard-batch-worker
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: worker
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: worker
+    spec:
+      containers:
+      - name: worker
+        image: gcr.io/openssf/scorecard-batch-worker
+        imagePullPolicy: Always
+        env:
+        - name: GITHUB_AUTH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: github
+              key: testToken
+        - name: USE_BLOB_CACHE
+          value: "true"
+        - name: BLOB_URL
+          value: "gs://ossf-scorecards-cache"
+        resources:
+          requests:
+            memory: 10Gi
+            cpu: 1
+          limits:
+            memory: 15Gi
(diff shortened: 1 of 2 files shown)


cron/k8s/worker.yaml
```diff
@@ -0,0 +1,49 @@
...
+          value: "true"
+        - name: BLOB_URL
+          value: "gs://ossf-scorecards-cache"
+        resources:
+          requests:
+            memory: 10Gi
+            cpu: 100
+          limits:
+            memory: 15Gi
+            cpu: 220
```
oliverchang, line 49: that's quite beefy.. any reason why we need 220 CPUs machines, rather than more (smaller) machines? this workload seems like something that inherently better suits horizontal scaling instead.

oliverchang (approved):

cron/k8s/worker.yaml
```diff
@@ -0,0 +1,49 @@
...
+    metadata:
+      labels:
+        app.kubernetes.io/name: worker
+    spec:
+      containers:
+      - name: worker
+        image: gcr.io/openssf/scorecard-batch-worker
+        imagePullPolicy: Always
+        env:
+        - name: GITHUB_AUTH_TOKEN
```
naveensrinivasan, line 34: The question I have about the token is that. In the existing single binary without scaling horizontally, the round-robin token https://github.com/ossf/scorecard/blob/587f41117bc9588bf940330195331d5049312c96/roundtripper/github.go#L62 will pick the next one based on the `counter`.

How is this going to work with horizontal scaling? 

If the 5 of them start and everything's going to use the first token. Have we tested this model and seen how are hitting the limits in the GitHub API? 

Ideally, we would need to test the sample set with enough size and a new set of tokens to compare the existing one with this.

cron/k8s/worker.yaml
naveensrinivasan: +1

azeemshaikh38: Yes they are, was hoping to get some feedback here on it. I simply did some rough back of the hand calculations and set them to pretty high numbers. Basically, each repo runs 14 threads in parallel and we are hoping to process maybe 10 repos in parallel in a single machine. I have assumed 1vCPU~=1 thread, so set the CPU requirements to ~150 CPUs.

What do you guys think? Whats a reasonable value to set here.

cron/k8s/worker.yaml
azeemshaikh38: Good question. This is a concern that did come up in our design discussions. TLDR is that for the sake of simplicity of the design, we chose to start out with this, where multiple jobs share tokens (although it might cause us some issues with rate limits). We can look into improving this once we have a more stable framework.

cron/k8s/worker.yaml
oliverchang: These threads are mostly IO bound right? We can probably get by with 1 CPU core per repo (at most). We can start with a very small number of CPUs and monitor the CPU usage and adjust as needed. 220 seems a bit overkill to me.

Also, it seems more beneficial to run 10 separate workers rather than 10 repos in parallel. Are there benefits to running so many on the one worker?

azeemshaikh38: Nope, no benefit IMO. Like I said, I was just doing some rough calculations. I have reduced the CPU numbers to req:8 and limit:12. Should be a good starting point?

oliverchang: In that case, honestly I think we can get back with (req:1, limit:1.5). Scorecards shouldn't be too CPU intensive. We can always monitor and increase as needed.

azeemshaikh38: Done.

PR #500, opened 2021-05-24
🐛  Limit number of batched data points in Stackdriver

azeemshaikh38:

Fix the resource exceeded errors showing up for Stackdriver: 
image

No.

### DIFF
--- a/cron/monitoring/exporter.go
+++ b/cron/monitoring/exporter.go
@@ -22,6 +22,8 @@ import (
 	"github.com/ossf/scorecard/cron/config"
 )
 
+const stackdriverTimeSeriesQuota = 100
+
 func NewStackDriverExporter() (*stackdriver.Exporter, error) {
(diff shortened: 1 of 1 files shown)


cron/monitoring/exporter.go
```diff
@@ -30,6 +30,8 @@ func NewStackDriverExporter() (*stackdriver.Exporter, error) {
 	exporter, err := stackdriver.NewExporter(stackdriver.Options{
 		ProjectID:    projectID,
 		MetricPrefix: "scorecard-cron",
+		// https://cloud.google.com/monitoring/quotas
```
inferno-chromium, line 33: please add metrics name as well, there are many things on that page.

inferno-chromium (approved):
Thanks for the fix!

cron/monitoring/exporter.go
azeemshaikh38: Done.

PR #507, opened 2021-05-25
🐛  Bug fixes and additional logs for PubSub worker

azeemshaikh38:

i) Adds more logs to the PubSub worker for better debuggability.
ii) Removes parallelism during repo runs. Threads are stuck indefinitely, will debug and fix this later. Removing it for now to unblock myself.

No.

### DIFF
--- a/cron/worker/main.go
+++ b/cron/worker/main.go
@@ -17,11 +17,10 @@ package main
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
-	"sync"
 
 	"github.com/google/go-github/v32/github"
 	"github.com/shurcooL/githubv4"
@@ -39,10 +38,6 @@ import (
 func processRequest(ctx context.Context,
 	batchRequest *data.ScorecardBatchRequest, bucketURL string,
 	httpClient *http.Client, githubClient *github.Client, graphClient *githubv4.Client) error {
-	repoURLs := make([]repos.RepoURL, 0, len(batchRequest.GetRepos()))
-	var resultCh chan repos.RepoResult
-	var buffer bytes.Buffer
-
 	checksToRun := checks.AllChecks
 	// nolint
 	// FIXME :- deleting branch-protection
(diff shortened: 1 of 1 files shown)


cron/worker/main.go
```diff
@@ -61,20 +57,12 @@ func processRequest(ctx context.Context,
 		}
 		repoURLs = append(repoURLs, repoURL)
 	}
-	go func() {
```
naveensrinivasan, line 64: Are these requests going to be sequential? Can you please let me know the reason for this change?

cron/worker/main.go
```diff
@@ -51,6 +46,7 @@ func processRequest(ctx context.Context,
 	// This will reduce usage of the API.
 	delete(checksToRun, "Branch-Protection")
 
+	repoURLs := make([]repos.RepoURL, 0, len(batchRequest.GetRepos()))
```
inferno-chromium, line 49: nit: initialize closer to the code using it. (before line 62)

cron/worker/main.go
```diff
@@ -84,11 +72,15 @@ func processRequest(ctx context.Context,
 	filename := data.GetBlobFilename(
 		fmt.Sprintf("shard-%05d", batchRequest.GetShardNum()),
 		batchRequest.GetJobTime().AsTime())
-	return errors.Unwrap(data.WriteToBlobStore(ctx, bucketURL, filename, buffer.Bytes()))
+	if err := data.WriteToBlobStore(ctx, bucketURL, filename, buffer.Bytes()); err != nil {
+		return fmt.Errorf("error during WriteToBlobStore: %w", err)
+	}
+	log.Printf("Write to shard file successful: %s", filename)
```
inferno-chromium, line 78: is this a debug message, will this show in regular runs?

inferno-chromium (approved):

cron/worker/main.go
azeemshaikh38: Yes, at least for now. With the parallel-thread approach something was off and the job was stuck indefinitely. I needed to unblock myself to further test this, so I changed this to sequential. Also adding a TODO here for this.

naveensrinivasan: OK, Thanks

naveensrinivasan (approved):

cron/worker/main.go
azeemshaikh38: Yes, it will show in regular run. Useful to find the file dumped out by the worker.

cron/worker/main.go
azeemshaikh38: repoURLs is populated in the below for loop - line 58.

cron/worker/main.go
inferno-chromium: I meant does it show when running the scorecard tool on command line ?

azeemshaikh38: No, only run during the cron job.

Issue #512, opened 2021-05-26
BUG: Cron job looping


azeemshaikh38:
**Describe the bug**
The cron job seems to be looping. After completing  a few hundred repos, it starts again from the top. 

![image](https://user-images.githubusercontent.com/4825078/119745626-5f8b8f00-be43-11eb-8614-b4fef4829d95.png)

azeemshaikh38:
FYI, we still do not have a successful run for >1 week. The job just ends for some reason and there are no logs/events which specify why.

oliverchang:
Those logs are aggregated so it's possible they're from two separate pods? i.e. the first one exited and then a second one started? I can't see the past pods info anymore on that though, was this cronjob recently recreated?

oliverchang:
Searching for some of these previously run pod IDs in the logs, I see:

```
{
  "jsonPayload": {
    "reportingComponent": "",
    "involvedObject": {
      "uid": "443fbe7e-34c5-4a09-8147-f9a4728922fd",
      "apiVersion": "v1",
      "resourceVersion": "106807510",
      "fieldPath": "spec.containers{run-score}",
      "name": "daily-score-1621908000-6qwtc",
      "kind": "Pod",
      "namespace": "default"
    },
    "reportingInstance": "",
    "message": "Stopping container run-score",
    "type": "Normal",
    "source": {
      "component": "kubelet",
      "host": "gke-openssf-default-pool-aeaa7a9c-wcz1"
    },
    "apiVersion": "v1",
    "metadata": {
      "uid": "4ad5f83e-33e9-4467-b07b-c59da7fa5e49",
      "name": "daily-score-1621908000-6qwtc.16822f7b9e0be112",
      "managedFields": [
            ...OMITTED....
      ],
      "resourceVersion": "180503",
      "creationTimestamp": "2021-05-25T02:50:23Z",
      "namespace": "default"
    },
    "eventTime": null,
    "reason": "Killing",
    "kind": "Event"
  },
  "resource": {
    "type": "k8s_pod",
    "labels": {
      "namespace_name": "default",
      "location": "us-central1-c",
      "cluster_name": "openssf",
      "pod_name": "daily-score-1621908000-6qwtc",
      "project_id": "openssf"
    }
  },
  "timestamp": "2021-05-25T02:50:23Z",
  "severity": "INFO",
  "logName": "projects/openssf/logs/events",
  "receiveTimestamp": "2021-05-25T02:50:28.893360951Z"
}
```

[...]

PR #518, opened 2021-05-27
🌱  Export registered check names 

azeemshaikh38:

In context of https://github.com/ossf/scorecard/issues/512, removing checks which use ArchiveURL.

No.

### DIFF
--- a/checks/active.go
+++ b/checks/active.go
@@ -23,31 +23,32 @@ import (
 )
 
 const (
-	activeStr    = "Active"
+	// CheckActive is the registered name for IsActive.
+	CheckActive  = "Active"
 	lookbackDays = 90
 )
 
 //nolint:gochecknoinits
 func init() {
-	registerCheck(activeStr, IsActive)
+	registerCheck(CheckActive, IsActive)
 }
 
 func IsActive(c *checker.CheckRequest) checker.CheckResult {
 	commits, _, err := c.Client.Repositories.ListCommits(c.Ctx, c.Owner, c.Repo, &github.CommitsListOptions{})
 	if err != nil {
-		return checker.MakeRetryResult(activeStr, err)
+		return checker.MakeRetryResult(CheckActive, err)
 	}
 
 	tz, err := time.LoadLocation("UTC")
 	if err != nil {
-		return checker.MakeRetryResult(activeStr, err)
+		return checker.MakeRetryResult(CheckActive, err)
 	}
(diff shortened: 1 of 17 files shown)


naveensrinivasan:
Is this based on concrete evidence on these rollbacks?

azeemshaikh38:
> Is this based on concrete evidence on these rollbacks?
> 
> I think we would want to investigate further before rolling back changes.
> 
> Thoughts?

I agree. However, I'm a bit desperate to get the cron job running again and since the long weekend is coming up I'm not sure if anyone will be looking into the root cause. So I want to see if this can help.

laurentsimon:
> > Is this based on concrete evidence on these rollbacks?
> > I think we would want to investigate further before rolling back changes.
> > Thoughts?
> 
> I agree. However, I'm a bit desperate to get the cron job running again and since the long weekend is coming up I'm not sure if anyone will be looking into the root cause. So I want to see if this can help.

shall we create a different branch for this sort of test? This way it does not affect the main branch and it will help merges, etc

azeemshaikh38:
> > > Is this based on concrete evidence on these rollbacks?
> > > I think we would want to investigate further before rolling back changes.
> > > Thoughts?
> > 
> > 
> > I agree. However, I'm a bit desperate to get the cron job running again and since the long weekend is coming up I'm not sure if anyone will be looking into the root cause. So I want to see if this can help.
> 
> shall we create a different branch for this sort of test? This way it does not affect the main branch and it will help merges, etc

Fair enough. I updated the PR just to be a refactoring. Will push a locally built binary to remove these checks.

naveensrinivasan (approved):

naveensrinivasan:
> > > > Is this based on concrete evidence on these rollbacks?
> > > > I think we would want to investigate further before rolling back changes.
> > > > Thoughts?
> > > 
> > > 
> > > I agree. However, I'm a bit desperate to get the cron job running again and since the long weekend is coming up I'm not sure if anyone will be looking into the root cause. So I want to see if this can help.
> > 
> > 
> > shall we create a different branch for this sort of test? This way it does not affect the main branch and it will help merges, etc
> 
> Fair enough. I updated the PR just to be a refactoring. Will push a locally built binary to remove these checks.

Thanks @azeemshaikh38 !

laurentsimon:
It may be useful to have a header to the latest.json files we produce, to indicate which checks were run or not.

PR #528, opened 2021-06-02
✨ Use prod tokens for the PubSub cron

azeemshaikh38:

Exchange the GitHub tokens used for PubSub cron job with the daily cron job. 

No.

### DIFF
--- a/cron/k8s/cron.yaml
+++ b/cron/k8s/cron.yaml
@@ -32,7 +32,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: github
-                  key: token
+                  key: test_token
             - name: GCS_BUCKET
               value: ossf-scorecards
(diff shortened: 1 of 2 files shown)


cron/k8s/cron.yaml
```diff
@@ -32,7 +32,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: github
-                  key: token
+                  key: test_token
```
inferno-chromium, line 35: can even start deleting all old cron code, after first iteration of pubsub works.

inferno-chromium (approved):
Exciting!

cron/k8s/cron.yaml
azeemshaikh38: Agreed.

PR #533, opened 2021-06-02
✨  Use lease extension for PubSub worker

azeemshaikh38:

Use lease extension in PubSub cron worker.

No.

### DIFF
--- a/cron/pubsub/subscriber_gcs.go
+++ b/cron/pubsub/subscriber_gcs.go
@@ -1,0 +1,137 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pubsub
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	pubsub "cloud.google.com/go/pubsub/apiv1"
+	pubsubpb "google.golang.org/genproto/googleapis/pubsub/v1"
+
+	"github.com/ossf/scorecard/cron/data"
+)
+
+const (
+	maxMessagesToPull         = 1
+	ackDeadlineExtensionInSec = 600
+	gracePeriodInSec          = 60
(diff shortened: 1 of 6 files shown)


cron/worker/main.go
```diff
@@ -168,7 +168,8 @@ func main() {
 			break
 		}
 		if err := processRequest(ctx, req, checksToRun, bucketURL, httpClient, githubClient, graphClient); err != nil {
-			panic(err)
+			// Nack the message so that another worker can retry.
+			subscriber.Nack()
```
oliverchang, line 172: if you haven't done so already, would be good to set up a dead letter topic (https://cloud.google.com/pubsub/docs/handling-failures#forwarding_to) which lets you set a cap on the number of delivery attempts (to prevent queue blowup if something keeps failing).

cron/pubsub/subscriber_gcs.go
```diff
@@ -0,0 +1,137 @@
...
+			ackDeadline := ackDeadlineExtensionInSec * time.Second
+			err := subscriber.client.ModifyAckDeadline(subscriber.ctx, &pubsubpb.ModifyAckDeadlineRequest{
+				Subscription:       subscriber.subscriptionURL,
+				AckIds:             subscriber.recvdAckIDs,
+				AckDeadlineSeconds: int32(ackDeadline.Seconds()),
+			})
+			if err != nil {
+				log.Fatal(err)
+			}
+			subscriber.delay = ackDeadline - gracePeriodInSec*time.Second
```
oliverchang, line 83: Any reason we need to store/recompute this all the time rather than storing them as consts?

cron/pubsub/subscriber_gcs.go
```diff
@@ -0,0 +1,137 @@
...
+	gcpPubsubPrefix           = "gcppubsub://"
+)
+
+type gcsSubscriber struct {
+	ctx             context.Context
+	done            chan bool
+	client          *pubsub.SubscriberClient
+	pullRequest     *pubsubpb.PullRequest
+	subscriptionURL string
+	recvdAckIDs     []string
```
oliverchang, line 43: Any reason why this needs to be a slice? The way the code is structured right now is pretty tied to the a maximum of 1 message at a time.

oliverchang (approved):
nice!

cron/pubsub/subscriber_gcs.go
azeemshaikh38: I want to have it in such a way that the first time we call `extendDeadline()`, `delay = 0`. After which it is a const.

cron/pubsub/subscriber_gcs.go
azeemshaikh38: True. Updated.

cron/worker/main.go
azeemshaikh38: Done.

PR #534, opened 2021-06-03
✨ check for read-only permissions of github token

laurentsimon:

feature

we don't check for permission in github workflows

we check for permissions in github workflows

no

* **Other information**:
The check currently enforces that the workflow declares no permission (none)
```
permissions:
```
or read permission only for the `contents`
```
permissions:
  contents: read
```

Adding more permission can then be done at the `job` level, e.g.,:
```
// https://docs.github.com/en/actions/reference/authentication-in-a-workflow#example-1-passing-the-github_token-as-an-input
jobs:
  create_commit:
    runs-on: ubuntu-latest 
    permissions:
      issues: write 
```

 This check is somewhat restrictive, since it is correct for a workflow to declare the `write` permission at the start of the file rather than at the job level:
```
permissions:
  issues: write
```

However, since nobody uses this permission today, and it's by default `read/write`, I don't think we will encounter projects with this config. This means we can ask them to follow the pattern where they add the `write` permission at the job level.

Note that there is a github [config](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) we could query in the future, but it's not available via REST API atm.


checks/checks.yaml
```diff
@@ -15,6 +15,31 @@
 # This is the source of truth for all check descriptions and remediation steps.
 # Run `cd checks/main && go run /main` to generate `checks.json` and `checks.md`.
 checks:
+  Token-Permissions:
+    description: >-
+      This check tries to determine if a project's github workflows
+      follow the principle of least privilege, i.e. if the githib tokens
```
oliverchang, line 21: nit: s/github/GitHub and s/githib/GitHub/ (typo)

checks/checks.yaml
```diff
@@ -50,10 +75,11 @@ checks:
       (Javascript),  requirements.txt, pipfile.lock (Python), gemfile.lock
       (Ruby), cargo.lock (Rust), yarn.lock (package manager), composer.lock
       (PHP), vendor/, third_party/, third-party/; (2) look for github actions
-      under .github/workflows/ and verifies they are pinned by hash. If one of
-      the files in (1) AND all the dependencies in (2) are pinned,  the check
-      succeds. This check does not currently look for docker image pinning and
-      shell script pinning.
+      under .github/workflows/ and verifies they are pinned by hash; (3) looks 
+      for docker files and verifies FROM dependencies are pinned by hash. If one of
```
oliverchang, line 79: s/docker files/Dockerfiles/

checks/permissions.go
```diff
@@ -0,0 +1,169 @@
...
+	registerCheck(CheckPermissions, leastPrivilegedTokens)
+}
+
+func leastPrivilegedTokens(c *checker.CheckRequest) checker.CheckResult {
+	return CheckFilesContent(CheckPermissions, ".github/workflows/*", true, c, validateGitHubActionTokenPermissions)
+}
+
+type GitHubActionWorkflowConfig struct {
+	Permissions *struct {
+		Contents       *string `yaml:"contents"`
```
oliverchang, line 37: nit: no need for these to be pointers. we can probably simplify things a bit by just using plain `string`s here and checking for `""` instead.

checks/permissions.go
```diff
@@ -0,0 +1,169 @@
...
+	return permission != nil &&
+		*permission != "read" &&
+		*permission != "none"
+}
+
+func permissionToString(permission *string) string {
+	if permission != nil {
+		return *permission
+	}
+	return "read"
```
oliverchang, line 62: should this be "none" instead?

checks/permissions.go
```diff
@@ -0,0 +1,169 @@
...
+		SecurityEvents *string `yaml:"security events"`
+		Statuses       *string `yaml:"statuses"`
+	} `yaml:"permissions"`
+	Name string `yaml:"name"`
+}
+
+func isPermissionWrite(permission *string) bool {
+	return permission != nil &&
+		*permission != "read" &&
+		*permission != "none"
```
oliverchang, line 55: why not check `== "write"` instead?

checks/permissions.go
oliverchang: Actually thinking about this again, we don't even have to specify all of these explicitly. If we just use a `map[string]string` here we can just iterate over the values here and below without having the list field names.

We probably also need to make `GitHubActionWorkflowConfig` itself a `map[string]interface{}` too and do some runtime type checks to catch the `permissions: read-all|write-all` case (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions)

checks/permissions.go
```diff
@@ -0,0 +1,169 @@
...
+
+func validateDefaultReadPermissions(config GitHubActionWorkflowConfig, path string,
+	logf func(s string, f ...interface{})) bool {
+	if config.Permissions == nil {
+		logf("!! token-permissions/token - no permission defined in %v", path)
+		return false
+	}
+
+	r := true
+	if isPermissionWrite(config.Permissions.Contents) {
```
oliverchang, line 73: as above let's deserialize the permissions field into a `map[string]string` here instead and just iterate here.

checks/permissions.go
laurentsimon: what are the benefits of not using pointers? I used them because I wanted to differentiate between the field not being present and being set to empty string. (Might not matter in this case, though). But I don't know if there are consequences to using pointers. 

Thanks for the `read-all|write-all` link. I could not find the link yesterday :/.
Is there a way for the yaml parser to understand a union, ie that a field is either a string or a structure? Using a map will work but feels a bit like a hack. What if the structure is more complicated? Is using a map the only way to support unions?

checks/permissions.go
laurentsimon: yep!

azeemshaikh38 (approved):

laurentsimon:
@naveensrinivasan do you know why this is failing?

inferno-chromium:
> @naveensrinivasan do you know why this is failing?

lint failure is fixed, see my trial fix above.

but some file probably forgot to git add

```
--- FAIL: TestGithubTokenPermissions (0.00s)
    --- FAIL: TestGithubTokenPermissions/None_test (0.00s)
panic: cannot read file: open ./testdata/github-workflow-permissions-none.yaml: no such file or directory [recovered]
	panic: cannot read file: open ./testdata/github-workflow-permissions-none.yaml: no such file or directory
```

checks/permissions.go
```diff
@@ -0,0 +1,153 @@
...
+
+		// Check the type of our values.
+		switch val := value.(type) {
+		// Empty string is nil type.
+		// It defaults to 'none'
+		case nil:
+
+		// String type.
+		case string:
+			fmt.Println(val)
```
naveensrinivasan, line 78: Nit- Do we need this?

checks/permissions.go
```diff
@@ -0,0 +1,153 @@
...
+			if val != "read-all" && val != "" {
+				logf("!! token-permissions/github-token - permission set to '%v' in %v", val, path)
+				return false, nil
+			}
+
+		// Map type.
+		case map[interface{}]interface{}:
+
+			// Iterate over the permission, verify keys and values are strings.
+			for k, v := range val {
```
naveensrinivasan, line 88: I saw the comments from @oliverchang  about static binding. 
But this custom marshaling could cause errors in the future and more than that readability is an issue. 

In `go`  `Clear is better than clever.` https://go-proverbs.github.io/  . I would have voted for static binding.

naveensrinivasan (commented):
Thanks!

checks/permissions.go
laurentsimon: no!

checks/permissions.go
laurentsimon: while re-writing it, I almost reverted to the old code because it was hard to read. I think readability is OKish.
what do you mean by `custom marshaling could cause errors in the future`?
do you know how to use union with golang yaml parser? That's what I would have liked to use, but I don't know it it's feasible.

naveensrinivasan: > while re-writing it, I almost reverted to the old code because it was hard to read. I think readability is OKish.
> what do you mean by `custom marshaling could cause errors in the future`?

we assume that if it is of type `string` then `validatePermission` would work along with that. This is an assumption and there isn't a contract by the provider that it would in the future

> do you know how to use union with golang yaml parser? That's what I would have liked to use, but I don't know it it's feasible. If it is, I could use that instead.

TBH I don't know. I think I would have still keep it simple.

laurentsimon: > > while re-writing it, I almost reverted to the old code because it was hard to read. I think readability is OKish.
> > what do you mean by `custom marshaling could cause errors in the future`?
> 
> we assume that if it is of type `string` then `validatePermission` would work along with that. This is an assumption and there isn't a contract by the provider that it would in the future

We verify it is a string before calling the function.

laurentsimon: an alternative approach to handle unions could be to first try to unmarshal statically with permissions defined as a `string`; and if it fails try again with the permission defined as a map (contents, pullrequest, etc) as in the original commit.

WDUT?

oliverchang: I would disagree that custom marshalling is less readable. The previous approach resulted in many bits of copy-pasted code which is hard to read through and easy to introduce mistakes in. 

Given that Go doesn't handle unions or dynamically typed structs very well (as in this case), we probably need something like this anyway.

checks/permissions.go
```diff
@@ -0,0 +1,166 @@
...
+
+func validateMapPermissions(values map[interface{}]interface{}, path string,
+	logf func(s string, f ...interface{})) (bool, error) {
+	permissionRead := true
+	var r bool
+	var err error
+
+	// Iterate over the permission, verify keys and values are strings.
+	for k, v := range values {
+		switch key := k.(type) {
```
oliverchang, line 62: probably simpler as just:

```go
key, ok := k.(string)
if !ok {
  return false, ErrInvalidGitHubWorkflowFile
}

if r, err = validatePermission(key, v, path, logf); err != nil {
    return false, err
}
...
```

checks/permissions.go
```diff
@@ -0,0 +1,166 @@
...
+	registerCheck(CheckPermissions, leastPrivilegedTokens)
+}
+
+func leastPrivilegedTokens(c *checker.CheckRequest) checker.CheckResult {
+	return CheckFilesContent(CheckPermissions, ".github/workflows/*", true, c, validateGitHubActionTokenPermissions)
+}
+
+func validatePermission(key string, value interface{}, path string,
+	logf func(s string, f ...interface{})) (bool, error) {
+	switch val := value.(type) {
```
oliverchang, line 42: can just do `val, ok := value.(string)`

checks/permissions.go
laurentsimon: Going to merge for now with the existing code. Happy to revisit once I have feedback on the above.

checks/permissions.go
laurentsimon: sorry, did not see before merging. I created #539 to address it.

checks/permissions.go
laurentsimon: same, see #539 .

PR #539, opened 2021-06-03
✨ Cleanup code for github tokens #534

laurentsimon:

cleanup

the same.

the same.

no

* **Other information**:
This addresses some comments I missed before merging, see #534

### DIFF
--- a/checks/permissions.go
+++ b/checks/permissions.go
@@ -17,6 +17,7 @@ package checks
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 
@@ -44,7 +45,7 @@ func validatePermission(key string, value interface{}, path string,
 		return false, ErrInvalidGitHubWorkflowFile
 	}
 
-	if val == "write" {
+	if strings.EqualFold(val, "write") {
 		logf("!! token-permissions/github-token - %v permission set to '%v' in %v", key, val, path)
 		return false, nil
 	}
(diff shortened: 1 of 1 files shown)


checks/permissions.go
```diff
@@ -39,15 +39,16 @@ func leastPrivilegedTokens(c *checker.CheckRequest) checker.CheckResult {
...
-			logf("!! token-permissions/github-token - %v permission set to '%v' in %v", key, val, path)
-			return false, nil
-		}
-	default:
+	val, ok := value.(string)
+	if !ok {
 		return false, ErrInvalidGitHubWorkflowFile
 	}
+
+	if val == "write" {
```
naveensrinivasan, line 47: instead of the direct comparison can we do `strings.EqualFold`? which is case insensitive

checks/permissions.go
```diff
@@ -59,20 +60,18 @@ func validateMapPermissions(values map[interface{}]interface{}, path string,
...
+		key, ok := k.(string)
+		if !ok {
 			return false, ErrInvalidGitHubWorkflowFile
 		}
+
+		if r, err = validatePermission(key, v, path, logf); err != nil {
+			return false, err
+		}
+
+		if !r {
```
naveensrinivasan, line 72: Shouldn't this be `else if`?

naveensrinivasan (approved):

oliverchang (approved):

checks/permissions.go
laurentsimon: done. Did the same in other comparison places too. I don't have unit tests for this though. I actually don't know if github enforces lower-case or not.

checks/permissions.go
laurentsimon: I want to be able to log all offending lines. I think it helps users to fix everything at once. Otherwise, I'm worried some users will fix the one line that scorecard reports, and then realize they should fix the rest.

Issue #432, opened 2021-05-11
BUG: restrict github token for scorecard repo


laurentsimon:
Le's reduce all our tokens' scope to be [read-only by default](https://github.com/ossf/scorecard/settings/actions), unless needed otherwise.

azeemshaikh38:
Adding this our bi-weekly agenda. Let's discuss there.

naveensrinivasan:
@azeemshaikh38  Included a label needs discussion. I have been adding that to the issues that need one.

azeemshaikh38:
Ah I see, nice. I did not notice it. You could also update the agenda doc as and when something comes to your mind. Some topics may not need a GitHub issue.

laurentsimon:
result from scorecard:
```
!! token-permissions/github-token - no permission defined in .github/workflows/codeql-analysis.yml
!! token-permissions/github-token - no permission defined in .github/workflows/gitcache-docker.yaml
!! token-permissions/github-token - no permission defined in .github/workflows/goreleaser.yaml
!! token-permissions/github-token - no permission defined in .github/workflows/integration.yml
!! token-permissions/github-token - no permission defined in .github/workflows/main.yml
!! token-permissions/github-token - no permission defined in .github/workflows/ok-to-test.yml
!! token-permissions/github-token - no permission defined in .github/workflows/stale.yml
!! token-permissions/github-token - no permission defined in .github/workflows/verify.yml

```

laurentsimon:
Once this is fixed, also create an e2e test in scorecard.

naveensrinivasan:
![image](https://user-images.githubusercontent.com/172697/121049960-3396d480-c786-11eb-8f95-373b1c7c122d.png)

FYI. I am changing the permissions to read-only and fixing it while testing the changes.

inferno-chromium:
> ![image](https://user-images.githubusercontent.com/172697/121049960-3396d480-c786-11eb-8f95-373b1c7c122d.png)
> 
> FYI. I am changing the permissions to read-only and fixing it while testing the changes.

Great, i dont know if the check looks for read-only enforcement in settings (it looks on per-file basis), maybe you need to update the check to account for that as well.

naveensrinivasan:
I agree. The code fix does not have that and AFAIK there isn't an API. The existing fixes will provide false positives. Till we have an API to check for these settings we should not run this check to avoid false positives. 

Thoughts?

laurentsimon:
[...]

PR #551, opened 2021-06-07
🌱 Allows write permissions for actions

naveensrinivasan:

Bug fix

#432

Allowed write permissions to action for commenting on the status of the PR.

None

* **Other information**:
None

### DIFF
--- a/.github/workflows/ok-to-test.yml
+++ b/.github/workflows/ok-to-test.yml
@@ -18,7 +18,8 @@ name: Ok To Test
 on:
   issue_comment:
     types: [created]
-
+permissions:
+  pull-requests: write
 jobs:
   ok-to-test:
     runs-on: ubuntu-latest
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -16,6 +16,8 @@ name: PR Verifier
 on:
   pull_request_target:
     types: [opened, edited, reopened]
+permissions:
+  pull-requests: write
 jobs:
   verify:
     name: Verify PR contents


naveensrinivasan:
I can't test this locally, so probably going to be multiple PR's for this fix

naveensrinivasan:
The checks are going to fail because they don't have permission.

laurentsimon (approved):

.github/workflows/ok-to-test.yml
```diff
@@ -18,7 +18,8 @@ name: Ok To Test
 on:
   issue_comment:
     types: [created]
-
+permissions:
+  pull-requests: write
```
laurentsimon, line 22: my bad: I LGTM'ed the CL but that's not going to pass the scorecard check. The check encourages devs to use `permission:read-all` or `permissions: XXX:read` or `permissions:` globally at the top of the file; and use `write` permission in each job individually. The motivation here is that if we encourage users to set write permission globally, any time they add new actions in the file it will inherit this permission. We d'like to avoid this situation.

naveensrinivasan: > my bad: I LGTM'ed the CL but that's not going to pass the scorecard check. The check encourages devs to use `permission:read-all` or `permissions: XXX:read` or `permissions:` globally at the top of the file; and use `write` permission in each job individually. The motivation here is that if we encourage users to set write permission globally, any time they add new actions in the file it will inherit this permission. We d'like to avoid this situation.
> 
> I'm going to start looking at known actions and the permissions they need, to see if we can go further than that and flag write permission when they are not actually needed.
> 
> Advice welcome for this.

The issue with this is a repository could have global reads within the repository setting and explicit write requests for each file and scorecard would fail. Most of the users would turn off this setting as it is more work for them to do this.

Issue #549, opened 2021-06-04
BUG: PubSub worker not Ack-ing requests


azeemshaikh38:
**Describe the bug**
The PubSub job is not ack-ing the received messages, resulting in messages being re-delivered multiple times. The requests are being successfully ack-ed according to the logged metrics (image below), but the number of un-acked messages doesn't seem to be going down at all. Not sure what might be the issue. I even tried a local test where I pull the message and immediately `Ack` it, and it still doesn't seem to work. Any help with this would be appreciated - @inferno-chromium @oliverchang @naveensrinivasan 

![image](https://user-images.githubusercontent.com/4825078/120858318-7878fc00-c537-11eb-85e3-1b6f34dc3f0a.png)

inferno-chromium:
My guess is some error in https://github.com/ossf/scorecard/blob/3b86d5721784b035669d13e6b8afddfbbeddef98/cron/worker/main.go#L172
which is not logged and hence we cant tell what is reason for it.
```
if err := processRequest(ctx, req, checksToRun, bucketURL, httpClient, githubClient, graphClient); err != nil {			
// Nack the message so that another worker can retry.			
subscriber.Nack()		
}
```
my usual process of debugging is to add more logging and then check those log messages/error in stackdriver.

azeemshaikh38:
Hmm, no I ruled that out by some local testing. I'm going to send out a PR to make sure we don't re-process a request for which a shard already exists. That'll save us some tokens. Will keep this issue open to see how to solve the `Ack` issue.

oliverchang:
Have you tried reverting to gocloud to see if the behaviour is the same here?

azeemshaikh38:
Yeah gocloud also has the same behaviour. I have started an internal bug with the Cloud PubSub oncall to get more help here.

azeemshaikh38:
A weird thing happened here, after worker finished processing all requests the subscription immediately `ack-ed` all requests at once - 
![image](https://user-images.githubusercontent.com/4825078/121275198-febb7800-c880-11eb-9358-147d02e1b004.png)

Keeping this bug open still understand what happened here, but on a lower priority.

azeemshaikh38:
Closing this issue. Following up internally on a low priority basis. So far the internal team doesn't seem to find any cause for concern and thinks it is working as expected.

PR #553, opened 2021-06-07
🐛 Fix issues with PubSub worker

azeemshaikh38:

Fixes 2 issues with the PubSub worker:
i) increases memory and CPU limits
ii) early return if shard already exists (temporary solution to overcome https://github.com/ossf/scorecard/issues/549)

No

### DIFF
--- a/cron/k8s/worker.yaml
+++ b/cron/k8s/worker.yaml
@@ -42,6 +42,8 @@ spec:
           value: "gs://ossf-scorecards-cache"
         resources:
           requests:
-            memory: 1Gi
+            cpu: 5
+            memory: 5Gi
           limits:
-            memory: 1.5Gi
+            cpu: 12
+            memory: 12Gi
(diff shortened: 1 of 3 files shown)


naveensrinivasan (approved):

cron/k8s/worker.yaml
```diff
@@ -42,6 +42,8 @@ spec:
           value: "gs://ossf-scorecards-cache"
         resources:
           requests:
-            memory: 1Gi
+            cpu: 3
```
inferno-chromium, line 45: can go higher, at this point, we want stability, so can try 5 with 5gi here and limit can be even higher like 10 or 15 or whatever.

inferno-chromium (approved):

cron/k8s/worker.yaml
azeemshaikh38: Done.

Issue #445, opened 2021-05-12
Feature: Branch-Protection: add support for release branches


laurentsimon:
Currently we only check the default branch - which we assume is the dev branch.
We want to add protection for release branches. A bonus would be to verify the default branch has activity and is likely the dev branch.

laurentsimon:
An idea @jeffmendoza and I brainstormed today to identify branches that need branch protection is to list the releases and see which branch they points to.
This may require multiple API requests, though (1 for each different `target_commitish`) using this [release API](https://docs.github.com/en/rest/reference/repos#list-releases ).

Example:
```
$ curl -isL  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/envoyproxy/envoy/releases | grep target_commitish
    "target_commitish": "release/v1.15",
    "target_commitish": "release/v1.16",
    "target_commitish": "release/v1.17",
    "target_commitish": "release/v1.18",
    "target_commitish": "main",
    "target_commitish": "main",
    "target_commitish": "main",
    "target_commitish": "release/v1.17",
    "target_commitish": "release/v1.16",
    "target_commitish": "release/v1.15",
    "target_commitish": "release/v1.14",
    "target_commitish": "release/v1.17",
    "target_commitish": "release/v1.13",
    "target_commitish": "master",
    "target_commitish": "release/v1.16",
    "target_commitish": "release/v1.15",
    "target_commitish": "release/v1.14",
    "target_commitish": "release/v1.13",
    "target_commitish": "release/v1.16",
    "target_commitish": "8fb3cb86082b17144a80402f5367ae65f06083bd",
    "target_commitish": "release/v1.15",
    "target_commitish": "release/v1.13",
    "target_commitish": "release/v1.15",
    "target_commitish": "release/v1.14",
    "target_commitish": "release/v1.13",
    "target_commitish": "release/v1.12",
    "target_commitish": "master",
    "target_commitish": "master",
    "target_commitish": "master",
    "target_commitish": "master",
```

asraa:
I can try this :) Commenting in case that will allow assign-ability :P

laurentsimon:
yaaay... it worked! Thanks @asraa

PR #554, opened 2021-06-08
:sparkles: add release branch protection check

asraa:

* **Other information**:
Fixes https://github.com/ossf/scorecard/issues/445
(although it does not do detection of activity on default branch)


asraa:
Probably an e2e test is best for this, have to check in on how to enable that. The unit tests probably suffice for the check code.

asraa:
Added unit tests with a nice suggestion from @jeffmendoza. Ready for review!

checks/branch_protected.go
```diff
@@ -31,47 +33,88 @@ func init() {
...
+	repoStr string) checker.CheckResult {
+	// Get release branches
+	releases, _, err := r.ListReleases(ctx, ownerStr, repoStr, &github.ListOptions{})
+	if err != nil {
+		return checker.MakeRetryResult(CheckBranchProtection, err)
+	}
+
+	var checks []checker.CheckResult
+	for _, release := range releases {
+		if release.TargetCommitish != nil {
```
laurentsimon, line 62: do you know when TargetCommitish may be nil? Shall we fail if it happens?

checks/branch_protected.go
```diff
@@ -31,47 +33,88 @@ func init() {
...
-		GetBranchProtection(c.Ctx, c.Owner, c.Repo, *repo.DefaultBranch)
 	const fileNotFound = 404
 	if resp.StatusCode == fileNotFound {
 		return checker.MakeRetryResult(CheckBranchProtection, err)
 	}
 
 	if err != nil {
-		c.Logf("!! branch protection not enabled")
+		l("!! branch protection not enabled for branch %s", branch)
 		const confidence = 10
```
laurentsimon, line 90: can use MaxResultConfidence

laurentsimon (approved):

checks/branch_protected.go
asraa: doesn't seem like it. actually another point i thought of is that some of these branches may have overlap (e.g. be the dev branch). it won't affect the result since the result is a multicheckand, but it will reduce calls to get the branch protection.

asraa: actually i think it can be nil https://github.com/google/go-github/blob/b235769d16066e228821a3e0fae66305b49021d6/github/github-accessors.go#L13295

it says "Unused if the Git tag already exists."

checks/branch_protected.go
```diff
@@ -31,47 +33,94 @@ func init() {
...
+	}
+	if !checkedBranches[*repo.DefaultBranch] {
+		res := getProtectionAndCheck(ctx, r, l, ownerStr, repoStr, *repo.DefaultBranch)
+		checks = append(checks, res)
+	}
+
+	return checker.MultiCheckResultAnd(checks...)
+}
+
+func getProtectionAndCheck(ctx context.Context, r repositories, l logger, ownerStr, repoStr,
```
laurentsimon, line 85: do you know if this will work when the commitish is a commit hash and not a branch name?
For envoy, for example, one of the releases has commitish `"target_commitish": "8fb3cb86082b17144a80402f5367ae65f06083bd"` (see #445)

checks/branch_protected_test.go
```diff
@@ -15,7 +15,10 @@
 package checks
```
laurentsimon, line 15: you can also add e2e tests in the e2e/ folder. You can launch `make e2e` with an GITHUB_AUTH_TOKEN set for scorecard repo. But you should be able to locally test only the branch protection e2e test (I forgot the exact go command :-))

checks/branch_protected.go
asraa: Nice catch! You get a "404 Branch not found" (nil response and an error) as opposed to if there's no protection, which gives "404 Branch not protected" (resp status not found and an error). 

The existing code probably assumed the default branch was always there. Updating so this skips when branches are not found.

checks/branch_protected.go
```diff
@@ -31,47 +33,94 @@ func init() {
...
 
-	protection, resp, err := c.Client.Repositories.
-		GetBranchProtection(c.Ctx, c.Owner, c.Repo, *repo.DefaultBranch)
+	var checks []checker.CheckResult
+	checkedBranches := map[string]bool{}
+	for _, release := range releases {
+		// may be nil if release tag already exists
+		commitish := release.TargetCommitish
+		if commitish != nil && !checkedBranches[*commitish] {
+			res := getProtectionAndCheck(ctx, r, l, ownerStr, repoStr, *commitish)
```
laurentsimon, line 66: we could reduce the number of API calls by returning early if we find a non-protected branch. In most checks, though, I've taken the decision to continue anyway: this way we can show users the entire list of offending branches they need to fix. I suppose a better way would be to return early if the option `--show-details` was not provided by the user, and iterate thru all branches if the option was not provided (in a different CL).

asraa: Actually yes. I just a nice way of resolving this and https://github.com/ossf/scorecard/pull/554#discussion_r650037369 together.

We can make one call to ListBranches at the start. This tells us which branches exist and which are protected in a single call https://github.com/google/go-github/blob/edaee9aa26ea0b2fedd961398433b8940b2afef1/github/repos.go#L912

Then from the release target_commitish and dev branch, check this list first and only check the ones with protections

checks/branch_protected.go
laurentsimon: I think you could check if a commitish is a hash. If it is, retrieve the name of the branch the commits belongs to, and use the branch name to call `getProtectionAndCheck()`.

checks/branch_protected.go
laurentsimon: Do you know what the comment actually mean? I don't :-)

How about failing and returning a specific error (e.g. `ErrCommitishNil`), so we can catch it upstream, log it and investigate when/how often this happens in practice. We can look at the logs from the cron job.

@azeemshaikh38 would this be ok? In the dockerfile pinning [check](https://github.com/ossf/scorecard/blob/main/checks/frozen_deps.go#L33), we return  `ErrInvalidDockerfile` for this reason. I don't think we're currently checking if this happens, though. Azeem, do you know? Maybe now is the time to start adding keyword to monitor the logs. wdut?

azeemshaikh38 (approved):
Nice, thanks!

checker/check_runner.go
```diff
@@ -126,3 +126,23 @@ func MultiCheckAnd(fns ...CheckFn) CheckFn {
 		return minResult
 	}
 }
+
+// MultiCheckResultAnd means all check results must succeed. This returns a conservative result
+// where the worst result is returned.
+func MultiCheckResultAnd(res ...CheckResult) CheckResult {
```
azeemshaikh38, line 132: Could move this to `checker/check_result.go` where there are already a bunch of fns to make CheckResult. 

Nit: rename to `MakeAndResult`?

checks/branch_protected_test.go
```diff
@@ -32,12 +35,302 @@ func (l *log) Logf(s string, f ...interface{}) {
...
+				Pass:        true,
+				Details:     nil,
+				Confidence:  10,
+				ShouldRetry: false,
+				Error:       nil,
+			},
+		},
+	}
+
+	get = func(context.Context, string, string) (*github.Repository,
```
azeemshaikh38, line 283: The `get`, `listReleases` and `getBranchProtection` fn vars are confusing. Instead, update `mockRepos` struct to hold the data passed in through the testcase. Something like:

```
type mockRepos struct {
  defaultBranch string
  releases []*github.RepositoryRelease
  protections map[string]*github.Protection
}
```

The interface fns can directly work on these internal vars. Then, your `mockRepo` initialization would simply look like:

```
mockrepo := mockRepo {
  defaultBranch: tt.defaultBranch,
  releases: tt.releases,
  protections: tt.protections,
}
```

checks/branch_protected.go
asraa: I refactored to check whether the branch exists with ListBranches. If it doesn't exist/is deleted, it skips.

I'm not sure I can check which branch the commit belongs to. I found references to the github search api, which seems like it's in preview mode.

checks/branch_protected.go
azeemshaikh38: Yes, it should be easy to setup monitoring to figure out how many errors (grouped by error type) we are seeing. Also, always a big +1 to returning specific errors that can be logged upstream.

However, I personally don't prefer enabling BranchProtection check in the cron job right away (it is currently disabled). I don't have confidence on what affect this check will have on our tokens. So we could do some local testing and build some confidence before enabling it in the cron job.

checks/branch_protected.go
laurentsimon: > I refactored to check whether the branch exists with ListBranches. If it doesn't exist/is deleted, it skips.

shall we fail when the commitish points to a non-existing branch, since it provides no information about the protection of the branch that was used? Wdut?

> I'm not sure I can check which branch the commit belongs to. I found references to the github search api, which seems like it's in preview mode.

you're right, does not seem to work. An alternative would be to clone the repo and use `git branch --all --contains <commit>`. I think it's better to wait for the API above to be more broadly available. Let's check if the commitish looks like a hash and ignore this case for now then. Let's create a github issue to track this for future improvement.

Any better ideas?

checker/check_runner.go
asraa: Thanks! added in a separate PR since I noticed a bug in testing this one.
https://github.com/ossf/scorecard/pull/571

checks/branch_protected.go
laurentsimon: > > I refactored to check whether the branch exists with ListBranches. If it doesn't exist/is deleted, it skips.
> 
> shall we fail when the commitish points to a non-existing branch, since it provides no information about the protection of the branch that was used? Wdut?

while looking at envoy, I realized you renamed the `master` branch to `main` (the new naming that GitHub recommends). With what I suggested above, we would fail the test since we would not find the master branch no longer exists. If you go with the solution I suggested, you'll have to treat main and master as being the same branch. 

> > I'm not sure I can check which branch the commit belongs to. I found references to the github search api, which seems like it's in preview mode.
> 
> you're right, does not seem to work. An alternative would be to clone the repo and use `git branch --all --contains <commit>`. I think it's better to wait for the API above to be more broadly available. Let's check if the commitish looks like a hash and ignore this case for now then. Let's create a github issue to track this for future improvement.
> 
> Any better ideas?

asraa: Sorry for the delay, was doing a bunch of local testing and ran into some rate limits. To answer the questions:
* Agreed, let's ignore the hash commitish for now (I added a comment TODO to go back later for improvement)
* About the branch rename. Ugh! the Github REST API gives you a 301 moved permanently when you try to get a moved branch (like master). BUT! Go-github does NOT from my testing. It gives you a 404 Not Found, giving me no way to distinguish between a moved and a deleted/non-existing branch. It feels wrong to fail on that. Still trying to see if I can get moved information from go-github.

laurentsimon: > Sorry for the delay, was doing a bunch of local testing and ran into some rate limits. To answer the questions:
> 
> * Agreed, let's ignore the hash commitish for now (I added a comment TODO to go back later for improvement)
+1

> * About the branch rename. Ugh! the Github REST API gives you a 301 moved permanently when you try to get a moved branch (like master). BUT! Go-github does NOT from my testing. It gives you a 404 Not Found, giving me no way to distinguish between a moved and a deleted/non-existing branch. It feels wrong to fail on that. Still trying to see if I can get moved information from go-github.

nice finding. I did not know about that! While you investigate, we could file an github issue for go-github to ask about the behavior and whether it'd be a breaking change for them to fix it. 
Last resort would be for us to make the http request ourselves :/ Kinda of sad to have to do this, though

asraa: Yep, just did! https://github.com/google/go-github/issues/1895

For now just to keep this PR going I'll manually handle the master->main change, and otherwise we'll just fail when the commitish returns a non-existing branch.

asraa: done... as much as can do.

checks/branch_protected.go
asraa: added a distinct error!

asraa:
Tried this on my own silly repo, that included a main -> master rename

```
$ ./scorecard --show-details --checks Branch-Protection --repo=github.com/asraa/asraa.github.io
Starting [Branch-Protection]
Finished [Branch-Protection]

RESULTS
-------
Repo: github.com/asraa/asraa.github.io
Branch-Protection: Fail 10
!! branch protection not enabled for branch branch1
!! branch protection - EnforceAdmins should be enabled on main
!! branch protection - Linear history should be enabled on main
!! branch protection - Status checks for merging should be enabled on main
!! branch protection - Stale review dismissal should be enabled on main
!! branch protection - Owner review should be enabled on main
```

checks/branch_protected_test.go
laurentsimon: want to do this in this PR or in a follow-up one? Are there blockers for this? (I know sometimes it's hard to find a repo that actually passes all the tests)

PR #556, opened 2021-06-09
🐛  Reduce cardinality of OpenCensus stats

azeemshaikh38:

To reduce the number of `too many time series` errors from Stackdriver:
i) Ensure that Stackdriver logs are updated at the end of every request.
ii) Reduce the cardinality of the metrics by not logging `Repo` and 'HttpPath' as labels.

No.

### DIFF
--- a/stats/views.go
+++ b/stats/views.go
@@ -15,7 +15,6 @@
 package stats
 
 import (
-	"go.opencensus.io/plugin/ochttp"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
 )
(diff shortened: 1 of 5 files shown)


stats/views.go
```diff
@@ -24,11 +23,12 @@ var (
 	// CheckRuntime tracks CPU runtime stats.
 	CheckRuntime = view.View{
 		Name:        "CheckRuntime",
-		Description: "CPU runtime stats per repo per check",
+		Description: "CPU runtime stats per check",
 		Measure:     CPURuntimeInSec,
-		TagKeys:     []tag.Key{Repo, CheckName},
```
oliverchang, line 29: do we want a separate view with just the "Repo" as the tag? Maybe some repos result in more with CPU/HTTP request usage and we want to know about them?

oliverchang (approved):

stats/views.go
azeemshaikh38: Good point. I've left a TODO here for this. If this PR does help with the logs, then can add/enable per-repo CPU runtime stats.

Issue #400, opened 2021-05-05
New check: Complain on binary artifacts in source repos, e.g. jars, esp for third_party use


inferno-chromium:
**Is your feature request related to a problem? Please describe.**
Third_party libraries might be checked in as binary artifact instead of building from source. These might be compromised artifacts, so discourage this bad security pattern.

**Describe the solution you'd like**
Check fail with details on binary artifacts used.

**Describe alternatives you've considered**
NA

**Additional context**
NA

laurentsimon:
Possible ideas:
- OS binaries: .o, .so, .ko, elf (Linux), same for Windows/Macos
- distro packages: .deb, .rpm, etc
- language packages: .whl, etc
- bytecode: .pyc, .jar

naveensrinivasan:
Looking for file types with the extension could be one way. 


The right way to do it is to parse the file header and figure out the type. This is work has been done at https://github.com/h2non/filetype

We could start with this.
1. https://github.com/h2non/filetype#application
2. https://github.com/h2non/filetype#archive

I am not a fan of figuring based on extension alone without checking the contents of the file using the header. The confidence in our results would be higher we actually investigate the header which is done by `filetype`

Thoughts?

inferno-chromium:
> Looking for file types with the extension could be one way.
> 
> The right way to do it is to parse the file header and figure out the type. This is work has been done at https://github.com/h2non/filetype
> 
> We could start with this.
> 
> 1. https://github.com/h2non/filetype#application
> 2. https://github.com/h2non/filetype#archive
> 
> I am not a fan of figuring based on extension alone without checking the contents of the file using the header. The confidence in our results would be higher we actually investigate the header which is done by `filetype`
> 
> Thoughts?

probably both. i think extensions is quick, but we should probably do libmagic based checking as well - https://man7.org/linux/man-pages/man3/libmagic.3.html [https://github.com/h2non/filetype is probably similar idea?]

azeemshaikh38:
Similar idea to `filetype` and `libmagic` is `file -i` command in Unix. 

![image](https://user-images.githubusercontent.com/4825078/118322219-a37dac00-b4b3-11eb-8720-c9d2adb6f83b.png)

naveensrinivasan:
> > Looking for file types with the extension could be one way.
> > The right way to do it is to parse the file header and figure out the type. This is work has been done at https://github.com/h2non/filetype
> > We could start with this.
> > 
[...]

PR #563, opened 2021-06-09
✨  Binary Artifact check

naveensrinivasan:

Feature

#400

    * Implemented binary artifact checks

    * Didn't include the filetype check as it requires extracting of the 
      tarball and getting the file content which is a larger effort.

None

* **Other information**:
None

### DIFF
--- a/checks/automatic_dependency_update.go
+++ b/checks/automatic_dependency_update.go
@@ -20,16 +20,16 @@ import (
 	"github.com/ossf/scorecard/checker"
 )
 
-const check = "Automatic-Dependency-Update"
+const CheckAutomaticDependencyUpdate = "Automatic-Dependency-Update"
 
 //nolint
 func init() {
-	registerCheck(check, AutomaticDependencyUpdate)
+	registerCheck(CheckAutomaticDependencyUpdate, AutomaticDependencyUpdate)
 }
 
 // AutomaticDependencyUpdate will check the repository if it contains Automatic dependency update.
 func AutomaticDependencyUpdate(c *checker.CheckRequest) checker.CheckResult {
-	result := CheckIfFileExists(check, c, fileExists)
+	result := CheckIfFileExists(CheckAutomaticDependencyUpdate, c, fileExists)
 	if !result.Pass {
 		result.Confidence = 3
 	}
--- a/checks/binary_artifact.go
+++ b/checks/binary_artifact.go
@@ -1,0 +1,93 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checks
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/h2non/filetype"
+	"github.com/h2non/filetype/types"
+
+	"github.com/ossf/scorecard/checker"
+)
+
+//nolint
+func init() {
+	registerCheck(CheckBinaryArtifacts, BinaryArtifacts)
+}
+
+const CheckBinaryArtifacts string = "Binary-Artifacts"
+
+// BinaryArtifacts  will check the repository if it contains binary artifacts.
+func BinaryArtifacts(c *checker.CheckRequest) checker.CheckResult {
+	return CheckFilesContent(CheckBinaryArtifacts, "*", false, c, checkBinaryFileContent)
+}
+
+func checkBinaryFileContent(path string, content []byte,
+	logf func(s string, f ...interface{})) (bool, error) {
+	binaryFileTypes := map[string]bool{
+		"crx":     true,
+		"deb":     true,
+		"dex":     true,
+		"dey":     true,
+		"elf":     true,
+		"bin":     true,
+		"o":       true,
+		"so":      true,
+		"iso":     true,
+		"class":   true,
+		"jar":     true,
+		"bundle":  true,
+		"dylib":   true,
+		"lib":     true,
+		"msi":     true,
+		"acm":     true,
+		"ax":      true,
+		"cpl":     true,
+		"dll":     true,
+		"drv":     true,
+		"efi":     true,
+		"exe":     true,
+		"mui":     true,
+		"ocx":     true,
+		"scr":     true,
+		"sys":     true,
+		"tsp":     true,
+		"pyc":     true,
+		"pyo":     true,
+		"par":     true,
+		"rpm":     true,
+		"swf":     true,
+		"torrent": true,
+		"cab":     true,
+	}
+	var t types.Type
+	var err error
+	if t, err = filetype.Get(content); err != nil {
+		return false, fmt.Errorf("failed in getting the content type %w", err)
+	}
+
+	if _, ok := binaryFileTypes[t.Extension]; ok {
+		logf("!! binary-artifact %s", path)
+		return false, nil
+	} else if _, ok := binaryFileTypes[filepath.Ext(path)]; ok {
+		// falling back to file based extension.
+		logf("!! binary-artifact %s", path)
+		return false, nil
+	}
+
+	return true, nil
+}
--- a/checks/checks.md
+++ b/checks/checks.md
@@ -23,9 +23,16 @@ This check tries to determine if a project has dependencies automatically update
 **Remediation steps**
 - Signup for automatic dependency updates with dependabot or renovatebot and place the config file in the locations that are recommended by these tools.
 
+## Binary-Artifacts 
+
+This check tries to determine if a project has binary artifacts in the source repository. These binaries could be compromised artifacts.Building from the source is recommended. 
+
+**Remediation steps**
+- Remove the binary artifacts from the repository. 
+
 ## Branch-Protection 
 
-Branch protection allows defining rules to enforce certain workflows for branches, such as requiring a review or passing certain status checks. This check would work only when the token has [Admin access](https://github.community/t/enable-branch-protection-get-api-without-admin/14197) to the repository. This check determines if the default and release branches are protected. More specifically, the checks for AllowForcePushes (disabled), AllowDeletions (disabled),  EnforceAdmins (enabled), RequireLinearHistory (enabled), RequiredStatusChecks (enabled and must have non-empty context enabled), RequiredPullRequestReviews (>=1), DismissStaleReviews (enabled), RequireCodeOwnerReviews (enabled).
+Branch protection allows defining rules to enforce certain workflows for branches, such as requiring a review or passing certain status checks. This check would work only when the token has [Admin access](https://github.community/t/enable-branch-protection-get-api-without-admin/14197) to the repository. This check determines if the default and release branches are protected. More specifically, the checks for AllowForcePushes (disabled), AllowDeletions (disabled),  EnforceAdmins (enabled), RequireLinearHistory (enabled), RequiredStatusChecks (enabled and must have non-empty context enabled), RequiredPullRequestReviews (>=1), DismissStaleReviews (enabled), RequireCodeOwnerReviews (enabled). 
 
 **Remediation steps**
 - Enable branch protection settings in your source hosting provider to avoid force pushes or deletion of your important branches.
--- a/checks/checks.yaml
+++ b/checks/checks.yaml
@@ -51,6 +51,13 @@ checks:
       - >-
         Signup for automatic dependency updates with dependabot or renovatebot and place the config
         file in the locations that are recommended by these tools.
+  Binary-Artifacts:
+    description: >-
+      This check tries to determine if a project has binary artifacts in the source repository.
+      These binaries could be compromised artifacts.Building from the source is recommended.
+    remediation:
+      - >-
+        Remove the binary artifacts from the repository. 
   Contributors:
     description: >-
       This check tries to determine if a project has a set of contributors from
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ bin
 cache
 output
 gitfolder.*
+githubrepo.tar.gz
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/google/go-github/v32 v32.1.0
 	github.com/google/go-github/v33 v33.0.0
 	github.com/google/go-querystring v1.1.0 // indirect
+	github.com/h2non/filetype v1.1.1
 	github.com/jszwec/csvutil v1.5.0
 	github.com/moby/buildkit v0.8.3
 	github.com/naveensrinivasan/httpcache v1.2.2
--- a/go.sum
+++ b/go.sum
@@ -736,6 +736,8 @@ github.com/grpc-ecosystem/grpc-gateway v1.9.2/go.mod h1:vNeuVxBJEsws4ogUvrchl83t
 github.com/grpc-ecosystem/grpc-gateway v1.9.5/go.mod h1:vNeuVxBJEsws4ogUvrchl83t/GYV9WGTSLVdBhOQFDY=
 github.com/grpc-ecosystem/grpc-gateway v1.12.1/go.mod h1:8XEsbTttt/W+VvjtQhLACqCisSPWTxCZ7sBRjU6iH9c=
 github.com/grpc-ecosystem/grpc-opentracing v0.0.0-20180507213350-8e809c8a8645/go.mod h1:6iZfnjpejD4L/4DwD7NryNaJyCQdzwWwH2MWhCA90Kw=
+github.com/h2non/filetype v1.1.1 h1:xvOwnXKAckvtLWsN398qS9QhlxlnVXBjXBydK2/UFB4=
+github.com/h2non/filetype v1.1.1/go.mod h1:319b3zT68BvV+WRj7cwy856M2ehB3HqNOt6sy1HndBY=
 github.com/hanwen/go-fuse v1.0.0/go.mod h1:unqXarDXqzAk0rt98O2tVndEPIpUgLD9+rwFisZH3Ok=
 github.com/hanwen/go-fuse/v2 v2.0.3/go.mod h1:0EQM6aH2ctVpvZ6a+onrQ/vaykxh2GH7hy3e13vzTUY=
 github.com/hashicorp/consul/api v1.1.0/go.mod h1:VmuI/Lkw1nC05EYQWNKwWGbkg+FbDBtguAZLlVdkD9Q=


inferno-chromium:
too small of a list for binary extensions, any better library or a better list we can pull from somewhere.

checks/binary_artifact.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"strings"
+
+	"github.com/ossf/scorecard/checker"
+)
+
+//nolint
+func init() {
+	registerCheck("Binary-Artifact", BinaryArtifacts)
```
azeemshaikh38, line 25: Export `Binary-Artifact` string like in other checks.

azeemshaikh38:
> too small of a list for binary extensions, any better library or a better list we can pull from somewhere.

+1. Weren't we considering using https://github.com/h2non/filetype?

checks/binary_artifact.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// BinaryArtifacts  will check the repository if it contains binary artifacts.
+func BinaryArtifacts(c *checker.CheckRequest) checker.CheckResult {
+	return CheckIfFileExists(check, c, binaryfileExists)
```
azeemshaikh38, line 30: `check` is declared here - https://github.com/ossf/scorecard/blob/main/checks/automatic_dependency_update.go#L23, we shouldn't be using that. 

On another note, could you also update `automatic_dependency_update` to not use  a generic var like `check` and instead update it to a exported var?

naveensrinivasan:
> > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> 
> +1. Weren't we considering using https://github.com/h2non/filetype?

How about we start off with this list? https://github.com/h2non/filetype#archive

azeemshaikh38:
> > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > 
> > 
> > +1. Weren't we considering using https://github.com/h2non/filetype?
> 
> How about we start off with this list? https://github.com/h2non/filetype#archive

Sounds good.

checks/binary_artifact.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2020 Security Scorecard Authors
```
laurentsimon, line 1: 2021

checks/checks.md
```diff
@@ -23,6 +23,13 @@ This check tries to determine if a project has dependencies automatically update
 **Remediation steps**
 - Signup for automatic dependency updates with dependabot or renovatebot and place the config file in the locations that are recommended by these tools.
 
+## Binary-Artifacts 
+
+This check tries to determine if a project has binary artifacts in the source repository. The files with following extensions are checked ".o", ".so", ".ko", ".elf", ".exe", ".dll", ".deb",  ".rpm", ".jar", ".pyc". These binaries could be compromised artifacts. 
+
+**Remediation steps**
+- The remove the binary artifacts from the repository.
```
laurentsimon, line 31: s/The remove/Remove/

inferno-chromium:
> > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > 
> > > > > 
> > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > 
> > > > 
> > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > 
> > > 
> > > Sounds good.
> > 
> > 
> > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> 
> here is a list of extensions that are executables: "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab".
> 
> Here's a list of image magic's types we could aim for (maybe next PR) "CAB_ARCHIVE", "CRX", "DEB", "DEX", "DEY", "ELF", "ISO", "JAR", "JAVA_BYTE_CODE", "MACH_O", "MSI", "PEBIN", "PE_NSIS", "PYTHON_BYTECODE", "PYTHON_PAR", "RPM", "SWF", "TORRENT"

https://golang.org/pkg/mime/#TypeByExtension uses /etc/mime.types when available and that list is crazy big (my machine > 2k entries)

checks/checks.yaml
```diff
@@ -51,6 +51,14 @@ checks:
       - >-
         Signup for automatic dependency updates with dependabot or renovatebot and place the config
         file in the locations that are recommended by these tools.
+  Binary-Artifacts:
+    description: >-
+      This check tries to determine if a project has binary artifacts in the source repository.
+      The files with following extensions are checked ".o", ".so", ".ko", ".elf", ".exe", ".dll", ".deb", 
+      ".rpm", ".jar", ".pyc". These binaries could be compromised artifacts.
```
laurentsimon, line 58: we can also add that binaries don't provide provenance information and are hard to patch when vulnerabilities are discovered - since we don't know what they contain.

checks/binary_artifact.go
naveensrinivasan: 👍

azeemshaikh38:
> > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > 
> > > > 
> > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > 
> > > 
> > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > 
> > 
> > Sounds good.
> 
> Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?

IIUC, with https://github.com/h2non/filetype we shouldn't need `TypeByExtension` since the library looks at the file content itself and not filename extension to decide its `FileType`.

laurentsimon:
> > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > 
> > > > > > 
> > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > 
> > > > > 
> > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > 
> > > > 
> > > > Sounds good.
> > > 
> > > 
> > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > 
> > 
> > here is a list of extensions that are executables: "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab".
> > Here's a list of image magic's types we could aim for (maybe next PR) "CAB_ARCHIVE", "CRX", "DEB", "DEX", "DEY", "ELF", "ISO", "JAR", "JAVA_BYTE_CODE", "MACH_O", "MSI", "PEBIN", "PE_NSIS", "PYTHON_BYTECODE", "PYTHON_PAR", "RPM", "SWF", "TORRENT"
> 
> https://golang.org/pkg/mime/#TypeByExtension uses /etc/mime.types when available and that list is crazy big (my machine > 2k entries). but here is the problem, like having images, video is common in repos. so yes just need binary extensions

what about svg/pdf files? Do we consider these binaries? They can contain javascripts. I think this would create many false positives, though. So I don't think we can include them.

I also thought of language serialization formats: pickle, etc but I think there's no standard naming and it's better to catch these via SATS by looking for deserialization APIs

checks/checks.yaml
azeemshaikh38, line 58: IMO, we shouldn't mention the extensions/binaries we look for. This can easily change in the future. Also, this is tied to the implementation detail (looking for extensions vs. looking at file content) which can also change. Instead, let's just mention the high-level issue here (as @laurentsimon has suggested).

laurentsimon:
> > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > 
> > > > > 
> > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > 
> > > > 
> > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > 
> > > 
> > > Sounds good.
> > 
> > 
> > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> 
> IIUC, with https://github.com/h2non/filetype we shouldn't need `TypeByExtension` since the library looks at the file content itself and not filename extension to decide its `FileType`.

we won't need `TypeByExtension` but I think we still need an extension list since `h2non/filetype` may not support everything. At list that gives us some redundancy in case.

checks/checks.yaml
inferno-chromium: +1, lets not put this in docs.

naveensrinivasan:
> > > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > > 
> > > > > > > 
> > > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > > 
> > > > > > 
> > > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > > 
> > > > > 
> > > > > Sounds good.
> > > > 
> > > > 
> > > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > > 
> > > 
> > > here is a list of extensions that are executables: "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab".
> > > Here's a list of image magic's types we could aim for (maybe next PR) "CAB_ARCHIVE", "CRX", "DEB", "DEX", "DEY", "ELF", "ISO", "JAR", "JAVA_BYTE_CODE", "MACH_O", "MSI", "PEBIN", "PE_NSIS", "PYTHON_BYTECODE", "PYTHON_PAR", "RPM", "SWF", "TORRENT"
> > 
> > 
> > https://golang.org/pkg/mime/#TypeByExtension uses /etc/mime.types when available and that list is crazy big (my machine > 2k entries). but here is the problem, like having images, video is common in repos. so yes just need binary extensions
> 
> what about svg/pdf files? Do we consider these binaries? They can contain javascripts. I think this would create many false positives, though. So I don't think we can include them.
> 
> I also thought of language serialization formats: pickle, etc but I think there's no standard naming and it's better to catch these via SATS by looking for deserialization APIs

How about this? This is based on https://github.com/h2non/filetype but removed pdf and rtf
```
     "epub", "zip", "tar", "rar", "gz", "bz2", "7z", "xz", "zstd",
      "exe", "swf",  "iso", "eot", "ps", "sqlite", "nes", "crx",
    "cab", "deb", "ar", "Z", "lz", "rpm", "elf", "dcm":
  ```

**Comment by inferno-chromium** [comment]
> > > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > > 
> > > > > > > 
> > > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > > 
> > > > > > 
> > > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > > 
> > > > > 
> > > > > Sounds good.
> > > > 
> > > > 
> > > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > > 
> > > 
> > > here is a list of extensions that are executables: "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab".
> > > Here's a list of image magic's types we could aim for (maybe next PR) "CAB_ARCHIVE", "CRX", "DEB", "DEX", "DEY", "ELF", "ISO", "JAR", "JAVA_BYTE_CODE", "MACH_O", "MSI", "PEBIN", "PE_NSIS", "PYTHON_BYTECODE", "PYTHON_PAR", "RPM", "SWF", "TORRENT"
> > 
> > 
> > https://golang.org/pkg/mime/#TypeByExtension uses /etc/mime.types when available and that list is crazy big (my machine > 2k entries). but here is the problem, like having images, video is common in repos. so yes just need binary extensions
> 
> what about svg/pdf files? Do we consider these binaries? They can contain javascripts. I think this would create many false positives, though. So I don't think we can include them.
> 
> I also thought of language serialization formats: pickle, etc but I think there's no standard naming and it's better to catch these via SATS by looking for deserialization APIs

taking a step back, thinking the goal of this check is to prevent built artifacts from being checked in source, like exe, dll, so, etc. so maybe just a list of those extensions is good enough ? like the list that laurent pointed. we cant exclude image, videos, pdf, svg, icons, etc since they dont pose risk. thoughts - @azeemshaikh38 @naveensrinivasan @laurentsimon

**Comment by azeemshaikh38** [comment]
> > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > 
> > > > > > 
> > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > 
> > > > > 
> > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > 
> > > > 
> > > > Sounds good.
> > > 
> > > 
> > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > 
> > 
> > IIUC, with https://github.com/h2non/filetype we shouldn't need `TypeByExtension` since the library looks at the file content itself and not filename extension to decide its `FileType`.
> 
> we won't need `TypeByExtension` but I think we still need an extension list since `h2non/filetype` may not support everything. At list that gives us some redundancy in case.

`h2non/filetype` allows for custom matchers - https://github.com/h2non/filetype#add-additional-file-type-matchers. So we can easily add new extensions if needed. My point being, we can start with something as simple as below without needing more logic/libraries:

```
if filetype.Match(filecontent, []filtype.Types{<list of extensions we fail on>}) {
  return Fail
}
return Pass
```

**Comment by azeemshaikh38** [comment]
> > > > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > > > 
> > > > > > > 
> > > > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > > > 
> > > > > > 
> > > > > > Sounds good.
> > > > > 
> > > > > 
> > > > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > > > 
> > > > 
> > > > here is a list of extensions that are executables: "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab".
> > > > Here's a list of image magic's types we could aim for (maybe next PR) "CAB_ARCHIVE", "CRX", "DEB", "DEX", "DEY", "ELF", "ISO", "JAR", "JAVA_BYTE_CODE", "MACH_O", "MSI", "PEBIN", "PE_NSIS", "PYTHON_BYTECODE", "PYTHON_PAR", "RPM", "SWF", "TORRENT"
> > > 
> > > 
> > > https://golang.org/pkg/mime/#TypeByExtension uses /etc/mime.types when available and that list is crazy big (my machine > 2k entries). but here is the problem, like having images, video is common in repos. so yes just need binary extensions
> > 
> > 
> > what about svg/pdf files? Do we consider these binaries? They can contain javascripts. I think this would create many false positives, though. So I don't think we can include them.
> > I also thought of language serialization formats: pickle, etc but I think there's no standard naming and it's better to catch these via SATS by looking for deserialization APIs
> 
> taking a step back, thinking the goal of this check is to prevent built artifacts from being checked in source, like exe, dll, so, etc. so maybe just a list of those extensions is good enough ? like the list that laurent pointed. we cant exclude image, videos, pdf, svg, icons, etc since they dont pose risk. thoughts - @azeemshaikh38 @naveensrinivasan @laurentsimon

+1. This is what I meant in my other comment. We can start with a handful of extensions that we fail on for our first iteration and improve on this over time.

**Comment by laurentsimon** [comment]
> > > > > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > > > > 
> > > > > > > > 
> > > > > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > > > > 
> > > > > > > 
> > > > > > > Sounds good.
> > > > > > 
> > > > > > 
> > > > > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > > > > 
> > > > > 
> > > > > here is a list of extensions that are executables: "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab".
> > > > > Here's a list of image magic's types we could aim for (maybe next PR) "CAB_ARCHIVE", "CRX", "DEB", "DEX", "DEY", "ELF", "ISO", "JAR", "JAVA_BYTE_CODE", "MACH_O", "MSI", "PEBIN", "PE_NSIS", "PYTHON_BYTECODE", "PYTHON_PAR", "RPM", "SWF", "TORRENT"
> > > > 
> > > > 
> > > > https://golang.org/pkg/mime/#TypeByExtension uses /etc/mime.types when available and that list is crazy big (my machine > 2k entries). but here is the problem, like having images, video is common in repos. so yes just need binary extensions
> > > 
> > > 
> > > what about svg/pdf files? Do we consider these binaries? They can contain javascripts. I think this would create many false positives, though. So I don't think we can include them.
> > > I also thought of language serialization formats: pickle, etc but I think there's no standard naming and it's better to catch these via SATS by looking for deserialization APIs
> > 
> > 
> > taking a step back, thinking the goal of this check is to prevent built artifacts from being checked in source, like exe, dll, so, etc. so maybe just a list of those extensions is good enough ? like the list that laurent pointed. we cant exclude image, videos, pdf, svg, icons, etc since they dont pose risk. thoughts - @azeemshaikh38 @naveensrinivasan @laurentsimon
> 
> +1. This is what I meant in my other comment. We can start with a handful of extensions that we fail on for our first iteration and improve on this over time.

+1.

**Comment by naveensrinivasan** [comment]
> > > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > > 
> > > > > > > 
> > > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > > 
> > > > > > 
> > > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > > 
> > > > > 
> > > > > Sounds good.
> > > > 
> > > > 
> > > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > > 
> > > 
> > > IIUC, with https://github.com/h2non/filetype we shouldn't need `TypeByExtension` since the library looks at the file content itself and not filename extension to decide its `FileType`.
> > 
> > 
> > we won't need `TypeByExtension` but I think we still need an extension list since `h2non/filetype` may not support everything. At list that gives us some redundancy in case.
> 
> `h2non/filetype` allows for custom matchers - https://github.com/h2non/filetype#add-additional-file-type-matchers. So we can easily add new extensions if needed. My point being, we can start with something as simple as below without needing more logic/libraries:
> 

> ```
> if filetype.Match(filecontent, []filtype.Types{<list of extensions we fail on>}) {
>   return Fail
> }
> return Pass
> ```

OK, going to use this as a list without the actual filecontent. 
"crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab"

Agreed?

**Comment by inferno-chromium** [comment]
> > > > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > > > 
> > > > > > > 
> > > > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > > > 
> > > > > > 
> > > > > > Sounds good.
> > > > > 
> > > > > 
> > > > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > > > 
> > > > 
> > > > IIUC, with https://github.com/h2non/filetype we shouldn't need `TypeByExtension` since the library looks at the file content itself and not filename extension to decide its `FileType`.
> > > 
> > > 
> > > we won't need `TypeByExtension` but I think we still need an extension list since `h2non/filetype` may not support everything. At list that gives us some redundancy in case.
> > 
> > 
> > `h2non/filetype` allows for custom matchers - https://github.com/h2non/filetype#add-additional-file-type-matchers. So we can easily add new extensions if needed. My point being, we can start with something as simple as below without needing more logic/libraries:
> 
> > ```
> > if filetype.Match(filecontent, []filtype.Types{<list of extensions we fail on>}) {
> >   return Fail
> > }
> > return Pass
> > ```
> 
> OK, going to use this as a list without the actual filecontent.
> "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab"
> 
> Agreed?

sgtm!

**Comment by naveensrinivasan** [comment]
> > > > > > > > > > too small of a list for binary extensions, any better library or a better list we can pull from somewhere.
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > +1. Weren't we considering using https://github.com/h2non/filetype?
> > > > > > > > 
> > > > > > > > 
> > > > > > > > How about we start off with this list? https://github.com/h2non/filetype#archive
> > > > > > > 
> > > > > > > 
> > > > > > > Sounds good.
> > > > > > 
> > > > > > 
> > > > > > Also, i just saw https://golang.org/pkg/mime/#TypeByExtension , we can even try that and exclude text/ ?
> > > > > 
> > > > > 
> > > > > IIUC, with https://github.com/h2non/filetype we shouldn't need `TypeByExtension` since the library looks at the file content itself and not filename extension to decide its `FileType`.
> > > > 
> > > > 
> > > > we won't need `TypeByExtension` but I think we still need an extension list since `h2non/filetype` may not support everything. At list that gives us some redundancy in case.
> > > 
> > > 
> > > `h2non/filetype` allows for custom matchers - https://github.com/h2non/filetype#add-additional-file-type-matchers. So we can easily add new extensions if needed. My point being, we can start with something as simple as below without needing more logic/libraries:
> > 
> > 
> > > ```
> > > if filetype.Match(filecontent, []filtype.Types{<list of extensions we fail on>}) {
> > >   return Fail
> > > }
> > > return Pass
> > > ```
> > 
> > 
> > OK, going to use this as a list without the actual filecontent.
> > "crx", "deb", "dex", "dey", "elf", "bin", "o", "so", "iso", "class", "jar", "bundle", "dylib", "lib", "msi", "acm", "ax", "cpl", "dll", "drv", "efi", "exe", "mui", "ocx", "scr", "sys", "tsp", "pyc", "pyo", "par", "rpm", "swf", "torrent", "cab"
> > Agreed?
> 
> sgtm! using the filetype lib, we should add some fileheader magic

for doing that we would have to Duplicate the `CheckIfFileExists` tar reading into another func and which will read the file content and making it generic enough to pass the contents to the validating func with fileheader magic.

**Comment by azeemshaikh38** [comment]
> for doing that we would have to Duplicate the `CheckIfFileExists` tar reading into another func and which will read the file content and making it generic enough to pass the contents to the validating func with fileheader magic.

Don't we already have `CheckFilesContent` for this? We just need to call `CheckFilesContent` with the callback function which uses `filetype` lib?

**Comment by naveensrinivasan** [comment]
> CheckFilesContent

I agree we already have `CheckFilesContent`. 

But checking the content for every file will slow a lot in large repositories.

**Comment by azeemshaikh38** [comment]
> > CheckFilesContent
> 
> I agree we already have `CheckFilesContent`.
> 
> But checking the content for every file will slow a lot in large repositories.

Agreed, but we are already doing that today in multiple checks so I think adding this is fine. Also, the actual `filetype.Match()` call may not be too compute expensive, so I'm not too worried.

**Comment by inferno-chromium** [comment]
friendly ping @naveensrinivasan , we want to do scorecards v2 blog post next-to-next week, this will be a good check to mention there. Thanks a lot!

**Comment by naveensrinivasan** [comment]
> friendly ping @naveensrinivasan , we want to do scorecards v2 blog post next-to-next week, this will be a good check to mention there. Thanks a lot!

I apologize for the delay. I will get this done by the end of this week. Thanks for the reminder!

File: checks/automatic_dependency_update.go [hunk 6]
```diff
@@ -20,16 +20,16 @@ import (
 	"github.com/ossf/scorecard/checker"
 )
 
-const check = "Automatic-Dependency-Update"
+const CheckDependencyUpdate = "Automatic-Dependency-Update"
```
inferno-chromium, line 23: why not AutomaticDependencyUpdateCheck, reads better.

checks/binary_artifact.go
```diff
@@ -0,0 +1,58 @@
...
+		"scr": true, "sys": true, "tsp": true, "pyc": true, "pyo": true, "par": true,
+		"rpm": true, "swf": true, "torrent": true, "cab": true,
+	}
+	var t types.Type
+	var err error
+	if t, err = filetype.Get(content); err != nil {
+		return false, fmt.Errorf("failed in getting the content type %w", err)
+	}
+	if _, ok := binaryFileTypes[t.Extension]; ok {
+		logf("!! binary-artifacts %s", path)
```
inferno-chromium, line 54: s/binary-artifacts/found binary artifact 
(without s)

checks/binary_artifact.go
```diff
@@ -0,0 +1,58 @@
...
+		"elf": true, "bin": true, "o": true, "so": true, "iso": true, "class": true, "jar": true,
+		"bundle": true, "dylib": true, "lib": true, "msi": true, "acm": true, "ax": true,
+		"cpl": true, "dll": true, "drv": true, "efi": true, "exe": true, "mui": true, "ocx": true,
+		"scr": true, "sys": true, "tsp": true, "pyc": true, "pyo": true, "par": true,
+		"rpm": true, "swf": true, "torrent": true, "cab": true,
+	}
+	var t types.Type
+	var err error
+	if t, err = filetype.Get(content); err != nil {
+		return false, fmt.Errorf("failed in getting the content type %w", err)
```
inferno-chromium, line 51: Should we fallback to actual file extension here instead of failing. Reason is some content matchers are not in filetype e.g. iso, etc, so we should still try the direct extension.

Also, should leave a TODO to add file content matchers for those types that are not in filetype, e.g. .so, .iso.

checks/automatic_dependency_update.go
azeemshaikh38: The convention we've been using so far is to have check names start with `Check`. I'd vote for keeping the name same or probably making it `CheckAutomaticDependencyUpdate`.

checks/binary_artifact.go
```diff
@@ -0,0 +1,58 @@
...
+const CheckBinaryArtifacts string = "Binary-Artifacts"
+
+// BinaryArtifacts  will check the repository if it contains binary artifacts.
+func BinaryArtifacts(c *checker.CheckRequest) checker.CheckResult {
+	return CheckFilesContent(CheckBinaryArtifacts, "*", false, c, checkBinaryFileContent)
+}
+
+func checkBinaryFileContent(path string, content []byte,
+	logf func(s string, f ...interface{})) (bool, error) {
+	binaryFileTypes := map[string]bool{
```
azeemshaikh38, line 40: Make this a const defined out the fn definition. Also, maybe specify one filetype per-line for better readability:

```
const binaryFileTypes := map[string]bool {
  "crx": true,
  "deb": true,
  ...
}
```

azeemshaikh38 (approved):

checks/binary_artifact.go
naveensrinivasan: golang does not allow const maps. https://www.reddit.com/r/golang/comments/993ujz/why_is_not_possible_to_create_a_constant_map/

naveensrinivasan: formatting the filetype per line. Thanks

checks/binary_artifact.go
naveensrinivasan: I agree with the suggestion of falling back to file-based extension checks. I will add that.

I am not in favor of adding TODO's because most of the time people don't get to it. I am going to open an issue for it to be solved.

checks/binary_artifact.go
naveensrinivasan: Thanks!

checks/automatic_dependency_update.go
naveensrinivasan: I am going with Azeem's suggestion to keep it consistent.  Thanks

inferno-chromium: SGTM

checks/binary_artifact.go
inferno-chromium: as more people contribute to it, todos should get fixed. opening an issue is fine as well.

naveensrinivasan: I am opening issues with "help wanted" so that we can attract more new contributors.

naveensrinivasan: https://github.com/ossf/scorecard/issues/596

checks/checks.md
```diff
@@ -23,6 +23,13 @@ This check tries to determine if a project has dependencies automatically update
 **Remediation steps**
 - Signup for automatic dependency updates with dependabot or renovatebot and place the config file in the locations that are recommended by these tools.
 
+## Binary-Artifacts 
+
+This check tries to determine if a project has binary artifacts in the source repository. These binaries could be compromised artifacts.
```
inferno-chromium, line 28: can add that building from source is recommended.

inferno-chromium (approved):
Thanks!

checks/checks.md
naveensrinivasan: I will include that. Thanks

checks/checks.yaml
```diff
@@ -2,6 +2,8 @@
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
+		logf("!! binary-artifact %s", path)
```
inferno-chromium, line 5: weird paste ?

naveensrinivasan: OMG!! 🤦

naveensrinivasan: Thank you! took care of it.

inferno-chromium:
Should we disable that integration test so that we can merge this. 
@laurentsimon on the integration test failure in https://github.com/ossf/scorecard/pull/563/checks?check_run_id=2876786197

naveensrinivasan:
> Should we disable that integration test so that we can merge this.
> @laurentsimon on the integration test failure in https://github.com/ossf/scorecard/pull/563/checks?check_run_id=2876786197

The check has already been disabled. We can merge this.

naveensrinivasan:
> > Should we disable that integration test so that we can merge this.
> > @laurentsimon on the integration test failure in https://github.com/ossf/scorecard/pull/563/checks?check_run_id=2876786197
> 
> The check has already been disabled. We can merge this.

Build/validate should succeed to merge.

PR #576, opened 2021-06-14
✨ Add CPU profile to cron worker

azeemshaikh38:

Adds CPU profile tracking to Cron worker.

No.

### DIFF
--- a/cron/worker/main.go
+++ b/cron/worker/main.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"runtime/pprof"
 
 	"github.com/google/go-github/v32/github"
 	"github.com/shurcooL/githubv4"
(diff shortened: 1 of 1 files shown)


cron/worker/main.go
```diff
@@ -79,11 +85,18 @@ func processRequest(ctx context.Context,
...
 		return fmt.Errorf("error during WriteToBlobStore: %w", err)
 	}
 	log.Printf("Write to shard file successful: %s", filename)
+
+	pprof.StopCPUProfile()
+	cpuProfFilename := data.GetBlobFilename(
+		fmt.Sprintf("pprof.shard-%05d.cpu", batchRequest.GetShardNum()),
+		batchRequest.GetJobTime().AsTime())
+	if err := data.WriteToBlobStore(ctx, bucketURL, cpuProfFilename, cpuProf.Bytes()); err != nil {
+		return fmt.Errorf("error during WriteToBlobStore: %w", err)
```
inferno-chromium, line 98: sometime in future, good to think about these error message, e.g. WriteToBlobStore is called in multiple places, so need to add more specific content in them, that it is cpu profiler specific etc.

inferno-chromium (approved):

cron/worker/main.go
azeemshaikh38: Acknowledged. Overall, the way we bubble up errors and log them needs to be improved.

PR #577, opened 2021-06-14
✨ Use RollingUpdate strategy for cron worker job

azeemshaikh38:

Adds `RollingUpdate` strategy to cron worker job to allow for auto updates.

No.

### DIFF
--- a/cron/k8s/worker.yaml
+++ b/cron/k8s/worker.yaml
@@ -47,3 +47,8 @@ spec:
           limits:
             cpu: 12
             memory: 12Gi
+  strategy:
+    type: "RollingUpdate"
+    rollingUpdate:
+      maxUnavailable: 1
+      maxSurge: 1


cron/k8s/worker.yaml
```diff
@@ -47,3 +47,8 @@ spec:
           limits:
             cpu: 12
             memory: 12Gi
+  strategy:
+    type: "RollingUpdate"
+    rollingUpdate:
+      maxUnavailable: 1
```
naveensrinivasan, line 53: If there is only one replica this will not work. We need at least 2 to rolling update to work.

azeemshaikh38: Doesn't this imply a `n+1` RollingUpdate strategy here? Basically, what I want to have here is a strategy that brings up a new instance (`maxSurge` = 1) and kills an older one (`maxUnavailable` = 1).

Let me know if this isn't doing what I expect it to do and what change I need to make here for it work that way.

naveensrinivasan: Have you tried this? I am not aware of the max surge with this option. As long as you have tried it and it works I am fine.

azeemshaikh38: I tried it through the UI and it worked. But then again, we only have 1 worker instance running. 

Maybe we keep this as is for now? My main aim is unblock auto deployment of new binaries when they become available. If the behavior breaks with >1 worker instances, we can revisit then. Wdyt?

naveensrinivasan: Sounds good.

naveensrinivasan (approved):

Issue #403, opened 2021-05-05
Frozen Deps does not check frozen dockerfile image use, CI/CD tools


inferno-chromium:
**Is your feature request related to a problem? Please describe.**
1) Check for either hardcoded sha in FROM.*dockerfile use or best to see if signature verification happens (with sigstore)
2) for CI/CD discourage bad patterns, e.g. curl | bash

**Describe the solution you'd like**
Improve frozen deps check to account for these.

**Describe alternatives you've considered**
NA

**Additional context**
NA

inferno-chromium:
Needs documentation update as well.

dlorenc:
Here's one tool to help pin Dockerfiles: https://github.com/safe-waters/docker-lock

Another WIP one, designed to help with signatures: https://github.com/sigstore/cosign/pull/189

laurentsimon:
1 is done as #469 
2 `(curl | bash)` is going to take a little bit more work. We'll use [shellcheck](https://www.shellcheck.net/) for shell scripts, I think. We need to check bad patterns not only in shell scripts, but also in makefiles, and even in code `system(...)`, `os.popen()`, so we need some language support for this one.
I'll start with shell script.

inferno-chromium:
> 1 is done as #469
> 2 `(curl | bash)` is going to take a little bit more work. We'll use [shellcheck](https://www.shellcheck.net/) for shell scripts, I think. We need to check bad patterns not only in shell scripts, but also in makefiles, and even in code `system(...)`, `os.popen()`, so we need some language support for this one.
> I'll start with shell script.

+1 shell script is a good start.

laurentsimon:
@inferno-chromium @oliverchang @naveensrinivasan I'm starting to work on `2. repo clones without pinning, e.g. from github.`
I worry this is going to throw false positives whenever CI/CD runs integration tests or releases - scripts will fetch the latest code. So I'm thinking of detecting if the `git clone reponame` is for the same repo scorecard is analyzing or not.
WDUT?

inferno-chromium:
> @inferno-chromium @oliverchang @naveensrinivasan I'm starting to work on `2. repo clones without pinning, e.g. from github.`
> I worry this is going to throw false positives whenever CI/CD runs integration tests or releases - scripts will fetch the latest code. So I'm thinking of detecting if the `git clone reponame` is for the same repo scorecard is analyzing or not.
> WDUT?

[...]

PR #584, opened 2021-06-17
✨ Check: detect downloads of scripts/binaries in docker's RUN

laurentsimon:

feature. Addresses part of #403 

we don't do detections.

We detect the download of scripts/binaries (wget, curl, gsutil, aws) and executed in the `RUN` command.
The check works across `RUN` commands, e.g. `RUN curl xxx > /tmp/file` and `RUN /tmp/file`
We also support the use of `sudo`. We try to detect various constructs:
`curl | bash`, `curl > /tmp/file && /tmp/file`, `curl > /tmp/file; bash /tmp/file`, `curl > /tmp/file; /tmp/file`, `curl > /tmp/file && 
 /tmp/file`, `bash <(wget -qO- http://website.com/my-script.sh)`, 

no

* **Other information**:
No e2e test in this PR. I'll add them in the final PR when's it's LGTM'ed.


checks/frozen_deps.go
```diff
@@ -46,14 +46,102 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+		if cmdType != "run" {
+			continue
+		}
+
+		var valueList []string
+		for n := child.Next; n != nil; n = n.Next {
+			valueList = append(valueList, n.Value)
+		}
+
+		if len(valueList) != 1 {
```
oliverchang, line 83: Does this capture both: 

```
RUN bash blah...
```

and 

```
RUN ["/bin/bash", "-c", "..."] 
```

?

checks/frozen_deps.go
```diff
@@ -46,14 +46,102 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+			return false, ErrParsingDockerfile
+		}
+
+		// Validate it's not downloading and piping into a shell, like
+		// `curl | bash` (supports `sudo`).
+		r, err := validateCommandIsNotFetchPipeExecute(valueList[0], pathfn, logf)
+		if err != nil {
+			return false, err
+		} else if !r {
```
oliverchang, line 92: nit: not need for else after return.

checks/frozen_deps.go
```diff
@@ -46,14 +46,102 @@ func FrozenDeps(c *checker.CheckRequest) checker.CheckResult {
...
+		r, err = validateCommandIsNotFetchToStdinExecute(valueList[0], pathfn, logf)
+		if err != nil {
+			return false, err
+		} else if !r {
+			ret = false
+		}
+
+		// TODO(laurent): add check for cat file | bash
+		// TODO(laurent): detect downloads of zip/tar files containing scripts.
+		// TODO(laurent): detect command being an env variable
```
oliverchang, line 120: I think we should just try to capture the most common anti patterns, rather than trying to detect every single possible case, which doesn't seem too feasible or maintainable.

oliverchang (commented):
nice! just a cursory look so far, will take a closer look later too.

azeemshaikh38 (approved):

checks/frozen_deps.go
laurentsimon: Thanks. I did not even know this was valid for `RUN` command. Added support.

checks/frozen_deps.go
laurentsimon: agreed. I've added them for future enhancements.

checks/frozen_deps.go
laurentsimon: I thought it was cleaner because it groups the checks and makes it more obvious both checks are for `validateCommandIsNotFetchPipeExecute`. Let me know if you'd rather have me change it anyway.

laurentsimon:
@inferno-chromium please merge. `security/snyk` is failing and I cannot see why (it requires login)

checks/testdata/Dockerfile-curl-file-sh
```diff
@@ -0,0 +1,33 @@
...
+RUN curl exe2 2>&1 > /tmp/exe3 ; \
+    /tmp/exe3
+
+RUN curl file 2>&1 > /tmp/file1 && sh /tmp/file1
+RUN curl file2 2>&1 > /tmp/file2 ; sh /tmp/file2
+RUN curl file2 2>&1 > /tmp/file4 ; sh /tmp/file4
+RUN curl file2 2>&1 > /tmp/file3 ; \
+    bash /tmp/file3
+
+RUN ["curl", "file", "2>&1", ">", "/tmp/file111", "&&", "sh", "/tmp/file111"]
```
oliverchang, line 30: Per https://docs.docker.com/engine/reference/builder/#run

I don't think this is a correct example. This form does not invoke a shell by default, so we'd want to look for something like:

```
RUN ["/bin/bash", "-c", "CHECK THIS"] 
```

instead.

This second form does not invoke a shell by default.

laurentsimon: I see what you mean.
The command example would be invalid and would not work in practice, right? (since redirection implies a shell is running).

I'll update the example. Note that the second form is detected with the current implementation. I simply `Join` the array to a string and give it to the parser. As far as the check is concerned, `RUN []` and `RUN string` are equivalent.

I've taken some other shortcuts in the implementation: it will detect things that can never be valid, but this allowed me to simplify the implementation. Example: `wget <> | bash` would not work because wget saves to a file, yet will be detected by the current implementation. I don't think this should be a problem since the example above should never appear in real scripts.

Let me know what you think though. I can refine things if needed.

laurentsimon: actually, I was wrong: Im not checking for `bash -c ""` at all. I'll send you a followup PR

laurentsimon: see https://github.com/ossf/scorecard/pull/600

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+
+// see https://serverfault.com/questions/226386/wget-a-script-and-run-it/890417.
+func isDownloadUtility(cmd []string) bool {
+	if len(cmd) == 0 {
+		return false
+	}
+	// Note: we won't be catching those if developers have re-named
+	// the utility.
+	// Note: wget -O - <website>, but we don't check for that explicitly.
+	utils := [3]string{"curl", "wget", "gsutil"}
```
oliverchang, line 53: nit: put this as a global (similar to "interpreters")

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+	// https://docs.aws.amazon.com/AmazonS3/latest/userguide/download-objects.html.
+	if isBinaryName("aws", cmd[0]) {
+		if len(cmd) >= 3 && strings.EqualFold("s3api", cmd[1]) && strings.EqualFold("get-object", cmd[2]) {
+			return true
+		}
+	}
+	return false
+}
+
+func getWgetOututFile(cmd []string) (pathfn string, ok bool, err error) {
```
oliverchang, line 70: fix spelling. same with others (e.g. getGsutilOututFile)

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+	f, err := syntax.NewParser().Parse(in, "")
+	if err != nil {
+		return false, ErrParsingShellCommand
+	}
+
+	cmdValidated := true
+	syntax.Walk(f, func(node syntax.Node) bool {
+		// Check if we're calling a file we previously downloaded.
+		if isExecuteFiles(node, cmd, pathfn, files, logf) {
+			cmdValidated = false
```
oliverchang, line 620: do we need to continue walking in this case?

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+			return false
+		} else if b {
+			files[fn] = true
+		}
+
+		// Continue walking the node graph.
+		return true
+	})
+
+	return files, true, err
```
oliverchang, line 605: why do we need the boolean as the second return value? callers can just check `files` right?

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+		if strings.EqualFold("su", c) {
+			isSu = true
+		} else if isSu && strings.HasPrefix(c, "-") && strings.Contains(c, "c") {
+			return true
+		}
+	}
+	return false
+}
+
+func extractSuCommand(args []*syntax.Word) (string, bool) {
```
oliverchang, line 452: some more comments on what the expected input/output is, what the functions are doing in this file would make this much more easier to understand in general.

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+	if !isDownloadUtility(cmd) {
+		return "", false, nil
+	}
+
+	fn, ok := getRedirectFile(ss.Redirs)
+	if !ok {
+		return getOutputFile(cmd)
+	}
+
+	return fn, true, nil
```
oliverchang, line 346: do we need to the boolean at all? can callers just check that `fn != ""` ?

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+		if len(r.Word.Parts) != 1 {
+			continue
+		}
+
+		lit, ok := r.Word.Parts[0].(*syntax.Lit)
+		if ok {
+			return lit.Value, true
+		}
+	}
+	return "", false
```
oliverchang, line 278: boolean not needed? callers can just check return value is not `""` ?

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+	return false
+}
+
+func isInterpreterWithFile(cmd []string, fn string) bool {
+	if len(cmd) == 0 {
+		return false
+	}
+
+	for _, b := range interpreters {
+		if isBinaryName(b, cmd[0]) {
```
oliverchang, line 187: nit: avoid extra indentation for lines 188- by inverting this condition and continuing

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+			}
+
+			// File provided.
+			return pathfn, true, nil
+		}
+	}
+	return "", false, nil
+}
+
+func getAwsOututFile(cmd []string) (pathfn string, ok bool, err error) {
```
oliverchang, line 118: Aws should be AWS per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

checks/shell_download_validate.go
```diff
@@ -0,0 +1,628 @@
...
+	}
+
+	files := make(map[string]bool)
+	syntax.Walk(f, func(node syntax.Node) bool {
+		if err != nil {
+			return false
+		}
+		var b bool
+		var fn string
+		fn, b, err = recordFetchFileFromNode(node)
```
oliverchang, line 594: can just be
```fn, b, err := recordFetchFileFromNode(node)```

Go lets you do this even though `err` was defined previously. Same in other places (e.g. line 550)

oliverchang (commented):
LGTM with some suggestions

checks/shell_download_validate.go
laurentsimon: yes, as I want to provide all offending lines to the user. Otherwise they'll see a single error, fix it.. and scorecard will continue reporting errors.

checks/shell_download_validate.go
laurentsimon: I'll add comments in a follow-up PR. I agree I was light on comments :/

checks/shell_download_validate.go
laurentsimon: the caller could do that, yes. I was not sure what's the `golang` way of doing things. Having a separate `bool` means we're not conflating the output string with the `is found` information. But you're right in this function, the empty string indicates it was not found. Let me know if I should change it.

checks/shell_download_validate.go
laurentsimon: what will be the output of this? Will `err` be an alias to the global `err` previously defined or is it a local variable with the same name? (Note: I've changed this in a. follow-up-CL to look like `fn, b, e := recordFetchFileFromNode(node)`)

checks/shell_download_validate.go
laurentsimon: so I think it may be better to keep the current version, as my concern is the same as https://github.com/golang/go/wiki/CodeReviewComments#in-band-errors. wdut?

checks/shell_download_validate.go
laurentsimon: done

checks/shell_download_validate.go
laurentsimon: done.

checks/shell_download_validate.go
oliverchang: per https://golang.org/ref/spec#Short_variable_declarations

"Redeclaration does not introduce a new variable; it just assigns a new value to the original."

checks/frozen_deps.go
oliverchang: I think it's cleaner to not have the else after if, but your choice since there's no real guideline for Go. You can still group these together by not having a newline in between them.

checks/shell_download_validate.go
oliverchang: I think this still feels unnecessarily complicated, because there is both a boolean and an error. 

are there ambiguities here with empty value being a possibly valid return value? If not, probably cleaner to remove. I think that is the point that https://github.com/golang/go/wiki/CodeReviewComments#in-band-errors is making.

checks/shell_download_validate.go
oliverchang: thanks!

checks/shell_download_validate.go
oliverchang: ah actually in this case, because you're in another block the scope is different and you'll create a new one instead.

Issue #578, opened 2021-06-14
BUG: CheckFileExists/CheckFileContent performance issues 


azeemshaikh38:
The `CheckFileExists` and `CheckFileContent` functions are used by multiple checks.  These fns seem to be really slowing down the Scorecard run though. On some runs these checks can take >1hr to run.

![image](https://user-images.githubusercontent.com/4825078/121938037-2fc3fe80-cd00-11eb-9278-3d4efff0b67d.png)

The CPU profile confirms that the TAR decompression is the real culprit here: 
![image](https://user-images.githubusercontent.com/4825078/121938600-dc05e500-cd00-11eb-882b-e92d33c3f327.png)

We should update these fns (atleast the `CheckFileExists` fn) to use this API endpoint instead - https://api.github.com/repos/ossf/scorecard/git/trees/main?recursive=1. We just need to get the default branch of a repo and then make this call to get a list of all filepaths. More context: https://stackoverflow.com/questions/25022016/get-all-file-names-from-a-github-repo-through-the-github-api

@laurentsimon @naveensrinivasan do either of you have cycles to look into this? We also need a similar solution for `CheckFileContents` too.

inferno-chromium:
if we shift to endpoint, that will work only for github repos, how about other repos if we want to support (gitlab, etc?) or even just directories as input (git-on-borg case). do these checks share the archive or keep redownloading+unpacking it again and again (that could be a nice optimization to do it once per repo). another thing could be cloning the repo in a shared place (e..g cloud filestore) and just updating it from time to time (start of cron or something ?)

azeemshaikh38:
> if we shift to endpoint, that will work only for github repos

Not really, it would be easier to support other usecases with this endpoint in comparison to what we currently have. In terms of a generic `RepoInterface` - today we expect clients to support an equivalent of `GetRepoTarball` call. Instead, with the new endpoint we would expect clients to support `ListFiles` and `GetFileContent` call which is far more generic (especially in a local directories usecase).

> do these checks share the archive or keep redownloading+unpacking it again and again (that could be a nice optimization to do it once per repo)

As of now, they don't share the archive and the unpacking happens repeatedly. We definitely need to optimize this too.

inferno-chromium:
> > if we shift to endpoint, that will work only for github repos
> 
[...]

Issue #462, opened 2021-05-17
Feature: repo/file interface for Scorecard library


azeemshaikh38:
**Is your feature request related to a problem? Please describe.**
Scorecard today is limited to run on GitHub repositories only. We should expand this support to handle local files and other VCS like GitLab, BitBucket etc.

**Describe the solution you'd like**
Create an interface `RepoClient`:

```
type RepoClient interface {
  Commits()
  File()
  ...
}
```

We need to look into our current code and sketch out all the required methods needed in this interface. Update our code bit-by-bit to use this interface instead of directly relying on GitHub API. 

Under the hood use GitHubClient/LocalClient/GitLabClient which implements this interface and can fit clients needs.

azeemshaikh38:
@inferno-chromium @oliverchang @laurentsimon @naveensrinivasan thoughts?

oliverchang:
+1. Sounds great!

inferno-chromium:
+1, this would be awesome, and we definitely need localclient support. some checks might not be applicable when provided a local repo, but we need to add some skip state for those (instead of fail).

naveensrinivasan:
> +1, this would be awesome, and we definitely need localclient support. some checks might not be applicable when provided a local repo, but we need to add some skip state for those (instead of fail).

+1. Great!!

laurentsimon:
+1

PR #587, opened 2021-06-17
✨  Add v0 of RepoClient interface

azeemshaikh38:

Adds [`RepoClient interface`](#462) v0. Also improves upon the performance issues mentioned in #578 

No.

### DIFF
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -1,0 +1,78 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package githubrepo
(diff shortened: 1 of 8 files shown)


clients/repo_client.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clients
+
+type RepoClient interface {
+	InitRepo(owner, repo string) error
+	GetTarball() []byte
```
inferno-chromium, line 19: nit: since we are creating interface, might be nice to use generic name like GetRepoArchive. when you expand to gitlab or others, archive can be tar, zip or whatever. same for client.targz

inferno-chromium (approved):

clients/repo_client.go
azeemshaikh38: Done.

clients/githubrepo/client.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+	defer resp.Body.Close()
+
+	respData, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+	client.archiveData = respData
```
oliverchang, line 65: rather than keeping this in memory, should we think about storing this temporarily on disk ? repos can get quite large (think chromium, linux kernel)

we can even store it on disk uncompressed to avoid constantly decompressing this in the various checks.

PR #590, opened 2021-06-18
🐛 Fix cron worker OOM-ing

azeemshaikh38:

Fixes OOM-ing issues of PubSub cron by writing Repo tarball to file instead of memory.

No.

### DIFF
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -17,8 +17,9 @@ package githubrepo
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/google/go-github/v32/github"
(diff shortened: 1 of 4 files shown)


clients/githubrepo/client.go
```diff
@@ -47,27 +50,36 @@ func (client *Client) InitRepo(owner, repoName string) error {
...
 	}
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return fmt.Errorf("%w", err)
+		return fmt.Errorf("error during HTTP call: %w", err)
 	}
 	defer resp.Body.Close()
 
-	respData, err := ioutil.ReadAll(resp.Body)
+	repoFile, err := os.OpenFile(repoFilename, os.O_CREATE|os.O_WRONLY, 0o644)
```
oliverchang, line 63: how about storing this on disk unpacked as well? the disk space savings probably aren't worth the repeated unpack/decompressions that the checks are doing.

oliverchang (approved):

clients/githubrepo/client.go
azeemshaikh38: Yes, that is on my mind too. Just don't want to attempt it now since changing this would require me to update `CheckIfFileExists` and `CheckFileContent`.

Eventually, I plan to expose `ListFiles()` and `GetFileContent()` APIs instead of a `GetRepoArchive()` API. When I make this change I will update the underlying data to be stored uncompressed.

azeemshaikh38:
Note to reviewers: I tested this locally and something seems off - the gzip/tar readers complain on reading the data from file. Will debug this further and update the PR.

azeemshaikh38:
> Note to reviewers: I tested this locally and something seems off - the gzip/tar readers complain on reading the data from file. Will debug this further and update the PR.

Fixed. I was basically creating a single reader which got shared between threads. Updated it to read from same file but create a new reader per `GetRepoArchiveReader()` call.

Issue #573, opened 2021-06-14
The integration tests are broken.


naveensrinivasan:
> The integration tests are broken.

Unrelated to this PR and is happening on all PRs. Not sure how to fix it though, `go mod download` or `go mod tidy` doesn't seem to have any effect.

_Originally posted by @azeemshaikh38 in https://github.com/ossf/scorecard/issues/569#issuecomment-860347109_

naveensrinivasan:
Fixed by #593

PR #593, opened 2021-06-20
🌱  Fixes GitHub workflow failures

naveensrinivasan:

  Bug fix

  e2e are failing #573

    The validate and the e2e are failing because of the bug in golang https://github.com/golang/go/issues/44129

    This fix is a temporary workaround.

None

* **Other information**:

None


azeemshaikh38 (approved):
Thanks. Looks like this has exposed other breakages which we weren't seeing so far. @laurentsimon

naveensrinivasan:
> Thanks. Looks like this has exposed other breakages which we weren't seeing so far. @laurentsimon

Thanks! I am going to open an issue with the failing e2e.

Issue #594, opened 2021-06-20
BUG - Some e2e tests are failing


naveensrinivasan:
**Describe the bug**
```
Summarizing 5 Failures:

[Panic!] E2E TEST:SecurityPolicy E2E TEST:Validating security policy [It] Should return valid security policy 
/opt/hostedtoolcache/go/1.16.5/x64/src/runtime/panic.go:212

[Panic!] E2E TEST:Automatic-Dependency-Update E2E TEST:Validating dependencies are automatically updated [It] Should return deps are automatically updated for dependabot 
/opt/hostedtoolcache/go/1.16.5/x64/src/runtime/panic.go:212

[Panic!] E2E TEST:Automatic-Dependency-Update E2E TEST:Validating dependencies are automatically updated [It] Should return deps are automatically updated for renovatebot 
/opt/hostedtoolcache/go/1.16.5/x64/src/runtime/panic.go:212

[Panic!] E2E TEST:FrozenDeps E2E TEST:Validating deps are frozen [It] Should return deps are not frozen 
/opt/hostedtoolcache/go/1.16.5/x64/src/runtime/panic.go:212

[Panic!] E2E TEST:FrozenDeps E2E TEST:Validating deps are frozen [It] Should return deps are not frozen 
/opt/hostedtoolcache/go/1.16.5/x64/src/runtime/panic.go:212

Ran 19 of 22 Specs in 23.564 seconds
FAIL! -- 14 Passed | 5 Failed | 0 Pending | 3 Skipped
```
https://github.com/ossf/scorecard/runs/2870164426?check_suite_focus=true

azeemshaikh38:
@laurentsimon

naveensrinivasan:
@azeemshaikh38 Looks like this one broke some of the `e2e` https://github.com/ossf/scorecard/commit/c41f0682233fe2b6bb2772949c919bdcf9d7e586

It assumes the file will be there on the disk.

laurentsimon:
there's another issue:
```
Test Panicked
    runtime error: invalid memory address or nil pointer dereference
    /usr/lib/google-golang/src/runtime/panic.go:212

    Full Stack Trace
    github.com/ossf/scorecard/checks.CheckIfFileExists(0x1073b69, 0x1b, 0xc0000eb1e8, 0x10a00f0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[...]

PR #598, opened 2021-06-22
🐛  Fix for e2e failures

laurentsimon:

bug fix

see https://github.com/ossf/scorecard/issues/594

addresses https://github.com/ossf/scorecard/issues/594

no

* **Other information**:
I've commented out the `t.Parallel()` because we need to share the tarballs. If we run the tests in parallel, we don't know when the download is finished: so a check may start reading from an incomplete tarball. An alternative approach would be to have a special `InitRepoForE2E` that downloads the tarball with a random name. This way we could run the tests in parallel, but the tarballs would not be shared, which I think would be ok.

### DIFF
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -35,6 +35,7 @@ type Client struct {
 	ctx        context.Context
 	owner      string
 	repoName   string
+	tarball    string
 }
 
 func (client *Client) InitRepo(owner, repoName string) error {
@@ -55,15 +56,16 @@ func (client *Client) InitRepo(owner, repoName string) error {
 	if err != nil {
 		return fmt.Errorf("http.NewRequestWithContext: %w", err)
 	}
+
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("http.DefaultClient.Do: %w", err)
 	}
 	defer resp.Body.Close()
 
-	tarfile := fmt.Sprintf(repoFilename, owner, repoName)
-	repoFile, err := os.OpenFile(tarfile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_EXCL, 0o644)
-	if err != nil && !os.IsExist(err) {
+	client.tarball = fmt.Sprintf(repoFilename, owner, repoName)
+	repoFile, err := os.OpenFile(client.tarball, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_EXCL, 0o644)
+	if err != nil {
 		return fmt.Errorf("os.OpenFile: %w", err)
 	}
 	defer repoFile.Close()
@@ -75,8 +77,7 @@ func (client *Client) InitRepo(owner, repoName string) error {
 }
 
 func (client *Client) ReleaseRepo() error {
-	tarfile := fmt.Sprintf(repoFilename, client.owner, client.repoName)
-	err := os.Remove(tarfile)
+	err := os.Remove(client.tarball)
 	if !os.IsNotExist(err) {
 		return fmt.Errorf("os.Remove: %w", err)
 	}
@@ -84,8 +85,7 @@ func (client *Client) ReleaseRepo() error {
 }
 
 func (client *Client) GetRepoArchiveReader() (io.ReadCloser, error) {
-	tarfile := fmt.Sprintf(repoFilename, client.owner, client.repoName)
-	archiveReader, err := os.OpenFile(tarfile, os.O_RDONLY, 0o644)
+	archiveReader, err := os.OpenFile(client.tarball, os.O_RDONLY, 0o644)
 	if err != nil {
 		return archiveReader, fmt.Errorf("os.OpenFile: %w", err)
 	}
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -28,7 +28,7 @@ import (
 	"github.com/ossf/scorecard/clients"
 )
 
-const repoFilename = "%v.%v.githubrepo.tar.gz"
+const repoFilename = "githubrepo*.tar.gz"
 
 type Client struct {
 	repo       *github.Repository
@@ -65,7 +65,7 @@ func (client *Client) InitRepo(owner, repoName string) error {
 	defer resp.Body.Close()
 
 	// Create a temp file. This automaticlly appends a random number to the name.
-	repoFile, err := ioutil.TempFile("", fmt.Sprintf(repoFilename, owner, repoName))
+	repoFile, err := ioutil.TempFile("", repoFilename)
 	if err != nil {
 		return fmt.Errorf("ioutil.TempFile: %w", err)
 	}
@@ -75,14 +75,6 @@ func (client *Client) InitRepo(owner, repoName string) error {
 
 	if _, err := io.Copy(repoFile, resp.Body); err != nil {
 		return fmt.Errorf("io.Copy: %w", err)
-	}
-	return nil
-}
-
-func (client *Client) ReleaseRepo() error {
-	err := os.Remove(client.tarball)
-	if !os.IsNotExist(err) {
-		return fmt.Errorf("os.Remove: %w", err)
 	}
 	return nil
 }
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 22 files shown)


clients/githubrepo/client.go
```diff
@@ -27,19 +27,21 @@ import (
 	"github.com/ossf/scorecard/clients"
 )
 
-const repoFilename = "./githubrepo.tar.gz"
+const repoFilename = "/tmp/%v.%v.githubrepo.tar.gz"
```
naveensrinivasan, line 30: The right way to create temp dir https://golang.org/pkg/io/ioutil/#TempDir

naveensrinivasan (approved):

clients/repo_client.go
```diff
@@ -40,4 +40,5 @@ func NewRepoUnavailableError(err error) error {
 type RepoClient interface {
 	InitRepo(owner, repo string) error
 	GetRepoArchiveReader() (io.ReadCloser, error)
+	ReleaseRepo() error
```
azeemshaikh38, line 43: Is there a reason to add this API? If not, I would prefer keeping this as is.

e2e/e2e_suite_test.go
```diff
@@ -45,7 +53,7 @@ func (l *log) Logf(s string, f ...interface{}) {
 }
 
 func TestE2e(t *testing.T) {
-	t.Parallel()
+	// t.Parallel()
```
azeemshaikh38, line 56: Remove?

clients/githubrepo/client.go
```diff
@@ -27,19 +27,21 @@ import (
...
 
 type Client struct {
 	repo       *github.Repository
 	repoClient *github.Client
 	ctx        context.Context
 	owner      string
 	repoName   string
+	tarball    string
```
azeemshaikh38, line 38: IIUC, the problem currently is when you create multiple instances of `githubrepo.Client` they all write to the same file. Seems like that's what you are trying to get around with the `ReleaseRepo` call.

Instead, we can simply allocate a single file per instance of `githubrepo.Client` and continue its usage as it is today. We can use a randomly generated value to do this. Basically, each instance will own 
`/tmp/githubrepo-randomvalue.tar.gz` and only do its I/O ops on that. Wdyt?

azeemshaikh38 (changes requested):
LGTM, apart from the API change.

clients/githubrepo/client.go
laurentsimon: thanks for the link. I was about to make this change to be OS-agnostic.

clients/githubrepo/client.go
laurentsimon: > IIUC, the problem currently is when you create multiple instances of `githubrepo.Client` they all write to the same file. Seems like that's what you are trying to get around with the `ReleaseRepo` call.

I implemented `ReleaseRepo` to try to clean up the user's machine and leave as little as possible on disk. `ReleaseRepo` actually does not solve the problem of concurrent runs because if one run deletes the file, the others will fail when reading the file.

> Instead, we can simply allocate a single file per instance of `githubrepo.Client` and continue its usage as it is today. We can use a randomly generated value to do this. Basically, each instance will own
> `/tmp/githubrepo-randomvalue.tar.gz` and only do its I/O ops on that. Wdyt?

I think I'll go with that. So long as we don't want to re-use tarballs across different scorecard runs, it will work. Thanks!

laurentsimon: Using `ioutil.TempFile` which automatically appends a random number to the filename.

laurentsimon:
please take a second look. I'm only creating `repoClient = githubrepo.CreateGithubRepoClient(ctx, ghClient)` in `e2e_suite_test.go` and moved the `InitRepo()` call to each of the unit tests, e.g. `frozen_deps_test.go`. Tell me what you think.

clients/githubrepo/client.go
```diff
@@ -27,19 +28,21 @@ import (
 	"github.com/ossf/scorecard/clients"
 )
 
-const repoFilename = "./githubrepo.tar.gz"
+const repoFilename = "%v.%v.githubrepo.tar.gz"
```
azeemshaikh38, line 31: We don't need this to be repo-specific anymore. So maybe just:

```
const repoFilename = "githubrepo*.tar.gz"
```

And `ioutil.TempFile` will replace the `*` with the random string.

clients/repo_client.go
azeemshaikh38: If the reason for having `ReleaseRepo` is to delete the tarball file after use, then let's call this method `Close` instead. It will be inline with other common Golang APIs.

And instead of calling `Close` at the end of every `RepoInit`, we simply call it once after we are done using the `RepoClient`. For example, add `defer repoClient` here - https://github.com/ossf/scorecard/blob/main/cron/worker/main.go#L174

laurentsimon: > If the reason for having `ReleaseRepo` is to delete the tarball file after use, then let's call this method `Close` instead. It will be inline with other common Golang APIs.

other golang APIs use `Open` and `Close`, but ours is called `InitRepo`. What's the opposite of `init`? I could not find a better word.

> And instead of calling `Close` at the end of every `RepoInit`, we simply call it once after we are done using the `RepoClient`. For example, add `defer repoClient` here - https://github.com/ossf/scorecard/blob/main/cron/worker/main.go#L174

Note sure I follow. I currently call `defer repoClient.ReleaseRepo()` after calling `InitRepo()`, see https://github.com/ossf/scorecard/pull/598/files#diff-afbd0f7e33a55279ebab362790883a75392a3f020c4c7d2de49e8322f08e2317R89.

azeemshaikh38:
> please take a second look. I've simplified this implementation by only creating `repoClient = githubrepo.CreateGithubRepoClient(ctx, ghClient)` in `e2e_suite_test.go` and I moved the `InitRepo()` call to each of the unit tests, e.g. `frozen_deps_test.go`. Tell me what you think.

This is tricky. `githubrepo.Client` is stateful and not thread-safe.  I'm not sure if these tests run in parallel, but if they do, that'll be an issue.

clients/githubrepo/client.go
laurentsimon: > We don't need this to be repo-specific anymore. So maybe just:
> 
> ```
> const repoFilename = "githubrepo*.tar.gz"
> ```

sure, I can do that. Not sure why it's necessary though.

> And `ioutil.TempFile` will replace the `*` with the random string.

I left `%v.%v` to help with debugging. it does not hurt to have it, I think. wdut?

e2e/e2e_suite_test.go
```diff
@@ -27,15 +27,21 @@ import (
...
 var (
 	ghClient    *github.Client
 	graphClient *githubv4.Client
-	client      *http.Client
+	httpClient  *http.Client
+	repoClient  clients.RepoClient
 )
 
+// List of repos to fetch source code of.
+var repoList = []string{"ossf/scorecard", "netlify/netlify-cms", "tensorflow/tensorflow"}
```
naveensrinivasan, line 43: After some thought, this approach feels hacky.  The present implementation expects to run the scorecard binary for things to work. No other library (for example another go project) cannot refer to scorecard as a dependency and use it because the file downloading is within the main package. 

I mean checks package cannot be used independently.

laurentsimon: this has been removed. I've moved everything under the tests themselves to avoid sharing the tarball and concurrency issues.

laurentsimon:
@azeemshaikh38 ptal. I addressed your comments.

clients/githubrepo/client.go
```diff
@@ -52,31 +55,34 @@ func (client *Client) InitRepo(owner, repoName string) error {
...
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return fmt.Errorf("error during HTTP call: %w", err)
+		return fmt.Errorf("http.DefaultClient.Do: %w", err)
 	}
 	defer resp.Body.Close()
 
-	repoFile, err := os.OpenFile(repoFilename, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o644)
+	// Create a temp file. This automaticlly appends a random number to the name.
+	repoFile, err := ioutil.TempFile("", repoFilename)
```
azeemshaikh38, line 68: You'll hate me for this , but this will end up creating multiple new files :) Either:

1. Delete old files at the beginning of `InitRepo` call
2. Keep a single instance of the TempFile and use it across all `InitRepo` calls
3. If absolutely necessary go back to the old `ReleaseRepo` API call and we can figure how to clean up the API at a later point.

azeemshaikh38 (approved):

clients/githubrepo/client.go
laurentsimon: my bad. You're right. I'll take option 1, since 2 requires closing the file in a `ReleaseRepo()` function.

laurentsimon: see https://github.com/ossf/scorecard/pull/599. (Sorry, the PR got auto-merged)

PR #604, opened 2021-06-23
✨  Support for package manager's unpinned downloads

laurentsimon:

feature

no support

check for `go install/get pkname` where `pkgname` is not pinned
check for `pip install` where there's no `-r requirements.txt`

* **Other information**:

### DIFF
--- a/checks/shell_download_validate.go
+++ b/checks/shell_download_validate.go
@@ -334,8 +334,8 @@ func isGoUnpinnedDownload(cmd []string) bool {
 
 	// `Go install` will automatically look up the
 	// go.mod and go.sum, so we don't flag it.
-	l := 2
-	if len(cmd) <= l {
+	// nolinter
+	if len(cmd) <= 2 {
 		return false
 	}
 
@@ -355,7 +355,8 @@ func isGoUnpinnedDownload(cmd []string) bool {
 		pkg := cmd[i+1]
 		// Verify pkg = name@hash
 		parts := strings.Split(pkg, "@")
-		if len(parts) != l {
+		// nolinter
+		if len(parts) != 2 {
 			continue
 		}
 		hash := parts[1]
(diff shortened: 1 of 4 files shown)


azeemshaikh38 (approved):

checks/shell_download_validate.go
```diff
@@ -316,6 +323,111 @@ func isExecuteFiles(node syntax.Node, cmd, pathfn string, files map[string]bool,
...
+		return false
+	}
+
+	if !isBinaryName("go", cmd[0]) {
+		return false
+	}
+
+	// `Go install` will automatically look up the
+	// go.mod and go.sum, so we don't flag it.
+	l := 2
```
oliverchang, line 337: why do we need this var? I see this is used both here and on line 358 in an overloaded way.

in general for constants, better to do:

```go
const descriptiveName = 2
```

in this case though, I think we don't need this and can just explain this with a comment

```go
if len(cmd) < 2 {
  // either `go install` (which checks go.sum) or an unknown command.
  return false
}
```

checks/shell_download_validate.go
```diff
@@ -316,6 +323,111 @@ func isExecuteFiles(node syntax.Node, cmd, pathfn string, files map[string]bool,
...
+		}
+
+		if !found {
+			continue
+		}
+
+		pkg := cmd[i+1]
+		// Verify pkg = name@hash
+		parts := strings.Split(pkg, "@")
+		if len(parts) != l {
```
oliverchang, line 358: just `!= 2` here should be fine.

checks/shell_download_validate.go
```diff
@@ -316,6 +323,111 @@ func isExecuteFiles(node syntax.Node, cmd, pathfn string, files map[string]bool,
...
+		}
+
+		if !isInstalled || !strings.EqualFold("-r", cmd[i]) {
+			continue
+		}
+
+		requirements := cmd[i+1]
+		if strings.EqualFold(path.Base(requirements), "requirements.txt") {
+			return false
+		}
```
oliverchang, line 394: what about `pip install package==1.0.0` ?

oliverchang (approved):

checks/shell_download_validate.go
laurentsimon: I added this variable because the linter complains that's `2` is an arbitrary number. I'll add a `//nolint` instead.

checks/shell_download_validate.go
laurentsimon: thanks. I will extend the check for `pip install package==hash`

PR #605, opened 2021-06-23
🐛 Fix truncated file extraction from tarball

laurentsimon:

bug fix

tarball files were truncated if the call to `tr.Read` did not return all the bytes at once.

fix the above.

no

* **Other information**:

### DIFF
--- a/checks/checkforcontent.go
+++ b/checks/checkforcontent.go
@@ -90,7 +90,7 @@ func getTarReader(in io.Reader) (*tar.Reader, error) {
 	return tr, nil
 }
 
-func readEntireFile(tr *tar.Reader, hdr *tar.Header) (content []byte, err error) {
+func readEntireFile(tr *tar.Reader) (content []byte, err error) {
 	var buf bytes.Buffer
 	_, err = io.Copy(&buf, tr)
 	if err != nil {
@@ -161,9 +161,9 @@ func CheckFilesContent(checkName, shellPathFnPattern string,
 			continue
 		}
 
-		content, err := readEntireFile(tr, hdr)
+		content, err := readEntireFile(tr)
 		if err != nil {
-			return checker.MakeFailResult(checkName, err)
+			return checker.MakeRetryResult(checkName, err)
 		}
 
 		// We should have reached the end of files AND
@@ -175,7 +175,6 @@ func CheckFilesContent(checkName, shellPathFnPattern string,
 			return checker.MakeRetryResult(checkName, ErrReadFile)
 		}
 
-		// We truncate the file to remove trailing 0 (sparse format).
 		rr, err := onFileContent(fullpath, content, c.Logf)
 		if err != nil {
 			return checker.MakeFailResult(checkName, err)
--- a/checks/checkforcontent.go
+++ b/checks/checkforcontent.go
@@ -16,6 +16,7 @@ package checks
 
 import (
 	"archive/tar"
+	"bytes"
 	"compress/gzip"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


checks/checkforcontent.go
```diff
@@ -150,22 +161,22 @@ func CheckFilesContent(checkName, shellPathFnPattern string,
 			continue
 		}
 
-		content := make([]byte, hdr.Size)
-		n, err := tr.Read(content)
-		if err != nil && err != io.EOF {
-			return checker.MakeRetryResult(checkName, err)
+		content, err := readEntireFile(tr, hdr)
+		if err != nil {
+			return checker.MakeFailResult(checkName, err)
```
inferno-chromium, line 166: Why fail, why not retry.

checks/checkforcontent.go
```diff
@@ -150,22 +161,22 @@ func CheckFilesContent(checkName, shellPathFnPattern string,
...
 		// the number of bytes should be the same as number
 		// indicated in header, unless the file format supports
 		// sparse regions. Only USTAR format does not support
 		// spare regions -- see https://golang.org/pkg/archive/tar/
-		if b := headerSizeMatchesFileSize(hdr, n); !b {
+		if b := headerSizeMatchesFileSize(hdr, len(content)); !b {
 			return checker.MakeRetryResult(checkName, ErrReadFile)
 		}
 
 		// We truncate the file to remove trailing 0 (sparse format).
```
inferno-chromium, line 178: is this comment still valid ?

checks/frozen_deps_test.go
```diff
@@ -90,12 +92,14 @@ func TestGithubWorkflowPinning(t *testing.T) {
...
+			r, err := validateGitHubActionWorkflow(tt.args.Filename, content, tt.args.Log.Logf)
 
 			if !errors.Is(err, tt.want.Error) ||
-				r != tt.want.Result {
-				t.Errorf("TestGithubWorkflowPinning:\"%v\": %v (%v,%v) want (%v, %v)",
-					tt.name, tt.args.Filename, r, err, tt.want.Result, tt.want.Error)
+				r != tt.want.Result ||
+				len(tt.args.Log.messages) != tt.want.NumberOfErrors {
+				t.Errorf("TestDockerfileScriptDownload:\"%v\": %v (%v,%v,%v) want (%v, %v, %v)\n%v",
+					tt.name, tt.args.Filename, r, err, len(tt.args.Log.messages), tt.want.Result, tt.want.Error, tt.want.NumberOfErrors,
```
inferno-chromium, line 101: nit: Do we need that complicated message, similar is better ?

inferno-chromium (approved):

checks/checkforcontent.go
```diff
@@ -89,6 +90,16 @@ func getTarReader(in io.Reader) (*tar.Reader, error) {
 	return tr, nil
 }
 
+func readEntireFile(tr *tar.Reader, hdr *tar.Header) (content []byte, err error) {
```
azeemshaikh38, line 93: tar.Header is not being used. Consider removing.

azeemshaikh38 (approved):

checks/checkforcontent.go
laurentsimon: good point. Mistake from me :-)

checks/checkforcontent.go
laurentsimon: you have lynx eyes!

checks/checkforcontent.go
laurentsimon: good catch!

checks/frozen_deps_test.go
laurentsimon: this gives all information we need to fix the problem. It only triggers when there's a failure.

PR #606, opened 2021-06-23
✨ Check for shell script's insecure download

laurentsimon:

feature

we don't support.

we support `curl | bash` type detection for shell scripts.

no

* **Other information**:
A lot of refactoring has happened in the code to be able to share the detection for docker, shell, and later makefiles/github workflows. The refactoring simplifies things, though.

### DIFF
--- a/checks/frozen_deps.go
+++ b/checks/frozen_deps.go
@@ -74,7 +74,7 @@ func isShellScriptFreeOfInsecureDownloads(c *checker.CheckRequest) checker.Check
 func validateShellScriptDownloads(pathfn string, content []byte,
 	logf func(s string, f ...interface{})) (bool, error) {
 	// Validate the file type.
-	if !IsShellScriptFile(pathfn, content) {
+	if !isShellScriptFile(pathfn, content) {
 		return false, nil
 	}
 
--- a/checks/frozen_deps.go
+++ b/checks/frozen_deps.go
@@ -78,6 +78,7 @@ func validateDockerfileDownloads(pathfn string, content []byte,
 		return false, fmt.Errorf("cannot read dockerfile content: %w", err)
 	}
 
+	// nolinter:prealloc
 	var bytes []byte
 
 	// Walk the Dockerfile's AST.
--- a/checks/frozen_deps.go
+++ b/checks/frozen_deps.go
@@ -17,7 +17,6 @@ package checks
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


checks/shell_download_validate.go
```diff
@@ -552,57 +459,69 @@ func extractInterpreterCommandFromNode(node syntax.Node) (string, bool) {
...
 }
 
-func extractInterpreterCommandFromString(cmd string) (c string, res bool, err error) {
-	in := strings.NewReader(cmd)
+func nodeToString(p *syntax.Printer, node syntax.Node) (string, error) {
+	// https://github.com/mvdan/sh/blob/24dd9930bc1cfc7be025f8b75b2e9e9f04524012/syntax/printer.go#L135.
+	var buf bytes.Buffer
+	err := p.Print(&buf, node)
+	// This is ugly, but the parser does not have a defined error type :/.
+	if err != nil && !strings.Contains(err.Error(), "unsupported node type") {
```
oliverchang, line 467: what about other non-nil errors? do we not care?

checks/frozen_deps.go
```diff
@@ -84,26 +97,12 @@ func validateDockerfileDownloads(pathfn string, content []byte,
...
-		fn, b, err := recordFetchFileFromString(cmd)
-		if err != nil {
-			return false, err
-		} else if b {
-			for f := range fn {
-				files[f] = true
-			}
-		}
+		bytes = append(bytes, cmd...)
+		bytes = append(bytes, []byte("\n")...)
```
oliverchang, line 103: just `append(bytes, '\n')` ?

oliverchang (approved):

checks/shell_download_validate.go
laurentsimon: this line will catch all errors that are *not* of the type `unsupported type`. We're only discarding the `unsupported type` because those won't need to be printed, ie they will never contain insecure download code.

checks/frozen_deps.go
laurentsimon: done

checks/shell_download_validate.go
oliverchang: I see, I misunderstood the intent here. Would be good to add a comment here to explain why we discard "unsupported node type" :)

laurentsimon: will do. Thanks

PR #610, opened 2021-06-24
✨ check insecure downloads in github workflows

laurentsimon:

feature

we don't support

we support `curl | bash` in github workflow scripts.

no

* **Other information**:


checks/shell_download_validate.go
```diff
@@ -680,7 +678,6 @@ func isShellScriptFile(pathfn string, content []byte) bool {
 		}
 
 		for _, name := range shellNames {
-			line = line[2:]
```
laurentsimon, line 683: this was a bug I introduced during rebase :/

checks/shell_download_validate.go
```diff
@@ -679,8 +677,8 @@ func isShellScriptFile(pathfn string, content []byte) bool {
 			continue
 		}
 
+		line = line[2:]
 		for _, name := range shellNames {
-			line = line[2:]
```
laurentsimon, line 683: that's a bug due to a bad merge from me :-(

checks/frozen_deps.go
```diff
@@ -188,40 +213,80 @@ func validateDockerfile(path string, content []byte,
...
+		if job.Defaults.Run.Shell != "" {
+			defaultShell = job.Defaults.Run.Shell
+		}
+
+		for i, step := range job.Steps {
+			if step.Run == "" {
+				continue
+			}
+
+			shell := defaultShell
```
oliverchang, line 247: i think we can simplify this by not having this `defaultShell` var.

Just something like:

```go
if step.Shell != "" && !isShell(step.Shell) {
  continue
}
```

since the default is bash, we only need to check if the user specified shell is valid?

checks/frozen_deps.go
```diff
@@ -188,40 +213,80 @@ func validateDockerfile(path string, content []byte,
...
+
+	var workflow gitHubActionWorkflowConfig
+	err := yaml.Unmarshal(content, &workflow)
+	if err != nil {
+		return false, fmt.Errorf("!! frozen-deps - cannot unmarshal file %v\n%v: %w", pathfn, string(content), err)
+	}
+
+	githubVarRegex := regexp.MustCompile(`{{[^{}]*}}`)
+	validated := true
+	// https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell.
```
oliverchang, line 234: do we need to check the workflow's platform to make sure it's not windows?

otherwise the windows default shell is `pwsh` but we continue trying to analyze it under current logic (and potentially add unhelpful error messages)?

oliverchang (approved):
LGTM with some suggestions for improvement.

checks/frozen_deps.go
```diff
@@ -188,40 +213,80 @@ func validateDockerfile(path string, content []byte,
...
+
+			shell := defaultShell
+			if step.Shell != "" {
+				shell = step.Shell
+			}
+
+			// https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun.
+			// Skip unsupported shells.
+			// Note: this may be python.
+			if !isShellScriptFile("file", []byte("#!"+path.Join("/bin/", shell))) {
```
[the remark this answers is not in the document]
laurentsimon: done. I created `isShell()` which calls `isShellScriptFile ()` since it's the function that detects shell script. let me know if you think I should change the check entirely.

checks/frozen_deps.go
```diff
@@ -188,40 +213,80 @@ func validateDockerfile(path string, content []byte,
...
+				continue
+			}
+
+			run := step.Run
+			script := githubVarRegex.ReplaceAll([]byte(run), []byte(fmt.Sprintf("GITHUB_REDACTED_VAR_%v", i)))
```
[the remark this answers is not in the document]
laurentsimon: done, thanks.

checks/frozen_deps.go
laurentsimon: there's an option to change the default shell for all runs in the workflow, see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iddefaults.
That's why I have this defaultShell variable. Resolving for now, feel free to re-open.

checks/frozen_deps.go
laurentsimon: good point. `pwsh` is not supported as a shell, see https://github.com/ossf/scorecard/pull/610/files#diff-5345e5b2866b919ef20545abf37043d1ee8ba7a30b8e8d27a1f7a9ded375232aR50
I'll add a comment

PR #611, opened 2021-06-24
✨ Detect python -m pip pkg

laurentsimon:

feature

no support

we support

no

* **Other information**:
There are some false positives (same for `pip install`) if someone genuinely tries to do `pip install *.whl` for `*.whl` for tests. Example: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/builds/docker_cpu_pip.sh#L23

### DIFF
--- a/checks/shell_download_validate.go
+++ b/checks/shell_download_validate.go
@@ -237,17 +237,31 @@ func extractCommand(cmd interface{}) ([]string, bool) {
 	if !ok {
 		return nil, ok
 	}
-
 	var ret []string
 	for _, w := range c.Args {
 		if len(w.Parts) != 1 {
 			continue
 		}
-		lit, ok := w.Parts[0].(*syntax.Lit)
-		if ok && !strings.EqualFold(lit.Value, "sudo") {
-			ret = append(ret, lit.Value)
+		switch v := w.Parts[0].(type) {
+		default:
+			continue
+		case *syntax.SglQuoted:
+			ret = append(ret, "'"+v.Value+"'")
+		case *syntax.DblQuoted:
+			if len(v.Parts) != 1 {
+				continue
+			}
+			lit, ok := v.Parts[0].(*syntax.Lit)
+			if ok {
+				ret = append(ret, "\""+lit.Value+"\"")
+			}
+		case *syntax.Lit:
+			if !strings.EqualFold(v.Value, "sudo") {
+				ret = append(ret, v.Value)
+			}
 		}
 	}
+
 	return ret, true
 }
 
@@ -382,6 +396,7 @@ func isUnpinnedPipInstall(cmd []string) bool {
 	}
 
 	isInstall := false
+	hasWhl := false
 	for i := 1; i < len(cmd); i++ {
 		// Search for install commands.
 		if strings.EqualFold(cmd[i], "install") {
@@ -397,9 +412,19 @@ func isUnpinnedPipInstall(cmd []string) bool {
 		if strings.EqualFold("-r", cmd[i]) {
 			return false
 		}
+
+		// Exclude *.whl
+		if strings.HasSuffix(cmd[i], ".whl") {
+			hasWhl = true
+			continue
+		}
+
+		// Any other arguments are considered unpinned.
+		return true
 	}
 
-	return isInstall
+	// We get here only for `pip install [bla.whl ...]`.
+	return isInstall && !hasWhl
 }
 
 func isPythonCommand(cmd []string) bool {
@@ -431,7 +456,6 @@ func isUnpinnedPythonPipInstall(cmd []string) bool {
 	if !isPythonCommand(cmd) {
 		return false
 	}
-
 	pipCommand, ok := extractPipCommand(cmd)
 	if !ok {
 		return false
@@ -583,20 +607,23 @@ func extractInterpreterCommandFromArgs(args []*syntax.Word) (string, bool) {
 			continue
 		}
 		part := arg.Parts[0]
-		v, ok := part.(*syntax.DblQuoted)
-		if !ok {
-			continue
-		}
-		if len(v.Parts) != 1 {
-			continue
-		}
+		switch v := part.(type) {
+		case *syntax.DblQuoted:
+			if len(v.Parts) != 1 {
+				continue
+			}
 
-		lit, ok := v.Parts[0].(*syntax.Lit)
-		if !ok {
-			continue
+			lit, ok := v.Parts[0].(*syntax.Lit)
+			if !ok {
+				continue
+			}
+			return lit.Value, true
+
+		case *syntax.SglQuoted:
+			return v.Value, true
 		}
-		return lit.Value, true
 	}
+
 	return "", false
 }
 
--- a/checks/shell_download_validate.go
+++ b/checks/shell_download_validate.go
@@ -376,7 +376,7 @@ func isGoUnpinnedDownload(cmd []string) bool {
 	return found
 }
 
-func isPipInstall(cmd []string) bool {
+func isUnpinnedPipInstall(cmd []string) bool {
 	if !isBinaryName("pip", cmd[0]) && !isBinaryName("pip3", cmd[0]) {
 		return false
 	}
@@ -411,14 +411,13 @@ func isPythonCommand(cmd []string) bool {
 	return false
 }
 
-func isPythonPipInstall(cmd []string) bool {
+func isUnpinnedPythonPipInstall(cmd []string) bool {
 	if !isPythonCommand(cmd) {
 		return false
 	}
 
 	isInstall := false
 	hasPip := false
-
 	for i := 1; i < len(cmd); i++ {
 		// Search for pip module.
 		if strings.EqualFold(cmd[i], "-m") &&
@@ -428,14 +427,23 @@ func isPythonPipInstall(cmd []string) bool {
 			continue
 		}
 
+		if !hasPip {
+			continue
+		}
+
 		// Search for install commands.
 		if strings.EqualFold(cmd[i], "install") {
 			isInstall = true
 			continue
 		}
 
-		if isInstall && hasPip {
-			return true
+		if !isInstall {
+			continue
+		}
+
+		// Search for `-r file`.
+		if strings.EqualFold(cmd[i], "-r") {
+			return false
 		}
 	}
 
@@ -447,11 +455,11 @@ func isPipUnpinnedDownload(cmd []string) bool {
 		return false
 	}
 
-	if isPipInstall(cmd) {
+	if isUnpinnedPipInstall(cmd) {
 		return true
 	}
 
-	if isPythonPipInstall(cmd) {
+	if isUnpinnedPythonPipInstall(cmd) {
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


checks/shell_download_validate.go
```diff
@@ -401,7 +400,63 @@ func isPipUnpinnedDownload(cmd []string) bool {
...
+func isPythonCommand(cmd []string) bool {
+	for _, pi := range pythonInterpreters {
+		if isBinaryName(pi, cmd[0]) {
+			return true
+		}
+	}
+	return false
+}
+
+func isPythonPipInstall(cmd []string) bool {
```
laurentsimon, line 415: There are some false positives if someone genuinely tries to do `pip install *.whl` for `*.whl` for tests. Example: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/builds/docker_cpu_pip.sh#L23

checks/shell_download_validate.go
```diff
@@ -35,8 +35,11 @@ var ErrParsingDockerfile = errors.New("file cannot be parsed")
 var ErrParsingShellCommand = errors.New("shell command cannot be parsed")
 
 // List of interpreters.
+var pythonInterpreters = []string{"python", "python3", "python2.7"}
+
 var interpreters = []string{
```
azeemshaikh38, line 40: Change to:

```
var interpreters = append([]string{"sh", <>}, pythonInterpreters...)
```

laurentsimon: thank you! Was looking for some better way!

checks/shell_download_validate.go
oliverchang: I'm not sure if it's common at all for people to install a *.whl unless its in a testing context. Perhaps we should just exclude this.

checks/shell_download_validate.go
```diff
@@ -401,7 +400,63 @@ func isPipUnpinnedDownload(cmd []string) bool {
...
+			continue
+		}
+
+		// Search for install commands.
+		if strings.EqualFold(cmd[i], "install") {
+			isInstall = true
+			continue
+		}
+
+		if isInstall && hasPip {
```
oliverchang, line 438: what's actually checking here that the install is pinned?

checks/shell_download_validate.go
```diff
@@ -401,7 +400,63 @@ func isPipUnpinnedDownload(cmd []string) bool {
...
+	for _, pi := range pythonInterpreters {
+		if isBinaryName(pi, cmd[0]) {
+			return true
+		}
+	}
+	return false
+}
+
+func isPythonPipInstall(cmd []string) bool {
+	if !isPythonCommand(cmd) {
```
oliverchang, line 415: can also rewrite this to share more logic between `python -m pip` and `pip`.

e.g. something like:

```go
func isPythonPipInstall(cmd []string) bool {
  pipCommand, ok := extractPipCommand(cmd)
  if !ok {
    return false
  }

  return isPipInstall(pipCommand) 
}
```

checks/shell_download_validate.go
laurentsimon: gotcha. Will remove this and maybe re-introduce it with more accuracy when working on https://github.com/ossf/scorecard/issues/613

checks/shell_download_validate.go
laurentsimon: you're absolutely right. I lost track of `python -m pip install -r _some_file`. Thank yoou!
Fixed and added it to unit tests.

checks/shell_download_validate.go
```diff
@@ -401,7 +399,71 @@ func isPipUnpinnedDownload(cmd []string) bool {
...
+		if strings.EqualFold(cmd[i], "install") {
+			isInstall = true
+			continue
+		}
+
+		if !isInstall {
+			continue
+		}
+
+		// Search for `-r file`.
```
oliverchang, line 444: did we add an exclusion for `pkg==ver` somewhere as well?

checks/shell_download_validate.go
```diff
@@ -401,7 +400,63 @@ func isPipUnpinnedDownload(cmd []string) bool {
...
+		}
+
+		// Search for install commands.
+		if strings.EqualFold(cmd[i], "install") {
+			isInstall = true
+			continue
+		}
+
+		if isInstall && hasPip {
+			return true
```
azeemshaikh38: Maybe just `break` here?

checks/shell_download_validate.go
```diff
@@ -401,7 +400,63 @@ func isPipUnpinnedDownload(cmd []string) bool {
...
+	if !isPythonCommand(cmd) {
+		return false
+	}
+
+	isInstall := false
+	hasPip := false
+
+	for i := 1; i < len(cmd); i++ {
+		// Search for pip module.
+		if strings.EqualFold(cmd[i], "-m") &&
```
azeemshaikh38, line 423: Maybe make this `if {}` and `else if {}` without `continue`, so that, we check for `isInstall && hasPip` at the end of every for loop and break if `true`.

azeemshaikh38 (approved):

checks/shell_download_validate.go
laurentsimon: great suggestion. Done

checks/shell_download_validate.go
laurentsimon: i've updated the function as per https://github.com/ossf/scorecard/pull/611/files/5f1fdea08e75ccf719ccb959e40af0c0e16021c5#r658350655, so this function is simpler now. please ptal.

checks/shell_download_validate.go
laurentsimon: `pkg==ver` is not considered pinning, is it? I'm not aware of `pkg=hash` being valid, so I consider anything as not pinned. wdut?

checks/shell_download_validate.go
laurentsimon: function re-written and split in two. ptal

checks/shell_download_validate.go
oliverchang: `-r requirements.txt` is also not really pinning in that case either then, because it does *not* include all transitive dependencies.

The only real python solution is to use Pipfile.lock, but requires pipenv and is not supported everywhere.

checks/shell_download_validate.go
laurentsimon: done.

checks/shell_download_validate.go
laurentsimon: > `-r requirements.txt` is also not really pinning in that case either then, because it does _not_ include all transitive dependencies, nor does it do any kind of checksumming.

`requirements.txt` does support pinning/checksumming, see https://davidwalsh.name/hashin. I'm going to improve the check to check the content of the file, see my TODO https://github.com/ossf/scorecard/blob/main/checks/frozen_deps.go#L273

Assuming I implement the check above, `requirements.txt` will be assumed pinned in this check PR. That's why I'm assuming that anything with `-r <>` is fine.

There's a caveat to the check https://github.com/ossf/scorecard/blob/main/checks/frozen_deps.go#L273, which is that there's no guarantee the file is even called `requirements.txt`. It's a convention, but if devs have multiple files, they'll certainly rename them. Only way would be to parse the file. 

@naveensrinivasan to comment on that.

> The only real python solution is to use Pipfile.lock, but requires pipenv and is not supported everywhere.

given the above, I think requirements.txt would work.  If you think that's too many assumptions, maybe we should consider any use of `pip` as unpinned? wdut?

> 
> Pip packages are supposed to be immutable though, so a published version should never change (unless there's a bug in the infra).

we'd like to mitigate this threat with this check.

laurentsimon: @inferno-chromium FYI

oliverchang: I see! I wasn't aware of hashin. It doesn't handle/lock transitive dependencies though right? I'm not sure if that's enough of an improvement in that case to consider this "safe".

Do others have thoughts on this?

laurentsimon: I'm starting to think it's brittle too. I've added a `need-discussion` label to bring this up in next meeting too.

laurentsimon: I had a chat w/ @rsprabery. Results:
One way to ensure the hashes are checked is thru `--require-hashes`, see https://pip.pypa.io/en/stable/cli/pip_install/#hash-checking-mode. This is simple.

Indirect dependencies are not necessarily pulled into the requirements.txt files. According to @rsprabery, one needs to use `pip-compile` to generate them from a `requirements.ini`. To verify that indirect deps are declared, we'd need to do that ourselves in scorecard by calling pypi's API or run `pip-tools` to tell us.

If we want to continue having a check for insecure uses of `pip install`, we could start by

1. Check for presence of `--require-hashes` and
2. Provide a `bootstrap-requirement.txt` containing `pip-tools` and its dependencies (about 4 of them only) along with their corresponding hash. We can rely on dependabot to bump the hashes.
2. Provide good remediation steps to explain how do use `pip-compile` properly.

We need not verify that indirect dependencies are declared for now.

For the next iteration of the check, we could update the check to:

1. Checks for indirect dependencies

What do you think?

I'll let @rsprabery comment on the above.

oliverchang: I think we should just recommend `pipenv` instead, and consider all direct `pip install` commands as unpinned . It's officially supported by Python, and seems like much less effort to set up compared to all the hoops to get this working for `requirements.txt`. If we see a lot of complaints here, or false positives, perhaps we can explore this more then. Having to check for the contents of  a `requirements (or whatever).txt` file will also be difficult to do in all contexts, as it's not always going to be obvious where to find it.

We could also expand this to other Python package management solutions (e.g. Poetry) with locking supported out of the box since they are likely much easier to check (because e.g. we can almost assume pinning is done if there is a .lock file)

laurentsimon: Thanks everyone for the feedback. I'll consider all `pip install` as unpinned for now. We can re-visit and maybe enable it with `--require-hashes` if 1) we get push back from community OR 2) we have a good story/tooling/remediation steps. Let's keep things simple for now, we can iterate in the next version.

laurentsimon:
@naveensrinivasan @inferno-chromium failing check 
```
make: *** [Makefile:88: cron/data/request.pb.go] Error 127
```

Can someone help fix or merge this PR?

checks/shell_download_validate.go
rsprabery: @oliverchang - I didn't realize that `Pipenv` was officially supported by Python these days. Is it being bundled like `venv` and `pip` in the python3 distributions now?

In the long term we'll probably want to support `poetry`, `pip-tools`, and `Pipenv` as I don't think the Python community has settled on an approach yet, but maybe that's changed recently?

laurentsimon:
friendly ping.

naveensrinivasan:
> friendly ping.

I will take a look at it. Apologies for the delay!

checks/shell_download_validate.go
evverx: > I think we should just recommend pipenv instead, and consider all direct pip install commands as unpinned

I think it's overkill though when all I need is to install a package in a test script. I'm planning to switch to `-r` in https://github.com/systemd/systemd/pull/21438 and I can theoretically add `--require-hashes` there since `Dependabot` supports it but switching to `pipenv` or any other tool creating virtual python environments is absolutely unwarranted in that particular case I think

evverx: I opened https://github.com/ossf/scorecard/issues/1306 so that it won't get lost

PR #628, opened 2021-06-29
:sparkles: Add a Vulnerabilities check.

### LINKED ISSUE #52: New check: ADD OSV: Show current state w.r.t open security vulnerabilities (e.g. CVEs, OSS-Fuzz, etc)

oliverchang:
Uses OSV to check this.

Fixes #52.

New check for unfixed vulnerabilities using OSV. 

* **Other information**:

### DIFF
--- a/checks/unfixed_vulnerabilities.go
+++ b/checks/unfixed_vulnerabilities.go
@@ -1,3 +1,17 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package checks
 
 import (
--- a/checks/unfixed_vulnerabilities.go
+++ b/checks/unfixed_vulnerabilities.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"net/http"
 
 	"github.com/google/go-github/v32/github"
@@ -77,7 +76,6 @@ func HasUnfixedVulnerabilities(c *checker.CheckRequest) checker.CheckResult {
 	}
 
 	if len(osvResp.Vulns) > 0 {
-		fmt.Printf("%v\n", osvResp)
 		for _, vuln := range osvResp.Vulns {
 			c.Logf("HEAD is vulnerable to %s", vuln.ID)
 		}
--- a/checks/unfixed_vulnerabilities.go
+++ b/checks/unfixed_vulnerabilities.go
@@ -1,0 +1,88 @@
+package checks
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/google/go-github/v32/github"
+
+	"github.com/ossf/scorecard/checker"
+)
+
+const (
+	// UnfixedVulnerabilities is the registered name for the OSV check.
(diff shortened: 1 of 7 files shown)


checks/unfixed_vulnerabilities.go
```diff
@@ -0,0 +1,100 @@
...
+	"net/http"
+
+	"github.com/google/go-github/v32/github"
+
+	"github.com/ossf/scorecard/checker"
+)
+
+const (
+	// UnfixedVulnerabilities is the registered name for the OSV check.
+	UnfixedVulnerabilities = "Unfixed-Vulnerabilities"
```
azeemshaikh38, line 30: Rename to `CheckUnfixedVulnerabilities` in sync with the convention of check names starting with `Check`.

checks/unfixed_vulnerabilities.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	defer resp.Body.Close()
+
+	var osvResp osvResponse
+	decoder := json.NewDecoder(resp.Body)
+	if err := decoder.Decode(&osvResp); err != nil {
+		return checker.MakeRetryResult(UnfixedVulnerabilities, err)
+	}
+
+	if len(osvResp.Vulns) > 0 {
```
azeemshaikh38, line 92: We can make a function:

```
func (resp osvResponse) GetVulnerabilties() []string {}
```

which we call here to log and pass to the `MakeFailResult` fn too.

checks/unfixed_vulnerabilities.go
```diff
@@ -0,0 +1,100 @@
...
+		for _, vuln := range osvResp.Vulns {
+			c.Logf("HEAD is vulnerable to %s", vuln.ID)
+		}
+		return checker.MakeFailResult(UnfixedVulnerabilities, nil)
```
azeemshaikh38, line 96: Lets create a OSV-check specific error which gets passed to `MakeFailResult`.

azeemshaikh38 (approved):
Awesome, thanks!

checks/unfixed_vulnerabilities.go
oliverchang: Shouldn't `err` be used when a check fails to run because of some issue with the infra, rather than be a negative result of a successful check on a repo?

In this case we have a successful run, and determined that there are open vulnerabilities.

checks/unfixed_vulnerabilities.go
oliverchang: Done! Renamed to `CheckVulnerabilities` (per @inferno-chromium's suggestion to shorten the check name)

checks/unfixed_vulnerabilities.go
oliverchang: Do you mean:

```go

func (resp osvResponse) GetVulnerabilties() []string {
  return resp.Vulns
}

vulns := osvResp.GetVulnerabilities()
if len(vulns) > 0 {
  // ... as-is ...
}
```

This adds more lines but doesn't seem to give too much benefit? (because there are no other vuln sources). 

Did you mean something else?

checks/unfixed_vulnerabilities.go
azeemshaikh38: I wrongly thought that not capturing the `vluns.ID` would mean this info will be lost. Although I now realize we capture this in `Details` through `Logf` so should be fine.

FYI, though, as of today, we rely on the confidence score to distinguish between a `Fail` result due to infra issues vs. a negative result. So there are instances in the code where we `Fail` and the error is not an infra error. We don't really have a good guideline on how we should be handling these 2 types of errors.

checks/unfixed_vulnerabilities.go
azeemshaikh38: I did mean this. No benefit as such, just wanted to avoid the multiple `Logf` calls and instead have a single `Logf` which reads something like: `Vulnerabilities found at HEAD: vuln1, vuln2 ...`.

oliverchang: I see, a single Logf makes sense! Just did this.

checks/unfixed_vulnerabilities.go
oliverchang: Yea, we should clean this up in the future. resolving for now, thanks!

PR #637, opened 2021-06-30
✨  Add `ListFiles` and `GetFileContent` APIs

azeemshaikh38:

Adds `ListFiles` and `GetFileContent` APIs to the RepoClient interface. The underlying data is decompressed avoiding multiple decompressions during the run.

No.


clients/githubrepo/client.go
```diff
@@ -15,48 +15,89 @@
...
+func extractAndValidateArchivePath(path, dest string) (string, error) {
+	const splitLength = 2
+	names := strings.SplitN(path, "/", splitLength)
+	if len(names) < splitLength {
+		log.Printf("Unable to split path: %s", path)
+		return "", nil
+	}
+	// Check for ZipSlip: https://snyk.io/research/zip-slip-vulnerability
+	cleanpath := filepath.Join(dest, names[1])
+	if !strings.HasPrefix(cleanpath, dest) {
```
oliverchang, line 52: does `dest` end with "/" ? 

Otherwise if `dest` is `/tmp`, we can write to `/tmpblah` as well? 

probably better to clean `dest` too and add a "/".

Also, how about returning the joined value here (if it's valid), rather than discarding the joined value (`dest` is not really necessary at this point), rather than having the caller join this again?

clients/githubrepo/client.go
```diff
@@ -15,48 +15,89 @@
...
+const (
+	repoDir      = "repo*"
+	repoFilename = "githubrepo*.tar.gz"
+)
+
+var errZipSlip = errors.New("ZipSlip path detected")
+
+func extractAndValidateArchivePath(path, dest string) (string, error) {
+	const splitLength = 2
+	names := strings.SplitN(path, "/", splitLength)
```
oliverchang, line 45: so the archive always has a top level directory containing all the actual files? good to leave a comment here to explain this.

clients/githubrepo/client.go
```diff
@@ -72,27 +113,114 @@ func (client *Client) InitRepo(owner, repoName string) error {
...
+				continue
+			}
+			filenamepath, err := extractAndValidateArchivePath(header.Name, client.tempDir)
+			if err != nil {
+				return fmt.Errorf("error extracting file path: %w", err)
+			}
+			if filenamepath == "" {
+				continue
+			}
+			outFile, err := os.Create(filepath.Join(client.tempDir, filenamepath))
```
oliverchang, line 174: are we guaranteed that any required dirs will come before the file entries?

cmd/root.go
```diff
@@ -135,6 +135,8 @@ or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show
 		githubClient := github.NewClient(httpClient)
 		graphClient := githubv4.NewClient(httpClient)
 		repoClient := githubrepo.CreateGithubRepoClient(ctx, githubClient)
+		defer repoClient.Close()
```
oliverchang, line 138: The error gets discarded and never checked right?

oliverchang (approved):
nice!

clients/githubrepo/client.go
azeemshaikh38: > does `dest` end with "/" ?
> 
> Otherwise if `dest` is `/tmp`, we can write to `/tmpblah` as well?
> 
> probably better to clean `dest` too and add a "/".
> 

Nice catch! Done.

> Also, how about returning the joined value here (if it's valid), rather than discarding the joined value (`dest` is not really necessary at this point), rather than having the caller join this again?

Done.

clients/githubrepo/client.go
azeemshaikh38: Yes. Done.

clients/githubrepo/client.go
azeemshaikh38: Good point. Done.

cmd/root.go
azeemshaikh38: Yes, the error is never checked in the current implementation.

PR #649, opened 2021-07-02
🐛 Fix bug causing performance issues

azeemshaikh38:

Not cleaning `files` list causes performance issues.

No.

### DIFF
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -228,6 +228,8 @@ func (client *Client) cleanup() error {
 	if err := os.RemoveAll(client.tempDir); err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("os.Remove: %w", err)
 	}
+	// Remove old files so we don't iterate through them.
+	client.files = nil
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


clients/githubrepo/client.go
```diff
@@ -228,6 +228,7 @@ func (client *Client) cleanup() error {
 	if err := os.RemoveAll(client.tempDir); err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("os.Remove: %w", err)
 	}
+	client.files = nil
```
naveensrinivasan, line 231: How is this causing performance issues? It would be helpful if you can add comments to the code so that in the future it can help others when they are fixing issues.

azeemshaikh38: If we don't clean this up, the list of files we iterate through keeps growing. Added comment.

laurentsimon: I suspect it's because the code keeps appending to the `files`. So the code ends up iterating over all previous's project files, instead of just the one it should be looking at.

laurentsimon (approved):

clients/githubrepo/client.go
naveensrinivasan: Thanks! @azeemshaikh38

naveensrinivasan (approved):

PR #650, opened 2021-07-03
✨ Improve check failure/success details

laurentsimon:

feature

it's hard for user to have information about failures, and where to find remediation steps.

WARNING: I'madding everyone as reviewer to get enough feedback!

This PR adds more information for each check failure. This PR only shows a PoC for 4 of our current checks. The functions that needed change have been duplicated as `funcName2`; and variables changed to `varName2` for this PoC. The changes will happen progressively. Once all the checks have been updated, we can rename the variables to their original names.

For each failure, an error number is shown. The user can use this number to get more information about the failure via, for example:
`scorecard --explain Automatic-Dependency-Update-E01`

(This is inspired by the Rust toolchain).

I have not implemented this `--explain` functionality yet. I'll do that when we are in agreement about everything else in this PR.

An error number is returned for every failure in a check: this allows us to have different error numbers for sub-checks: dockerfile pinning, curl | bash, etc within a single check. Note that some checks have a single failure. Checks that use `MultiCheckAnd` will have multiple sub-checks.

Note that if the check only perform one check, it may feel award to have specific error numbers. We could have 
```
`scorecard --explain Automatic-Dependency-Update`
```
for certain checks. The current code already supports this. it suffices to never add log details in the check, and the error message above will automatically be updated (see `pkg/scorecard_result.go` file). The yaml file, as currently updated, will also make this work automatically.

Example of output:
```
Repo: github.com/a1ive/grub2-filemanager
Frozen-Deps: Fail (Confidence=10)
 FAIL/LockFile: no lock file found in the repo.
   For more information about this failure, try `scorecard --explain Automatic-Dependency-Update/LockFile`
 FAIL/GitHubActions: .github/workflows/build.yml has non-pinned dependency 'actions/checkout@v1' (job 'build').
   For more information about this failure, try `scorecard --explain Automatic-Dependency-Update/GitHubActions`
 FAIL/GitHubActions: .github/workflows/build.yml has non-pinned dependency 'marvinpinto/action-automatic-releases@latest' (job 'build').
 FAIL/GitHubActions: .github/workflows/sync_gitee.yml has non-pinned dependency 'actions/checkout@v1' (job 'build').
 PASS/Dockerfile: Dockerfile dependencies are pinned.
   For more information about this result, try `scorecard --explain Automatic-Dependency-Update/Dockerfile`
 PASS/BinaryDownload: no binary downloads found in Dockerfiles.
   For more information about this result, try `scorecard --explain Automatic-Dependency-Update/BinaryDownload`
 PASS/BinaryDownload: no binary download found in GitHub workflows.
```

```
go run . --repo=https://github.com/ossf/scorecard --checks Automatic-Dependency-Update --show-details
Starting [Automatic-Dependency-Update]
Finished [Automatic-Dependency-Update]

RESULTS
-------
Repo: github.com/ossf/scorecard
Automatic-Dependency-Update: Pass (Confidence=10)
 PASS: dependabot config found: .github/dependabot.yml.
For more information about this result, try `scorecard --explain Automatic-Dependency-Update`
```

```
go run . --repo=https://github.com/ossf/scorecard --checks Binary-Artifacts --show-details
Starting [Binary-Artifacts]
Finished [Binary-Artifacts]

RESULTS
-------
Repo: github.com/ossf/scorecard
Binary-Artifacts: Pass (Confidence=10)
 PASS: no binary files found in the repo.
For more information about this result, try `scorecard --explain Binary-Artifacts`
```

It's easy to adapt the code to save into a SARIF format or Json, since there is enough information available in the details and in the yaml file. If we want to display full remediation steps, e.g. for `scorecard --explain Automatic-Dependency-Update-E01`, we can load the info from the yaml file.

Please read some of the comments I added in the PR.

### DIFF
--- a/checker/check_request.go
+++ b/checker/check_request.go
@@ -30,6 +30,8 @@ type CheckRequest struct {
 	GraphClient *githubv4.Client
 	HTTPClient  *http.Client
 	RepoClient  clients.RepoClient
+	// Note: Ultimately Log will be removed and replaced by
+	// CLogger.
 	Logf        func(s string, f ...interface{})
 	CLogger     CheckLogger
 	Owner, Repo string
--- a/checker/check_request.go
+++ b/checker/check_request.go
@@ -31,6 +31,6 @@ type CheckRequest struct {
 	HTTPClient  *http.Client
 	RepoClient  clients.RepoClient
 	Logf        func(s string, f ...interface{})
-	Logf2       func(typ int, code, desc string, f ...interface{})
+	CLogger     CheckLogger
 	Owner, Repo string
 }
--- a/checker/check_request.go
+++ b/checker/check_request.go
@@ -31,5 +31,6 @@ type CheckRequest struct {
 	HTTPClient  *http.Client
 	RepoClient  clients.RepoClient
 	Logf        func(s string, f ...interface{})
+	Logf2       func(typ int, code, desc string, f ...interface{})
 	Owner, Repo string
 }
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -16,6 +16,7 @@ package checker
 
 import (
 	"errors"
+	"fmt"
 
 	scorecarderrors "github.com/ossf/scorecard/errors"
 )
@@ -25,10 +26,33 @@ const MaxResultConfidence = 10
 // ErrorDemoninatorZero indicates the denominator for a proportional result is 0.
 var ErrorDemoninatorZero = errors.New("internal error: denominator is 0")
 
+// Types of details.
+const (
+	DetailFail = 0
+	DetailPass = 1
+	DetailInfo = 2
+)
+
+// CheckDetail contains information for each detail.
+//nolint:govet
+type CheckDetail struct {
+	Type int    // Any of DetailFail, DetailPass, DetailInfo.
+	Code string // A 4 digit string identifying the code, e.g. for remediation.
+	Desc string // A short string representation of the information.
+}
+
+func (cd *CheckDetail) Validate() {
+	if cd.Type < DetailFail ||
+		cd.Type > DetailInfo {
+		panic(fmt.Sprintf("invalid CheckDetail type: %v", cd.Type))
+	}
+}
+
 type CheckResult struct {
 	Error       error `json:"-"`
 	Name        string
 	Details     []string
+	Details2    []CheckDetail
 	Confidence  int
 	Pass        bool
 	ShouldRetry bool `json:"-"`
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -28,9 +28,10 @@ var ErrorDemoninatorZero = errors.New("internal error: denominator is 0")
 
 // Types of details.
 const (
-	DetailFail = 0
-	DetailPass = 1
-	DetailInfo = 2
+	DetailFail    = 0
+	DetailPass    = 1
+	DetailInfo    = 2
+	DetailWarning = 3
 )
 
 // CheckDetail contains information for each detail.
@@ -43,7 +44,7 @@ type CheckDetail struct {
 
 func (cd *CheckDetail) Validate() {
 	if cd.Type < DetailFail ||
-		cd.Type > DetailInfo {
+		cd.Type > DetailWarning {
 		panic(fmt.Sprintf("invalid CheckDetail type: %v", cd.Type))
 	}
 }
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -27,8 +27,8 @@ const (
 	MinResultConfidence  = 0
 )
 
-// UPGRADEv2: to remove.
 // ErrorDemoninatorZero indicates the denominator for a proportional result is 0.
+// UPGRADEv2: to remove.
 var ErrorDemoninatorZero = errors.New("internal error: denominator is 0")
 
 //nolint
@@ -172,6 +172,8 @@ func MakeAndResult2(checks ...CheckResult) CheckResult {
 
 	worseResult := checks[0]
 
+	// UPGRADEv2: will go away after old struct is removed.
+	//nolint
 	for _, result := range checks[1:] {
 		if result.Score2 < worseResult.Score2 {
 			worseResult = result
@@ -248,6 +250,7 @@ func MakeProportionalResult(name string, numerator int, denominator int,
 }
 
 // Given a min result, check if another result is worse.
+//nolint
 func isMinResult(result, min CheckResult) bool {
 	if Bool2int(result.Pass) < Bool2int(min.Pass) {
 		return true
@@ -267,7 +270,8 @@ func MakeAndResult(checks ...CheckResult) CheckResult {
 		Pass:       true,
 		Confidence: MaxResultConfidence,
 	}
-
+	// UPGRADEv2: will go away after old struct is removed.
+	//nolint
 	for _, result := range checks {
 		if minResult.Name == "" {
 			minResult.Name = result.Name
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -59,7 +59,7 @@ type CheckResult struct {
 	Error      error `json:"-"`
 	Name       string
 	Details    []string
-	Details2   []CheckDetail json:"-"
+	Details2   []CheckDetail `json:"-"`
 	Confidence int
 	// Note: Pass2 will ultimately be renamed
 	// as Pass.
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -173,7 +173,6 @@ func MakeAndResult2(checks ...CheckResult) CheckResult {
 	}
 
 	worseResult := checks[0]
-
 	// UPGRADEv2: will go away after old struct is removed.
 	//nolint
 	for _, result := range checks[1:] {
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -28,10 +28,10 @@ var ErrorDemoninatorZero = errors.New("internal error: denominator is 0")
 
 // Types of details.
 const (
-	DetailFail    = 0
-	DetailPass    = 1
-	DetailInfo    = 2
-	DetailWarning = 3
+	DetailFail = 0
+	DetailPass = 1
+	DetailInfo = 2
+	DetailWarn = 3
 )
 
 // CheckDetail contains information for each detail.
@@ -44,10 +44,17 @@ type CheckDetail struct {
 
 func (cd *CheckDetail) Validate() {
 	if cd.Type < DetailFail ||
-		cd.Type > DetailWarning {
+		cd.Type > DetailWarn {
 		panic(fmt.Sprintf("invalid CheckDetail type: %v", cd.Type))
 	}
 }
+
+// Types of results.
+const (
+	ResultPass     = 0
+	ResultFail     = 1
+	ResultDontKnow = 2
+)
 
 type CheckResult struct {
 	Error       error `json:"-"`
@@ -56,6 +63,7 @@ type CheckResult struct {
 	Details2    []CheckDetail
 	Confidence  int
 	Pass        bool
+	Pass2       int
 	ShouldRetry bool `json:"-"`
 }
 
@@ -64,6 +72,7 @@ func MakeInconclusiveResult(name string, err error) CheckResult {
 		Name:       name,
 		Pass:       false,
 		Confidence: 0,
+		Pass2:      ResultDontKnow,
 		Error:      scorecarderrors.MakeZeroConfidenceError(err),
 	}
 }
@@ -72,6 +81,7 @@ func MakePassResult(name string) CheckResult {
 	return CheckResult{
 		Name:       name,
 		Pass:       true,
+		Pass2:      ResultPass,
 		Confidence: MaxResultConfidence,
 	}
 }
@@ -80,6 +90,7 @@ func MakeFailResult(name string, err error) CheckResult {
 	return CheckResult{
 		Name:       name,
 		Pass:       false,
+		Pass2:      ResultFail,
 		Confidence: MaxResultConfidence,
 		Error:      err,
 	}
@@ -89,11 +100,14 @@ func MakeRetryResult(name string, err error) CheckResult {
 	return CheckResult{
 		Name:        name,
 		Pass:        false,
+		Pass2:       ResultDontKnow,
 		ShouldRetry: true,
 		Error:       scorecarderrors.MakeRetryError(err),
 	}
 }
 
+// TODO: update this function to return a ResultDontKnow
+// if the confidence is low?
 func MakeProportionalResult(name string, numerator int, denominator int,
 	threshold float32) CheckResult {
 	if denominator == 0 {
@@ -103,6 +117,7 @@ func MakeProportionalResult(name string, numerator int, denominator int,
 		return CheckResult{
 			Name:       name,
 			Pass:       false,
+			Pass2:      ResultFail,
 			Confidence: MaxResultConfidence,
 		}
 	}
@@ -111,6 +126,7 @@ func MakeProportionalResult(name string, numerator int, denominator int,
 		return CheckResult{
 			Name:       name,
 			Pass:       true,
+			Pass2:      ResultPass,
 			Confidence: int(actual * MaxResultConfidence),
 		}
 	}
@@ -118,6 +134,7 @@ func MakeProportionalResult(name string, numerator int, denominator int,
 	return CheckResult{
 		Name:       name,
 		Pass:       false,
+		Pass2:      ResultFail,
 		Confidence: MaxResultConfidence - int(actual*MaxResultConfidence),
 	}
 }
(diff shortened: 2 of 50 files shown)


checker/check_request.go
```diff
@@ -31,5 +31,6 @@ type CheckRequest struct {
 	HTTPClient  *http.Client
 	RepoClient  clients.RepoClient
 	Logf        func(s string, f ...interface{})
```
laurentsimon, line 33: `Logf` will ultimately be removed.

checker/check_result.go
```diff
@@ -25,12 +26,44 @@ const MaxResultConfidence = 10
...
+type CheckDetail struct {
+	Type int    // Any of DetailFail, DetailPass, DetailInfo.
+	Code string // A 4 digit string identifying the code, e.g. for remediation.
+	Desc string // A short string representation of the information.
+}
+
+func (cd *CheckDetail) Validate() {
+	if cd.Type < DetailFail ||
+		cd.Type > DetailWarn {
+		panic(fmt.Sprintf("invalid CheckDetail type: %v", cd.Type))
```
laurentsimon, line 48: this should never happen and indicate a programming error.

checker/check_result.go
```diff
@@ -64,11 +100,14 @@ func MakeRetryResult(name string, err error) CheckResult {
...
 		Name:        name,
 		Pass:        false,
+		Pass2:       ResultDontKnow,
 		ShouldRetry: true,
 		Error:       scorecarderrors.MakeRetryError(err),
 	}
 }
 
+// TODO: update this function to return a ResultDontKnow
+// if the confidence is low?
```
laurentsimon, line 110: should this decision be left to the check itself, rather than trying to decide for all checks here? Since checks are very different, I think it's hard to do that in this function in practice.

checker/check_result.go
```diff
@@ -64,11 +100,14 @@ func MakeRetryResult(name string, err error) CheckResult {
 	return CheckResult{
 		Name:        name,
 		Pass:        false,
```
laurentsimon, line 102: `Pass` will ultimately be removed.

checker/check_runner.go
```diff
@@ -40,7 +40,36 @@ type CheckFn func(*CheckRequest) CheckResult
 type CheckNameToFnMap map[string]CheckFn
 
 type logger struct {
-	messages []string
+	messages  []string
```
laurentsimon, line 43: `messages` will ultimately be removed.

checks/automatic_dependency_update.go
```diff
@@ -20,32 +20,38 @@ import (
...
-	if !result.Pass {
-		result.Confidence = 3
+	r, err := CheckIfFileExists2(checkAutomaticDependencyUpdate, c, fileExists)
+	if err != nil {
+		return checker.MakeInconclusiveResult(checkAutomaticDependencyUpdate, err)
 	}
-	return result
+	if !r {
+		c.CLogger.Fail("E01", "no configuration file found in the repo")
+		return checker.MakeInconclusiveResult(checkAutomaticDependencyUpdate, nil)
```
laurentsimon, line 38: I'm thinking of updating the signature of `MakeInconclusiveResult ` to take a `CLogger` and code number as input: this way we can simply call `MakeInconclusiveResult ` and get a code number added to the check details. That will reduce cognitive load on our side.

checks/checks2.yaml
```diff
@@ -0,0 +1,57 @@
...
+checks:
+  Binary-Artifacts:
+    description: >-
+      This check tries to determine if a project has binary artifacts in the source repository.
+      These binaries could be compromised artifacts. Building from the source is recommended.
+    url: https://github.com/ossf/scorecard/blob/main/checks/checks.md#binary-artifacts
+    remediation:
+      - >-
+        Remove the binary artifacts from the repository. 
+    errors:
```
laurentsimon, line 26: these error list may seem akward (as I describe in the PR description). It's needed to support more complex checks like the pinning check that contains many sub-checks. It's easy to leave the error list empty if we want AND not call `Fail("E01)` in the code. The display of errors will automatically adjust.

checker/check_request.go
azeemshaikh38: Super nit: instead of review comments will help if you added these as code comments :)

checker/check_result.go
```diff
@@ -25,12 +26,44 @@ const MaxResultConfidence = 10
...
+}
+
+// Types of results.
+const (
+	ResultPass     = 0
+	ResultFail     = 1
+	ResultDontKnow = 2
+)
+
 type CheckResult struct {
```
azeemshaikh38, line 59: FYI, updating `CheckResult` schema will require a significant amount of work. At minimum, doing any changes here will require:

1. Updating cron job code which outputs JSON.
2. Updating BQ schema.
3. Following up with existing clients of BQ.

checker/check_result.go
azeemshaikh38: We can pass 2 extra args here `resultWhenBelowThreshold` and `resultWhenAboveThreshold`.

checks/automatic_dependency_update.go
```diff
@@ -20,32 +20,38 @@ import (
 	"github.com/ossf/scorecard/checker"
 )
 
-const CheckAutomaticDependencyUpdate = "Automatic-Dependency-Update"
+const checkAutomaticDependencyUpdate = "Automatic-Dependency-Update"
```
azeemshaikh38, line 23: Why are choosing to stop exporting these check names?

checker/check_result.go
azeemshaikh38: Spoke too soon. Just add `json:"-"` to the new fields to avoid doing any of this.

checker/check_result.go
```diff
@@ -25,12 +26,44 @@ const MaxResultConfidence = 10
 // ErrorDemoninatorZero indicates the denominator for a proportional result is 0.
 var ErrorDemoninatorZero = errors.New("internal error: denominator is 0")
 
+// Types of details.
```
azeemshaikh38, line 29: IMO, we can capture all info here (from line 29-58) using custom errors.

```
type ScorecardError struct {
  ErrorLevel int  // enum values: Info, Warn, Log
  ErrorType  int  // enum values: Pass, Fail, NA
}
func (e ScorecardError) Error() string {}
func (e ScorecardError) Unwrap() error {}
```

We then update `CheckResult` to have a new field`Errors []ScorecardError`  or `Errors []error`. This will contain all failures from a single check. 

This is also useful when we expose Scorecard as a library, since the caller can propagate/handle these like regular Golang errors.

checker/check_result.go
```diff
@@ -25,12 +26,44 @@ const MaxResultConfidence = 10
...
+const (
```
azeemshaikh38, line 30: FYI, a good way to emulate enums in Golang is to use custom types and `iota`. See example [here](https://github.com/ossf/scorecard/blob/main/checks/ci_tests.go#L27). You can avoid the validation logic also with this.

checker/check_result.go
azeemshaikh38: See https://github.com/ossf/scorecard/blob/main/errors/types.go#L44 for example.

oliverchang (commented):
One high level comment: There's no context passed in between the scorecard run and the proposed `--explain` apart from an error code. 

It seems beneficial for the detailed error/remediation messages to be a template etc, where context from the actual check run is passed through? Some context might be included in the log message with the `CLogger.Fail` call, but it seems better for UX if we just merge this together with the `--explain` output to provide a single cohesive set of remediation steps rather than 2 separate ones that a human has to mentally connect? 

Do we need a separate step (`--explain`) to show the detailed error?  We can also avoid non-descriptive error codes if we don't have this indirection.

checks/checks2.yaml
```diff
@@ -0,0 +1,57 @@
...
+      E01:
+        description: >-
+          A binary file is found in the repository.
+        url: https://github.com/ossf/scorecard/blob/main/checks/checks.md#binary-artifacts-E01
```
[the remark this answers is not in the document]
laurentsimon: they can. I've removed them from the yaml file.

checks/checks2.yaml
laurentsimon: yes, thanks for the suggestion. I've used better names. PTAL.

checks/automatic_dependency_update.go
laurentsimon: these variables are not exported and not used outside this file. They are passed as arguments to `registerCheck` only, so I've renamed them.

checker/check_result.go
laurentsimon: > IMO, we can capture all info here (from line 29-58) using custom errors.
> 
> ```
> type ScorecardError struct {
>   ErrorLevel int  // enum values: Info, Warn, Log
>   ErrorType  int  // enum values: Pass, Fail, NA

I'm not sure this always works: some warning are intended for the user and don't indicate failures or success per-se. For example, if we log every commit that is reviewed, eg [code_review.go](https://github.com/ossf/scorecard/blob/main/checks/code_review.go#L127): it indicates neither failure, nor success, nor NA.
> }
> func (e ScorecardError) Error() string {}
> func (e ScorecardError) Unwrap() error {}
> ```
> 
> We then update `CheckResult` to have a new field`Errors []ScorecardError` or `Errors []error`. This will contain all failures from a single check.
> 
> This is also useful when we expose Scorecard as a library, since the caller can propagate/handle these like regular Golang errors.

I did not do it this way because using errors for failed passes conflates runtime errors with results. What I've implemented is equivalent to:
```
details, err := runCheck()
if err != nil {
   // we could not run the check
}
// check was run, we can now inspect the results
```

AFAIK, errors should indicate a runtime error to perform the check: for example we cannot parse a file, we cannot parse an API response, the repo does not exist, the API response indicates an error, etc

Check failures/passes are just results the API caller can look at, but there's no need to 'propagate' them like errors. Currently, we propagate all errors we encounter: for example [frozen_deps.go](https://github.com/ossf/scorecard/blob/main/checks/frozen_deps.go#L102), [binary_artifact.go](https://github.com/ossf/scorecard/blob/main/checks/binary_artifact.go#L80), [code_review.go](https://github.com/ossf/scorecard/blob/main/checks/code_review.go#L91) But I think this is wrong because it exposes low-level implementation details the caller should not rely upon. This is called out in https://blog.golang.org/go1.13-errors.
If we ever change the implementation, our callers will all break. To be clear, I started doing this myself so I'm the main culprit :-)

There are additional problems to take into account when exposing failures/success to the caller (with or without the use. of `error`): some checks have multiple sub-checks and use `MultiCheckOr` or `MultiCheckAnd`: the caller needs to know how to interpret these: if a FAIL is present in the result, it may not indicate a failure of the check overall if `MultiCheckOr` is used. This PR does not fix all these problems: It only tries to provide enough information to clarify to the user what the check does and why it fails or has low confidence.

I'll send a follow-up PoC PR with suggestions on error handling so we can discuss it more. (by error, I only mean runtime errors).
After that we'll need to think more of the API/results we want to expose to callers.

checker/check_result.go
laurentsimon: Thanks for the tip! I updated the code using `iota`

laurentsimon:
> One high level comment: There's no context passed in between the scorecard run and the proposed `--explain` apart from an error code.
> 
> It seems beneficial for the detailed error/remediation messages to be a template etc, where context from the actual check run is passed through? Some context might be included in the log message with the `CLogger.Fail` call, but it seems better for UX if we just merge this together with the `--explain` output to provide a single cohesive set of remediation steps rather than 2 separate ones that a human has to mentally connect?
> 
> Do we need a separate step (`--explain`) to show the detailed error? We can also avoid non-descriptive error codes if we don't have this indirection.

I don't know the answer. Having all in one place has it pros like you mention, but it also bloats the output. We could have leave it up to the user to decide what they want:
`scorecard --repo ... --show-remediation` or 
```
scorecard --repo ... 
scorecard --explain CheckName/FailName
```

Note that even checks that pass may need some explanation: users don't always understand why the check succeeds or has low confidence. The logging we add is explanatory to the user, and `scorecard --explain` lets users get further details on the check without the need for a web browser or a url to visit (like a `man scorecard`).

Let's discuss more in the next meeting.

checker/check_result.go
laurentsimon: Thanks

checker/check_request.go
laurentsimon: agreed. I pushed the PR and then realized it needed some explanation. Thanks for reminding me of adding the comments in the code!

oliverchang:
> > One high level comment: There's no context passed in between the scorecard run and the proposed `--explain` apart from an error code.
> > It seems beneficial for the detailed error/remediation messages to be a template etc, where context from the actual check run is passed through? Some context might be included in the log message with the `CLogger.Fail` call, but it seems better for UX if we just merge this together with the `--explain` output to provide a single cohesive set of remediation steps rather than 2 separate ones that a human has to mentally connect?
> > Do we need a separate step (`--explain`) to show the detailed error? We can also avoid non-descriptive error codes if we don't have this indirection.
> 
> I don't know the answer. Having all in one place has it pros like you mention, but it also bloats the output.
> To reduce the bloat, we could color-code the information: red for failures, green for passes, purple for NA, some other color for remediations steps, etc. Wdut?

Yeah, colour coding or some other way of organising info should work here. I'm not sure bloat is a strong argument against this. It seems much more annoying for an end user if they have to issue many commands (1 + N explain commands) and mentally connect the results to get a full picture of what's going on.

laurentsimon:
@oliverchang @azeemsgoogle I've addressed the comments and implemented what we discussed offline. I'm working on unit tests, but you can review already. Example of output:
```
go run . --repo=github.com/googleapis/gaxios --checks Frozen-Deps,Binary-Artifacts,Automatic-Dependency-Update,Code-Review --show-details --v2
Starting [Code-Review]
Starting [Frozen-Deps]
Starting [Binary-Artifacts]
Starting [Automatic-Dependency-Update]
Finished [Automatic-Dependency-Update]
Finished [Code-Review]
Finished [Frozen-Deps]
Finished [Binary-Artifacts]

RESULTS
-------
|-------|--------------------------------|-----------------------------|----------------------------------------|------------------------------------------------------------------------------------------|
| SCORE |             REASON             |            NAME             |                DETAILS                 |                                 DOCUMENTATION/REMDIATION                                 |
|-------|--------------------------------|-----------------------------|----------------------------------------|------------------------------------------------------------------------------------------|
| 10    | tool detected                  | Automatic-Dependency-Update |  Info: renovate detected:              | https://github.com/ossf/scorecard/blob/main/checks/checks.md#automatic-dependency-update |
|       |                                |                             | renovate.json                          |                                                                                          |
|-------|--------------------------------|-----------------------------|----------------------------------------|------------------------------------------------------------------------------------------|
| 10    | no binaries found in the repo  | Binary-Artifacts            |                                        | https://github.com/ossf/scorecard/blob/main/checks/checks.md#binary-artifacts            |
|-------|--------------------------------|-----------------------------|----------------------------------------|------------------------------------------------------------------------------------------|
| 10    | GitHub code reviews found for  | Code-Review                 |                                        | https://github.com/ossf/scorecard/blob/main/checks/checks.md#code-review                 |
|       | 30 commits out of the last 30  |                             |                                        |                                                                                          |
|       | -- code normalized to 10       |                             |                                        |                                                                                          |
|-------|--------------------------------|-----------------------------|----------------------------------------|------------------------------------------------------------------------------------------|
| 0     | GitHub actions are not pinned  | Frozen-Deps                 |  Warn: unpinned dependency detected    | https://github.com/ossf/scorecard/blob/main/checks/checks.md#frozen-deps                 |
|       |                                |                             | in .github/workflows/ci.yaml:          |                                                                                          |
|       |                                |                             | 'actions/checkout@v2' (job 'lint')     |                                                                                          |
|       |                                |                             | Warn: unpinned dependency detected     |                                                                                          |
|       |                                |                             | in .github/workflows/ci.yaml:          |                                                                                          |
|       |                                |                             | 'actions/setup-node@v1' (job 'lint')   |                                                                                          |
|       |                                |                             | Warn: unpinned dependency detected     |                                                                                          |
|       |                                |                             | in .github/workflows/ci.yaml:          |                                                                                          |
```

checks/checks2.yaml
```diff
@@ -0,0 +1,108 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This is the source of truth for all check descriptions and remediation steps.
+# Run `cd checks/main && go run /main` to generate `checks.json` and `checks.md`.
+checks:
+  Binary-Artifacts:
```
laurentsimon, line 18: @inferno-chromium @oliverchang wdut of the risk levels I gave?

checks/automatic_dependency_update.go
azeemshaikh38: We do use these check names outside:

https://github.com/ossf/scorecard/blob/main/cron/worker/main.go#L183

laurentsimon: thanks. Will revert this hen.

laurentsimon:
closing since it's superseded by other commits.

PR #656, opened 2021-07-04
:sparkles: Included security.rst as SecurityPolicy

### LINKED ISSUE #655: Feature - accept alternate suffixes for security policy namely `.rst`
The script does not detect a security policy named `security.rst`.

Currently only `security.md` is accepted as a policy. Please consider also detecting at least `security.rst` as RST is a common text format which is directly supported on GH etc.

I could rename the policy file to match what the script expects. But as the policy is one of the inputs that are fed into Sphinx to generate the project website and documentation, then the policy would be omitted. Which would not be very helpful.

naveensrinivasan:

  Feature

 Fixes #655

  * Included security.rst as name check for security policy.

None

* **Other information**:
None

### DIFF
--- a/checks/security_policy.go
+++ b/checks/security_policy.go
@@ -34,6 +34,9 @@ func SecurityPolicy(c *checker.CheckRequest) checker.CheckResult {
 		if strings.EqualFold(name, "security.md") {
 			logf("security policy : %s", name)
 			return true, nil
+		} else if isSecurityrstFound(name) {
+			logf("security policy : %s", name)
+			return true, nil
 		}
 		return false, nil
 	}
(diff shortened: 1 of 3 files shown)


inferno-chromium (approved):
LGTM with one change needed.
There is one more callsite on line 52, so best to create a helper function.

checks/security_policy.go
```diff
@@ -34,6 +34,9 @@ func SecurityPolicy(c *checker.CheckRequest) checker.CheckResult {
 		if strings.EqualFold(name, "security.md") {
 			logf("security policy : %s", name)
 			return true, nil
+		} else if strings.EqualFold(name, "**/security.rst") {
```
inferno-chromium, line 37: we should not allow arbitrary paths here since security policy should be discoverable. i think that helper function i am saying can account for docs/ directory in root, but lets not allow any other paths ?

naveensrinivasan, line 37: @azeemshaikh38  The file glob pattern doesn't seem to work with the refactor. `security.rst` could be `docs` or other folders and it is not consistent across repositories.

inferno-chromium: we are all out this week, can discuss in next week sync.

naveensrinivasan: Took care of it! Now it will just look at either `doc` or `docs` folder

naveensrinivasan: OK

Issue #661, opened 2021-07-05
Feature: Reorder/sort/color default output format for better legibility


katrinleinweber:
**Is your feature request related to a problem? Please describe.**

In the `default` output ([examples](https://github.com/ossf/scorecard/blame/v2.0.0/README.md#L107-L120)) it's difficult to visually find the `fail`ing items.

**Describe the solution you'd like**

Swap the column order to for example:

```
Fail  3 Automatic-Dependency-Update
Pass 10 Binary-Artifacts
Fail  0 Branch-Protection
Pass 10 CI-Tests
Pass 10 CII-Best-Practices
Pass 10 Code-Review
Pass 10 Contributors
Fail 10 Frozen-Deps
Pass 10 Fuzzing
Fail  0 Packaging
Pass 10 Pull-Requests
Fail 10 SAST
Fail  5 Security-Policy
Fail 10 Signed-Releases
Fail 10 Signed-Tags
Pass 10 Token-Permissions
```

**Describe alternatives you've considered**

- coloring: red/green or only red for `Fail`
- sorting the `Fail` to either the top of bottom of the list.

Both "alternatives" could be combined with the re-ordering, though.

**Additional context**

CSV & JSON are cool for reporting, but sometimes one just wants to interactively find a problem to work on. An easier-to-read default output would help a lot.

naveensrinivasan:
@katrinleinweber  Thanks for the suggestion! In the initial fix, we aren't including the color.

katrinleinweber:
Wow, that's fast :-) Thank you!

PR #663, opened 2021-07-05
:sparkles: Sorted and grouped the output results

naveensrinivasan:

  Feature

  #661

  * Grouped the output results to pass and fail
  * Sorted them by confidence
  
The output would display like this.
  
```
  RESULTS
-------
Repo: github.com/ossf/scorecard
Pass 10 Active
Pass 10 Automatic-Dependency-Update
Pass 10 Binary-Artifacts
Pass 10 CI-Tests
Pass 9 Code-Review
Pass 10 Contributors
Pass 10 Frozen-Deps
Pass 10 Packaging
Pass 10 Pull-Requests
Pass 10 SAST
Pass 10 Security-Policy
Pass 10 Vulnerabilities
Fail 10 Branch-Protection
Fail 10 CII-Best-Practices
Fail 10 Fuzzing
Fail 3 Signed-Releases
Fail 4 Signed-Tags
Fail 10 Token-Permissions
```

 None

* **Other information**:
None

### DIFF
--- a/pkg/scorecard_result.go
+++ b/pkg/scorecard_result.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


naveensrinivasan:
As a follow-up to this would be something like this https://github.com/olekukonko/tablewriter

pkg/scorecard_result.go
```diff
@@ -86,9 +87,27 @@ func (r *ScorecardResult) AsCSV(showDetails bool, writer io.Writer) error {
 }
 
 func (r *ScorecardResult) AsString(showDetails bool, writer io.Writer) error {
-	fmt.Fprintf(writer, "Repo: %s\n", r.Repo)
+	type result struct {
```
oliverchang, line 90: Rather than mostly duplicate an existing struct, could we do something like:

```go
        sortedChecks := make([]checker.CheckResult, len(r.Checks))
        copy(sortedChecks, r.Checks)
        sort.Slice(sortedChecks, func(i, j int) bool {
                if sortedChecks[i].Pass == sortedChecks[j].Pass {
                        return sortedChecks[i].Name < sortedChecks[j].Name
                }
                return sortedChecks[i].Pass
        })
```

oliverchang (approved):
Nice! LGTM with a suggestion.

pkg/scorecard_result.go
naveensrinivasan: Thanks, I agree. Fixed it 👍

PR #681, opened 2021-07-12
:book: Included docker documentation in README

naveensrinivasan:

    * Included docker run for easier consumption of scorecard.

* **Other information**:


README.md
```diff
@@ -69,6 +70,16 @@ the [checks documentation page](checks/checks.md).
 
 ## Usage
 
+### Docker
```
azeemshaikh38, line 73: Move this section to after `Authentication` header.

azeemshaikh38 (approved):

README.md
naveensrinivasan: I think it should be before because as a consumer, I would first look for a docker container to test and go with that. That was the reason to have it prior to it. 

Having a docker container makes it a lot easier for consumers to try.

azeemshaikh38: Hmm, in terms of readability it makes more sense to talk about the command line first and then explain that there is a docker option too. Also, from all the feedback we have gotten so far, everyone seems to be using the command line more frequently.

README.md
```diff
@@ -69,6 +70,16 @@ the [checks documentation page](checks/checks.md).
...
+
+The `GITHUB_AUTH_TOKEN` has to be set to a valid [token](#Authentication)
+
+``` shell
+docker run -e GITHUB_AUTH_TOKEN=token gcr.io/openssf/scorecard:latest --show-details --repo=https://github.com/ossf/scorecard
+```
+
 ### Using repository URL
 
 The program can run using just one argument, the URL of the repo:
```
azeemshaikh38, line 85: Unrelated, but could you also update the command line instructions to use either:

```
go get github.com/ossf/scorecard
scorecard --repo=...
```

or 

```
make build-scorecard
./scorecard --repo=...
```

README.md
naveensrinivasan: They are using the command line option because the docker instructions were removed from README.md

README.md
naveensrinivasan: I can do that. 👍

README.md
azeemshaikh38: Ok, upto you.

naveensrinivasan:
> Integration tests failure for [30a1045407080c8b88004ec6cd9e43d8d81346e6](https://github.com/ossf/scorecard/actions/runs/1031188264)

I am fixing this with another PR. We removed the docker push from our packaging. 🤦

PR #685, opened 2021-07-13
:bug: Updated checks arguments to be case insensitive

### LINKED ISSUE #674: BUG - Checks are case sensitive
**Describe the bug**
The `checks` are case-sensitive. `go run .  --repo=github.com/ossf/scorecard  --checks=code-review`. This causes the command line to ignore the check that was supposed to `Code-Review`

**Expected behavior**
The checks have to be case-insensitive. 

**Additional context**
Related code https://github.com/ossf/scorecard/blob/main/cmd/root.go#L115

notanton:

Fixes #674

`checks` arguments are case sensitive

`checks` arguments are no longer case sensitive

No

* **Other information**:


checks/all_checks.go
```diff
@@ -14,11 +14,13 @@
 
 package checks
 
+import "strings"
```
naveensrinivasan, line 17: Can you please combine these imports?
```
import(
"strings"
"github.com/ossf/scorecard/checker")
```

This will get rid of the lint failures.

naveensrinivasan (commented):
Thanks, @notanton for the contribution!  Can you please resolve the conflict and the import statement? Then it should be good to be merged.

azeemshaikh38 (changes requested):
This PR will break this line here: https://github.com/ossf/scorecard/blob/main/cron/worker/main.go#L183

cmd/root.go
```diff
@@ -118,8 +118,9 @@ or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show
 		enabledChecks := checker.CheckNameToFnMap{}
 		if len(checksToRun) != 0 {
 			for _, checkToRun := range checksToRun {
-				if checkFn, ok := checks.AllChecks[checkToRun]; ok {
-					enabledChecks[checkToRun] = checkFn
+				checkIndex := strings.ToLower(checkToRun)
```
azeemshaikh38, line 121: We can update this code to:

```
for _, checkToRun := range checksToRun {
  for key, checkFn := range checks.AllChecks {
    if strings.ToLower(checkToRun) == strings.ToLower(key) {
     enabledChecks[key] = checkFn
   }
  }  
}
```

Then the rest of the code can remain the same.

notanton: Thanks @azeemshaikh38. I have made the above change and removed the need to index check functions with the lower case key so that constants like `checks.CheckBranchProtection` remain to work if we need to delete functions from the map, as in the case https://github.com/ossf/scorecard/blob/main/cron/worker/main.go#L183

cmd/root.go
```diff
@@ -117,12 +117,15 @@ or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show
 
 		enabledChecks := checker.CheckNameToFnMap{}
 		if len(checksToRun) != 0 {
+		loop:
 			for _, checkToRun := range checksToRun {
-				if checkFn, ok := checks.AllChecks[checkToRun]; ok {
-					enabledChecks[checkToRun] = checkFn
-				} else {
-					log.Fatalf("Invalid check: %s", checkToRun)
+				for key, checkFn := range checks.AllChecks {
```
azeemshaikh38, line 122: Instead of `continue loop`, put the nested `for` inside a fn:

```
for _, checkToRun := range checksToRun {
  if b := parseCheckName(checkToRun, &enabledChecks); b == false {
    log.Fatalf("Invalid check: %s", checkToRun)
  }
}
```

azeemshaikh38 (approved):

naveensrinivasan (approved):

PR #694, opened 2021-07-15
✨ Add more github token names for env variable

laurentsimon:

feature

we only use GITHUB_AUTH_TOKEN

we add GITHUB_TOKEN and GH_TOKEN. This is helpful because some existing tools use different naming, so re-using the names requires users to set the env variable once only across multiple tools.

no

* **Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,7 @@ For example, `--checks=CI-Tests,Code-Review`.
 Before running Scorecard, you need to, either:
 
 -   [create a GitHub access token](https://docs.github.com/en/free-pro-team@latest/developers/apps/about-apps#personal-access-tokens)
-    and set it in environment variable `GITHUB_AUTH_TOKEN`. This helps to avoid
+    and set it in environment variable called `GITHUB_AUTH_TOKEN`, `GITHUB_TOKEN`, `GH_AUTH_TOKEN` or `GH_TOKEN`. This helps to avoid
     the GitHub's
     [api rate limits](https://developer.github.com/v3/#rate-limiting) with
     unauthenticated requests.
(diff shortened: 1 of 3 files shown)


roundtripper/roundtripper.go
```diff
@@ -39,27 +40,28 @@ const (
...
+}
+
 // NewTransport returns a configured http.Transport for use with GitHub.
 func NewTransport(ctx context.Context, logger *zap.SugaredLogger) http.RoundTripper {
 	transport := http.DefaultTransport
 
-	if token := os.Getenv(GithubAuthToken); token != "" {
-		// Use GitHub PAT
-		transport = githubrepo.MakeGitHubTransport(transport, strings.Split(token, ","))
-	} else if keyPath := os.Getenv(GithubAppKeyPath); keyPath != "" { // Also try a GITHUB_APP
```
laurentsimon, line 49: if I understand comment https://github.com/ossf/scorecard/commit/7cf56e04afc54d1f894a130270dfecf951ae23e0, we ca no longer support app token because of graphQl.

azeemshaikh38: What I meant was that you cannot use graphQL without authorization headers of some sort. It could be PATs or GH app. So we should continue supporting GH apps.

README.md
```diff
@@ -249,19 +249,6 @@ set GITHUB_AUTH_TOKEN=<your access token>
 Multiple `GITHUB_AUTH_TOKEN` can be provided separated by comma to be utilized
 in a round robin fashion.
 
--   create a GitHub App Installations for higher rate-limit quotas. If you have
```
azeemshaikh38, line 252: Undo this if keeping GH apps.

README.md
```diff
@@ -249,19 +249,6 @@ set GITHUB_AUTH_TOKEN=<your access token>
 Multiple `GITHUB_AUTH_TOKEN` can be provided separated by comma to be utilized
```
azeemshaikh38, line 249: Update README to mention other env vars are also supported.

roundtripper/roundtripper.go
```diff
@@ -19,18 +19,19 @@ import (
...
 	"strings"
 
-	"github.com/bradleyfalzon/ghinstallation"
 	"go.uber.org/zap"
 
 	"github.com/ossf/scorecard/clients/githubrepo"
 )
 
+// nolinter
+// GithubAuthToken is for making requests to GiHub's API.
```
azeemshaikh38, line 30: `GithubAuthToken` should be plural in comments to match the var name. Also remove `// nolinter`, which you shouldn't need after updating to plural.

azeemshaikh38 (approved):

roundtripper/roundtripper.go
laurentsimon: gocha.

roundtripper/roundtripper.go
laurentsimon: done.

PR #696, opened 2021-07-15
:seedling: Fix failing e2e tests

naveensrinivasan:

  Bug Fix

   Failing e2e tests

    The packaging docker image for scorecard has been removed from github
    workflow to gcr.io.

    This was causing the e2e check failing.

    This fix will remove that check and address the failing e2e.

  None

* **Other information**:
  None


inferno-chromium (commented):
Slightly worried here, so will let @azeemsgoogle @azeemshaikh38 review this. we might forget to reenable this, so lets see if we can fix it somehow.

naveensrinivasan:
> Slightly worried here, so will let @azeemsgoogle @azeemshaikh38 review this. we might forget to reenable this, so lets see if we can fix it somehow.

We will need to have a GitHub workflow with docker push. TBH I find that not a valid check. As a project, we don't use GitHub action to do any packaging.

naveensrinivasan:
Cannot merge any PR's till we make a decision for this.

inferno-chromium (approved):
Ah there is a test already for apache/orc, then definitely fine to remove scorecard one.

PR #716, opened 2021-07-20
:seedling: Fix the go.mod with v2 upgrade

### LINKED ISSUE #711: Can't import scorecard
I believe go.mod need to be updated from:
```
module github.com/ossf/scorecard
```
to:
```
module github.com/ossf/scorecard/v2
```
To correspond to the v2.0.0 release, which indicates a breaking change in semver, which go uses for imports.
see: https://github.com/golang/go/issues/35732

naveensrinivasan:

bug fix

Fixes #711 

The `go.mod` wasn't updated with the v2 upgrade.https://github.com/ossf/scorecard/issues/711

This fix will address the issue.

None

* **Other information**:
None


jeffmendoza:
Do all self-imports in code need to be changed to v2 as well? Not sure.

naveensrinivasan:
> Do all self-imports in code need to be changed to v2 as well? Not sure.

Agreed, looks like every replacement would require a go.mod https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive

So we would need multiple modules. 

cc @azeemshaikh38

azeemshaikh38:
> > Do all self-imports in code need to be changed to v2 as well? Not sure.
> 
> Agreed, looks like every replacement would require a go.mod https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive
> 
> So we would need multiple modules.
> 
> cc @azeemshaikh38

Looking at a Google repo, the way this seems to be done is by updating all references to self-imports: https://github.com/google/go-github/commit/05e95d3c51abdf832f9430209ccb46b494ac7be8#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6.

I would prefer this way in contrast to the `replace` directive to stay in sync with Google guidelines.

naveensrinivasan:
Apologies for the large PR. This cannot be broken down into smaller chunks as it is changing the version across many files.

azeemshaikh38 (approved):
Thanks!

@jeffmendoza FYI

PR #734, opened 2021-07-22
✨ Add e2e release tests for cron job

azeemshaikh38:

Adds e2e release tests for cron job.

No.


cron/controller/Dockerfile
```diff
@@ -27,6 +26,7 @@ ARG TARGETARCH
 RUN CGO_ENABLED=0 make build-pubsub
 
 FROM gcr.io/distroless/base:nonroot@sha256:bc84925113289d139a9ef2f309f0dd7ac46ea7b786f172ba9084ffdb4cbd9490
+USER 0
```
oliverchang, line 29: what's this needed for?

oliverchang (approved):
nice! what's the expected workflow for determining if the e2e passed and we can promote this to prod?

cron/controller/Dockerfile
azeemshaikh38: So I'm running into a weird issue where Go is unable to open "cron/data/projects.csv" file due to permission error. According to Dockerfile reference, the default user for the `COPY` command is `0`. So I tried setting `USER` to `0` before `ENTRYPOINT` and that seems to solve this. Not sure if this is the right solution though. Any ideas?

oliverchang: That's weird. How did it work before?

It does seem a bit hacky to set the user to root given that we're using a "nonroot" image. Can you try using another distroless image (not the "nonroot" one) and see if that fixes things?

azeemshaikh38: So this issue happened due to https://github.com/ossf/scorecard/pull/706. I only caught it now when writing these e2e tests. Before that we were embedding the file and not reading it through `os.Open`.

Will try the "nonroot" image and get back to you.

azeemshaikh38: Fixed it, thanks to @dlorenc. The permission of the file in the repo only allowed `root` to have read access. Updated that now.

cron/releasetest/k8s/controller.yaml
```diff
@@ -0,0 +1,41 @@
...
+  # Daily at midnight PST.
+  schedule: "0 7 * * *"
+  concurrencyPolicy: "Forbid"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: controller
+            image: gcr.io/openssf/scorecard-batch-controller-releasetest:latest
```
laurentsimon, line 29: is it possible to pin this? Does dependabot support updating it?
I have a tracking issue to add support for this in our Frozen-Deps check.

cron/worker/main.go
```diff
@@ -127,7 +127,7 @@ func createNetClients(ctx context.Context) (
 func startMetricsExporter() (monitoring.Exporter, error) {
 	exporter, err := monitoring.GetExporter()
 	if err != nil {
-		return nil, fmt.Errorf("error during NewStackDriverExporter: %w", err)
+		return nil, fmt.Errorf("error during monitoring.GetExporter: %w", err)
```
laurentsimon, line 130: shall we start using the new command to create errors?

azeemshaikh38:
> nice! what's the expected workflow for determining if the e2e test passed and we can promote this to prod?

Right now, its manual. This is mostly for sanity checking that the cron job doesn't fail for some reason.

I still need to figure out the process for prod releases. Will discuss this in the meeting.

cron/worker/main.go
azeemshaikh38: Yes, we should. Forgot about it in this PR.

cron/releasetest/k8s/controller.yaml
azeemshaikh38: Hmm, the idea is that this should always pull the most recent image to test. Pinning doesn't really make sense here.

PR #736, opened 2021-07-22
✨ catch all errors in BP check

laurentsimon:

bug fix

catch only 404 error

catch all errors

no

* **Other information**:


asraa (approved):
I think this makes sense! At this point just checking error is the same as 404 since we know is protected (otherwise this can sometimes return "unprotected")

inferno-chromium (approved):

naveensrinivasan (approved):

PR #763, opened 2021-07-27
📖 Update local install instructions to use v2

loosebazooka:
The docs instructions for installing locally would `go get` an older version of scorecards. This updates to `v2`.

Signed-off-by: Appu Goundan <appu@google.com>


azeemshaikh38 (approved):
Thanks Appu!

PR #764, opened 2021-07-27
✨Add version cli subcommand

loosebazooka:
`scorecard version` will print out something like

```
GitVersion:     v2.0.0-73-g7fd331a-dirty
GitCommit:      7fd331adf2a0f6578b78d951e16646da03e6a6f5
GitTreeState:   dirty
BuildDate:      2021-07-27T14:14:34Z
GoVersion:      go1.16.4
Compiler:       gc
Platform:       linux/amd64
```

Signed-off-by: Appu Goundan <appu@google.com>

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,0 +1,59 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
+	"runtime"
+
+	"github.com/spf13/cobra"
+)
+
+// Base version information.
+//
+// This is the fallback data used when version information from git is not
+// provided via go ldflags in the Makefile. See version.mk.
+var (
+	// Output of "git describe". The prerequisite is that the branch should be
+	// tagged using the correct versioning strategy.
+	gitVersion = "unknown"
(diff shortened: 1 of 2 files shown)


version.mk
```diff
@@ -0,0 +1,20 @@
+# Set version variables for LDFLAGS
```
loosebazooka, line 1: sorry about all this shell nonsense. this is slightly modified version of https://github.com/sigstore/cosign/blob/main/Makefile#L26-L43

cmd/version.go
```diff
@@ -0,0 +1,58 @@
...
+func init() {
+	rootCmd.AddCommand(versionCmd)
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print version information",
+	Long:  ``,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Printf("GitVersion:\t%s\n", gitVersion)
```
loosebazooka, line 50: should this use the logger?

version.mk
azeemshaikh38: Can we move this logic to `Makefile`?

Also, why a modified version? What was different here compared to cosign?

cmd/version.go
azeemshaikh38: To be consistent with `root.go`, use `log.Printf` here.

azeemshaikh38 (approved):
Thanks!

version.mk
loosebazooka: Didn't want to crowd the makefile, but I can move it.

1. so `PKG` is now calculated using `go list -m`, this allows for automatic package detection if the module changes from v2 -> v3 or whatever.
1. "clean" -> clean, "dirty" -> dirty, these seems unnecessarily quoted
1. VERSION_FLAGS is not quoted, because it needs to be appended to other ldflag options

cmd/version.go
loosebazooka: so `log.Printf` includes a bunch of timing information.

version.mk
azeemshaikh38: Makes sense. Thanks for explaining.

cmd/version.go
azeemshaikh38: Ah ok. Let's keep this then.

Issue #741, opened 2021-07-22
BUG: e2e tests always break


azeemshaikh38:
We have setup our e2e tests such that they query real repositories. If anything changes on their end, our e2e tests break. Thats super flaky and very unproductive since we have to keep fixing them. We need to find a better solution to this. Lets discuss this in the upcoming meet.

laurentsimon:
yep, that's one we need badly. I've left a few comments in the code for that as well.
btw, we already have https://github.com/ossf/scorecard/issues/435

naveensrinivasan:
We add new features which break the existing e2e.  We would have to tend to e2e to keep up with the features that we add.

For example 

This is failing  https://github.com/ossf/scorecard/runs/3156521297?check_suite_focus=true because our expectation of the e2e result changed. 

```
➜  scorecard git:(main) ginkgo --focus="E2E TEST:CITests" ./e2e/...
Running Suite: E2e Suite
========================
Random Seed: 1627250142
Will run 1 of 24 specs

SSSSSSSSSSSSSSSSS
------------------------------
• Failure [8.870 seconds]
E2E TEST:CITests
/home/sammy/go/src/github.com/ossf/scorecard/e2e/ci_tests_test.go:28
  E2E TEST:Validating use of CI tests
  /home/sammy/go/src/github.com/ossf/scorecard/e2e/ci_tests_test.go:29
    Should return use of CI tests [It]
    /home/sammy/go/src/github.com/ossf/scorecard/e2e/ci_tests_test.go:30

    Expected
        <bool>: false
    to be true

    /home/sammy/go/src/github.com/ossf/scorecard/e2e/ci_tests_test.go:55
------------------------------
SSSSSS

Summarizing 1 Failure:

[Fail] E2E TEST:CITests E2E TEST:Validating use of CI tests [It] Should return use of CI tests
/home/sammy/go/src/github.com/ossf/scorecard/e2e/ci_tests_test.go:55

Ran 1 of 24 Specs in 8.872 seconds
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 23 Skipped
--- FAIL: TestE2e (8.88s)
```

azeemshaikh38:
But if new features are breaking the e2e, how are we able to submit PRs which cause these breakages?

laurentsimon:
The e2e tests break because the repos we rely on change over time.

[...]

PR #765, opened 2021-07-27
✨  rename Frozen-Deps to Pinned-Dependencies

laurentsimon:

feature

name is frozen-deps

name is pinned-dependencies

no

* **Other information**:
I also fix https://github.com/ossf/scorecard/issues/741 along the way.
Put all file-related functions into new file `fileutils.go`


checks/fileutils.go
```diff
@@ -15,6 +15,7 @@
 package checks
```
inferno-chromium, line 15: add _ in filename, file_utils.go

checks/fileutils.go
```diff
@@ -106,3 +130,23 @@ func CheckFilesContent(shellPathFnPattern string,
...
+	if len(content) == 0 {
+		return false
+	}
+
+	r := strings.NewReader(string(content))
+	scanner := bufio.NewScanner(r)
+	s := ""
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		s += "'" + line + "'\n"
```
inferno-chromium, line 146: s is unused ?

checks/fileutils.go
```diff
@@ -106,3 +130,23 @@ func CheckFilesContent(shellPathFnPattern string,
 
 	return res, nil
 }
+
+// ContainsInstructions checks if the file contains instructions or not.
+// `comment` is the string or character that indicates a comment.
+func ContainsInstructions(content []byte, comment string) bool {
```
inferno-chromium, line 136: ContainsInstructions is slightly wrong name, since it is not contains but startsWith, maybe lineStartsWith ?
also, comment var name seems incorrect, since you are creating a generic function, so maybe rename to start or something like that ?

inferno-chromium (approved):

checks/fileutils.go
laurentsimon: oops, that was debug code! removed.

checks/fileutils.go
laurentsimon: done.

checks/fileutils.go
laurentsimon: Maybe `ContainsCommands()`?

azeemshaikh38 (approved):

checks/file_utils.go
```diff
@@ -106,3 +130,21 @@ func CheckFilesContent(shellPathFnPattern string,
 
 	return res, nil
 }
+
+// ContainsCommands checks if the file content contains commands or not.
+// `comment` is the string or character that indicates a comment.
```
inferno-chromium, line 135: change |comment| to |command| in this function and comments.

laurentsimon: it's confusing. The function is used to see if the function contains commands. But it does so by checking if it contains only comments or empty lines. If we use comments instead of commands, the function should be called `ContainsOnlyCommentsOrEmptyLines` and I should change the return values false <-> true

laurentsimon: changed to `CheckFileContainsCommands`. Can update in follow-up PR if necessary.

laurentsimon:
@naveensrinivasan how to I bypass the failing check? I've added empty files on purpose for unit tests.

naveensrinivasan:
> @naveensrinivasan how to I bypass the failing check? I've added empty files on purpose for unit tests.
> 
> @inferno-chromium


I don’t understand, so you want to bypass the integration test check?

Why not actually fix the issue? 

Can you please explain?

laurentsimon:
> I've added empty files on purpose for unit tests.

I've added empty files on purpose for unit tests (need to verify that our checks work properly on empty files). The check is failing because it wants every file to contain a copyright notice. 

Please advise. Thanks.

Issue #727, opened 2021-07-21
Feature: use proportional result for Token-Permission check


laurentsimon:
the check is now ported to using a score. But the result are still binary: 0 or 10. 
For example, we may have a distinction between `contents`, `checks` etc. `contents`/`packages`/`actions` seem the most dangerous since it allows pushing code to the repo.

Good repo to try on: github.com/apache/airflow

Issue #646, opened 2021-07-02
Bug - Random test failures


naveensrinivasan:
```
protoc --go_out=../../../ cron/data/request.proto
/bin/bash: protoc: command not found
make: *** [Makefile:88: cron/data/request.pb.go] Error 127
```

_Originally posted by @laurentsimon in https://github.com/ossf/scorecard/issues/611#issuecomment-872455510_

laurentsimon:
@azeemsgoogle

laurentsimon:
happening on another PR https://github.com/ossf/scorecard/pull/766

laurentsimon:
Copying the link @azeemsgoogle provided in https://github.com/ossf/scorecard/pull/766#issuecomment-887962649: https://lukasjoswiak.com/github-actions-protobuf/
@azeemsgoogle do you have cycles to fix this issue? (It's part of the cron job IIUC)

naveensrinivasan:
I can take this.

azeemshaikh38:
Thanks Naveen!

PR #773, opened 2021-07-28
✨  Make Token-Permission check more granular

laurentsimon:

feature

see https://github.com/ossf/scorecard/issues/727

see https://github.com/ossf/scorecard/issues/727

* **Other information**:

### DIFF
--- a/checks/file_utils.go
+++ b/checks/file_utils.go
@@ -173,12 +173,12 @@ func CheckFilesContent2(shellPathFnPattern string,
 			return err
 		}
 
-		contIterate, err := onFileContent(file, content, c.Dlogger, data)
+		continueIter, err := onFileContent(file, content, c.Dlogger, data)
 		if err != nil {
 			return err
 		}
 
-		if !contIterate {
+		if !continueIter {
 			break
 		}
 	}
--- a/checks/file_utils.go
+++ b/checks/file_utils.go
@@ -131,6 +131,60 @@ func CheckFilesContent(shellPathFnPattern string,
 	return res, nil
 }
 
+// FileCbData is any data the caller can act upon
+// to keep state.
+type FileCbData interface{}
+
+// FileContentCb is the callback.
+// The bool returned indicates whether the CheckFilesContent2
+// should continue iterating over files or not.
+type FileContentCb func(path string, content []byte,
+	dl checker.DetailLogger, data FileCbData) (bool, error)
+
+func CheckFilesContent2(shellPathFnPattern string,
+	caseSensitive bool,
+	c *checker.CheckRequest,
+	onFileContent FileContentCb,
+	data FileCbData,
+) error {
+	predicate := func(filepath string) (bool, error) {
+		// Filter out Scorecard's own test files.
+		if isScorecardTestFile(c.Owner, c.Repo, filepath) {
+			return false, nil
+		}
+		// Filter out files based on path/names using the pattern.
+		b, err := isMatchingPath(shellPathFnPattern, filepath, caseSensitive)
+		if err != nil {
+			return false, err
+		}
+		return b, nil
+	}
+
+	matchedFiles, err := c.RepoClient.ListFiles(predicate)
+	if err != nil {
+		// nolint: wrapcheck
+		return err
+	}
+	for _, file := range matchedFiles {
+		content, err := c.RepoClient.GetFileContent(file)
+		if err != nil {
+			//nolint
+			return err
+		}
+
+		ct, err := onFileContent(file, content, c.Dlogger, data)
+		if err != nil {
+			return err
+		}
+
+		if !ct {
+			break
+		}
+	}
+
+	return nil
+}
+
 // CheckFileContainsCommands checks if the file content contains commands or not.
 // `comment` is the string or character that indicates a comment:
 // for example for Dockerfiles, it would be `#`.
--- a/checks/file_utils.go
+++ b/checks/file_utils.go
@@ -165,7 +165,7 @@ func CheckFilesContent2(shellPathFnPattern string,
 		// nolint: wrapcheck
 		return err
 	}
-	fmt.Printf("list %v\n", matchedFiles)
+
 	for _, file := range matchedFiles {
 		content, err := c.RepoClient.GetFileContent(file)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


checks/file_utils.go
```diff
@@ -131,6 +131,61 @@ func CheckFilesContent(shellPathFnPattern string,
...
+// to keep state.
+type FileCbData interface{}
+
+// FileContentCb is the callback.
+// The bool returned indicates whether the CheckFilesContent2
+// should continue iterating over files or not.
+type FileContentCb func(path string, content []byte,
+	dl checker.DetailLogger, data FileCbData) (bool, error)
+
+func CheckFilesContent2(shellPathFnPattern string,
```
laurentsimon, line 144: note: once this PR is merged, I will rename it to `CheckFilesContent` and update other uses of it. Same of `CheckForFile`.

laurentsimon:
failing with https://github.com/ossf/scorecard/issues/646
```
```

checks/permissions.go
```diff
@@ -24,149 +24,235 @@ import (
...
-	return true, nil
+	return nil
+}
+
+// Calculate the score.
+func calculateScore(result permissionCbData) int {
+	// See list https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/.
+	// Note: there are legitimate reasons to use some of the permissions like checks, deployments, etc.
+	// in CI/CD systems https://docs.travis-ci.com/user/github-oauth-scopes/.
+	if _, ok := result.permissions["all"]; ok {
```
laurentsimon, line 156: That's the interesting part I need feedback on

checks/file_utils.go
```diff
@@ -131,6 +131,61 @@ func CheckFilesContent(shellPathFnPattern string,
...
+		}
+		// Filter out files based on path/names using the pattern.
+		b, err := isMatchingPath(shellPathFnPattern, filepath, caseSensitive)
+		if err != nil {
+			return false, err
+		}
+		return b, nil
+	}
+
+	matchedFiles, err := c.RepoClient.ListFiles(predicate)
```
inferno-chromium, line 163: Would this not burn the api quota like crazy ? When we get the archive, we do it once and then everything is read from that unpacked repo archive. this is like git clone, but one call. now you are doing one per file request, that will choke quota.

laurentsimon: the repo client is initialized once only, so the tarball is only downloaded once. Not this code is just a copy of the existing `CheckFilesContent()`

checks/permissions.go
laurentsimon: @inferno-chromium comments on this function? What do you think of the score calculation?

checks/file_utils.go
```diff
@@ -131,6 +131,61 @@ func CheckFilesContent(shellPathFnPattern string,
...
+			return err
+		}
+
+		ct, err := onFileContent(file, content, c.Dlogger, data)
+		if err != nil {
+			return err
+		}
+
+		if !ct {
+			break
```
oliverchang, line 182: Why do we break now instead of continuing (as we did before)?

checks/permissions.go
```diff
@@ -24,149 +24,251 @@ import (
...
 }
 
-func leastPrivilegedTokens(c *checker.CheckRequest) checker.CheckResult {
-	r, err := CheckFilesContent(".github/workflows/*", false, c, validateGitHubActionTokenPermissions)
-	return createResultForLeastPrivilegeTokens(r, err)
+// Holds stateful data to pass thru callbacks.
+// Each field correpsonds to a GitHub permission type, and
+// will hold true if declared non-write, false otherwise.
+type permissionCbData struct {
+	permissions map[string]bool
```
oliverchang, line 38: nit: s/permissions/writePermissions/ or something similar makes this easier to understand.

checks/permissions.go
```diff
@@ -24,149 +24,251 @@ import (
...
+	// status: https://docs.github.com/en/rest/reference/repos#statuses.
+	// May allow an attacker to change the result of pre-submit and get a PR merged.
+	// Low risk: -0.5.
+	if _, ok := result.permissions["statuses"]; ok {
+		score -= 0.5
+	}
+
+	// checks.
+	// May allow an attacker to edit checks to remove pre-submit and introduce a bug.
+	// Low risk: -1.
```
oliverchang, line 170: nit: comment does not match code.

checks/permissions.go
```diff
@@ -24,149 +24,251 @@ import (
...
+	}
+
+	// actions.
+	// May allow an attacker to steal GitHub secrets by adding a malicious workflow/action.
+	// High risk: -10
+	if _, ok := result.permissions["actions"]; ok {
+		score -= checker.MaxResultScore
+	}
+
+	if score < 0 {
```
oliverchang, line 212: nit: use `checker.MinResultScore` in place of both 0s here?

oliverchang (approved):

checks/file_utils.go
laurentsimon: depends on what the caller returns. This gives the caller the chance to decide. I think it's more flexible. Resolving the comment. Please re-open if you disagree.

oliverchang: I see! In that case it would've been much clearer if `ct` was `continue` :) I thought it stood for "content" or something similar before.

checks/permissions.go
laurentsimon: done.

checks/permissions.go
laurentsimon: thanks

checks/permissions.go
laurentsimon: thanks.

checks/file_utils.go
laurentsimon: i see. continue is a language keyword so I could not use it. let me find a better name then. maybe `cont`?

oliverchang: Perhaps `continueIter` or `continueMatch` ?

laurentsimon: maybe `iterate` or `continueIter`

laurentsimon: i'll go with `continueIter` since we had the same idea!

PR #778, opened 2021-07-29
✨ Update release test cron job

azeemshaikh38:

Update cron job controller to `rollout restart` the worker jobs. This way, it is guaranteed that both controller and worker are at the latest build when starting the release tests. If this works well, will also update the weekly cron jobs to use similar logic.

This PR also updates Dockerfiles in the repo to use consistent file naming.

No.


inferno-chromium (approved):

cron/releasetest/k8s/controller.yaml
```diff
@@ -40,4 +65,10 @@ spec:
                 memory: 1Gi
               requests:
                 memory: 1Gi
-          restartPolicy: OnFailure
+          - name: worker-update
+            image: bitnami/kubectl
```
naveensrinivasan, line 69: I would pin the SHA for this.

azeemshaikh38: Could you tell me how I would do that? Looks like this is their release version commit - https://github.com/bitnami/bitnami-docker-kubectl/commit/c1bf66c18a2dde1c3ad7f070f3fad15611862e6d.

So the commit SHA `c1bf66c18a2dde1c3ad7f070f3fad15611862e6d` would be what we pin it to?

azeemshaikh38: Nvm, used `docker inspect` to get it.

PR #782, opened 2021-07-29
✨  Cleanup errors and log

laurentsimon:

cleanup

we use old error handling
structure has logf

we use new error handling
structure no longer has logf

no

* **Other information**:


roundtripper/census.go
```diff
@@ -43,18 +44,21 @@ type censusTransport struct {
 func (ct *censusTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 	ctx, err := tag.New(r.Context(), tag.Upsert(stats.RequestTag, "requested"))
 	if err != nil {
-		return nil, fmt.Errorf("error during tag.New: %w", err)
+		//nolint:wrapcheck
+		return nil, sce.Create(sce.ErrScorecardInternal, fmt.Sprintf("tag.New: %v", err))
```
azeemshaikh38, line 48: Ok for this PR, but FYI - these errors will get bubbled up and will eventually get wrapped in a `sce.Create(ErrScorecardInternal)` anyways. Maybe we don't need to wrap them with `sce.Create` here. Something to consider in the future.

azeemshaikh38 (approved):

roundtripper/census.go
laurentsimon: where do they get wrapped into `sce.Create(ErrScorecardInternal)`?

azeemshaikh38: In the `checks` package which calls `roundtripper`

repos/repo_url.go
```diff
@@ -74,11 +78,14 @@ func (r *RepoURL) ValidGitHubURL() error {
 	case "github.com":
 		break
 	default:
-		return fmt.Errorf("%w: %s", ErrorUnsupportedHost, r.Host)
+		//nolint:wrapcheck
+		return sce.Create(ErrorUnsupportedHost, r.Host)
```
laurentsimon, line 82: a bunch of errors (e.g., `ErrorUnsupportedHost `) are declared at https://github.com/ossf/scorecard/blob/c02df073a9f39e6914d7f934eb9f5af131ff2361/repos/repo_url.go#L27:L32

They are also used in a cron job file `cron/data/iterator_test.go`.

We have the main `errors/` directory with error definitions used by checks.

I think we need to start a thread on how to harmonize everything?

checker/check_runner.go
```diff
@@ -61,19 +60,15 @@ func (l *logger) Debug(desc string, args ...interface{}) {
...
-	l.messages = append(l.messages, fmt.Sprintf(s, f...))
-}
-
 func logStats(ctx context.Context, startTime time.Time, result *CheckResult) error {
 	runTimeInSecs := time.Now().Unix() - startTime.Unix()
 	opencensusstats.Record(ctx, stats.CheckRuntimeInSec.M(runTimeInSecs))
 
 	if result.Error != nil {
-		ctx, err := tag.New(ctx, tag.Upsert(stats.ErrorName, scorecarderrors.GetErrorName(result.Error)))
+		ctx, err := tag.New(ctx, tag.Upsert(stats.ErrorName, result.Error2.Error()))
```
laurentsimon, line 68: @azeemsgoogle are you OK with this change? If errors contain temp file name or repo names, it will show as different tags.

roundtripper/census.go
laurentsimon: talked offline.

checker/check_runner.go
azeemshaikh38: Oh, I didn't realize `Error2.Error` will print out the whole error string. That will be an issue. Can we use the error mapping logic here: https://github.com/ossf/scorecard/blob/main/errors/names.go.

It'll dump mostly dump out everything as `UnknownError`, but we can improve on it over time. Its better than logging the Error string.

azeemshaikh38 (changes requested):
Let me know when you are done making changes to this PR. I can take another look at it.

laurentsimon:
> Let me know when you are done making changes to this PR. I can take another look at it.

ptal I'm done with the changes (besides linter error :-))

checker/check_result.go
```diff
@@ -31,11 +30,7 @@ const (
 // UPGRADEv2: to remove.
 const migrationThresholdPassValue = 8
 
-// ErrorDemoninatorZero indicates the denominator for a proportional result is 0.
-// UPGRADEv2: to remove.
-var ErrorDemoninatorZero = errors.New("internal error: denominator is 0")
-
-// DetailType defines types of details.
+// Types of details.
```
azeemshaikh38, line 33: Comments should be of form `DetailType ...`, else linter will complain.

clients/githubrepo/tarball.go
```diff
@@ -70,23 +72,18 @@ type tarballHandler struct {
 func (handler *tarballHandler) init(ctx context.Context, repo *github.Repository) error {
 	// Cleanup any previous state.
 	if err := handler.cleanup(); err != nil {
-		return fmt.Errorf("error during githubrepo cleanup: %w", err)
+		return err
```
azeemshaikh38, line 75: Why remove the `fmt.Errorf` here? Will be useful to have some context for the errors.

repos/repo_url.go
azeemshaikh38: Ack.

azeemshaikh38 (approved):
LGTM. Thanks!

clients/githubrepo/tarball.go
laurentsimon: ah, I see what you mean. My brain inaccurately flagged this because of `Errorf` + `%w`. I agree this usage is fine. Do you think I should add a dedicated function in `errors.go` for this case? I don't know. if that's even useful. Le me know what you think.

azeemshaikh38: We can skip the dedicated errors for this case. These anyways get exposed here: https://github.com/ossf/scorecard/blob/main/pkg/scorecard.go#L87

laurentsimon: gocha!

PR #791, opened 2021-07-30
✨ Catch RuntimeErrors in release testing

azeemshaikh38:

Catch errors of type `ErrScorecardInternal` during release testing. We will continue to ignore these in prod for now, so that the weekly jobs don't break.

No.

### DIFF
--- a/cron/worker/main.go
+++ b/cron/worker/main.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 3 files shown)


cron/worker/main.go
```diff
@@ -39,13 +40,17 @@ import (
...
+	sce "github.com/ossf/scorecard/v2/errors"
 	"github.com/ossf/scorecard/v2/pkg"
 	"github.com/ossf/scorecard/v2/repos"
 	"github.com/ossf/scorecard/v2/roundtripper"
 	"github.com/ossf/scorecard/v2/stats"
 )
 
-var errIgnore *clients.ErrRepoUnavailable
+var (
+	ignoreRuntimeErrors = flag.Bool("ignoreRuntimeErrors", true, "if set to true any runtime errors will be ignored")
```
inferno-chromium, line 51: Shouldn't the default be false, and we explicitly set it to true in cron job (maybe with a FIXME) ?

inferno-chromium (approved):

cron/worker/main.go
azeemshaikh38: Done.

oliverchang (commented):
nice! LGTM

PR #800, opened 2021-08-02
✨ Improve token permission check

laurentsimon:

feature

we only support top-level permissions

we support run-level permissions

no

* **Other information**:
1. We don't flag the `packages` permission if the workflow is a packaging workflow. This is good enough for now. Ideally we should only check for packaging workflows that publish to GitHub registries (since the `packages` permission is only needed in this case).
2. We don't check for `security-events` permission if the workflow is codeql. We may add other SAST tool in the future
3. I added support to the Packaging check for some missing registries/language like go, ruby's gem and nuget.


checks/packaging.go
```diff
@@ -77,49 +77,75 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
...
+
+	// Python packages.
+	if strings.Contains(s, "actions/setup-python@") && strings.Contains(s, "pypa/gh-action-pypi-publish@master") {
+		dl.Info("candidate python packaging workflow using pypi: %s", fp)
 		return true
 	}
 
-	if strings.Contains(s, "docker push") {
-		c.Dlogger.Info("candidate docker publishing workflow: %s", fp)
+	// Go packages.
```
oliverchang, line 142: Let's add support for Rust/Cargo as well?

checks/permissions.go
```diff
@@ -72,41 +78,46 @@ func validatePermission(key string, value interface{}, path string,
...
+	ignoredPermissions map[string]bool,
+	acceptUndefined bool) error {
 	var permissions interface{}
 
 	// Check if permissions are set explicitly.
 	permissions, ok := config["permissions"]
 	if !ok {
-		dl.Warn("no permission defined in %v", path)
-		recordAllPermissionsWrite(pdata)
+		if !acceptUndefined {
```
oliverchang, line 117: would be helpful to add a short comment here to explain why there's this condition. e.g. "Explicit permissions are optional for run level" or something similar. 

Also might be clearer to move the `permissions, ok := config["permissions"]` bits and this check to where they're called from (i.e. validateRunLevelPermissions / validateTopLevelPermissions) and pass the resulting `permissions` to this function instead. This makes the special case (Warn + recordAllPermissionsWrite) closer to where it's needed and easier to follow.

i.e. 

```go
func validateTopLevelPermissions(config map[interface{}]interface{}, path string,
	dl checker.DetailLogger, pdata *permissionCbData) error {
        permissions, ok := config["permissions"]
        // comment explaining why.
        if !ok {
             ... // warn + recordAllPermissionsWrite
            return
        }
	return validatePermissions(permissions, path, dl,
		pdata.topLevelWritePermissions, map[string]bool{},
		false)
}
```

oliverchang (approved):

checks/packaging.go
laurentsimon: done.

checks/permissions.go
laurentsimon: good idea. I started adding more and more arguments without realizing it could be simplified. Thanks!

PR #806, opened 2021-08-03
✨  fix/enhance pinned-dependencies

laurentsimon:

bug+feature

lack of details when there's no files to check: e.g. no dockerfile -> we still need to display a message
wrong results because result is shared across files

fix the above

no

* **Other information**:

### DIFF
--- a/checks/pinned_dependencies.go
+++ b/checks/pinned_dependencies.go
@@ -121,25 +121,60 @@ func maxScore(s1, s2 int) int {
 	return s2
 }
 
-func createReturnValues(r bool, infoMsg string, dl checker.DetailLogger, err error) (int, error) {
+type pinnedResult int
+
+const (
+	pinnedUndefined pinnedResult = iota
+	pinned
+	notPinned
+)
+
+func addPinnedResult(r *pinnedResult, to bool) {
+	// If the result is `noPinned`, we keep it.
+	// In other casees, we always update the result.
+	if *r == pinnedUndefined || *r == pinned {
+		switch to {
+		case true:
+			*r = pinned
+		case false:
+			*r = notPinned
+		}
+	}
+}
+
+func dataAsResultPointer(data FileCbData) *pinnedResult {
+	pdata, ok := data.(*pinnedResult)
+	if !ok {
+		// This never happens.
+		panic("invalid type")
+	}
+	return pdata
+}
+
+func createReturnValues(r pinnedResult, infoMsg string, dl checker.DetailLogger, err error) (int, error) {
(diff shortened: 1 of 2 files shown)


checks/pinned_dependencies.go
```diff
@@ -121,87 +121,123 @@ func maxScore(s1, s2 int) int {
...
+type pinnedResult int
+
+const (
+	pinnedUndefined pinnedResult = iota
+	pinned
+	notPinned
+)
+
+func addPinnedResult(r *pinnedResult, to bool) {
+	// If the result is `noPinned`, we keep it.
```
azeemshaikh38, line 133: Fix typo `noPinned`

checks/pinned_dependencies.go
```diff
@@ -121,87 +121,123 @@ func maxScore(s1, s2 int) int {
...
+const (
+	pinnedUndefined pinnedResult = iota
+	pinned
+	notPinned
+)
+
+func addPinnedResult(r *pinnedResult, to bool) {
+	// If the result is `notPinned`, we keep it.
+	// In other cases, we always update the result.
+	if *r == pinnedUndefined || *r == pinned {
```
azeemshaikh38, line 135: Nit: update to:

```
if r == nil || *r == notPinned {
  return
}
switch to {
...
}
```

azeemshaikh38 (approved):

checks/pinned_dependencies.go
```diff
@@ -121,87 +121,123 @@ func maxScore(s1, s2 int) int {
...
+}
+
+func createReturnValues(r pinnedResult, infoMsg string, dl checker.DetailLogger, err error) (int, error) {
 	if err != nil {
 		return checker.InconclusiveResultScore, err
 	}
-	if !r {
+
+	switch r {
+	default:
```
azeemshaikh38, line 160: Nit: make `default` as the last case in switch statement

laurentsimon: is this a golang rule? I found that having it first is less error-prone because it prevents forgetting it.

checks/pinned_dependencies.go
laurentsimon: done. I don't catch the `r = nil` case because I'd rather have it panic and tell me something is wrong in the code than gracefully returning.

checks/pinned_dependencies.go
azeemshaikh38: I don't think so. I'm just used to seeing `default` at the end because of C++ :) Feel free to leave it as is if you think this improves readability.

PR #810, opened 2021-08-03
✨ Use a single image for worker and controller

azeemshaikh38:

Only maintain a single image of worker/controller and use k8s args for running different jobs.

No.


cron/controller/Dockerfile
```diff
@@ -27,5 +27,6 @@ RUN CGO_ENABLED=0 make build-pubsub
 
 FROM gcr.io/distroless/base:nonroot@sha256:bc84925113289d139a9ef2f309f0dd7ac46ea7b786f172ba9084ffdb4cbd9490
 COPY ./cron/data/projects.csv cron/data/projects.csv
+COPY ./cron/data/projects.release.csv cron/data/projects.release.csv
```
oliverchang, line 30: nit: just `COPY ./cron/data/projects*csv cron/data/` to make it simpler and to avoid the extra layer.

oliverchang (approved):

cron/k8s/worker.release.yaml
```diff
@@ -29,6 +29,7 @@ spec:
       containers:
       - name: worker
         image: gcr.io/openssf/scorecard-batch-worker-releasetest:latest
```
oliverchang, line 31: can we share the worker image for release test as well?

cron/controller/Dockerfile
azeemshaikh38: Done.

cron/k8s/worker.release.yaml
azeemshaikh38: Ah forgot. Done.

PR #812, opened 2021-08-05
🐛 Ignore errors extracting corrupted tarball

azeemshaikh38:

Ignore errors while extracting a corrupted tarball.

No.

### DIFF
--- a/clients/githubrepo/tarball.go
+++ b/clients/githubrepo/tarball.go
@@ -39,8 +39,9 @@ const (
 )
 
 var (
-	errTarballNotFound = errors.New("tarball not found")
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
LGTM, but how does this happen? Any how often does it happen? Is this a similar issue to the previous issue?

azeemshaikh38:
> LGTM, but how does this happen? Any how often does it happen? Is this a similar issue to the previous issue?

Same issue as the previous one (and for the same repos). The HTTP download doesn't throw any error even though it is unsuccessful. And then when we try to extract it, it fails for various reasons.

PR #814, opened 2021-08-06
✨ Generate `.shard_metadata` file in cron job shard

azeemshaikh38:

* **Other information**:


cron/controller/Dockerfile
```diff
@@ -26,6 +26,8 @@ ARG TARGETARCH
 RUN CGO_ENABLED=0 make build-pubsub
 
 FROM gcr.io/distroless/base:nonroot@sha256:bc84925113289d139a9ef2f309f0dd7ac46ea7b786f172ba9084ffdb4cbd9490
+ARG COMMIT_SHA
+ENV SCORECARD_COMMIT_SHA=${COMMIT_SHA}
```
inferno-chromium, line 30: Why need to redeclare, can just use COMMIT_SHA ?

cron/controller/main.go
```diff
@@ -99,18 +102,37 @@ func main() {
...
+		NumShard:  new(int32),
+		ShardLoc:  new(string),
+		CommitSha: new(string),
+	}
+	*metadata.NumShard = (shardNum + 1)
+	*metadata.ShardLoc = bucket + "/" + data.GetBlobFilename("", t)
+	*metadata.CommitSha = os.Getenv(commitSHA)
+	metadataJSON, err := protojson.Marshal(&metadata)
+	if err != nil {
+		panic(err)
```
inferno-chromium, line 132: maybe add error message in these two panics ?

inferno-chromium (approved):

cron/controller/main.go
azeemshaikh38: Done.

cron/controller/Dockerfile
azeemshaikh38: I tested it without ARG, and it doesn't seem to work. I guess we need to declare it for it to accept `--build-arg` option.

Issue #725, opened 2021-07-21
Feature: add line number to Pinned-Dependencies detail reporting


laurentsimon:
add a line number to the `Pinned-Dependencies`, `Token-Permissions` checks. 
Today the details contain the line content only, e.g.:
``Warn: insecure (unpinned) download detected in .github/workflows/pylint-presubmit.yml: 'python -m pip install --upgrade pip'``

We may want to add a line number as well.

chrismcgehee:
I'll take this issue since I have some time now to work on scorecard now that I've settled in from my move. I'd like to have an easier issue to get back into the swing of things.

laurentsimon:
Thank you!

FYI I'm working on adding [SARIF support](https://docs.github.com/en/code-security/secure-coding/integrating-with-code-scanning/sarif-support-for-code-scanning) for scorecard, and for this I need to make some code update to support more structured data in details (line, filename, etc). 
ETA for the required update is ~10 days. You can start working on the PR for sure in the meantime, you'll just need to change the call to `Warn` when I land my update.

laurentsimon:
@chrismcgehee assigned issue to you, thanks again for taking this issue!

laurentsimon:
@chrismcgehee this issue is important for the v4 (~EOY) release. Are you still interested in helping for this issue?
I've finished the SARIF support and you can try it using https://github.com/ossf/scorecard/issues/1074#issuecomment-933653061

laurentsimon:
@chrismcgehee what's left to be done for this? Sorry, I've not kept up with all the commits :-)

chrismcgehee:
> @chrismcgehee what's left to be done for this?

I'm just about ready to submit a PR that will add numbers for more places. The line numbers for anything in shell code is going to be tricky because of the way we pass shell code throughout the program. I'd be open to having help for that part. 

> Do you know if other checks need updates as well?

I believe Pinned Dependencies and Token Permissions are all that need updating. Asra already added proper line numbering for Dangerous Workflow.

laurentsimon:
> > @chrismcgehee what's left to be done for this?
> 
> I'm just about ready to submit a PR that will add numbers for more places. The line numbers for anything in shell code is going to be tricky because of the way we pass shell code throughout the program. I'd be open to having help for that part.

sg. Let us know how we should help. I think you can start by adding the line numbers to entry functions and try to propagate them to callees? Then ask for help when things get tricky? Or would you like to proceed differently?

[...]

PR #821, opened 2021-08-08
✨ Add line number to unpinned dependency: GitHub workflow "uses" field

chrismcgehee:

Feature

Line numbers are not displayed in the results when there are unpinned dependencies.
#725 

When there is an unpinned `uses` field in a `.github/workflows` file , the line number for the `uses` field will be displayed in the detailed results.
I will create additional PRs to add line numbers for other places in the code that deal with unpinned dependencies.

No

* **Other information**:
I updated `gopkg.in/yaml` to v3 in `checks/pinned_dependencies.go` in order to use the newer `UnmarshalYAML()` in order to set the line number of the field. I did not update other files to go from `gopkg.in/yaml/v2` to ``gopkg.in/yaml/v3` because when I tried doing so, some of the test for Token-Permissions failed.


naveensrinivasan (commented):
Thanks,Could we add some tests to this?

chrismcgehee:
> Thanks,Could we add some tests to this?

Sure thing, I'll get some added.

chrismcgehee:
Tests have been added.

checks/pinned_dependencies.go
```diff
@@ -52,6 +52,23 @@ type gitHubActionWorkflowConfig struct {
 	Name string `yaml:"name"`
 }
 
+// stringWithLine is for when you want to know the line number of a field in a YAML file.
```
laurentsimon, line 55: since we want this line also for Dockerfiles, etc let's remove the end of the comment.

checks/pinned_dependencies_test.go
```diff
@@ -542,3 +543,78 @@ func TestGitHubWorflowRunDownload(t *testing.T) {
 		})
 	}
 }
+
+func TestGitHubWorkflowUsesLineNumber(t *testing.T) {
```
laurentsimon, line 547: is there a way we could re-use the existing `TestGithubWorkflowPinning` instead of creating a new `TestGitHubWorkflowUsesLineNumber `?

chrismcgehee: I personally think it's better having it in its own test because I prefer unit tests to focus on just one thing as opposed to several disparate things. Combining it with `TestGithubWorkflowPinning` would make it a bit too convoluted.

checks/pinned_dependencies.go
chrismcgehee: Comment updated

checks/pinned_dependencies_test.go
naveensrinivasan: I agree with @chrismcgehee

laurentsimon:
FYI, we can start migrating to using new structure. Example in https://github.com/ossf/scorecard/pull/883

naveensrinivasan:
@laurentsimon Is this good to merge?

laurentsimon:
I wanted to wait for @chrismcgehee to migrate the changes he made to the new structured results (https://github.com/ossf/scorecard/pull/883), if possible. 
@chrismcgehee what option do you prefer? Merge as-is or do you want to take a second pass?

chrismcgehee:
> @chrismcgehee what option do you prefer? Merge as-is or wait for you to update?

I'd like to update this PR to take advantage of the logging structure.

laurentsimon (approved):

PR #823, opened 2021-08-09
:sparkles: feature: branch protection without admin token

asraa:

- [x ] Tests for the changes have been added (for bug fixes / features)
- [x ] PR title follows the guidelines defined in  https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process

Enables a score for branch protection without an admin token for the repository. Only an "is protected" boolean is available via the Branches API. In this case, a score of 1 with a warning is added to the final score (with a detailed reason). Enabling branch protection without using any settings gives a score of 1.

Not found for non-admin token usage

Score of 1 with warning for protected branches

* **Other information**:


checks/branch_protection.go
```diff
@@ -113,7 +114,12 @@ func checkReleaseAndDevBranchProtection(ctx context.Context, r repositories, dl
 			// The branch is protected. Check the protection.
 			score, err := getProtectionAndCheck(ctx, r, dl, ownerStr, repoStr, b)
 			if err != nil {
-				return checker.CreateRuntimeErrorResult(CheckBranchProtection, err)
```
laurentsimon, line 116: how do we differentiate `true` errors and errors that lead to an `unknown` result?

asraa: Because we know this branch is protected, so calling `getProtectionAndCheck` will only return an error if we can't get the detailed protection information.

Redirects however: they did close my issue on getting the redirected branch when you get protection, but I will test this out in case I'm missing a way to use it.

laurentsimon: thanks. The assumption is that https://github.com/ossf/scorecard/blob/main/checks/branch_protection.go#L174 does not trigger a new request that could timeout or return a GitHub's internal error. Is that correct?

asraa: Added an explicit check to make sure it's a 404 Not Found and not a timeout, i think you're right.

We should have already resolved the branch name by matching it against known branch names, so we can't run into a problem that the branch isn't found because it doesn't exist.

laurentsimon: sg, thanks for double checking!

laurentsimon (approved):

PR #827, opened 2021-08-10
✨  Feature: Include Check Score When Outputting JSON

RexBelli:

**Feature**
When outputting the results in JSON, it can be helpful to know the score as determined by the specific check.

Currently, the score field is not included when the output format is JSON.

With these changes, the score field will be included in the check object.

I can't think of any scenarios this would break anything.  The only case is if someone depends on a missing score field in the check object, but I'm not sure what use case that would have.

* **Other information**:

### DIFF
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -78,7 +78,7 @@ type CheckResult struct {
 	Version  int           `json:"-"` // Default value of 0 indicates old structure.
 	Error2   error         `json:"-"` // Runtime error indicate a filure to run the check.
 	Details2 []CheckDetail `json:"-"` // Details of tests and sub-checks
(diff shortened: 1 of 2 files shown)


azeemshaikh38 (changes requested):
We are in process of migrating to the new result format. We have some jobs and clients which depend on the existing format and this PR would break that.

laurentsimon:
@RexBelli the new json format is implemented in `AsJSON2`, see https://github.com/ossf/scorecard/blob/main/pkg/scorecard_result.go#L91.

If you run the `scorecard ... --format json | jq`, you will get the score as part of the output.

As @azeemsgoogle mentioned, it's not used in the cron job so the BigQuery table does not expose it yet. We're working on it.

Closing this issue. Feel free to re-open if needed.

PR #830, opened 2021-08-10
✨  Upload cron results to a table with new format

laurentsimon:

feature

we store cron results in one table using old format only

we store cron results in both the old and new format in 2 tables.

no

* **Other information**:


checker/check_runner.go
```diff
@@ -101,7 +101,9 @@ func (r *Runner) Run(ctx context.Context, f CheckFn) CheckResult {
 		break
 	}
 	res.Details2 = l.messages2
-
+	for _, d := range l.messages2 {
```
laurentsimon, line 104: this will disappear from the PR once https://github.com/ossf/scorecard/pull/841 is merged.

cron/config/config.yaml
```diff
@@ -21,3 +21,6 @@ bigquery-table: scorecard
 shard-size: 10
 webhook-url: 
 metric-exporter: stackdriver
+# UPGRADEv2: to remove.
+result-data-bucket-url2: gs://ossf-scorecard-data2
```
azeemshaikh38, line 25: Nit: use `-v2` suffix to make it look cleaner.

cron/config/config.go
```diff
@@ -147,6 +147,17 @@ func GetBigQueryTable() (string, error) {
 	return getStringConfigValue(bigqueryTable, configYAML, "BigQueryTable", "bigquery-table")
 }
 
+// UPGRADEv2: to remove.
+// GetBigQueryTable2 returns the table name to transfer cron job results.
+func GetBigQueryTable2() (string, error) {
```
azeemshaikh38, line 152: You'll need to update the `config` struct to add `BigQueryTableV2` and `ResultDataBucketURLV2` fields. Also, need to defined `env` vars that correspond to these.

cron/config/config.yaml
```diff
@@ -21,3 +21,6 @@ bigquery-table: scorecard
...
+bigquery-table2: scorecard2
```
azeemshaikh38, line 26: Unit tests for `config_test.go` might fail with this change if you don't update `config_test.go`

cron/worker/main.go
```diff
@@ -71,6 +72,16 @@ func processRequest(ctx context.Context,
 		return nil
 	}
 
+	exists, err = data.BlobExists(ctx, bucketURL2, filename)
+	if err != nil {
+		return fmt.Errorf("error during BlobExists: %w", err)
+	}
+	if exists {
```
azeemshaikh38, line 79: If this `exists` is true, that means the data-v1 was not written but data-v2 was. In that case, we shouldn't early return. Instead this should be:

```
if existsV1 && existsV2 {
 return ...
}
```

pkg/scorecard_result.go
```diff
@@ -38,8 +38,23 @@ type ScorecardResult struct {
...
 }
 
+type jsonCheckResult struct {
+	Details []string
+	Score   int
+	Reason  string
+	Name    string
+}
+
+type jsonScorecardResult struct {
```
azeemshaikh38, line 48: Nit: to clarify what these structs are going to be used for, suffix with `V2`

cron/bq/main.go
```diff
@@ -148,6 +148,22 @@ func getBQConfig() (projectID, datasetName, tableName string, err error) {
 	return
 }
 
+func getBQConfig2() (projectID, datasetName, tableName string, err error) {
```
azeemshaikh38, line 151: Simply update `getBQConfig` to return `tableName` and `tableNameV2`?

azeemshaikh38: Nvm, I just realized something. You don't need to make any code changes in this file at all. Instead, you can simply run a different instance of BQ transfer job which transfer the V2 JSON data to the V2 table.

azeemshaikh38 (commented):
LGTM overall. Left some comments.

cron/config/config.go
laurentsimon: done. ptal

cron/config/config.yaml
laurentsimon: updated. tests pass. I'll keep an eye on it.

cron/bq/main.go
laurentsimon: so basically I would create a new k8's `transferV2.yaml` and set some env variables to define the name of tables and buckets `SCORECARD_BIGQUERY_TABLE: scorecard2` and `SCORECARD_DATA_BUCKET_URL : gs://ossf-scorecard-data2`.
Is my understanding correct?

azeemshaikh38: Yep, thats correct.

laurentsimon: thanks. And I suppose I'll also need to create a `transfer.releaseV2.yaml`

cron/config/config.yaml
```diff
@@ -21,3 +21,6 @@ bigquery-table: scorecard
 shard-size: 10
 webhook-url: 
 metric-exporter: stackdriver
+# UPGRADEv2: to remove.
```
azeemshaikh38, line 24: One more thing you'll need is to add `BUCKET_V2` env var [here](https://github.com/ossf/scorecard/blob/main/cron/k8s/controller.release.yaml#L60) and [here](https://github.com/ossf/scorecard/blob/main/cron/k8s/worker.release.yaml#L35). That way, the release tests will write to a different bucket and not to prod bucket.

laurentsimon: got it, thanks. I completely overlooked the release tests :-(

laurentsimon: ptal, updated

cron/k8s/transfer.release-v2.yaml
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: scorecard-bq-transfer-releasetest
```
azeemshaikh38, line 18: Update `name` to not confuse with the other transfer job

cron/k8s/transfer.release-v2.yaml
```diff
@@ -0,0 +1,41 @@
...
+  # Daily at 9am and 6pm PST.
+  schedule: "0 1,16 * * *"
+  concurrencyPolicy: "Forbid"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: OnFailure
+          containers:
+          - name: bq-transfer-releasetest
```
azeemshaikh38, line 29: Update name.

cron/k8s/transfer-v2.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2021 Security Scorecard Authors
```
azeemshaikh38, line 1: Delete file?

azeemshaikh38 (approved):

cron/k8s/transfer.release-v2.yaml
laurentsimon: done

cron/k8s/transfer.release-v2.yaml
laurentsimon: done

cron/k8s/transfer-v2.yaml
laurentsimon: sorry I'm lost. Why do I need to delete this file? The `bq/main.go` is unchanged so I think we need this new instance. Please correct me if I'm wrong.

azeemshaikh38: Oh sorry. I confused this with the v2.release.yaml file. No need to delete this. Just update the job name to have `-v2` suffix.

PR #842, opened 2021-08-11
✨ Update SAST and CITest with Repoclient API

azeemshaikh38:

Use Repoclient API for SAST and CITest checks.

No.

### DIFF
--- a/checks/ci_tests.go
+++ b/checks/ci_tests.go
@@ -21,6 +21,7 @@ import (
 	"github.com/google/go-github/v32/github"
 
 	"github.com/ossf/scorecard/v2/checker"
+	"github.com/ossf/scorecard/v2/clients"
 	sce "github.com/ossf/scorecard/v2/errors"
 )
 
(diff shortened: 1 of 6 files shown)


checks/ci_tests.go
```diff
@@ -43,19 +44,18 @@ func init() {
...
 		return checker.CreateRuntimeErrorResult(CheckCITests, e)
 	}
 
 	usedSystem := unknown
 	totalMerged := 0
 	totalTested := 0
-	for _, pr := range prs {
-		if pr.MergedAt == nil {
+	for index := range prs {
+		pr := &prs[index]
```
oliverchang, line 57: Any reason in particular for doing this (as opposed to how we were doing it before)? To avoid copying?

oliverchang (approved):

checks/ci_tests.go
azeemshaikh38: Yeah, go-linter wasn't happy with `pr` struct being copied every loop.

PR #843, opened 2021-08-11
✨ SARIF 1: add structured detail

laurentsimon:

feature

no sarif support

we add the underlying structure to details

no

* **Other information**:

### DIFF
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -48,11 +48,11 @@ type FileType int
 const (
 	// FileTypeNone is a default, not defined.
 	FileTypeNone FileType = iota
-	// FileTypeSource is for source code.
+	// FileTypeSource is for source code files.
 	FileTypeSource
-	// FileTypeBinary is fir binary files.
+	// FileTypeBinary is for binary files.
 	FileTypeBinary
-	// FileTypeText is text.
+	// FileTypeText is for text files.
 	FileTypeText
 	// FileTypeURL for URLs.
 	FileTypeURL
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -42,25 +42,11 @@ const (
 	DetailDebug
 )
 
-// CheckDetail contains information for each detail.
-//nolint:govet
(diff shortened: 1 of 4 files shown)


checker/check_result.go
```diff
@@ -56,6 +56,48 @@ type DetailLogger interface {
 	Debug(desc string, args ...interface{})
 }
 
+// FileType is the type of a file.
+type FileType int
```
naveensrinivasan, line 60: Is there a specific reason to create a custom type for int?

checker/check_result.go
```diff
@@ -56,6 +56,48 @@ type DetailLogger interface {
...
+	Text    string   // A short string explaining why the detail was recorded/logged.
+	Path    string   // Fullpath to the file.
+	Type    FileType // Type of file.
+	Offset  int      // Offset in the file of Path (line for source/text files).
+	Snippet string   // Snippet of code.
+}
+
+// CheckDetail3 contains information for each detail.
+// UPGRADEv3: rename to DetailLogger.
+type CheckDetail3 struct {
```
oliverchang, line 88: Rather than introduce yet another similar but different structure, can we just re-use a single structure? 

either extending the original `CheckDetail` without breaking compatibility

```go
type CheckDetail struct {
  Msg string
  Type DetailType
  Details struct {
    Path string
    Type FileType
    Offset int
    Snippet string
  }
}
```

Or just migrate it to what you have here, and have the non-detailed logger store `Msg.Text`  and keep everything else empty.

If you later need export this in a different JSON format, we can convert it to the structure structure then.

checker/check_result.go
laurentsimon: to have the compiler verify it's an enum

checker/check_result.go
laurentsimon: sg. done. ptal

oliverchang (approved):

azeemshaikh38 (approved):

PR #844, opened 2021-08-12
✨ Update SignedReleases to use RepoClient API

azeemshaikh38:

Use RepoClient API in `Signed-Releases` check.

No.

### DIFF
--- a/clients/githubrepo/graphql.go
+++ b/clients/githubrepo/graphql.go
@@ -25,10 +25,12 @@ import (
 )
 
 const (
-	pullRequestsToAnalyze = 30
-	reviewsToAnalyze      = 30
-	labelsToAnalyze       = 30
-	commitsToAnalyze      = 30
+	pullRequestsToAnalyze  = 30
+	reviewsToAnalyze       = 30
+	labelsToAnalyze        = 30
+	commitsToAnalyze       = 30
+	releasesToAnalyze      = 30
+	releaseAssetsToAnalyze = 30
 )
 
 // nolint: govet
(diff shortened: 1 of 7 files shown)


clients/githubrepo/graphql.go
```diff
@@ -152,6 +173,24 @@ func pullRequestFrom(data *graphqlData) []clients.PullRequest {
 	return ret
 }
 
+func releasesFrom(data *graphqlData) []clients.Release {
+	releases := make([]clients.Release, 0)
```
oliverchang, line 177: nit: do we need the ", 0" (or the make() at all for that matter)? https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices also says to prefer `var releases []clients.Release`

clients/githubrepo/graphql.go
```diff
@@ -152,6 +173,24 @@ func pullRequestFrom(data *graphqlData) []clients.PullRequest {
...
+	for _, r := range data.Repository.Releases.Nodes {
+		release := clients.Release{
+			TagName: string(r.TagName),
+			Assets:  make([]clients.ReleaseAsset, 0),
```
oliverchang, line 181: nit: no need for this. you can append to nil arrays (unless you want to preallocate the capacity here with `len(r.ReleaseAssets.Nodes)`

oliverchang (approved):

clients/githubrepo/graphql.go
azeemshaikh38: Done.

clients/githubrepo/graphql.go
azeemshaikh38: Done.

PR #848, opened 2021-08-12
✨ SARIF 2: add short description to checks.yml

laurentsimon:

feature

no short description for checks

add short description for checks

no

* **Other information**:
the SARIF format requires a short description for each result. This is use by GitHub to show a message for the offending line.
We will populate the short description in SARIF with the text from the yaml file.


docs/checks/checks.yaml
```diff
@@ -48,8 +50,9 @@ checks:
         file in the locations that are recommended by these tools.
   Binary-Artifacts:
     risk: High
+    short: determine if the project has binary artifacts in the source repository.
```
inferno-chromium, line 53: Determines with a capital D to match others (also missing s)

docs/checks/checks.yaml
```diff
@@ -32,8 +33,9 @@ checks:
         your project activity and maintenance commitment.
   Automatic-Dependency-Update:
     risk: High
+    short: Determines if a project has dependencies automatically updated.
```
inferno-chromium, line 36: maybe
project has enabled automatic dependency updates (both in short and full description)

docs/checks/checks.yaml
```diff
@@ -229,6 +240,7 @@ checks:
         or [renovate bot](https://github.com/renovatebot/renovate).
   SAST:
     risk: Medium
+    short: Determines if the project uses static code analysis systems.
```
inferno-chromium, line 243: uses a static code analysis system

docs/checks/checks.yaml
```diff
@@ -315,6 +331,7 @@ checks:
         GitHub's [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).
   Vulnerabilities:
     risk: High
+    short: Determines if there are open, known unfixed vulnerabilities for the project.
```
inferno-chromium, line 334: Determines if the project has open, known unfixed vulnerabilities ?

inferno-chromium (approved):

docs/checks/checks.yaml
laurentsimon: we could also change this check to be `dependency update tool` or more generally `dependency management tool`. The updates are not automatic and this seems to confuse some users, because the check's name seems to go against dependency pinning.
Can so that later... but curious what you think

inferno-chromium: yes "project uses a dependency update tool" is better.

laurentsimon: shall I update the check's name as well -> `Dependency-Update-Tool`?

inferno-chromium:
you would need to update this pr slightly since signed tags is getting removed soon, in auto-merge.

inferno-chromium (approved):

azeemshaikh38 (approved):
Also update REAMDE.md - https://github.com/ossf/scorecard#scorecard-checks

PR #854, opened 2021-08-13
✨ Use RepoClient API for Contributors check

azeemshaikh38:

Use RepoClient API for Contributors check.

No.

### DIFF
--- a/checks/contributors.go
+++ b/checks/contributors.go
@@ -18,8 +18,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/google/go-github/v38/github"
-
 	"github.com/ossf/scorecard/v2/checker"
 	sce "github.com/ossf/scorecard/v2/errors"
 )
@@ -38,7 +36,7 @@ func init() {
 
 // Contributors run Contributors check.
 func Contributors(c *checker.CheckRequest) checker.CheckResult {
-	contribs, _, err := c.Client.Repositories.ListContributors(c.Ctx, c.Owner, c.Repo, &github.ListContributorsOptions{})
+	contribs, err := c.RepoClient.ListContributors()
 	if err != nil {
 		e := sce.Create(sce.ErrScorecardInternal, fmt.Sprintf("Client.Repositories.ListContributors: %v", err))
 		return checker.CreateRuntimeErrorResult(CheckContributors, e)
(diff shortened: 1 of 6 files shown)


checks/contributors.go
```diff
@@ -38,31 +36,25 @@ func init() {
...
+
+		for _, org := range contrib.Organizations {
+			if org.Login != "" {
+				companies[org.Login] = struct{}{}
+			}
 		}
 
-		company := u.GetCompany()
+		company := contrib.Company
 		if company != "" {
```
inferno-chromium, line 58: is this null check even needed now ?

clients/githubrepo/client.go
```diff
@@ -44,7 +45,7 @@ func (client *Client) InitRepo(owner, repoName string) error {
 	}
 	client.repo = repo
 
-	// Init tarballHandler.
+	// Init tarballHandler
```
inferno-chromium, line 48: nit: i thought we keep periods at end, some mismatches in different messages.

clients/githubrepo/contributors.go
```diff
@@ -0,0 +1,65 @@
...
+		if err == nil {
+			for _, org := range orgs {
+				contributor.Organizations = append(contributor.Organizations, clients.User{
+					Login: org.GetLogin(),
+				})
+			}
+		}
+		user, _, err := handler.ghClient.Users.Get(ctx, contrib.GetLogin())
+		if err != nil {
+			return fmt.Errorf("error during Users.Get: %w", err)
```
inferno-chromium, line 55: can we land here if contrib.GetLogin() is "". do we need to continue and log error or can this not happen at all ?

inferno-chromium (approved):
Thanks a lot!

clients/githubrepo/contributors.go
azeemshaikh38: Good point. It shouldn't be empty, but added code to skip the entry in case it is.

checks/contributors.go
azeemshaikh38: It is. GitHub can return empty strings and not error.

clients/githubrepo/client.go
azeemshaikh38: My bad. We do keep periods at the end.

Issue #839, opened 2021-08-11
BUG: Parsing errors


azeemshaikh38:
Noticing many parsing errors during cron job. Not blocking cron job, but good to fix since otherwise we just return `ErrScorecardInternal`. Some examples to reproduce shown below:

```
scorecard --repo=github.com/ufal/udpipe --checks=Pinned-Dependencies

scorecard --repo=github.com/ubisoft/Sharpmake --checks=Pinned-Dependencies

scorecard --repo=github.com/uber/okbuck --checks=Pinned-Dependencies

scorecard --repo=github.com/uber/NullAway --checks=Pinned-Dependencies

scorecard --repo=github.com/aliyun/aliyun-odps-python-sdk --checks=Token-Permissions,Pinned-Dependencies

scorecard --repo=github.com/alibaba/GraphScope --checks=Token-Permissions

scorecard --repo=github.com/u-boot/u-boot --checks=Pinned-Dependencies
``` 

There might be more cases. Will add as I find them. Would be good to fix these and add these repos to `cron/data/projects.release.csv` as and when we fix them.

laurentsimon:
@chrismcgehee are you interested in helping for this issue? It's in the Pinned-Dependency check that you're looking at to add lines.

chrismcgehee:
Sure, I'll take this one.

chrismcgehee:
There are several different categories these parsing errors fall into. I'll list them and state how I plan to address them:
- Treating powershell code as shell code
          Example: [ubisoft/Sharpmake](https://github.com/ubisoft/Sharpmake/blob/62f11eb03f5730dc2e02d250ad275b9b77cc2815/.github/workflows/actions.yml#L115)
          How to fix: Check for `if: runner.os == 'Windows'`. Skip for now and add a todo to parse powershell code
- Valid bash that is not parsed correctly
          Example: [uber/NullAway](https://github.com/uber/NullAway/blob/master/config/hooks/pre-commit#L7)
          How to fix: Open issue in mvdan.cc/sh/v3/syntax.
- Bash files in `.github/workflows`
          Example: [alibaba/GraphScope](https://github.com/alibaba/GraphScope/blob/main/.github/workflows/hadoop_scripts/prepare_hadoop.sh)
          How to fix: Parse as shell code instead of yaml
- Non-shell files with ".sh" extension
          Example: [u-boot/u-boot](https://github.com/u-boot/u-boot/blob/master/scripts/ld-version.sh)
[...]

PR #862, opened 2021-08-17
🐛 Only validate shell scripts supported by our parser

chrismcgehee:

Bug fix

The pinned dependency check is failing on some repos because they contain shell scripts that our not supported by the parser we are using. For example, https://github.com/u-boot/u-boot/blob/master/scripts/ld-version.sh is an awk file that we are trying to parse as a shell script. 
Part of #839 

The parser we are using, https://github.com/mvdan/sh, supports parsing the following scripts: bash, sh, and mksh. For this PR, I am only parsing a file if it is using of those 3 shells. If a shebang is present in the file, I base the decision to parse entirely on it, otherwise I use the file extension.

No

* **Other information**:
This PR makes it so that some shell scripts (dash, ksh, etc.) will no longer be checked for pinned dependencies. I think this is the most correct thing to do since our parser does not support these shells. If we don't want to lose the checking of these shells, what we could do is attempt to parse these shells, and if parsing fails, swallow the error and continue with the check. This should work ~99% of the time, and in the 1% of times where it doesn't work, the check will run without giving any errors. Let me know if you'd prefer this approach instead.


checks/pinned_dependencies_test.go
```diff
@@ -515,9 +515,9 @@ func TestGitHubWorflowRunDownload(t *testing.T) {
 			filename: "testdata/github-workflow-wget-across-steps",
 			expected: scut.TestReturn{
 				Errors:        nil,
-				Score:         checker.MinResultScore,
-				NumberOfWarn:  2,
-				NumberOfInfo:  0,
+				Score:         checker.MaxResultScore,
+				NumberOfWarn:  0,
+				NumberOfInfo:  1,
```
chrismcgehee, line 520: This is because we are no longer parsing `ksh`: https://github.com/ossf/scorecard/blob/8960533b7b38916ad1c0a235643e9468ffbdd152/checks/testdata/github-workflow-wget-across-steps#L43

e2e/pinned_dependencies_test.go
```diff
@@ -49,7 +49,7 @@ var _ = Describe("E2E TEST:"+checks.CheckPinnedDependencies, func() {
 			expected := scut.TestReturn{
 				Errors:        nil,
 				Score:         checker.MinResultScore,
-				NumberOfWarn:  154,
+				NumberOfWarn:  152,
```
chrismcgehee, line 52: This is because we are no longer parsing `ksh`: 
https://github.com/ossf-tests/scorecard-check-pinned-dependencies-e2e/blob/91af031ed4e8f5f70d0db2bfd8c126cc27332e1d/.github/workflows/github-workflow-wget-across-steps#L43

azeemshaikh38:
Is this PR ready for review?

chrismcgehee:
> Is this PR ready for review?

Yes. Am I supposed to request reviewers or tag it with something?

azeemshaikh38:
> > Is this PR ready for review?
> 
> Yes. Am I supposed to request reviewers or tag it with something?

You can request reviewers to let them know PR is ready for review. Have added myself and @laurentsimon for now.

checks/shell_download_validate_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"io/ioutil"
+	"testing"
+)
+
+func TestIsSupportedShellScriptFile(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		filename string
```
azeemshaikh38, line 27: Nit: add a `name string` field here which can be used to identify the test. Currently we use `filename` to identify the test (line 66).

checks/shell_download_validate_test.go
```diff
@@ -0,0 +1,84 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt // Re-initializing variable so it is not changed while executing the closure below
+		t.Run(tt.filename, func(t *testing.T) {
+			t.Parallel()
+			var content []byte
+			var err error
+			if tt.filename == "" {
```
azeemshaikh38, line 70: Why do we need to test for empty filename here?

checks/shell_download_validate_test.go
```diff
@@ -0,0 +1,84 @@
...
+				content = make([]byte, 0)
+			} else {
+				content, err = ioutil.ReadFile(tt.filename)
+				if err != nil {
+					panic(fmt.Errorf("cannot read file: %w", err))
```
azeemshaikh38, line 75: Instead of `panic`, use `t.Failf`

checks/shell_download_validate_test.go
```diff
@@ -0,0 +1,84 @@
...
+				}
+			}
+			result := isSupportedShellScriptFile(tt.filename, content)
+			if result != tt.expected {
+				t.Fail()
```
azeemshaikh38, line 80: Add a failure message, showing the `expected` and `got` values. Something like: https://github.com/ossf/scorecard/blob/main/utests/utlib.go#L129

checks/shell_download_validate.go
```diff
@@ -35,6 +35,10 @@ var (
 	shellNames = []string{
 		"sh", "bash", "dash", "ksh", "mksh",
 	}
+	// supportedShells is the list of shells that are supported by mvdan.cc/sh/v3/syntax.
```
laurentsimon, line 38: you can declare `supportedShells `, and build `shellNames` from it. Take `shellInterpreters` as an example.

checks/pinned_dependencies_test.go
laurentsimon: instead of ignoring the file, change the shell name in the workflow. We want to make sure this test passes with a supported shell.

e2e/pinned_dependencies_test.go
laurentsimon: I've updated the file in https://github.com/ossf-tests/scorecard-check-pinned-dependencies-e2e/blob/main/.github/workflows/github-workflow-wget-across-steps to use `bash` so you need not update this test.

laurentsimon (approved):

azeemshaikh38 (approved):

checks/shell_download_validate_test.go
chrismcgehee: Ok, yeah, we don't need that here. I copy-pasted and didn't look to closely at what I copied.

checks/shell_download_validate_test.go
chrismcgehee: Yeah, I was unsure whether I should do that here since the only possibilities are true and false, but I do think it would be helpful.

checks/shell_download_validate.go
chrismcgehee: Good idea.

azeemshaikh38 (approved):

PR #866, opened 2021-08-17
✨ SARIF 4: Add support to output SARIF format

laurentsimon:

feature

no SARIF support

SARIF support

no

* **Other information**:


pkg/testdata/check1.sarif
```diff
@@ -0,0 +1,74 @@
+{
```
laurentsimon, line 1: add exception for license

laurentsimon: @naveensrinivasan I'm going to need to a few more exceptions for https://github.com/ossf/scorecard/blob/main/.github/workflows/main.yml#L70
Can I just modify as `"**/script-empty.sh,pkg/testsdata/*"`?
All my files under `pkg/testdata/` need to be excluded.

pkg/common.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/ossf/scorecard/v2/checker"
+	"go.uber.org/zap/zapcore"
+)
+
+// TODO: Verify this works in GitHub's dashboard.
+func textToHTML(s string) string {
+	return strings.ReplaceAll(s, "\n", "<br>")
```
oliverchang, line 27: Why do we need this? Because a single \n doesn't translate to \n in rendered markdown?

How about replacing the single newline with double?

pkg/common.go
```diff
@@ -0,0 +1,79 @@
...
+		switch v.Msg.Version {
+		//nolint
+		case 3:
+			if v.Type == checker.DetailDebug && logLevel != zapcore.DebugLevel {
+				continue
+			}
+			if v.Msg.Path != "" {
+				sa = append(sa, fmt.Sprintf("%s: %s: %s:%d", typeToString(v.Type), v.Msg.Text, v.Msg.Path, v.Msg.Offset))
+			} else {
+				sa = append(sa, fmt.Sprintf("%s: %s: %s", typeToString(v.Type), v.Msg.Text, v.Msg.Path))
```
oliverchang, line 48: Do we still want v.Msg.Path here even if it's "" ?

cmd/root.go
```diff
@@ -158,6 +166,17 @@ or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show
...
+			if !v3 {
+				log.Fatalf("sarif not supported yet")
+			}
+			checkDocs, e := docs.Read()
+			if e != nil {
+				log.Fatalf("cannot read yaml file: %v", err)
+			}
+			// TODO: support config files and update checker.MaxResultScore.
+			// TODO: update version.
+			err = repoResult.AsSARIF("1.2.3", showDetails, *logLevel, os.Stdout, checkDocs, checker.MaxResultScore)
```
oliverchang, line 179: What does "1.2.3" refer to? sarif version? better to make constant to make it clear what it's referring to.

pkg/sarif.go
```diff
@@ -0,0 +1,446 @@
...
+
+func detailsToLocations(details []checker.CheckDetail,
+	showDetails bool, logLevel zapcore.Level, minScore, score int) []location {
+	locs := []location{}
+
+	//nolint
+	// Populate the locations.
+	// Note https://docs.github.com/en/code-security/secure-coding/integrating-with-code-scanning/sarif-support-for-code-scanning#result-object
+	// "Only the first value of this array is used. All other values are ignored."
+	if showDetails {
```
oliverchang, line 226: nit: invert and early return. then we can avoid the indentation of lines 227 -244

pkg/sarif.go
```diff
@@ -0,0 +1,446 @@
...
+		Schema:  "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
+		Version: "2.1.0",
+		Runs: []run{
+			run{
+				Tool: tool{
+					Driver: driver{
+						Name:           strings.Title(name),
+						InformationURI: url,
+						SemVersion:     version,
+						Rules:          make([]rule, 1),
```
oliverchang, line 326: Is it intentional that we create a slice with a single empty `rule` here?  This gets overridden later anyway right? So we can just skip this line completely.

oliverchang (approved):

pkg/common.go
laurentsimon: good point. So it's not clear to me how GitHub handles this yet. But I can change to double space, for sure. I'm hoping they refuse html content when displaying results :-)

pkg/common.go
laurentsimon: good catch. Fixed.

pkg/sarif.go
laurentsimon: good catch. I split this function in two in the last CL, but now you're right it can be simplified!

pkg/sarif.go
laurentsimon: you're right. Fixed.

cmd/root.go
laurentsimon: it's the scorecard version. It needs to be a semantic version, something I need to work on. I'll add a variable to make it clearer. Thanks!

PR #877, opened 2021-08-22
🐛 Github workflow steps run on Windows should default to pwsh as its shell

chrismcgehee:

Bug fix

Github Action Workflow steps that are run only on Windows are currently treated as if they are using `bash` as their shell. Windows runners default to using `pwsh` as their shell. Since our shell parser cannot parse `pwsh`, we end up with syntax parsing errors.
Part of #839 

Now we will determine for each step if it is running on Windows or not. If it is, we will skip analyzing the commands within `run` because `pwsh` is not one of our supported shells.

No


chrismcgehee:
@laurentsimon On the linked issue, you suggested only treating a step as "Windows" if it a) contains `runner.os == 'Windows'`, and b) there is at least 1 Windows OS listed in `runs-on`. I ended up not doing b) because it would make the code more complicated and it could lead to incorrect results. Suppose a workflow file looked like:
```yaml
jobs:
  Job1:
    runs-on: [ubuntu-latest, macos-latest]
    steps:
      - if: runner.os == 'Windows'
        run: & $env:COMMAND C:\path\to\file
```
The step listed in this file would be dead code (perhaps `windows-latest` used to be in `runs-on`, but it was removed). If we tried to parse this step as bash, we would get a syntax error and fail this check.

checks/testdata/github-workflow-shells-runner-windows-ubuntu.yaml
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+
+jobs:
+  Job1:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-2019]
+        configuration: [debug, release]
+    steps:
```
chrismcgehee, line 22: I came up with these test cases by searching GitHub for various formulations used in the if clause in workflow files.

laurentsimon: sgtm

checks/pinned_dependencies.go
```diff
@@ -33,23 +33,60 @@ const CheckPinnedDependencies = "Pinned-Dependencies"
...
+}
+
+// stringOrSlice is for fields that can be a single string or a slice of strings. If the field is a single string,
+// this value will be a slice with a single string item.
+type stringOrSlice []string
+
+func (s *stringOrSlice) UnmarshalYAML(value *yaml.Node) error {
+	var stringSlice []string
+	err := value.Decode(&stringSlice)
+	if err != nil {
```
laurentsimon, line 78: is it worth refining the error checking by looking at a specific error that indicate wrong type?

checks/pinned_dependencies.go
```diff
@@ -477,6 +509,87 @@ func validateGitHubWorkflowIsFreeOfInsecureDownloads(pathfn string, content []by
 	return true, nil
 }
 
+// The only OS that this job runs on is Windows.
+func jobAlwaysRunsOnWindows(job *gitHubActionWorkflowJob) bool {
+	var jobOses []string
+	// The 'runs-on' field either lists the OS'es directly, or it can have an expression '${{ matrix.os }}' which
+	// is where the OS'es are actually listed
```
laurentsimon, line 516: nit: end with full dot.

laurentsimon (approved):
LGTM. Make sure to run the other unit tests and e2e tests (they are currently disabled in pre-submits).

Can you send me a PR to https://github.com/ossf-tests/scorecard-check-pinned-dependencies-e2e with additional tests you've added to this PR? This repo is used by the e2e test.

checks/pinned_dependencies.go
chrismcgehee: I tested out changing the `runs-on` field to something that wasn't a string or a slice. The error was:
```
internal error: invalid yaml file: internal error: error decoding stringOrSlice Value: yaml: unmarshal errors: line 9: cannot unmarshal !!map into string
```
I did the same to a different field (`shell`), and it gave the same error (just not wrapped):
```
internal error: invalid yaml file: yaml: unmarshal errors: line 13: cannot unmarshal !!map into string
```
So I don't see a need to indicate a special error when `run-on` is the wrong type.

chrismcgehee:
> LGTM. Make sure to run the other unit tests and e2e tests (they are currently disabled in pre-submits).

I ran these, and my changes did not cause any failures.

> Can you send me a PR to https://github.com/ossf-tests/scorecard-check-pinned-dependencies-e2e with additional tests you've added to this PR? This repo is used by the e2e test.

The tests I added here are just unit tests. Do I need to add anything to this repo if I didn't add e2e tests?

checks/pinned_dependencies_test.go
```diff
@@ -642,3 +644,130 @@ func TestGitHubWorkflowUsesLineNumber(t *testing.T) {
...
+				for _, step := range job.Steps {
+					step := step
+					shell, err := getShellForStep(&step, &job)
+					if err != nil {
+						t.Errorf("error getting shell: %w", err)
+					}
+					actualShells = append(actualShells, shell)
+				}
+			}
+			if !areSlicesEqual(tt.expectedShells, actualShells) {
```
azeemshaikh38, line 756: Use `cmp.Equal` here: https://pkg.go.dev/github.com/google/go-cmp/cmp

checks/pinned_dependencies.go
```diff
@@ -33,23 +33,60 @@ const CheckPinnedDependencies = "Pinned-Dependencies"
...
+	Uses  stringWithLine `yaml:"uses"`
+}
+
+// stringOrSlice is for fields that can be a single string or a slice of strings. If the field is a single string,
+// this value will be a slice with a single string item.
+type stringOrSlice []string
+
+func (s *stringOrSlice) UnmarshalYAML(value *yaml.Node) error {
+	var stringSlice []string
+	err := value.Decode(&stringSlice)
```
azeemshaikh38, line 77: Nit: use early return:

```
if err == nil {
  *s = stringSlice
  return nil
}
// Handle error
...
```

checks/pinned_dependencies.go
```diff
@@ -494,6 +526,87 @@ func validateGitHubWorkflowIsFreeOfInsecureDownloads(pathfn string, content []by
...
+func getShellForStep(step *gitHubActionWorkflowStep, job *gitHubActionWorkflowJob) (string, error) {
+	// https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell.
+	if step.Shell != "" {
+		return step.Shell, nil
+	}
+	if job.Defaults.Run.Shell != "" {
+		return job.Defaults.Run.Shell, nil
+	}
+
+	const defaultShellNonWindows = "bash"
```
azeemshaikh38, line 557: Define these consts at the top of the file

checks/pinned_dependencies.go
```diff
@@ -494,6 +526,87 @@ func validateGitHubWorkflowIsFreeOfInsecureDownloads(pathfn string, content []by
...
+		return defaultShellWindows, nil
+	}
+
+	return defaultShellNonWindows, nil
+}
+
+// isStepWindows returns true if the step will be run on Windows.
+func isStepWindows(step *gitHubActionWorkflowStep) (bool, error) {
+	// Looking for "if: runner.os == 'Windows'" (and variants)
+	matches, err := regexp.MatchString(`(?i)runner\.os\s*==\s*['"]windows['"]`, step.If)
```
azeemshaikh38, line 578: Define this regex (and the one below) as consts at top of the file.

checks/pinned_dependencies.go
```diff
@@ -494,6 +526,87 @@ func validateGitHubWorkflowIsFreeOfInsecureDownloads(pathfn string, content []by
...
+	if err != nil {
+		//nolint:wrapcheck
+		return false, sce.Create(sce.ErrScorecardInternal, fmt.Sprintf("error matching runner.os regex: %v", err))
```
azeemshaikh38, line 581: Instead of repeating same code multiple times run it inside a for loop?

azeemshaikh38 (approved):

azeemshaikh38:
> @azeemshaikh38 could you review this one?

Apologies for the delay. This got buried in my emails. Looks good to me with minor comments.

checks/pinned_dependencies.go
chrismcgehee: What's the benefit to this? When it's at the top of the file, I think it adds cognitive overhead because it's away from the context where it is used.

checks/pinned_dependencies.go
chrismcgehee: Oh, yeah, of course!

checks/pinned_dependencies.go
azeemshaikh38: My thinking was that it'll help improve readability to see all the different regex patterns in one place. But I agree with you, a bunch of complex regexes at the top of the file may not really help readability. Let's just define them inside the function.

chrismcgehee:
Ready to be merged

PR #882, opened 2021-08-23
✨ Remove token-heavy checks from cron job

azeemshaikh38:

Removes token heavy checks from the cron job to try and improve scaling.

No.

### DIFF
--- a/cron/worker/main.go
+++ b/cron/worker/main.go
@@ -217,12 +217,11 @@ func main() {
 	}()
 
 	checksToRun := checks.AllChecks
-	// nolint
-	// FIXME :- deleting branch-protection
-	// The branch protection check needs an admin access to the repository.
-	// All of the checks from cron would fail and uses another call to the API.
-	// This will reduce usage of the API.
-	delete(checksToRun, checks.CheckBranchProtection)
+	// TODO: Temporarily remove checks which require lot of GitHub API token.
+	delete(checksToRun, checks.CheckSAST)
+	delete(checksToRun, checks.CheckCITests)
+	// TODO: Re-add Contributors check after fixing: #859.
+	delete(checksToRun, checks.CheckContributors)
 	for {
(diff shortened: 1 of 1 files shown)


cron/worker/main.go
```diff
@@ -217,6 +217,11 @@ func main() {
 	}()
 
 	checksToRun := checks.AllChecks
+	// Remove checks which require lot of GitHub API token.
```
inferno-chromium, line 220: Add TODO: and say Remove checks temporarily

cron/worker/main.go
```diff
@@ -217,6 +217,11 @@ func main() {
...
+	delete(checksToRun, checks.CheckSAST)
+	delete(checksToRun, checks.CheckCITests)
+	// TOOD: Re-add Contributors check after fixing: #859.
+	delete(checksToRun, checks.CheckContributors)
```
inferno-chromium, line 224: Also, can you enable branch protection check below. that does not need to disabled since @asraa code went in.

inferno-chromium (approved):

cron/worker/main.go
azeemshaikh38: Done.

cron/worker/main.go
azeemshaikh38: I tested `Branch-Protection` locally on `kubernetes` and it returns `branch protection is enabled on development and all release branches but settings are unknown`. IIUC, this info is what we get through `Code-Review` too. Not sure if its worth enabling `Branch-Protection` yet. @laurentsimon @asraa thoughts?

inferno-chromium: that is expected, it will show score=1, the workaround we have won't give furthur info, but we atleast have a differentiator (1=where something is enabled, 0=not enabled)

azeemshaikh38: Done. Enabled.

PR #885, opened 2021-08-23
🐛 Add ephemeral-storage to cron worker

azeemshaikh38:

Cron workers were getting evicted since no ephemeral storage was requested. Explicitly requesting ephemeral-storage to avoid this.

No.

### DIFF
--- a/cron/k8s/worker.release.yaml
+++ b/cron/k8s/worker.release.yaml
@@ -50,9 +50,11 @@ spec:
           requests:
             cpu: 5
             memory: 5Gi
+            ephemeral-storage: 100Gi
           limits:
             cpu: 12
             memory: 12Gi
+            ephemeral-storage: 500Gi
   strategy:
     type: "RollingUpdate"
     rollingUpdate:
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):
Use more :) 100/500 is fine as well

naveensrinivasan (commented):
It is the https://kubernetes.io/docs/concepts/storage/volumes/#emptydir that is causing ephemeralstorage to get filled up. I agree with @inferno-chromium  we should probably go higher.

azeemshaikh38:
> Use more :) 100/500 is fine as well

Done.

azeemshaikh38:
> It is the https://kubernetes.io/docs/concepts/storage/volumes/#emptydir that is causing ephemeralstorage to get filled up. I agree with @inferno-chromium we should probably go higher.

That makes more sense. Thanks!

naveensrinivasan (approved):
Thanks!

oliverchang (commented):
LGTM

PR #900, opened 2021-08-24
🐛 Fixes for Branch Protection

azeemshaikh38:

Minor fixes for the `Branch-Protection` check.

No.

### DIFF
--- a/checks/branch_protection.go
+++ b/checks/branch_protection.go
@@ -17,6 +17,7 @@ package checks
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"regexp"
 
@@ -37,6 +38,7 @@ func init() {
 	registerCheck(CheckBranchProtection, BranchProtection)
 }
 
+// TODO: Use RepoClient interface instead of this.
 type repositories interface {
 	Get(context.Context, string, string) (*github.Repository,
 		*github.Response, error)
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):

pkg/scorecard.go
```diff
@@ -92,20 +92,16 @@ func RunScorecards(ctx context.Context,
...
-	if err != nil {
-		//nolint:wrapcheck
-		return ScorecardResult{}, err
-	}
-
-	if len(commits) == 0 {
-		//nolint:wrapcheck
-		return ScorecardResult{}, sce.Create(sce.ErrScorecardInternal, "no commits found")
+	var commitSHA string
+	if commits, err := repoClient.ListCommits(); err != nil && len(commits) > 0 {
```
laurentsimon, line 96: I think this should be `err == nil`?

pkg/scorecard.go
```diff
@@ -92,20 +92,16 @@ func RunScorecards(ctx context.Context,
...
+		commitSHA = commits[0].SHA
+	} else {
```
laurentsimon, line 98: let's differentiate between the "no commit" error and the case `ListCommits()` returned an error?

checks/branch_protection.go
```diff
@@ -84,40 +115,43 @@ func checkReleaseAndDevBranchProtection(ctx context.Context, r repositories, dl
...
-		if !p {
+		if !branch.GetProtected() {
 			protected = false
 			dl.Warn("branch protection not enabled for branch '%s'", b)
 		} else {
 			// The branch is protected. Check the protection.
 			score, err := getProtectionAndCheck(ctx, r, dl, ownerStr, repoStr, b)
 			if err != nil {
-				if errors.Is(err, errInternalBranchNotFound) {
+				if errors.As(err, &errInternalBranchNotFound) {
```
laurentsimon, line 154: can you tell me the difference between `As` and `Is` and why `As` is better?

laurentsimon (approved):
In the PR, maybe add what `Minor fixes` mean. What is a bug in the code or some cleanup?

pkg/scorecard.go
azeemshaikh38: Done.

pkg/scorecard.go
azeemshaikh38: Done.

checks/branch_protection.go
azeemshaikh38: Ah you are right! I always considered `As` being the equivalent of `check error type only and not value`. Need to update all occurring of `errors.As` in codebase.

PR #902, opened 2021-08-25
🐛 Use singleton pattern for OSS-Fuzz

azeemshaikh38:

Avoids `Init-Repo` which downloads the tarball everytime.

No.

### DIFF
--- a/checks/fuzzing.go
+++ b/checks/fuzzing.go
@@ -16,6 +16,7 @@ package checks
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/ossf/scorecard/v2/checker"
 	"github.com/ossf/scorecard/v2/clients"
@@ -26,6 +27,12 @@ import (
 // CheckFuzzing is the registered name for Fuzzing.
 const CheckFuzzing = "Fuzzing"
 
+var (
+	ossFuzzRepo    clients.RepoClient
+	errOssFuzzRepo error
+	once           sync.Once
+)
+
 //nolint:gochecknoinits
 func init() {
 	registerCheck(CheckFuzzing, Fuzzing)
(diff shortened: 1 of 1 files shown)


checks/fuzzing.go
```diff
@@ -26,16 +27,25 @@ import (
...
 }
 
 // Fuzzing runs Fuzzing check.
 func Fuzzing(c *checker.CheckRequest) checker.CheckResult {
-	ossFuzzRepo := githubrepo.CreateGithubRepoClient(c.Ctx, c.Client, c.GraphClient)
-	if err := ossFuzzRepo.InitRepo("google", "oss-fuzz"); err != nil {
-		e := sce.Create(sce.ErrScorecardInternal, fmt.Sprintf("InitRepo: %v", err))
+	once.Do(func() {
+		ossFuzzRepo = githubrepo.CreateGithubRepoClient(c.Ctx, c.Client, c.GraphClient)
+		errOssFuzzRepo <- ossFuzzRepo.InitRepo("google", "oss-fuzz")
```
oliverchang, line 45: don't we have to `make()` the channel first? otherwise we're writing to a nil channel, which blocks forever.

checks/fuzzing.go
```diff
@@ -26,16 +27,25 @@ import (
...
+	})
+	if errOssFuzzRepo != nil {
```
oliverchang, line 47: We need to read from this channel first too right? Rather than compare the channel itself to nil?

checks/fuzzing.go
azeemshaikh38: Ah nice catch. Done.

checks/fuzzing.go
azeemshaikh38: Didn't realize this is checking the channel itself and not the value. Updated, PTAL.

checks/fuzzing.go
```diff
@@ -26,16 +27,33 @@ import (
...
 	registerCheck(CheckFuzzing, Fuzzing)
 }
 
 // Fuzzing runs Fuzzing check.
 func Fuzzing(c *checker.CheckRequest) checker.CheckResult {
-	ossFuzzRepo := githubrepo.CreateGithubRepoClient(c.Ctx, c.Client, c.GraphClient)
-	if err := ossFuzzRepo.InitRepo("google", "oss-fuzz"); err != nil {
-		e := sce.Create(sce.ErrScorecardInternal, fmt.Sprintf("InitRepo: %v", err))
+	once.Do(func() {
+		ossFuzzRepo = githubrepo.CreateGithubRepoClient(c.Ctx, c.Client, c.GraphClient)
```
oliverchang, line 46: I think we can just set ossFuzzRepo and errOssFuzzRepo directly without involving any channels.

oliverchang (approved):
LGTM but I think we can make this a bit simpler and not involve channels at all :)

checks/fuzzing.go
azeemshaikh38: Done :)

PR #903, opened 2021-08-25
🐛 Use `arduino/setup-protoc` for installing Protoc

azeemshaikh38:

Use `arduino/setup-protoc` action to install protoc. Its way faster than doing it ourselves.

Note: if we choose to do protoc installation ourselves in the future, we should simply download the release archive which will have the `protoc` binary.

No.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,32 +16,11 @@ name: build
 on: [push, pull_request]
 jobs:
   validate:
-    name: Validate 
+    name: Validate
     runs-on: ubuntu-latest
     steps:
-     - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
-     - name: Cache protobuf library.
-       id: cache-protobuf
-       uses: actions/cache@d9747005de0f7240e5d35a68dca96b3f41b8b340
-       with:
-         path: protobuf
-         key: ${{ runner.os }}-protobuf
-     - name: Build protobuf library.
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -16,32 +16,11 @@ name: build
...
-         ./configure
-         make
-         make check
-     - name: Install protobuf library
-       run: |
-         cd protobuf
-         sudo make install
-         sudo ldconfig
+     - name: Install Protoc
+       uses: arduino/setup-protoc@v1
```
naveensrinivasan, line 23: Could we pin the  SHA instead of `@v1`?

naveensrinivasan (approved):
Thank you!

.github/workflows/main.yml
azeemshaikh38: Apologies, auto-merge was enabled so couldn't get to this in time. However, I was not able to find the release SHA. Do you know where I can find it?

naveensrinivasan: Not an issue  👍 
This is the location for the SHA's https://github.com/arduino/setup-protoc/releases and https://github.com/arduino/setup-protoc/commit/64c0c85d18e984422218383b81c52f8b077404d3

azeemshaikh38: Ah I see, so for GitHub releases we pin using commit SHA?

naveensrinivasan: Yes!

> Ah I see, so for GitHub releases we pin using commit SHA?

PR #934, opened 2021-08-30
✨  Improve JSON format

laurentsimon:

feature

lack of info in JSON format

1. JSON format uses lower-case letter for fields (OSV, GitHub API, etc all use lowercase fields)
2. Add scorecard's hash and version to JSON
3. Retrieve version/hash information automatically
4. Move build info under `pkg/`

no

* **Other information**:

### DIFF
--- a/pkg/json.go
+++ b/pkg/json.go
@@ -41,18 +41,28 @@ type jsonScorecardResult struct {
 
 //nolint
 type jsonCheckResultV2 struct {
-	Details []string
-	Score   int
-	Reason  string
-	Name    string
+	Details []string `json:"details"`
+	Score   int      `json:"score"`
+	Reason  string   `json:"reason"`
+	Name    string   `json:"name"`
+}
+
+type jsonRepoV2 struct {
+	Name   string `json:"name"`
+	Commit string `json:"commit"`
+}
+
+type jsonScorecardV2 struct {
+	Version string `json:"version"`
+	Commit  string `json:"commit"`
 }
 
 type jsonScorecardResultV2 struct {
-	Repo     string
-	Date     string
-	Commit   string
-	Checks   []jsonCheckResultV2
-	Metadata []string
+	Date      string              `json:"date"`
+	Repo      jsonRepoV2          `json:"repo"`
+	Scorecard jsonScorecardV2     `json:"scorecard"`
+	Checks    []jsonCheckResultV2 `json:"checks"`
+	Metadata  []string            `json:"metadata"`
 }
 
 // AsJSON exports results as JSON for new detail format.
@@ -60,7 +70,7 @@ func (r *ScorecardResult) AsJSON(showDetails bool, logLevel zapcore.Level, write
 	encoder := json.NewEncoder(writer)
 
 	out := jsonScorecardResult{
-		Repo:     r.Repo,
+		Repo:     r.Repo.Name,
 		Date:     r.Date.Format("2006-01-02"),
 		Metadata: r.Metadata,
 	}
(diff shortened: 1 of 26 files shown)


pkg/json.go
```diff
@@ -41,26 +41,36 @@ type jsonScorecardResult struct {
 
 //nolint
 type jsonCheckResultV2 struct {
-	Details []string
-	Score   int
-	Reason  string
-	Name    string
+	Details []string `json:"details"`
```
azeemshaikh38, line 44: The `json` annotations are not needed. Remove?

azeemshaikh38 (changes requested):
Can we decouple logic for cron job JSON from Scorecard JSON before this PR?

pkg/json.go
laurentsimon: I've renamed all JSON fields to be lower-case. Alternatively, I could try renaming the go structure's fields. wdut?

azeemshaikh38: IIUC, the Go struct fields need to be exposed to json encoder so cannot be lower case. Why do we want the JSON output to be lower case though?

laurentsimon: everywhere I look (OSV, GitHub API, etc), the fields are lower-case.

azeemshaikh38: I see. I'm ok with changing to lower case. These changes will affect the cron job (v2 version) though. So I would prefer holding off on this PR until we have decoupled the two.

laurentsimon: is there anything in particular that you think may break with this PR for v2? The BQ table needs to be updated, what else?

laurentsimon:
> Can we decouple logic for cron job JSON from Scorecard JSON before this PR?

I'll start working on it. Are there unit tests needs for the cron job/JSON? I don't see any. I think I'll add JSON unit tests/schema validator as I did for the scorecard JSON format. Sounds good?

pkg/json.go
azeemshaikh38: BQ schema change for sure, can't think of anything else.

azeemshaikh38:
> > Can we decouple logic for cron job JSON from Scorecard JSON before this PR?
> 
> I'll start working on it. Are there unit tests needs for the cron job/JSON? I don't see any. I think I'll add JSON unit tests/schema validator as I did for the scorecard JSON format. Sounds good?

SGTM

laurentsimon:
> > Can we decouple logic for cron job JSON from Scorecard JSON before this PR?
> 
> I'll start working on it. Are there unit tests needs for the cron job/JSON? I don't see any. I think I'll add JSON unit tests/schema validator as I did for the scorecard JSON format. Sounds good?

done with https://github.com/ossf/scorecard/pull/941

pkg/scorecard_result.go
```diff
@@ -30,19 +30,31 @@ import (
...
+	CommitSHA string
+}
+
 // ScorecardResult struct is returned on a successful Scorecard run.
 type ScorecardResult struct {
-	Repo      string
+	Repo      RepoInfo
 	Date      time.Time
-	CommitSHA string
+	Scorecard ScorecardInfo
```
azeemshaikh38, line 49: Nit: call this info?

pkg/scorecard_result.go
```diff
@@ -30,19 +30,31 @@ import (
 	sce "github.com/ossf/scorecard/v2/errors"
 )
 
+// ScorecardInfo contains information about the scorecard code that was run.
+type ScorecardInfo struct {
+	Version   string
+	CommitSHA string
```
azeemshaikh38, line 36: What is the difference between ScorecardInfo.CommitSHA vs. RepoInfo.CommitSHA?

pkg/scorecard_result.go
azeemshaikh38: Nvm.

pkg/scorecard_result.go
azeemshaikh38: Nvm.

azeemshaikh38 (approved):

Issue #783, opened 2021-07-29
Cleanup error handling


laurentsimon:
We have several places that define errors:

1. a bunch of errors (e.g., `ErrorUnsupportedHost `) are declared at https://github.com/ossf/scorecard/blob/c02df073a9f39e6914d7f934eb9f5af131ff2361/repos/repo_url.go#L27:L32
They are also used in a cron job file `cron/data/iterator_test.go`.

2. We have the main `errors/` directory with error definitions used by checks.

We need to harmonize this.

laurentsimon:
another thing is when we call a repoclient's API, shall we expect the repoclient create the `ErrScorecardInternal` or not?
If not, we need all `checks`, `checker` or `pkg` to create it.

PR #936, opened 2021-08-30
✨ Use RefUpdateRule in BranchProtection check

azeemshaikh38:

Uses graphQL API `refUpdateRule` to access non-admin branch protection fields.

No.


oliverchang (approved):

clients/githubrepo/branches.go
```diff
@@ -0,0 +1,203 @@
...
+	handler.once.Do(func() {
+		vars := map[string]interface{}{
+			"owner":         githubv4.String(handler.owner),
+			"name":          githubv4.String(handler.repo),
+			"refsToAnalyze": githubv4.Int(refsToAnalyze),
+			"refPrefix":     githubv4.String(refPrefix),
+		}
+		handler.data = new(branchesData)
+		if err := handler.graphClient.Query(handler.ctx, handler.data, vars); err != nil {
+			handler.errSetup = sce.Create(sce.ErrScorecardInternal, fmt.Sprintf("githubv4.Query: %v", err))
```
laurentsimon, line 95: since the repo implementation returns a `ErrScorecardInternal`, we can assume in the checks that errors from any `repo.*` functions can be returned as-is. Correct?

e2e/code_review_test.go
```diff
@@ -51,8 +51,8 @@ var _ = Describe("E2E TEST:CodeReview", func() {
 				Error:         nil,
 				Score:         checker.MaxResultScore,
 				NumberOfWarn:  0,
-				NumberOfInfo:  2,
-				NumberOfDebug: 30,
+				NumberOfInfo:  0,
```
laurentsimon, line 54: how about changing this tests to one of our test repos to avoid breaking? We can update it properly later.

laurentsimon (approved):

e2e/code_review_test.go
azeemshaikh38: I would prefer leaving this in explicitly broken state so that we don't miss this when we are fixing e2e tests.

clients/githubrepo/branches.go
azeemshaikh38: As of now, yes. But I think its best we don't keep this assumption. The `sce.Err*` errors should be returned by `checks` package. Internal packages like `gtihubrepo` should be free to return other internal errors.

laurentsimon: if we add `ErrScorecardInternal` in checks and the repo does too, we end up with multiple `ErrScorecardInternal` wrapping one another. Is there a better way to do it?

azeemshaikh38: Let's file an issue for cleaning this up because I think as of today, there is no clear guideline. IMO, only `checks` and `pkg` packages of Scorecard should return these `Internal` errors.

e2e/code_review_test.go
laurentsimon: there's one already https://github.com/ossf/scorecard/issues/783

clients/githubrepo/branches.go
laurentsimon: there's one already https://github.com/ossf/scorecard/issues/783

PR #949, opened 2021-09-01
🌱 Add and use RepoClient API for ListStatuses

azeemshaikh38:

Use a new ListStatuses for CITests check.

No.

### DIFF
--- a/clients/githubrepo/statuses.go
+++ b/clients/githubrepo/statuses.go
@@ -1,0 +1,61 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


clients/githubrepo/statuses.go
```diff
@@ -0,0 +1,61 @@
...
+	handler.ctx = ctx
+	handler.owner = owner
+	handler.repo = repo
+}
+
+func (handler *statusesHandler) listStatuses(ref string) ([]clients.Status, error) {
+	statuses, _, err := handler.client.Repositories.ListStatuses(handler.ctx, handler.owner, handler.repo, ref,
+		&github.ListOptions{})
+	if err != nil {
+		// nolint: wrapcheck
```
naveensrinivasan, line 44: Can we avoid this?

naveensrinivasan (approved):
Thanks

clients/githubrepo/statuses.go
azeemshaikh38: I plan to remove all these linters which we seem to run `nolint` on all the time. We have too many `nolint` comments.

naveensrinivasan: Sounds good.

Issue #744, opened 2021-07-23
Feature: Use `google/ko` instead of `Dockerfile`


azeemshaikh38:
Simplify our workflows and instead of Dockerfiles use `google/ko`.

naveensrinivasan:
@dlorenc Does ko support tags? The cork job uses couple of tags to “latest” and “stable” to differentiate which between test runs?

developer-guy:
> @dlorenc Does ko support tags? The cork job uses couple of tags to “latest” and “stable” to differentiate which between test runs?

hello @naveensrinivasan, yes `google/ko` work with tags, let me give you an example:

```shell
$ KO_DOCKER_REPO=docker.io/devopps ko publish -B --tags latest,0.3.3,v0 github.com/GoTurkiye/hello-world-cli
```

<details>

<summary> Output </summary

```shell
2021/10/10 18:09:30 Using base gcr.io/distroless/static:nonroot for github.com/GoTurkiye/hello-world-cli
2021/10/10 18:09:33 Building github.com/GoTurkiye/hello-world-cli for linux/amd64
2021/10/10 18:09:34 Publishing docker.io/devopps/hello-world-cli:latest
2021/10/10 18:09:36 existing blob: sha256:250c06f7c38e52dc77e5c7586c3e40280dc7ff9bb9007c396e06d96736cf8542
2021/10/10 18:09:36 existing blob: sha256:ec52731e927332d44613a9b1d70e396792d20a50bccfa06332a371e1c68d7785
2021/10/10 18:09:36 existing blob: sha256:24a21ecf55f58f84cd43d659032c2ecaad57faa9ec8ba82fdd9e3ab49a932d47
2021/10/10 18:09:37 existing blob: sha256:5a849f7331628c0129d61181f9d9f92dae6c664006cc827785072d1e23bd6442
2021/10/10 18:09:37 docker.io/devopps/hello-world-cli:latest: digest: sha256:ccb2d059265dad0c616f0b0b56bf74a14fa1242ab0cc2a49a7b08be784007d57 size: 952
2021/10/10 18:09:37 Tagging docker.io/devopps/hello-world-cli:0.3.3
2021/10/10 18:09:37 docker.io/devopps/hello-world-cli:0.3.3: digest: sha256:ccb2d059265dad0c616f0b0b56bf74a14fa1242ab0cc2a49a7b08be784007d57 size: 952
2021/10/10 18:09:37 Tagging docker.io/devopps/hello-world-cli:v0
[...]

PR #975, opened 2021-09-07
:seedling: ko publish images

naveensrinivasan:

 https://github.com/ossf/scorecard/issues/744

  https://github.com/ossf/scorecard/issues/744

no

* **Other information**:
no

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,3 +177,11 @@ ifndef GITHUB_AUTH_TOKEN
 	$(error GITHUB_AUTH_TOKEN is undefined)
 endif
 ###############################################################################
+
+
+###############################################################################
+
+.PHONY: ko-local
+ko-local:
+	KO_DOCKER_REPO=${KO_PREFIX}/scorecard CGO_ENABLED=0 ko publish --bare --local --tags stable --tags $(GIT_HASH) .
+


Makefile
```diff
@@ -177,3 +177,11 @@ ifndef GITHUB_AUTH_TOKEN
 	$(error GITHUB_AUTH_TOKEN is undefined)
 endif
 ###############################################################################
+
+
+###############################################################################
+
+.PHONY: ko-local
+ko-local:
+	KO_DOCKER_REPO=${KO_PREFIX}/scorecard CGO_ENABLED=0 ko publish --bare --local --tags stable --tags $(GIT_HASH) .
```
dlorenc, line 186: You can just set KO_DOCKER_REPO, you only need the prefix bit if you have multiple images.

naveensrinivasan: The reason I am prefixing it was we do have multiple images. These images are in the cron-job. Thanks

Makefile
```diff
@@ -179,3 +179,11 @@ ifndef GITHUB_AUTH_TOKEN
 	$(error GITHUB_AUTH_TOKEN is undefined)
 endif
 ###############################################################################
+
+
+###############################################################################
+
+.PHONY: ko-local
+ko-local:
```
azeemshaikh38, line 187: I don't fully understand `ko` so I have a bunch of questions/comments:

1. Where is `KO_PREFIX` defined? 
2. Should it be `$()` instead of `${}`?
3. Its missing flags defined by `make build-scorecard` like version flags - https://github.com/ossf/scorecard/blob/main/Makefile#L108
4. How does this impact CloudBuild? And should we remove the existing `Dockerfile` for `scorecard`?
5. How will `ko` handle more complex Dockerfile usecases like copying over data file (like in cron/controller)?

IMO, it'll be useful to discuss the full plan in #744 before proceeding with the PR.

naveensrinivasan: > I don't fully understand `ko` so I have a bunch of questions/comments:
> 
> 1. Where is `KO_PREFIX` defined?
Ko_PREFIX would be defined the github actions before pushing it.
> 2. Should it be `$()` instead of `${}`?

> 3. Its missing flags defined by `make build-scorecard` like version flags - https://github.com/ossf/scorecard/blob/main/Makefile#L108
This is initial for pushing to `ko-local` which means this is not for pushing to gcr.io yet. It is pushing to local docker.
> 4. How does this impact CloudBuild? And should we remove the existing `Dockerfile` for `scorecard`?
In following PR's I will include things to push to gcr.io and then the cloudbuild and the existing docker file will be removed.
> 5. How will `ko` handle more complex Dockerfile usecases like copying over data file (like in cron/controller)?
It will be able to handle copying over data https://github.com/google/ko#static-assets
> 
> IMO, it'll be useful to discuss the full plan in #744 before proceeding with the PR.

azeemshaikh38: I see, in that case can we either extend this PR to provide more e2e/wholistic usage of `ko` or we can make a design before starting with a series of PRs. Currently, this PR is adding dead code which will not be touched without follow up PRs, so I would prefer not submitting it.

Issue #942, opened 2021-08-31
Scorecard repo fails at `Token-Permissions` with score `0`


laurentsimon:
We need to fix this. A cursory read of our workflows:

- [x] Codeql -> `contents:read`, `statuses: write`and possibly `security-events: write`
- [ ] Goreleaser -> `contents:read`
- [ ] Integration -> `contents:read`, `pull-requests:write` for this.
- [ ] Main -> `contents:read`, `status:write` for license part.
- [ ] Ok-to-test -> ?
- [ ] Stale -> `contents:read` and `issues:write` and `pull-requests:write`
- [ ] Verify -> `contents:read` and `statuses:write`

We can define `permission:` to set all permissions to `none` at the start of the workflows, and then set what we want in each `runs`. It's fine to set `contents:read` at the start of workflows, if this helps.

See https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/ for more info

laurentsimon:
@naveensrinivasan , can you handle it?

laurentsimon:
As we fix this, let's keep a record to explain the permissions we need. Everyone struggles with GitHub permissions AFAIK

naveensrinivasan:
I will take this.

naveensrinivasan:
> As we fix this, let's keep a record to explain the permissions we need. Everyone struggles with GitHub permissions AFAIK

It would be great to have a tool that will produce the actual permissions needed for these actions.

azeemshaikh38:
Can we close this issue now? @laurentsimon @naveensrinivasan

laurentsimon:
yes. Scorecard still gives a score of 0 but it's a bug. Will create an issue for tracking https://github.com/ossf/scorecard/issues/1400

PR #985, opened 2021-09-09
:seedling: Fix integration githubaction permissions

naveensrinivasan:

 Bug Fix

   https://github.com/ossf/scorecard/issues/942

  * Changed the integration GitHub action permissions to contents:read, pull-requests:write

No

* **Other information**:
No


naveensrinivasan:
Not yet ready, figuring out the right permissions.

.github/workflows/integration.yml
```diff
@@ -22,6 +22,19 @@ name: Integration tests
 env:
   IMAGE_NAME: scorecard
   OUTPUT_PATH: output
+
+permissions:
+  actions: none
```
laurentsimon, line 27: note: you need to not mention `name: none`. By default, permission not declared in the list default to `none`.

laurentsimon (approved):

.github/workflows/integration.yml
naveensrinivasan: Do you see an issue in having them explicitly calling out as `none`? IMO this specifically encourages the practice of we have explicitly looked into those permissions and have defined them as not needed. This could be an example for other repositories to follow this as a practice. Thoughts?

laurentsimon: I prefer the condensed version, but I don't have a strong opinion on the matter. Omitting them is more condensed, but it's fine to have them all.

laurentsimon: thinking about this more. I think the condensed version is easier to read from a human's point of view. The full version makes it hard to see what's not set to `none`. The condensed version feels more like an `allow list`, where only what's needed is declared.

naveensrinivasan: Sounds good.

azeemshaikh38:
Any ETA here @naveensrinivasan ?

laurentsimon:
I came across this interesting app that claims to do that automatically https://github.com/apps/step-security, and it's currently free to try. (It requires some `write` permission, though)

naveensrinivasan:
> Any ETA here @naveensrinivasan ?

I have made a push now. Hopefully this will be good. It is hard to determine the exact permission.

laurentsimon:
Awesome, thanks for making that change @naveensrinivasan

Issue #962, opened 2021-09-03
Reduce number of `nolint` comments in code 


azeemshaikh38:
We have way too many `nolint` comments in the code. Tracking issue for improving this technical debt by either:

i) Removing unhelpful linters
ii) Improving coding practices

neil465:
I would like to work on this issue.

azeemshaikh38:
Great, thanks! Assigning this issue to you.

naveensrinivasan:
> We have way too many `nolint` comments in the code. Tracking issue for improving this technical debt by either:
> 
> i) Removing unhelpful linters

Which linters do we want to remove? 

> ii) Improving coding practices

azeemshaikh38:
On the top of my head, the ones I can think of are:

- gomnd (magic number detection. causes lot of false positives)
- prealloc (forces you to pre-allocate slices)
- goerr113 (disallows you from creating errors on the fly)
- dupl (lot of false positives in the `e2e` package)

Also, a lot of times we have `nolint` which disables all linters instead of a specific one. We should fix this too and make it our guideline to only disable specific linters in comment if/when needed.

naveensrinivasan:
> On the top of my head, the ones I can think of are:
> 
> * gomnd (magic number detection. causes lot of false positives)

I agree this can be a pain and wouldn't mind removing this. 

> * prealloc (forces you to pre-allocate slices)

prealloc has performance implications https://github.com/alexkohler/prealloc#purpose , so why do you think this should be removed?

> * goerr113 (disallows you from creating errors on the fly)

The go error in 1.13 https://go.dev/blog/go1.13-errors is the right way to create errors. If we don't handle errors correctly we are likely not nested call-stack of the error. What is your reasoning behind removing this>

> * dupl (lot of false positives in the `e2e` package)

I agree `dupl` is an issue for `e2e` package, but it also helps in avoiding duplicates in the future.  

> 
> Also, a lot of times we have `// nolint` which disables all linters instead of a specific one. We should fix this too and make it our guideline to only disable specific linters in comment if/when needed.

azeemshaikh38:
> prealloc has performance implications https://github.com/alexkohler/prealloc#purpose , so why do you think this should be removed?

Because this link seems to prefer empty slices over pre-allocated one: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices. I'm ok with either way as long as we agree with one guideline and apply it across the code without the need for `// nolint`. 

[...]

PR #995, opened 2021-09-10
🌱 sce.Create -> sce.WithMessage to comply with wrapcheck

azeemshaikh38:

`wrapcheck` linter ignores fn signature of form `Wrap` or `WithMessage`. See - https://golangci-lint.run/usage/linters/#wrapcheck. Change `sce.Create` -> `sce.WithMessage` to avoid `nolint` everytime it is used.

#962 

No.


docs/checks/checks.md
```diff
@@ -0,0 +1,166 @@
+
```
inferno-chromium, line 1: Is this intended for another PR ? it is fine if this is just fixing some path issue

inferno-chromium (approved):

docs/checks/checks.md
azeemshaikh38: This should be fixed with #996

Issue #998, opened 2021-09-10
Include URL for more information


david-a-wheeler:
**Is your feature request related to a problem? Please describe.**

Currently there's a very short description of the issues & what to do, but many users will need more information. e.g., on *why* that's a problem, the impact if not changed, and specifics of how to fix it for various circumstances.

**Describe the solution you'd like**

Please include, for the output of each check, a URL to more detail within ossf/scorecard/docs. You could use a link to https://github.com/ossf/scorecard/blob/main/docs/checks.md#CHECK-NAME , though I think it might be wise for each check to have its own *separate* file inside `docs`. The top-level README.md should also link to this docs/ information for each check. Then you can easily update the documentation, and users can easily see the current documentation (without updating the tool just to see better docs).

**Describe alternatives you've considered**

Links to pages *not* under the scorecard page seem like a terrible idea, because scorecard can't fix them. Link to scorecard, and then scorecard can link to those other pages as appropriate.

Don't include documentation in the tool itself; then people have to update the tool to update the docs.

**Additional context**

This should be relatively easy, you already have at least starting material. It's important to also improve the documentation about the checks, but this feature request would make those docs more widely used.

david-a-wheeler:
@naveensrinivasan - This was inspired by the CII Best Practices badge experience so far.

laurentsimon:
This is now supported:
```
go run . --repo=https://github.com/bhuvi11/ecom-service --checks Code-Review --show-details --format json | jq
{
  "date": "2021-09-10",
  "repo": {
    "name": "github.com/bhuvi11/ecom-service",
    "commit": "bb5488c6b99bcbb0954bf5cb7b569068c1cf89fb"
  },
  "scorecard": {
    "version": "unknown",
    "commit": "unknown"
  },
  "checks": [
    {
      "details": [
        "Info: Gerrit code reviews found for 0 commits out of the last 30",
        "Info: Prow code reviews found for 0 commits out of the last 5"
      ],
      "score": 10,
      "reason": "GitHub code reviews found for 5 commits out of the last 5 -- score normalized to 10",
      "name": "Code-Review",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/main/docs/checks.md#code-review",
        "short": "Determines if the project requires code review before pull requests are merged."
      }
    }
  ],
  "metadata": null
}

```

This does not address the other things like a dedicated doc file for each check, etc. We'll get there was we refine check's documentation and remediation.
[...]

Issue #978, opened 2021-09-08
Availability and freshness of data available on Google Storage


fridex:
**Is your feature request related to a problem? Please describe.**

Hi everyone,

I'm trying to consume scorecards from google storage available at `gs://ossf-scorecards/`. The last entry I can see is `gs://ossf-scorecards/05-17-2021.json` according to the date in object key. The date also corresponds to the `Date` field in scorecard entries available in `gs://ossf-scorecards/latest.json`.

I would like to ask how often are scorecards refreshed and available to the community.

Thanks in advance for any response.

naveensrinivasan:
scorecard has moved to Big Query https://github.com/ossf/scorecard#public-data. Please use that. Thanks

naveensrinivasan:
> > scorecard has moved to Big Query https://github.com/ossf/scorecard#public-data. Please use that. Thanks
> 
> If I understand it correctly, authentication is required to obtain the dataset. Is there a way to access the dataset without authentication? Or, are there any plans to provide the dataset in an open way? Thanks for your response.

AFAIK this is open and does not require authentication @azeemshaikh38  can you please chime in?

fridex:
I've tried the example from the README file:

```
$ bq query --nouse_legacy_sql 'SELECT partition_id FROM
openssf.scorecardcron.INFORMATION_SCHEMA.PARTITIONS WHERE table_name="scorecard"
ORDER BY partition_id DESC
LIMIT 1'
ERROR: (bq) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials.

If you have already logged in with a different account:

    $ gcloud config set account ACCOUNT

to select an already authenticated account to use.
```

Also, tried with Python client:

```python
from google.cloud import bigquery

client = bigquery.Client()
query_job = client.query("""SELECT partition_id FROM openssf.scorecardcron.INFORMATION_SCHEMA.PARTITIONS WHERE table_name="scorecard" ORDER BY partition_id DESC LIMIT 1""")
results = query_job.result()  
```

with the following exception raised:

```
Traceback (most recent call last):
  File "./t.py", line 5, in <module>
    client = bigquery.Client()
  File "/home/fpokorny/.local/share/virtualenvs/prescriptions-refresh-job-equYSKS1/lib/python3.8/site-packages/google/cloud/bigquery/client.py", line 215, in __init__
    super(Client, self).__init__(
[...]

PR #1001, opened 2021-09-10
✨ Update cron's JSON format

laurentsimon:

feature

older json

same json format as scorecard itself. It's a mere copy of files

yes

* **Other information**:


cron/format/testdata/check3.json
```diff
@@ -0,0 +1,51 @@
...
+   },
+   "checks": [
+      {
+         "details": [
+            "Warn: warn message: bin/binary.elf"
+         ],
+         "score": 0,
+         "reason": "min result reason",
+         "name": "Check-Name",
+         "documentation": {
```
oliverchang, line 19: These will be very repetitive in the BQ table. Is there any benefit to including this in every single row, as opposed to asking users to get this info from elsewhere?

azeemshaikh38 (approved):

cron/format/testdata/check3.json
laurentsimon: yep, I agree and thought about this too. BQ table supports column compression (Run Length Encoding, etc), so this should not take much space on disk. However, I just looked up pricing and we're charged based on [data size before compression](https://cloud.google.com/bigquery/pricing#data), so thats not ideal. 
The main reason I included it is ease for users to retrieve and for us to keep scorecard changes in sync with where the URL lives. Deps.dev already broke because the doc location changed.

Alternatives:
1. Create a url per json instead of per check. That still requires us to keep url#check-name non-breaking. This would break if we separate checks in their own doc file (as suggested in https://github.com/ossf/scorecard/issues/998).
2. URL that redirects to doc...

What are better options?

laurentsimon: @azeemsgoogle any thoughts on this?

laurentsimon:
some error
```
protoc --go_out=../../../ cron/data/metadata.proto
make: *** [Makefile:94: cron/data/metadata.pb.go] Error 127
```

@naveensrinivasan

cron/format/testdata/check3.json
azeemshaikh38: Do we want to launch Scorecard V2 (version with scores)? Given our conversations with ProjectsByIf, it is likely that our result formats will undergo massive overhaul soon. Do we gain anything by launching this version then?

laurentsimon: It helps users of deps.dev get a better format. I don't know how much traffic they get, but I think they are a source of discovery for scorecard we should not neglect. @kimsterv do you know?

In terms of projectsbyif, if we need to change scorecard/format a lot, then this means it will take quite some time to land the changes. In the meantime it's useful to have deps.dev use a score-based version.

wdut?

naveensrinivasan:
> some error
> 
> ```
> protoc --go_out=../../../ cron/data/metadata.proto
> /bin/bash: protoc: command not found
> make: *** [Makefile:94: cron/data/metadata.pb.go] Error 127
> ```
> 
> @naveensrinivasan

https://github.com/ossf/scorecard/blob/main/.github/workflows/main.yml#L23 this was fixed by @azeemshaikh38  and I was hoping this would fix the issue

cron/format/testdata/check3.json
laurentsimon: follow-up about pricing. https://cloud.google.com/bigquery/pricing says, under storage section: `$0.020 per GB`. If we reach 1M repos, considering a `300` characters/bytes per doc (url + description), given 10 checks, and with 4 runs per month, we get:
$0.020 * 300 * 10 * 1000000 * 4 / 1000000000 = $0.24/month.

Please chime in if I forgot something in the calculation above. In the calculation is correct, pricing is not an issue/blocker.

Given the use of compression by BQ, latency should not be an issue either.

So I think we can keep the doc fields per check in JSON. 

@oliverchang waiting for ur feedback

laurentsimon:
> some error
> 
> ```
> protoc --go_out=../../../ cron/data/metadata.proto
> /bin/bash: protoc: command not found
> make: *** [Makefile:94: cron/data/metadata.pb.go] Error 127
> ```
> 
> @naveensrinivasan

for some reason

> > some error
> > ```
> > protoc --go_out=../../../ cron/data/metadata.proto
> > /bin/bash: protoc: command not found
> > make: *** [Makefile:94: cron/data/metadata.pb.go] Error 127
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @naveensrinivasan
> 
> https://github.com/ossf/scorecard/blob/main/.github/workflows/main.yml#L23 this was fixed by @azeemshaikh38 and I was hoping this would fix the issue

for some reason this is now working. The problem seems to be happening sometimes only

cron/format/testdata/check3.json
azeemshaikh38: My concern is making deps.dev and OSSF metrics teams go through 2 different migrations. There might also be other teams (like Project Thoth - https://github.com/ossf/scorecard/issues/978#issuecomment-917955571) who are relying on the BQ data without us knowing about it. It might be worth figuring out a plan which minimizes the overhead on our clients for doing these migrations.  

re: pricing/storage - IMO its fine to have this data, even though it maybe redundant. For teams relying on our BQ data, it'll be useful to have this present in BQ itself.

oliverchang: Thanks for explaining. Ease of use + link stability seems like a good enough reason to keep this in the table.

oliverchang (approved):

cron/format/testdata/check3.json
laurentsimon: re: other teams relying on BQ dataset. I think we need to announce breaking changes/new versions on a dedicated channel. Maybe scorecard-announce@ mailing list or a dedicated section on pkg.go.dev, probably both. We cannot, ourselves, keep an inventory of *who* is using scorecard. I've added an item for the next scorecard sync. 

Note that updating the cron JSON does not force our users to use it. We can still support the old format for 6 months.

PR #1002, opened 2021-09-11
✨ Add policy file

laurentsimon:

feature

no policy support

policy file loaded - no effect yet

no

* **Other information**:


naveensrinivasan:
Sorry, I am still not clear with this feature. What does the policy file do?

laurentsimon:
> Sorry, I am still not clear with this feature. What does the policy file do?

PR still in Draft mode. TL:DR: the policy file will allow users to specify a minimum score for each check in order to customize the results and get a Pass/Fail per check.

The first use case is for SARIF in GitHub action: to allow users to define a minimum score; and be alerted if it's violated. Over time, we expect users to increase the minimum score to improve their security posture.

naveensrinivasan:
> > Sorry, I am still not clear with this feature. What does the policy file do?
> 
> PR still in Draft mode. TL:DR: the policy file will allow users to specify a minimum score for each check in order to customize the results and get a Pass/Fail per check.
> 
> The first use case is for SARIF in GitHub action: to allow users to define a minimum score; and be alerted if it's violated. Over time, we expect users to increase the minimum score to improve their security posture.

Thanks

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+	errRepeatingCheck = errors.New("check has multiple definitions")
+)
+
+var modes = map[string]bool{"enforced": true, "disabled": true, "logging": true}
+
+// CheckPolicy defines the policy for a check.
+//nolint:govet
+type CheckPolicy struct {
+	Score int    `yaml:"score"`
+	Mode  string `yaml:"mode"`
```
laurentsimon, line 41: I will add a `Risk` to this structure later. I will use the default doc risk or the policy risk if a policy is provided.
Main blocker here is doc link: to accommodate for dynamic risk, we need a proper web page and pass the risk as GET params. I'll add this to the doc soon.

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+	errInvalidScore   = errors.New("invalid score")
+	errInvalidMode    = errors.New("invalid mode")
+	errRepeatingCheck = errors.New("check has multiple definitions")
+)
+
+var modes = map[string]bool{"enforced": true, "disabled": true, "logging": true}
+
+// CheckPolicy defines the policy for a check.
+//nolint:govet
+type CheckPolicy struct {
```
azeemshaikh38, line 39: Let's define both `CheckPolicy` and `ScorecardPolicy` as proto bufs. We plan to use this data structures across various platforms (browsers, GitHub actions etc.), so let's keep the definition platform agnostic.

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+	errInvalidCheck   = errors.New("invalid check name")
+	errInvalidScore   = errors.New("invalid score")
+	errInvalidMode    = errors.New("invalid mode")
+	errRepeatingCheck = errors.New("check has multiple definitions")
+)
+
+var modes = map[string]bool{"enforced": true, "disabled": true, "logging": true}
+
+// CheckPolicy defines the policy for a check.
+//nolint:govet
```
azeemshaikh38, line 38: Invert the order of `Score` and `Mode` fields and remove `nolint` after that.

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+type CheckPolicy struct {
+	Score int    `yaml:"score"`
+	Mode  string `yaml:"mode"`
+}
+
+// ScorecardPolicy defines a policy.
+//nolint:govet
```
azeemshaikh38, line 45: Invert order of `Version` and `Policies` to avoid `nolint`.

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+type ScorecardPolicy struct {
+	Version  int                    `json:"version"`
+	Policies map[string]CheckPolicy `json:"policies"`
+}
+
+func (sp *ScorecardPolicy) Read(b []byte) error {
```
azeemshaikh38, line 51: Rename to `ParseFromYAML` to specify what this fn is doing.

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+	err := yaml.Unmarshal(b, sp)
+	if err != nil {
+		return sce.WithMessage(sce.ErrScorecardInternal, err.Error())
+	}
+
+	if sp.Version != 1 {
```
azeemshaikh38, line 57: Create an exported list `AllowedVersions` and update this line to check that `sp.Version` belongs to one of the `AllowedVersions`.

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+type CheckPolicy struct {
+	Score int    `yaml:"score"`
+	Mode  string `yaml:"mode"`
+}
+
+// ScorecardPolicy defines a policy.
+//nolint:govet
+type ScorecardPolicy struct {
+	Version  int                    `json:"version"`
+	Policies map[string]CheckPolicy `json:"policies"`
```
azeemshaikh38, line 48: Instead of `Policies` can we have check-specific fields like:

```
type ScorecardPolicy struct {
  BranchProtection CheckPolicy
  BinaryArtifacts CheckPolicy
  ...
}
```

It'll be useful when we want to customize and add more granularity to each check.

policy/policy.go
```diff
@@ -0,0 +1,85 @@
...
+		return sce.WithMessage(sce.ErrScorecardInternal, err.Error())
+	}
+
+	if sp.Version != 1 {
+		return sce.WithMessage(sce.ErrScorecardInternal, errInvalidVersion.Error())
+	}
+
+	checksFound := make(map[string]bool)
+	for n, p := range sp.Policies {
+		_, exists := checks.AllChecks[n]
```
azeemshaikh38, line 63: Nit: these `if` statements can be in single line:

```
if _, exists := checks.AllChecks[n]; !exists {
  ...
}
```

policy/policy_test.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"errors"
+	"io/ioutil"
+	"reflect"
+	"testing"
+
+	sce "github.com/ossf/scorecard/v2/errors"
+)
+
+//nolint
```
azeemshaikh38, line 26: Let's not add `nolint` which disables all linters. Ideally would be great to fix the failing lint issue, but if it isn't worth taking the effort to fix, specify the linter explicitly - `// nolint: <linterName>`.

main.go
```diff
@@ -17,6 +17,7 @@ package main
 
 import "github.com/ossf/scorecard/v2/cmd"
 
+//nolint:typecheck
```
azeemshaikh38, line 20: Why did we need to add this `nolint`?

cmd/root.go
```diff
@@ -59,17 +61,55 @@ const (
 	formatDefault = "default"
 )
 
+const (
+	scorecardLong = "A program that shows security scorecard for an open source software."
+	scorecardUse  = `./scorecard --repo=<repo_url> [--checks=check1,...] [--show-details]
```
azeemshaikh38, line 66: Add `[--policy=]` to `scorecardUse`

cmd/root.go
```diff
@@ -59,17 +61,55 @@ const (
...
+			log.Fatal(err)
+		}
+		err = sp.Read(data)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+}
+
+func ensureChecksHavePolicies(sp *spol.ScorecardPolicy, enabledChecks checker.CheckNameToFnMap) {
```
azeemshaikh38, line 84: Lets ignore `--checks` if `--policy` is passed? Policy file also has `mode: enabled/disabled` which is same as `--checks`.

cmd/root.go
```diff
@@ -59,17 +61,55 @@ const (
...
+	Use:   scorecardUse,
+	Short: scorecardShort,
+	Long:  scorecardLong,
 	Run: func(cmd *cobra.Command, args []string) {
 		// UPGRADEv3: remove.
 		var v3 bool
 		if _, v3 = os.LookupEnv("SCORECARD_V3"); v3 {
 			fmt.Printf("**** Using SCORECARD_V3 code ***** \n\n")
 		}
+		if !v3 {
```
azeemshaikh38, line 103: Should this be `!v3 && format==SARIF`?

policy/policy.go
laurentsimon: should be feasible.

cmd/root.go
laurentsimon: good point. I don't think we want this. Imagine a user has a policy on disk and they only want to run a subset of checks, or even a single: maybe for debugging, maybe to check why a specific check fails, etc. This would force users to create multiple policy files and becomes a usability concern. So I think we should keep the `--checks` to decide which checks to run, and verify we have a policy for it.

policy/policy.go
laurentsimon: will do. Note that various platforms could take a file as input; files are platform agnostic. I think protobuf is better too as it's easier to transport as a sequence of bytes  + files lead to all kinds of vulnerabilities.
@oliverchang in case you have something to comment/warn about

oliverchang: Using protobufs internally seems fine, as long as the actual format exposed to end users is JSON. Protobufs can be serialized to JSON, and I believe there is a stable ordering of properties (based on the field numbers).

laurentsimon: If I'm not mistaken @azeemsgoogle 's comment is using the protobuf for the exported structures `CheckPolicy` and `ScorecardPolicy`. Can you clarify?

cmd/root.go
azeemshaikh38: Will leave it upto you. My 2 cents - it'll make the code more complex and IMO its not worth optimizing for usability (at least for v0).

policy/policy.go
azeemshaikh38: To be clear, my reasoning to define these as protobufs is to guarantee that the data structure is consistently maintained across platforms. The intention is *not* to serialize these as protobufs. The best format to serialize will depend on the usecase, and in most cases we will probably stick to JSON. I believe this is what @oliverchang is also trying to say.

laurentsimon: I don't understand, probably better to take this offline over GVC. I thought the point of using  protobuf was that we can easily have clients use javascript or other languages to read it from a serialized string.
If it stays internal, isn't the limiting factor golang, ie everyone needs to use our golang APIs anyway...

laurentsimon: Talked to @azeemsgoogle offline. I'm fine with this. Thanks Azeem!

laurentsimon:
@azeemsgoogle I've addressed the comments. Can you take a second look?

policy/policy.go
laurentsimon: not done, as discussed offline.

cmd/root.go
```diff
@@ -59,17 +61,62 @@ const (
...
+	Short: scorecardShort,
+	Long:  scorecardLong,
 	Run: func(cmd *cobra.Command, args []string) {
 		// UPGRADEv3: remove.
 		var v3 bool
 		if _, v3 = os.LookupEnv("SCORECARD_V3"); v3 {
 			fmt.Printf("**** Using SCORECARD_V3 code ***** \n\n")
 		}
+		if format == formatSarif && !v3 {
+			log.Fatalf("sarif not supported yet")
```
azeemshaikh38, line 106: `Fatalf` -> `Fatal`

cmd/root.go
```diff
@@ -59,17 +61,62 @@ const (
...
 
+const (
+	scorecardLong = "A program that shows security scorecard for an open source software."
+	scorecardUse  = `./scorecard --repo=<repo_url> [--checks=check1,...] [--show-details] [--policy=file]
+or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show-details] [--policy=file]`
+	scorecardShort = "Security Scorecards"
+)
+
+func readPolicy() *spol.ScorecardPolicy {
+	if policyFile != "" {
```
azeemshaikh38, line 72: No need for checking if the file is empty. Its already done on caller side.

cmd/root.go
```diff
@@ -59,17 +61,62 @@ const (
...
 )
 
+const (
+	scorecardLong = "A program that shows security scorecard for an open source software."
+	scorecardUse  = `./scorecard --repo=<repo_url> [--checks=check1,...] [--show-details] [--policy=file]
+or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show-details] [--policy=file]`
+	scorecardShort = "Security Scorecards"
+)
+
+func readPolicy() *spol.ScorecardPolicy {
```
azeemshaikh38, line 71: Return `(*ScorecardPolicy, error)` ?

cmd/root.go
```diff
@@ -59,17 +61,62 @@ const (
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		return sp
+	}
+	return nil
+}
+
+func ensureChecksHavePolicies(sp *spol.ScorecardPolicy, enabledChecks checker.CheckNameToFnMap) {
+	for checkName := range enabledChecks {
```
azeemshaikh38, line 87: Rename to `checksHavePolicies` and return `bool` here.

cmd/root.go
```diff
@@ -59,17 +61,62 @@ const (
...
+			log.Fatal("policy not supported yet")
+		}
+
+		var policy *spol.ScorecardPolicy
+		if policyFile != "" {
+			policy = readPolicy()
+			if policy == nil {
+				log.Fatal("empty policy")
+			}
+		}
```
azeemshaikh38, line 119: Maybe move this down to line 173 instead of repeatedly checking for `policyFile != ""`

policy/policy.proto
```diff
@@ -0,0 +1,36 @@
...
+message CheckPolicy {
+    
+    // Mode definition.
+    enum Mode {
+        DISABLED = 0;
+        ENFORCED = 1;
+    }
+
+    Mode mode = 1;
+    sint32 score = 2;
```
azeemshaikh38, line 30: `sint32` -> `int32`?

azeemshaikh38 (approved):
LGTM with comments

policy/policy.proto
```diff
@@ -0,0 +1,36 @@
...
+
+message CheckPolicy {
+    
+    // Mode definition.
+    enum Mode {
+        DISABLED = 0;
+        ENFORCED = 1;
+    }
+
+    Mode mode = 1;
```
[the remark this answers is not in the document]
laurentsimon: what if we add more modes later? I think a bool is compatible with an int which is compatible with enum, so bool -> enum conversion would work if we ever need to add other modes. So I think it's safe to use bool. Agreed? It does not hurt to keep enum, though.

policy/policy.proto
laurentsimon: I think `sint32` serializes to smaller number of bytes, see https://developers.google.com/protocol-buffers/docs/encoding?csw=1#types, for a negative number `the resulting varint is always ten bytes long`

PR #1016, opened 2021-09-14
📖  Improve rationale for Binary-Artifacts

david-a-wheeler:
I'm fine with prohibiting binary executables, but
the *rationale* for doing this was completely unclear.

This commit rewrites the rationale to explain, in hopefully
a better way, why they can be a problem.

I prefer "executable" over "binary".
On digital computers, all data (including source code) are binaries :-).
In addition, some executables are simultaneously executables
and source code, e.g., shell scripts.
So I think what is meant here is a "generated binary".

I don't really think this merits a "High" level, but that's
a different dicussion.

* **Other information**:


docs/checks.md
```diff
@@ -11,11 +11,14 @@ If you have ideas for things to add, or new ways to detect things,
...
+It's fine to include, in the source repository, files that are simultaneously reviewable source code and executables, since these are reviewable. (Some interpretive systems, such as many operating system shells, don't have a format for separate generated executables.) Similarly, we allow including in the source repository source code generated by other tools (e.g., by bison, yacc, flex, and lex). There are potential downsides to generated source code, but generated source code tends to be much easier to review and thus presents a lower risk. Generated source code is also often difficult for external tools to detect.
```
naveensrinivasan, line 16: `separate` should be `separately`

naveensrinivasan (commented):
Few suggestions on typo's and not on the content :)

docs/checks.md
```diff
@@ -11,11 +11,14 @@ If you have ideas for things to add, or new ways to detect things,
...
+Including generated executables in a source repository increases user risks. Many programming language systems can generate executables from source code (e.g., C/C++ generated machine code, Java `.class` files, Python `.pyc` files, and minified JavaScript). Users will often directly use executables if they are included in the source repository, leading to many dangerous behaviors. One problem is that reviews becomes ineffective and thus enable the use of obsolete or maliciously-subverted executables. Reviews generally review source code, not executables, since it's difficult to audit executables to ensure that they correspond. Since it's hard to audit this, over time the included executables might not correspond to the source code. Another problem is that it allows the executable generation process to atrophy, which can lead to an inability to create working executables.  These problems can be countered with verified reproducible builds, but it's easier to implement verified reproducible builds when executables are not included in the source repository (since the executable generation process is less likely to have atrophied).
```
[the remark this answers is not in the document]
david-a-wheeler: > `Including generated executables in a source` should be `Including generated executables in a source,`

I don't think so. The next word is `repository`. That would produce `Including generated executables in a source, repository increases user risks. The first sentence uses the phrase `the source repository`, so let's use this:

> Including generated executables in the source repository increases user risks.




> `One problem is that reviews becomes` should be `One problem is that reviews become`

Agreed!

docs/checks.md
david-a-wheeler: I don't think so. `separately generated` focuses on generation, but we're not focused on generating source code files through some different process. The focus here is that the source files are *separate* from the executable files, so I think `separate` is correct. I'll reword the sentence to emphasize the point: in some cases there is no executable file that lives independently from the source file.

inferno-chromium (approved):

PR #1017, opened 2021-09-14
📖 Improve explanation about multiple reviewers (and their lack)

david-a-wheeler:
The current text oversells the value of multiple
reviewers, and falsely assumes it's always possible.
I'm a *huge* fan of having a second reviewer, but it
obviously *can't* be done when there's only 1 active participant.
Even projects with multiple active participants find it
difficult in practice if there aren't many participants.
Also, multiple reviewers guarantee nothing; the other
"reviewers" might be sock puppets or other subverted accounts.

So yes, encourage review, but let's make it clear that it
can't prevent all problems & that some projects cannot currently
do it. Put the details in Code-Review, where it best belongs.
Also, projects *can* try to remediate the lack of active participants,
so give them some practical remediation steps.

Finally: "pull request" is a GitHub-specific term.  GitLab, SourceForge,
and many other forges instead use the term "merge request".  So in the
interest of not locking into one specific proprietary service, let's
include a more generic term.


docs/checks/internal/checks.yaml
```diff
@@ -192,9 +249,19 @@ checks:
       [Prow](https://github.com/kubernetes/test-infra/tree/master/prow#readme)
       (labels "lgtm" or "approved") and Gerrit ("Reviewed-on" and "Reviewed-by").
     remediation:
+      - >-
+        If the project has only contributor, or does not have enough
```
laurentsimon, line 253: has only one contributor

docs/checks/internal/checks.yaml
```diff
@@ -204,17 +271,25 @@ checks:
...
+      Obviously projects with only one active participant
+      cannot meet it.
+      Many projects have a narrow scope and may not be able to attract
+      the interest of multiple organizations.
+      See [Code Reviews](#code-reviews) for added notes about
+      projects with a small number of participants.
+
       The check works by looking at the authors of recent commits
       and checking the `Company` field on the GitHub user profile. A contributor
       must have at least 5 commits in the last 30 commits. The check succeeds if
```
laurentsimon, line 291: note: you can remove the part about when the check succeeds. We don't use Pas/Fail anymore.

docs/checks/internal/checks.yaml
```diff
@@ -84,11 +84,35 @@ checks:
     short: Determines if the default and release branches are protected with GitHub's branch protection settings.
     description: >-
       [Branch protection](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches)
-      allows defining rules to enforce certain workflows for
-      branches, such as requiring a review or passing certain status checks.
+      can enable various rules to enforce certain workflows for
+      branches, such as preventing rewriting of public history
+      (e.g., a *force push*),
+      requiring review before acceptance into a main branch,
+      or passing certain status checks before acceptance into a main branch..
```
laurentsimon, line 91: nit: double dot

laurentsimon (approved):

laurentsimon:
@olivekl FYI

docs/checks/internal/checks.yaml
david-a-wheeler: Agreed!

docs/checks/internal/checks.yaml
david-a-wheeler: Ok, will do.

docs/checks/internal/checks.yaml
david-a-wheeler: Agreed.

docs/checks/internal/checks.yaml
```diff
@@ -326,8 +326,7 @@ checks:
 
       The check works by looking at the authors of recent commits
       and checking the `Company` field on the GitHub user profile. A contributor
-      must have at least 5 commits in the last 30 commits. The check succeeds if
-      all contributors span at least 2 different organizations (companies).
+      must have at least 5 commits in the last 30 commits.
```
laurentsimon, line 329: maybe say `a user is considered a contributor if they have at least 5 commits in the last 30 commits.`

laurentsimon (approved):

docs/checks/internal/checks.yaml
laurentsimon: note: the check currently gives the highest score if there are at least 3 contributors in the last 30 commits. Any number below that is score propportionally.

david-a-wheeler: I'll just fix the text for now. #1024 can be handled separately.

laurentsimon (approved):

PR #1019, opened 2021-09-15
📖 Warn when checks are prone to false negatives

david-a-wheeler:
Automated tools normally have some false negatives,
some false positives, or both. However, some scorecard criteria
are *especially* prone to false negatives (where a
project meets the criterion but the tool says it doesn't).

This commit adds warning text about false negatives for
criteria that are especially prone to false negatives.
In all cases the problem is that there are *many* ways to
implement the criterion, so while the tool may detect some
cases, there are countless other situations it will fail to detect.
While this doesn't *fix* the problem, warning the humans
will encourage them to double-check these criteria before
making decisions. Sometimes this is the best you can do, and
it's better than not having a warning.

### DIFF
--- a/docs/checks.md
+++ b/docs/checks.md
@@ -35,7 +35,7 @@ This check determines if the default and release branches are protected with Git
 
 This check tries to determine if the project runs tests before pull requests are merged.
 Running tests helps developers catch mistakes early on. A low score is considered 'Low' risk.
-The check works by looking for a set of well-known CI-system names in GitHub `CheckRuns` and `Statuses` among the recent commits (~30). A CI-system is considered well-known if its name contains any of the following: appveyor, buildkite, circleci, e2e, github-actions, jenkins, mergeable, test, travis-ci. The check succeeds if at least 75% of successful pull requests have at least one successful check associated with them. 
+The check works by looking for a set of well-known CI-system names in GitHub `CheckRuns` and `Statuses` among the recent commits (~30). A CI-system is considered well-known if its name contains any of the following: appveyor, buildkite, circleci, e2e, github-actions, jenkins, mergeable, test, travis-ci. The check succeeds if at least 75% of successful pull requests have at least one successful check associated with them. A project may meet this criterion yet have a failing scorecard report; there are many ways to implement this criterion and it's especially difficult for an automated tool (like scorecard) to detect them all. 
 
 **Remediation steps**
 - Check-in scripts that run all the tests in your repository.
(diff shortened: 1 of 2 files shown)


docs/checks.md
```diff
@@ -35,7 +35,7 @@ This check determines if the default and release branches are protected with Git
...
 Running tests helps developers catch mistakes early on. A low score is considered 'Low' risk.
-The check works by looking for a set of well-known CI-system names in GitHub `CheckRuns` and `Statuses` among the recent commits (~30). A CI-system is considered well-known if its name contains any of the following: appveyor, buildkite, circleci, e2e, github-actions, jenkins, mergeable, test, travis-ci. The check succeeds if at least 75% of successful pull requests have at least one successful check associated with them. 
+The check works by looking for a set of well-known CI-system names in GitHub `CheckRuns` and `Statuses` among the recent commits (~30). A CI-system is considered well-known if its name contains any of the following: appveyor, buildkite, circleci, e2e, github-actions, jenkins, mergeable, test, travis-ci. The check succeeds if at least 75% of successful pull requests have at least one successful check associated with them. A project may meet this criterion yet have a failing scorecard report; there are many ways to implement this criterion and it's especially difficult for an automated tool (like scorecard) to detect them all.
```
naveensrinivasan, line 38: Suggestion `A project may meet this criterion yet have a failing scorecard report; there are many ways to implement this criterion, and it's challenging for an automated tool (like scorecard) to detect them all. `

laurentsimon: 1. `The check succeeds if at least 75% of successful pull requests have at least one successful check associated with them.` please remove this sentence because our checks no longer fail or pass.  The score is calculated as number_reviews/reviewed_looked_at; and reivews_looked_at is currently 30.
2. The check only has false negatives and many can be removed by adding more tool names: please let's also drop a line saying that "if you think your system was not detected, open an issue at https://github.com/ossf/scorecard/issues/new/choose"

laurentsimon, line 38: maybe we should change `well-known` to `known` systems?

laurentsimon:
@olivekl FYI

docs/checks.md
david-a-wheeler: Will do.

david-a-wheeler: Neither `well-known` nor `known` are needed at all. Let's just remove the word entirely, the sentence is simpler and makes the intended point:

> The check works by looking for a set of CI-system names...

laurentsimon (approved):

PR #1026, opened 2021-09-16
🌱 Add counting of shell parsing errors

chrismcgehee:

Metrics

As per https://github.com/ossf/scorecard/issues/839#issuecomment-920072037, I am adding a metric to count the number of times that shell parsing fails. Based on how prevalent this is, we can decide how we want to handle these errors.

No

* **Other information**:
I have not tested this code on a Scorecard Kubernetes cluster. Someone who has experience with the Scorecard cluster may want to test this PR. I'm open to learning how to test changes in `cron` before pushing live code, but I'd need some instruction in that.

### DIFF
--- a/stats/measures.go
+++ b/stats/measures.go
@@ -27,4 +27,10 @@ var (
 		stats.UnitSeconds)
 	// HTTPRequests measures the count of HTTP requests.
 	HTTPRequests = stats.Int64("HTTPRequests", "Measures the count of HTTP requests", stats.UnitDimensionless)
+	// ShellParseErrors measures the count of errors when attempting to parse shell code.
+	ShellParseErrors = stats.Int64(
+		"ShellParseErrors",
+		"Measures the count of errors when attempting to parse shell code",
+		stats.UnitDimensionless,
+	)
 )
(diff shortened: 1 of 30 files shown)


stats/measures.go
```diff
@@ -27,4 +27,10 @@ var (
 		stats.UnitSeconds)
 	// HTTPRequests measures the count of HTTP requests.
 	HTTPRequests = stats.Int64("HTTPRequests", "Measures the count of HTTP requests", stats.UnitDimensionless)
+	// ShellParseErrors measures the count of errors when attempting to parse shell code.
+	ShellParseErrors = stats.Int64(
```
azeemshaikh38, line 31: Can we use existing `CheckErrors` to record these? You'll need to add a new error here: https://github.com/ossf/scorecard/blob/main/errors/public.go#L23 and update the GetName fn - https://github.com/ossf/scorecard/blob/main/errors/public.go#L44

naveensrinivasan:
@chrismcgehee  🆒

stats/measures.go
chrismcgehee: I considered that at first, but I wasn't sure if it would skew existing error count metrics. I've gone ahead and updated this PR to use `CheckErrors` since you've given me the confidence to do so.

azeemshaikh38 (approved):
Thanks!

Issue #1040, opened 2021-09-20
Need an aggregate risk score, ideally configurable (need for 3.0)


inferno-chromium:
**Is your feature request related to a problem? Please describe.**
https://www.sonatype.com/resources/state-of-the-software-supply-chain-2021
"OpenSSF does not
provide a mechanism for distilling this “Scorecard”
into a single metric and so we did not include it in the
quantitative analysis we describe below."

**Describe the solution you'd like**
An aggregate risk score at the end. Ideally could be made configurable by allowing a param name, param weight input via command line to override default. E.g. check https://github.com/ossf/criticality_score

**Describe alternatives you've considered**
Leave as-is. Not great, every organization will need this to gauge overall risk of a library. 

**Additional context**
NA

PR #1046, opened 2021-09-21
✨ Add aggregated score

laurentsimon:

feature

see https://github.com/ossf/scorecard/issues/1040

see https://github.com/ossf/scorecard/issues/1040

no

* **Other information**:
close https://github.com/ossf/scorecard/issues/1040

I am hesitant to make this parameterized right now because it will add a command we need to support/maintain. Once we've decided how to use config files, we can tackle it.

Please comment on how weights should be set.

Note: I re-run `make generate-docs`, which picked up David's changes from the YAML file not yet committed to the .md.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -157,11 +157,11 @@ or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show
 		}
 		switch format {
 		case formatDefault:
-			score, e := repoResult.AggregatedScore(checkDocs)
+			score, e := repoResult.AggregateScore(checkDocs)
 			if e != nil {
-				log.Fatalf("AggregatedScore: %v", e)
+				log.Fatalf("AggregateScore: %v", e)
 			}
-			fmt.Fprintf(os.Stdout, "Aggregated score: %d / %d\n", score, checker.MaxResultScore)
+			fmt.Fprintf(os.Stdout, "Aggregate score: %d / %d\n", score, checker.MaxResultScore)
 			fmt.Printf("\nCheck scores:\n")
 			err = repoResult.AsString(showDetails, *logLevel, checkDocs, os.Stdout)
 		case formatSarif:
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -155,14 +155,9 @@ or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show
 		if e != nil {
 			log.Fatalf("cannot read yaml file: %v", err)
 		}
+
 		switch format {
 		case formatDefault:
-			score, e := repoResult.AggregateScore(checkDocs)
-			if e != nil {
-				log.Fatalf("AggregateScore: %v", e)
-			}
-			fmt.Fprintf(os.Stdout, "Aggregate score: %d / %d\n", score, checker.MaxResultScore)
-			fmt.Printf("\nCheck scores:\n")
 			err = repoResult.AsString(showDetails, *logLevel, checkDocs, os.Stdout)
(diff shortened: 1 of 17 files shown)


pkg/scorecard_result.go
```diff
@@ -80,6 +81,36 @@ func (r *ScorecardResult) AsCSV(showDetails bool, logLevel zapcore.Level,
 	return nil
 }
 
+// AggregatedScore returns an aggregate score out of checker.MaxResultScore.
+func (r *ScorecardResult) AggregatedScore(checkDocs docs.Doc) (int, error) {
+	// TODO: calculate the score and make it a field
+	// of ScorecardResult
+	weights := map[string]int{"Critical": 7, "High": 5, "Medium": 3, "Low": 1}
```
laurentsimon, line 88: that needs some comments :-)

cmd/root.go
```diff
@@ -157,6 +157,12 @@ or ./scorecard --{npm,pypi,rubgems}=<package_name> [--checks=check1,...] [--show
 		}
 		switch format {
 		case formatDefault:
+			score, e := repoResult.AggregatedScore(checkDocs)
```
inferno-chromium: Why are we only calculating this only for default format, we should add it everywhere. Does adding a field in scorecard result hard (like does it need migration), otherwise just add it. we would need some aggregate result everywehre.

pkg/scorecard_result.go
inferno-chromium, line 88: Should we try equal distribution
critical: 10
high: 7.5
medium: 5
low: 2.5
easier to explain in our terms of 1-10 scale ?

pkg/scorecard_result.go
```diff
@@ -80,6 +81,36 @@ func (r *ScorecardResult) AsCSV(showDetails bool, logLevel zapcore.Level,
...
+	// Note: aggregate score changes depending on which checks are run.
+	total := 0
```
inferno-chromium, line 90: do these need to be float like 0.0 for total and score ?

inferno-chromium (approved):

cmd/root.go
laurentsimon: I can do that, I was mostly hesitant to update the JSON format. Do you want to have this reflected in the BQ table too?

pkg/scorecard_result.go
laurentsimon: the current scorecard scores are integers so I made this one float for consistency. If we want more granularity, we can make the aggregate score a float.

pkg/scorecard_result.go
inferno-chromium: yes main thing i want is aggregate score to show in 1-10 range.

cmd/root.go
laurentsimon: ack. That'll come in a follow-up PR.

pkg/scorecard_result.go
laurentsimon: sure, that's the case.

pkg/scorecard_result.go
laurentsimon: moved to float which gives more granularity.

laurentsimon:
@olivekl can you update your readme PR with some info on the aggregate score?
1. Update the run example output
2. Give a short sentence about it. The weights for calculating the aggregate scores are defined in https://github.com/ossf/scorecard/pull/1046/files#diff-9359395c0ce37bd4ecafda90038687c68381446dd8ed1c93674d9401faa7e494R98
The risk information is taken from the yaml config file.
Note that we currently do not have checks with risk set to `Critical`.

PR #1050, opened 2021-09-22
✨ Add aggregate score to cron JSON

laurentsimon:

feature

see https://github.com/ossf/scorecard/issues/1040 and https://github.com/ossf/scorecard/pull/1046

see https://github.com/ossf/scorecard/issues/1040 and https://github.com/ossf/scorecard/pull/1046

no

* **Other information**:

### DIFF
--- a/cron/format/json.go
+++ b/cron/format/json.go
@@ -69,10 +69,21 @@ type jsonScorecardV2 struct {
 	Commit  string `json:"commit"`
 }
 
+type jsonFloatScore float64
+
+func (s jsonFloatScore) MarshalJSON() ([]byte, error) {
+	if float64(s) == float64(int(s)) {
+		return []byte(fmt.Sprintf("%.0f", s)), nil
+	}
+	return []byte(fmt.Sprintf("%.1f", s)), nil
+}
+
+//nolint:govet
 type jsonScorecardResultV2 struct {
 	Date      string              `json:"date"`
 	Repo      jsonRepoV2          `json:"repo"`
 	Scorecard jsonScorecardV2     `json:"scorecard"`
+	AggScore  jsonFloatScore      `json:"score"`
 	Checks    []jsonCheckResultV2 `json:"checks"`
 	Metadata  []string            `json:"metadata"`
 }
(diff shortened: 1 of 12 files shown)


laurentsimon (commented):
test

cron/format/json.go
```diff
@@ -69,10 +69,21 @@ type jsonScorecardV2 struct {
 	Commit  string `json:"commit"`
 }
 
+type jsonFloatScore float64
+
+func (s jsonFloatScore) MarshalJSON() ([]byte, error) {
```
inferno-chromium, line 74: Might skip this, just X.0 is fine for even ints.

cron/format/json.go
```diff
@@ -128,6 +145,7 @@ func AsJSON2(r *pkg.ScorecardResult, showDetails bool,
 		},
 		Date:     r.Date.Format("2006-01-02"),
 		Metadata: r.Metadata,
+		AggScore: jsonFloatScore(score),
```
inferno-chromium, line 148: AggregateScore in full might be better to avoid confusion.

inferno-chromium (approved):

cron/format/json.go
oliverchang: +1. Given how floats are represented I'm unsure this will work reliability in every case either.

oliverchang (approved):

cron/format/json.go
laurentsimon: the BQ has a type `float`, and JSON sets this field with type `number`. (Note: score are 1-10)

inferno-chromium: why not keep everything float ?

cron/format/json.go
laurentsimon: sure. Note this is not the JSON field, but jut the structure field. The JSON field is currently called just `score`. Is that ok?

cron/format/json.go
laurentsimon: JSON does not understand the distinction between `float` and `int` in general. The schema we export (if users use it) has the field typed as `number` and not an integer.. so it's not an integer. Removing the `.X` was mostly esthetic when a human being looks at it. I'm fine to keep all `X.Y`... it may even be better: I can imagine users make it an `int` by mistake and only later realize it should be a `float`.

inferno-chromium: Yes that is what i meant, keep everything float.

inferno-chromium (approved):
Thanks for changes.

PR #1069, opened 2021-09-27
✨ Define scorecard GitHub action

laurentsimon:

feature

no GH action

GH action

no

* **Other information**:
Currently only supports push/cron events.


actions/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Use scorecard's official Docker image as base.
+ARG DOCKER_SHA
+FROM gcr.io/openssf/scorecard@sha256:${DOCKER_SHA} as base
+
+# Build our image and update the root certs.
+FROM debian:9.5-slim
```
naveensrinivasan, line 20: Why not build from distroless?

naveensrinivasan (approved):
This is cool! Thanks.

actions/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Copy the scorecard binary from the official scorecard image.
+COPY --from=base /scorecard /scorecard
+
+# Copy a test policy for local testing.
+COPY actions/testdata/policy.yml .
+
+# Our entry point.
+ADD actions/entrypoint.sh entrypoint.sh
+RUN chmod +x entrypoint.sh
```
azeemshaikh38, line 32: You shouldn't need this. It copies over the file permissions from the repository. So just commit the file with right permissions.

actions/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ENTRYPOINT ["/entrypoint.sh"]
```
azeemshaikh38, line 33: Can't the `ENTRYPOINT` itself be `scorecard --repo=<> ...`?

actions/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# Build our image and update the root certs.
+FROM debian:9.5-slim
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates
+
+# Copy the scorecard binary from the official scorecard image.
+COPY --from=base /scorecard /scorecard
+
+# Copy a test policy for local testing.
+COPY actions/testdata/policy.yml .
```
azeemshaikh38, line 28: Nit: since we deploy this policy file along with the Docker image, let's put it in a different folder, say `policies`, instead of `testdata` folder.

actions/Dockerfile
azeemshaikh38: +1

laurentsimon: AFAIK, we cannot use a shell in distroless. The entry point is a shell. For distrless, we'll need to new build `scorecard-action` as a golang app. If you don't mind, I'll create an issue for tracking and we can do that in a follow-up PR once everything else works.
Please let me know if you disagree.

actions/Dockerfile
laurentsimon: currently it's copied in the root directory. I will move to `policies`. Thanks for the suggestion

actions/Dockerfile
laurentsimon: So I think it's better to have a wrapper, se https://github.com/ossf/scorecard/pull/1069#discussion_r716982443
Long-term we can create a golang app to avoid using a shell and then we'll be able to move to distroless. Can do that in follow-up PR once the action flow works. I'll create an issue for tracking. 
Let me know if you disagree.

naveensrinivasan: Works for me. I think we shouldn’t use the Docker container for GitHub action. As part of that we can solve this issue. Thanks

azeemshaikh38: Could you add an action metadata file in this PR - https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action#creating-an-action-metadata-file?

I'm ok with changing the internals in follow-up PRs, if we can define the API that will be exposed to users in this PR.

actions/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+# https://docs.github.com/en/actions/learn-github-actions/environment-variables
+# GITHUB_EVENT_PATH contains the json file for the event.
+# GITHUB_SHA contains the commit hash.
+# GITHUB_WORKSPACE contains the repo folder.
+# GITHUB_EVENT_NAME contains the event name.
+
+export GITHUB_AUTH_TOKEN="$INPUT_REPO_TOKEN"
+export SCORECARD_V3=1
+export SCORECARD_POLICY_FILE="$INPUT_POLICY_FILE"
+export SCORECARD_SARIF_FILE="$INPUT_SARIF_FILE"
```
azeemshaikh38, line 25: Could you explain why users need to provide a SARIF file location? Isn't this something internal that will be used by GitHub and need not be exposed to users?

laurentsimon: see https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github
We can make it internal or let users decide. Note that SARIF option is public (`--format sarif`).
 
ShiftLeftSecurity does the same, see https://github.com/ShiftLeftSecurity/scan-action - I think they do it this way because it allows them to print results in a different format for users, which we could possibly too.

@naveensrinivasan any opinion on this?

actions/Dockerfile
laurentsimon: this is the API:
```
uses: docker://scorecard-action
        with:
          policy_file: .github/scorecard/policy.yml
          sarif_file: results.sarif
          repo_token: ${{ secrets.GITHUB_TOKEN }}
```

The `action.yml` is not used unless it defines a `composite` action, AFAIK. I have it defined but did not upload it because I found it nt to be necessary. Shall I upload it in this PR?

laurentsimon: > Works for me. I think we shouldn’t use the Docker container for GitHub action. As part of that we can solve this issue. Thanks

You mean https://docs.github.com/en/actions/creating-actions/creating-a-composite-action?
Can do that, yes.

actions/entrypoint.sh
azeemshaikh38: Right. But there isn't a need to ask the user to supply us with the SARIF file path. That can be internal. ShiftLeft seems to be doing the same thing.

actions/Dockerfile
laurentsimon: create https://github.com/ossf/scorecard/issues/1072 for tracking

azeemshaikh38: I wasn't suggesting a `composite action`. I just mean let's add the `action.yml` file even though it won't be used since it clarifies the API.

azeemshaikh38:
LGTM to submit this PR. Only open issue is exposing SARIF filepath to users.

actions/testdata/policy.yml
```diff
@@ -0,0 +1,50 @@
+version: 1
```
naveensrinivasan, line 1: This file needs a License header

actions/entrypoint.sh
laurentsimon: sorry, shiftleft was the wrong example, I mis-read upload-artifact for upload-sarif. It's up to us really. GH lets users do it with their built-in action in the workflow, as explained https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github
Nothing stopping us from doing inside our action. I just picked the simplest solution for the first iteration.

Issue #1079, opened 2021-09-28
BUG - Scorecard GitHub action is failing


naveensrinivasan:
**Describe the bug**

https://github.com/ossf/scorecard/runs/3734088142?check_suite_focus=true

laurentsimon:
https://github.com/ossf/scorecard/runs/3734678406 is failing

naveensrinivasan:
https://github.com/ossf/scorecard/runs/3749377785?check_suite_focus=true

naveensrinivasan:
@laurentsimon it is failing

laurentsimon:
I must have forgotten to push a new docker image. I pushed a new one.

naveensrinivasan:
Still failing https://github.com/ossf/scorecard/runs/3750175726?check_suite_focus=true

laurentsimon:
fixed https://github.com/ossf/scorecard/runs/3750344175?check_suite_focus=true

PR #1081, opened 2021-09-28
🐛 Fix failing GH action

laurentsimon:

bug

https://github.com/ossf/scorecard/issues/1079

https://github.com/ossf/scorecard/issues/1079

no

* **Other information**:
close https://github.com/ossf/scorecard/issues/1079

### DIFF
--- a/pkg/sarif.go
+++ b/pkg/sarif.go
@@ -149,34 +149,50 @@ type sarif210 struct {
 	Runs    []run  `json:"runs"`
 }
 
+func maxOffset(x, y int) int {
+	if x < y {
+		return y
+	}
+	return x
+}
+
 func detailToRegion(details *checker.CheckDetail) region {
 	var reg region
 	var snippet *text
 	if details.Msg.Snippet != "" {
 		snippet = &text{Text: details.Msg.Snippet}
 	}
+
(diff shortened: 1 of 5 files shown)


pkg/sarif.go
```diff
@@ -149,33 +149,49 @@ type sarif210 struct {
...
+		line := maxOffset(1, details.Msg.Offset)
 		reg = region{
-			StartLine: &details.Msg.Offset,
+			StartLine: &line,
 			Snippet:   snippet,
 		}
 	case checker.FileTypeText:
+		// Offset of 0 is acceptable here.
 		reg = region{
 			CharOffset: &details.Msg.Offset,
```
azeemshaikh38, line 190: To be safe, should we specify `maxOffset(0, Offset)`? In case `Offset` is < 0.

azeemshaikh38 (approved):

pkg/sarif.go
laurentsimon: `<0` means a bug in our checks code. If I do that it will go unnoticed. Shall I add a `panic` instead?

azeemshaikh38: Yeah, or atleast a log. Looks like right now, we silently fail which is not good.

PR #1083, opened 2021-09-28
:seedling: Reproducible go builds

naveensrinivasan:

  Feature

  Builds aren't reproducible 

 Builds are reproducible

 Builds are reproducible

* **Other information**:

### DIFF
--- a/scripts/version-ldflags
+++ b/scripts/version-ldflags
@@ -19,7 +19,9 @@
 # version info into cmd/version.go via ldflags
 GIT_VERSION=$(git describe --tags --always --dirty)
 GIT_HASH=$(git rev-parse HEAD)
-BUILD_DATE=$(date +'%Y-%m-%dT%H:%M:%SZ')
+# https://reproducible-builds.org/docs/source-date-epoch/
+# https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-log.html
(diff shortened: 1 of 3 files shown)


naveensrinivasan:
@loosebazooka FYI

loosebazooka:
looks good to me.

naveensrinivasan:
> looks good to me.

Thanks!

scripts/version-ldflags
```diff
@@ -19,7 +19,9 @@
 # version info into cmd/version.go via ldflags
 GIT_VERSION=$(git describe --tags --always --dirty)
 GIT_HASH=$(git rev-parse HEAD)
-BUILD_DATE=$(date +'%Y-%m-%dT%H:%M:%SZ')
+# https://reproducible-builds.org/docs/source-date-epoch/
+# https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-log.html
+SOURCE_DATE_EPOCH=$(git log --date=iso8601-strict -1 --pretty=%ct)
```
loosebazooka, line 24: This may not matter, but I wonder if this date should only be set if the tree state is "clean", if the tree state is "dirty" we just set current time?

naveensrinivasan: Sorry, I mean to answer and hit the resolve button.

I agree this is a good suggestion. We could do this later.

laurentsimon (approved):

PR #1085, opened 2021-09-28
✨ Start support for action on PR

laurentsimon:

feature

no support for action on PR

prepare for PR code

no

* **Other information**:

### DIFF
--- a/actions/entrypoint.sh
+++ b/actions/entrypoint.sh
@@ -34,8 +34,7 @@ if [[ "$GITHUB_EVENT_NAME" != "pull_request"* ]] && ! [[ "$GITHUB_REF" =~ ^refs/
     echo "Only the default branch is supported"
     exit 1
 fi
-echo "we're good"
-exit 0
+
 # It's important to change directories here, to ensure
 # the files in SARIF start at the source of the repo.
 # This allows GitHub to highlight the file.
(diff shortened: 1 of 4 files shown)


actions/entrypoint.sh
```diff
@@ -20,12 +20,19 @@ set -euo pipefail
...
 # GITHUB_WORKSPACE contains the repo folder.
 # GITHUB_EVENT_NAME contains the event name.
+# GITHUB_ACTIONS is true in GitHub env.
 
 export GITHUB_AUTH_TOKEN="$INPUT_REPO_TOKEN"
 export SCORECARD_V3=1
 export SCORECARD_POLICY_FILE="$INPUT_POLICY_FILE"
 export SCORECARD_SARIF_FILE="$INPUT_SARIF_FILE"
 
+if [[ "$GITHUB_EVENT_NAME" != "pull_request"* ]] && [[ "$GITHUB_REF" =~ ^refs/heads/(main|master)$ ]]; then
```
oliverchang, line 30: What is the "*" for after "pull_request" ?

actions/entrypoint.sh
```diff
@@ -20,12 +20,19 @@ set -euo pipefail
...
+    echo "$GITHUB_REF not supported with '$GITHUB_EVENT_NAME' event."
```
oliverchang, line 31: According to `scorecard-analysis.yml`, GITHUB_EVENT_NAME="push" and GITHUB_REF="refs/heads/main" is supported right? 

But this would fail this check.

oliverchang (commented):
nice!

actions/entrypoint.sh
laurentsimon: you're right. Thanks for catching this mistake :-)
I updated. PTAL, I don't trust myself in the evening!

oliverchang (approved):
LGTM!

actions/entrypoint.sh
laurentsimon: it means pull_request + something, so it would match `pull_request_target`, for example.

PR #1089, opened 2021-09-29
🐛  Fix failing action

laurentsimon:

bug

https://github.com/ossf/scorecard/runs/3750175726?check_suite_focus=true

https://github.com/ossf/scorecard/runs/3750175726?check_suite_focus=true

no

* **Other information**:

### DIFF
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -18,8 +18,6 @@ jobs:
     
     steps:
       - name: "Checkout code"
-        # Scorecard automatically uses the default branch for non-pull-request events.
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard-analysis.yml
```diff
@@ -18,8 +18,6 @@ jobs:
     
     steps:
       - name: "Checkout code"
-        # Scorecard automatically uses the default branch for non-pull-request events.
```
laurentsimon, line 21: I tried to be smart and I was not! We use the policy file as input but it's not present unless we check out the repo first!

naveensrinivasan (approved):

PR #1119, opened 2021-10-07
⚠️  remove CSV support

laurentsimon:

CSV support for Fail/Pass

support CSV

no CSV support

yes

* **Other information**:

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -128,6 +128,15 @@ func getEnabledChecks(sp *spol.ScorecardPolicy, argsChecks []string) (checker.Ch
 	return enabledChecks, nil
 }
 
+func validateFormat(format string) bool {
+	switch format {
+	case "json", "sarif", "default":
+		return true
+	default:
+		return false
+	}
+}
+
 var rootCmd = &cobra.Command{
 	Use:   scorecardUse,
 	Short: scorecardShort,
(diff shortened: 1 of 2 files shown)


lehors:
@laurentsimon Is there a discussion somewhere providing the rationale for this change?
Thanks.

lehors:
But how did you assess that "nobody seemed to be using it at the time"?
For what it's worth we've started using scorecard against a list of projects and csv is a convenient way to collect the results in a single table, even though it can clearly be argued that the format has its limits. For that reason I started adding a csv output option when I found out that you used to have it and removed it. That's why I'm asking.
Thanks.

azeemshaikh38:
> But how did you assess that "nobody seemed to be using it at the time"? What are people using instead? For what it's worth we've started using scorecard against a list of projects and csv is a convenient way to collect the results in a single table, even though it can clearly be argued that the format has its limits. For that reason I started adding a csv output option when I found out that you used to have it and removed it. That's why I'm asking. Thanks.

@lehors as you have mentioned, we found the CSV format very limiting especially when combined with options like `--show-details`. This gets more aggravated as Scorecard moves towards supporting granular results (see #1245). The JSON format is better suited to capture the output here. We recommend using ND-JSON or similar to collect results in a single table as @laurentsimon has mentioned. Let us know if that doesn't work for you.

PR #1123, opened 2021-10-09
📖 Signing your git commits as recommendation when contributing to scorecard

naveensrinivasan:

Git commits aren't signed 

This will encourage people to sign their git commits. It is not enforced.
More  information https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2014-May/005877.html

None

* **Other information**:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,7 @@ project. This document describes the contribution guidelines for the project.
 * [Adding New Checks](#adding-new-checks)
 * [Updating Docs](#updating-docs)
 * [Choosing checks to run](#choosing-checks-to-run)
+* [Sign your git commits](#sign-your-git-commits)
 
 <!-- vim-markdown-toc -->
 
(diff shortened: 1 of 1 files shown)


naveensrinivasan:
> I'm not sure signing commits with PGP is the right thing to recommend. PGP is hard to use and is being phased out, e.g. in cosign. It's not clear what security PGP gives you today in GitHub, for example:
> 
> 1. we don't know where the key is stored by users and key management is hard


> 2. key discovery is not good in GitHub, AFAIK. Where do verifiers get the public key to verify signatures? What if a user changes their key? How do consumers learn about it? GitHub provides no commit history of previous keys and changes, does it?

> 3. For the reasons above, only GitHub seems to be able to validate these signatures: so signing commits seems to add no value besides trusting GitHub to properly label commits/PRs with the correct users
> 4. If an attacker manages to break into someone's account, they can change the signature and it will still verify in the GitHub UI. (2FA saves us here)
> 
> sigstore/cosign makes key management easier but I don't know how they distribute keys: TUF? But where do users get the TUF today? There's also WIP to use ssh keys to sign commits on GitHub.
> 
> @asraa can you comment?

I agree with most of them. 

I could set git config user.name `laurentsimon` and git config user.email `laurentsimon@google.com` and push commits as you. It will look like you were pushing the commits. 

how do you mitigate this till we have a solution from TUF/Cosign? Even if TUF/cosign comes up with a solution, how long is it going to take to get into GitHub so that when we merge the code we are sure it is that author who is doing it?

laurentsimon:
> > I'm not sure signing commits with PGP is the right thing to recommend. PGP is hard to use and is being phased out, e.g. in cosign. It's not clear what security PGP gives you today in GitHub, for example:
> > 
> > 1. we don't know where the key is stored by users and key management is hard
> 
> > 1. key discovery is not good in GitHub, AFAIK. Where do verifiers get the public key to verify signatures? What if a user changes their key? How do consumers learn about it? GitHub provides no commit history of previous keys and changes, does it?
> 
> > 1. For the reasons above, only GitHub seems to be able to validate these signatures: so signing commits seems to add no value besides trusting GitHub to properly label commits/PRs with the correct users
> > 2. If an attacker manages to break into someone's account, they can change the signature and it will still verify in the GitHub UI. (2FA saves us here)
> > 
> > sigstore/cosign makes key management easier but I don't know how they distribute keys: TUF? But where do users get the TUF today? There's also WIP to use ssh keys to sign commits on GitHub.
> > @asraa can you comment?
> 
> I agree with most of them.
> 
> I could set git config user.name `laurentsimon` and git config user.email `laurentsimon@google.com` and push commits as you. It will look like you were pushing the commits.

I agree this is a problem. I don't understand why GitHub sometimes display the email/name, and sometimes display the GitHub username (depending whether we look at logs, merged commits, etc). Do you know?

> how do you mitigate this till we have a solution from TUF/Cosign? Even if TUF/cosign comes up with a solution, how long is it going to take to get into GitHub so that when we merge the code we are sure it is that author who is doing it?

tough question! This need not be built into GitHub, though. The main problem to solve is publishing a root of trust for verifiers to download. Could TUF data be added to Rekor or a registry using OIDC flow? Users would be free to decide which OIDC providers they want to use: GitHub, Google, etc

I agree with your initial comment though, it may take some time. My only concern is asking users to spend resource on something we'll ask them to undo in 1 year from now. @asraa how far are we from a solution to this? Is it still in flux? What help is needed?

> On a similar note, Scorecard has `signed-releases` checks which are using similar keys to validate the release, and cosign would be a better option up until then we have to make sure the releases are signed.

agreed. To be honest, I don't think anyone verifies these signatures. Where do we publish our public keys? We don't document how to verify the images, do we?

Let's add this to the discussion for next meeting. I feel we're taking some decisions for things that need discussion with other WG within OSSF, maybe `Best practices for open source developers`?

q464pp:
計分卡包含的需求討論標籤

PR #1125, opened 2021-10-12
📖  Update Install command for version 3

olivekl:
Change v2@latest to v@latest in README.md

Docs update

Old version in install command

* **Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ To install Scorecards:
 
 1. Visit our latest [release page](https://github.com/ossf/scorecard/releases/latest) and download the correct binary for your operating system
 2. Extract the binary file 
-3. Add the binary to your `GOPATH/bin` directory (use `go env gopath` to identify your directory if necessary)
+3. Add the binary to your `GOPATH/bin` directory (use `go env GOPATH` to identify your directory if necessary)
 
 ### Authentication
 


azeemshaikh38:
> Looks like it by design [golang/go#44840](https://github.com/golang/go/issues/44840). My suggestion would be is to change to `go get github.com/ossf/scorecard/v3@latest`
> 
> ```
>  go get github.com/ossf/scorecard/v3@latest
> go: downloading github.com/ossf/scorecard/v3 v3.0.1
> go: downloading github.com/ossf/scorecard v1.2.0
> ```

Hmm, it still fails for me:

```
go get github.com/ossf/scorecard/v3@latest
go get: installing executables with 'go get' in module mode is deprecated.
        Use 'go install pkg@version' instead.
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.
```

naveensrinivasan:
> > Looks like it by design [golang/go#44840](https://github.com/golang/go/issues/44840). My suggestion would be is to change to `go get github.com/ossf/scorecard/v3@latest`
> > ```
> >  go get github.com/ossf/scorecard/v3@latest
> > go: downloading github.com/ossf/scorecard/v3 v3.0.1
> > go: downloading github.com/ossf/scorecard v1.2.0
> > ```
> 
> Hmm, it still fails for me:
> 
> ```
> go get github.com/ossf/scorecard/v3@latest
> go get: installing executables with 'go get' in module mode is deprecated.
>         Use 'go install pkg@version' instead.
>         For more information, see https://golang.org/doc/go-get-install-deprecation
>         or run 'go help get' or 'go help install'.
> ```

It is not a failure. It is a warning. It will be installed. You can `rm scorecard` and figure out the path of `which scorecard`. After `go get` scorecard will be there.

azeemshaikh38:
> It is not a failure. It is a warning. It will be installed. You can `rm scorecard` and figure out the path of `which scorecard`. After `go get` scorecard will be there.

Yeah that did work. I wonder if that's what we should be asking our users to do though. What's the reason for having the `replace` directives? Can we remove them altogether?

azeemshaikh38:
> > > The replace is for not having these OSV's.
> > 
> > 
> > Makes sense. In that case, we should keep the `replace` directives.
> > From [golang/go#40276](https://github.com/golang/go/issues/40276), my understanding of the issue is this - `go install` _compiles_ the package. Hence, `replace` directives are not honored since they can cause inconsistencies between builds. In case of Scorecard, this is actually true - the binary available to be downloaded from the [release page](https://github.com/ossf/scorecard/releases/) has the correct `version` and `commit` info. But the binary from `go install` does not have this information.
> > Here's my suggestion - how about we ask users to download the `scorecard` binary from the release page (similar to what `protoc` does). That way we can continue using `replace` and also guarantee consistency on our binaries. Wdyt @naveensrinivasan ?
> 
> I concur we should ask users to download the binary from GitHub releases rather than `go install`.
> 
> Thanks!

Thanks. @olivekl can you update the documentation here to reflect this? Basically, we want users to visit our [latest release](https://github.com/ossf/scorecard/releases/latest) page. From there, they will need to download and extract the right binary (Linux, Mac or Windows) and add this to their `GOPATH/bin`.

Issue #1129, opened 2021-10-13
Relax token permission check or not


laurentsimon:
When calculating the score for the token permission check, we reduce the score when certain dangerous permissions are used, see [permissions.go#L262](https://github.com/ossf/scorecard/blob/main/checks/permissions.go#L262)

An alternative implementation would be to only check whether the permissions are defined explicitly or no via `permissions:` keyword. There are pros and cons to both approaches.

Checking only for the presence of permission definition `permissions:` is simpler but does not catch overly-broad permissions (e.g. if a user declared all permissions are `write`).

Reducing score based on dangerous permissions is more granular but prone to false positive, and will likely never be 100% accurate, see https://github.com/ossf/scorecard/issues/1099.

naveensrinivasan:
@laurentsimon  IMO some detailed notes on what needs to be fixed would be helpful for a `good rist issue`

laurentsimon:
> @laurentsimon IMO some detailed notes on what needs to be fixed would be helpful for a `good first issue`, Thanks

I've updated the description. I meant to tag this as `needs discussion`, not `good first issue`. I've updated the label too.

laurentsimon:
We discussed last meeting. We will separate out data/result gathering and policy/score calculation. So we'll keep our calculation as it is today and move to the policy calculation over time.

PR #1130, opened 2021-10-13
:book: Update Token-Permissions requirement in checks.yaml

varunsh-coder:

Docs update - to change the requirement for the token permissions check. 

#1129

New logic is in the PR

It will cause the score to increase for some repos. 

* **Other information**:
The logic for this change is the following:
1. If a developer has set explicit permissions for the token either at the workflow (top-level) or for each of the jobs (run-level), and it is not set to write-all, then at least some amount of effort has gone into deciding what the permissions should be. 
2. Calculating the exact permissions that the token should have is a hard problem and depends on the various different commands and actions being used in the workflow. Therefore, it may be best to check if a best effort has been put into deciding the permissions, rather than checking if the permissions are the exact minimum needed for each job. Otherwise it may happen that the assessment tool (Scorecards) miscalculates the minimum permissions, and as a result the developer may get an unfair score. 

The proposed change is a draft, and I expect it to evolve based on discussions...

### DIFF
--- a/docs/checks.md
+++ b/docs/checks.md
@@ -488,19 +488,20 @@ Note: The check does not verify the signatures.
 
 Risk: `High` (vulnerable to malicious code additions)
 
-This check determines whether the project's automated workflows tokens are set
-to read-only by default. It is currently limited to repositories hosted on
+This check determines whether permissions for the project's automated workflows tokens are set
+explicitly and are not set to grant all permissions. It is currently limited to repositories hosted on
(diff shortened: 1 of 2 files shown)


laurentsimon:
@olivekl any comment?

laurentsimon:
shall we close this PR? The code calculation based on the top-level vs run-level permission will b merged in soon thru https://github.com/ossf/scorecard/pull/1356

q464pp:
可以拉取合併嗎 謝謝

Issue #426, opened 2021-05-10
New check: Check for dangerous code practices in github workflows


laurentsimon:
There are several examples of github token leaks via `pull_request_target` event. It'd be nice to check for it - possibly filtering out known acceptable github actions that use it after we have reviewed their code (and it's pinned by hash). This check goes in the same direction as #414, ie harden github worflows.

inferno-chromium:
Dangerous permissions
or something :)

laurentsimon:
Thinking of broadening this check to "Dangerous workflow coding patterns":
1. dangerous events: `pull_request_target`
2. use of variable in `run` scripts: this lead to injection, see [here](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#understanding-the-risk-of-script-injections). The use is not always vulnerable, e.g. if the value is not attacker's controlled (hash commit) or if there's proper check in the workflow that it's only run on [maintainer's code](https://github.com/ossf/scorecard/blob/main/.github/workflows/integration.yml#L28). We can filter out non-attacker's controlled values to start with.
3.  I'm starting scouring for more :-)

I've left out `curl | bash` because it's part of another check around dependency pinning #403

laurentsimon:
related link https://securitylab.github.com/advisories/GHSL-2020-328-GoogleCloudPlatform-microservices-demo-workflow/

JarLob:
Hi, the important prerequisite for the dangerous pattern of `pull_request_target` is the usage of untrusted data, most commonly an explicit checkout of `github.event.pull_request.head` as in the example above. There are many ways to checkout the pull request code, but I think you can reduce false positives by checking for the most common - checkout action `ref:` parameter. Yet this still doesn't mean it is vulnerable if the code from pull request is treated as data, for example built scripts or tests are not run. But it may be hard to detect it programmatically to completely eliminate FPs. From my experience if `pull_request_target` is used with checkout `ref:` 95% chance the code is vulnerable.

laurentsimon:
Thanks for the information, that's super useful. If someone in your team is interested in taking a stab at this, please let me know.

laurentsimon:
Added a few more bad things to the list after reading https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

laurentsimon:
part 3 https://securitylab.github.com/research/github-actions-building-blocks/

laurentsimon:
[...]

PR #1168, opened 2021-10-27
:sparkles: Add dangerous workflow check with untrusted code checkout pattern

asraa:

- [ x] Tests for the changes have been added (for bug fixes / features)
- [ x] PR title follows the guidelines defined in  https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process

Adds a new check `Dangerous-Workflow` that checks if a github workflow uses dangerous coding patterns. 
The first pattern implemented is a check for an untrusted code checkout. 

This is described in https://github.com/ossf/scorecard/issues/426

No

* **Other information**:


azeemshaikh38:
Given recent discussion about Scorecard data change affecting clients, let's disable this check from running by default in CLI and cron job for now. If clients specifically want to invoke this check, that's ok.

laurentsimon:
> Given recent discussion about Scorecard data change affecting clients, let's disable this check from running by default in CLI and cron job for now. If clients specifically want to invoke this check, that's ok.

@asraa use an env variable to guard the access to this check. You can update the code in `cmd/root.go`, function `getEnabledChecks()`

checks/dangerous_workflow.go
```diff
@@ -0,0 +1,262 @@
...
+	if !CheckFileContainsCommands(content, "#") {
+		return true, nil
+	}
+
+	var workflow map[interface{}]interface{}
+	err := yaml.Unmarshal(content, &workflow)
+	if err != nil {
+		return false,
+			sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("yaml.Unmarshal: %v", err))
+	}
```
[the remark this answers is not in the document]
azeemshaikh38: Let's not block this PR on the parser discussion. We can discuss the parser issue in our upcoming meet and update code accordingly in a separate PR.

azeemshaikh38:
> > Given recent discussion about Scorecard data change affecting clients, let's disable this check from running by default in CLI and cron job for now. If clients specifically want to invoke this check, that's ok.
> 
> @asraa use an env variable to guard the access to this check. You can update the code in `cmd/root.go`, function `getEnabledChecks()`

Just `delete(enabledchecks, checks.CheckDangerousWorkflow)` [here](https://github.com/ossf/scorecard/blob/83649a799ec4a7953a3180297311c4427edd0474/cmd/root.go#L120) should be enough  :) 
Same for cron job [here](https://github.com/ossf/scorecard/blob/83649a799ec4a7953a3180297311c4427edd0474/cron/worker/main.go#L196).

laurentsimon:
> > > Given recent discussion about Scorecard data change affecting clients, let's disable this check from running by default in CLI and cron job for now. If clients specifically want to invoke this check, that's ok.
> > 
> > 
> > @asraa use an env variable to guard the access to this check. You can update the code in `cmd/root.go`, function `getEnabledChecks()`
> 
> Just `delete(enabledchecks, checks.CheckDangerousWorkflow)` [here](https://github.com/ossf/scorecard/blob/83649a799ec4a7953a3180297311c4427edd0474/cmd/root.go#L120) should be enough :) 

I want to be able to use and test the local repo interface with all checks (without providing a long list of `--checks`), so I'd prefer an env variable. 
 
Same for cron job [here](https://github.com/ossf/scorecard/blob/83649a799ec4a7953a3180297311c4427edd0474/cron/worker/main.go#L196).

azeemshaikh38:
> I want to be able to use and test the local repo interface with all checks (without providing a long list of `--checks`), so I'd prefer an env variable.
> 
> Same for cron job [here](https://github.com/ossf/scorecard/blob/83649a799ec4a7953a3180297311c4427edd0474/cron/worker/main.go#L196).

If its only needed for testing, you can simply remove the line locally? ;)

checks/dangerous_workflow.go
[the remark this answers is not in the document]
chrismcgehee: Agreed. Let's merge this PR and later consolidate our parsing with the library Naveen's suggested.

asraa:
I went with the ENV var solution, but only because I imagine future checks might just want to add to the list? And it would centralize where the code could be removed before the next cut but let me know if I should change that

checks/dangerous_workflow.go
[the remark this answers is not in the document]
naveensrinivasan: I concur, Let's merge this PR. Thanks

laurentsimon:
@asraa heads up. https://github.com/ossf/scorecard/pull/1195 got merged. You're going to have to update the checks.yaml file and add a `repos` field with a list of interfaces that are supported by your check. See example [checks.yaml#L78](https://github.com/ossf/scorecard/blob/main/docs/checks/internal/checks.yaml#L78).

The doc generation will also validate that your check does not use APIs that are inconsistent with the `repos` field. If you find a problem with the validation, let me know. (Note: validation happens automatically as you re-generate the .md)

checks/dangerous_workflow.go
```diff
@@ -0,0 +1,262 @@
...
+	data := patternCbData{
+		workflowPattern: make(map[string]bool),
+	}
+	err := CheckFilesContent(".github/workflows/*", false,
+		c, validateGitHubActionWorkflowPatterns, &data)
+	return createResultForDangerousWorkflowPatterns(data, err)
+}
+
+// Check file content.
+func validateGitHubActionWorkflowPatterns(path string, content []byte, dl checker.DetailLogger,
```
chrismcgehee, line 54: Hey Asra, some of the repos we check have random shell or python files in the .github/workflows directory. It would be good to check if this is a yaml file like in [pinned_dependencies.go](https://github.com/ossf/scorecard/blob/6562cc1f4488c7a018b5c6b4e031f990058d95a2/checks/pinned_dependencies.go#L525-L527). I'm creating a PR to fix this same issue in Token-Permissions.

docs/checks/internal/checks.yaml
```diff
@@ -628,3 +628,29 @@ checks:
...
+      the PR checkout, PR authors may compromise the repository, for example, by using 
+      build scripts controlled by the author of the PR or reading token in memory.
+
+      This check does not detect whether untrusted code checkouts are used safely, for
+      example, only on pull request that have been assigned a label.
+
+      The highest score is awarded when all workflows avoid the dangerous code patterns.
+    remediation:
+      - >-
+        Avoid the dangerous workflow patterns. See this [advisory](https://github.com/justinsteven/advisories/blob/master/2021_github_actions_checkspelling_token_leak_via_advice_symlink.md)
```
chrismcgehee, line 654: A lot of the information on this advisory for the `check-spelling` Github Action won't be relevant to people needing to remediate their workflow. Can we link to https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ instead?

checks/dangerous_workflow.go
```diff
@@ -0,0 +1,262 @@
...
+		// Uses defaults if not defined.
+		with, ok := mstep["with"]
+		if !ok {
+			continue
+		}
+		mwith, ok := with.(map[string]interface{})
+		if !ok {
+			return sce.WithMessage(sce.ErrScorecardInternal, errInvalidGitHubWorkflow.Error())
+		}
+		// Check for reference. If not defined, uses default main branch.
```
chrismcgehee, line 194: I don't think this is accurate for PR workflows. From https://github.com/actions/checkout#usage:
```
# The branch, tag or SHA to checkout. When checking out the repository that
# triggered a workflow, this defaults to the reference or SHA for that event.
# Otherwise, uses the default branch.
ref: ''
```
Sounds to be that when a PR triggers the workflow, then the latest commit in that PR will be checked out.

laurentsimon: How about tweaking the comment to "In the context of a `pull_request_target` event, the absence of `ref` checks out the target branch of the pull request" or something to this effect...?

asraa: I'm trying to read up on this, but I believe actually it checks out the base of the PR which is trusted
(See this thread https://github.community/t/can-workflow-changes-be-used-with-pull-request-target/178626

and GITHUB_REF in https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target)

docs/checks/internal/checks.yaml
asraa: that makes sense, done!

checks/dangerous_workflow.go
asraa: done!

asraa:
thanks for the review @chrismcgehee!

checks/dangerous_workflow.go
laurentsimon: Good catch. Im curious then: what if the base PR is cloned from a non-Branch-Protected branch on the original repo? Is it considered trusted?
Just curious... if there were testing workflows in that branch that auto-merge, etc it could be dangerous too.. probably not a very common threat anyway?

asraa: Right -- I think that if the base branch wasn't branch protected that would be a problem as well. But I assume that you should be using BP check in conjunction

laurentsimon: I mean, main branch is protected, but branch used by the PR is not. Say:
- repo has branch `main` and `test`. `main` is protected, but `test` is not.
- attacker clone the repo and update `test` branch locally.
- attacker creates PR to merge its local cloned `test` into our `main` branch.

Is this situation possible? If so, is there a way to detect it?

asraa: > attacker creates PR to merge its local cloned test into our main branch.

But then the target branch would be the main branch -- so the code being checked out in pull_request_target, GITHUB_REF, would be the main, not the untrusted code.

laurentsimon: ah, so I mis-understood the link. My original understanding was correct then. Sorry for the noise, ignore this comment :-)

asraa:
ping? anything else to do here?

laurentsimon:
> ping? anything else to do here?

license check is failing. apart from that, I think we're good to merge. If anyone disagrees, please reply.

asraa:
oops, something funky with the merge -- one second will update

olivekl:
> > ping? anything else to do here?
> 
> license check is failing. apart from that, I think we're good to merge. If anyone disagrees, please reply.

I'd just had a couple comments above on the documentation description. We could also address them after the merge, though.

laurentsimon:
> > > ping? anything else to do here?
> > 
> > 
> > license check is failing. apart from that, I think we're good to merge. If anyone disagrees, please reply.
> 
> I'd just had a couple comments above on the documentation description. We could also address them after the merge, though.

I think there's time to add your suggestions in this PR, since the pre-submit tests are still failing.

asraa:
Yeah! Happy to address more comments -- merge issue fixed

laurentsimon:
did you update the cron job https://github.com/ossf/scorecard/blob/83649a799ec4a7953a3180297311c4427edd0474/cron/worker/main.go#L196?

asraa:
> did you update the cron job

Oops -- good catch, I had assumed that it would call the root command, but it doesn't. Updated

olivekl (commented):
Just a few comments!

docs/checks.md
```diff
@@ -522,3 +522,18 @@ possible.
 **Remediation steps**
 - Fix the vulnerabilities. The details of each vulnerability can be found on <https://osv.dev>.
 
+## Dangerous
```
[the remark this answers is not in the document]
asraa: Yay thanks!!

> "This check determines whether the project has dangerous code patterns. ADD: Some examples of these patterns are [X] and [Y]."

added

> Is there a second part to the check that I'm not seeing?

Yep, I've only added one so far but the plan is to add a series of checks (https://github.com/ossf/scorecard/issues/426#issuecomment-856939499). I tried to rephrase this with `The first code pattern checked`. So I think it makes sense to keep this more general.

>  Or is this test only applicable to GitHub (in which case, let's add that)?

The check will have a lot of GitHub Actions dependent style and relies on it's particular default behavior, so I added this clarification ` Determines if the project's GitHub Action workflows avoid dangerous patterns.`

docs/checks.md
```diff
@@ -241,17 +241,18 @@ participants.
...
+logging github context and secrets, or use of potentially untrusted inputs in scripts.
+
+The first code pattern checked is the misuse of potentially dangerous triggers. 
+This checks if a `pull_request_target` workflow trigger was used in conjunction 
+with an explicit pull request checkout. Workflows triggered with `pull_request_target`
+have write permission to the target repository and access to target repository 
+secrets. With the PR checkout, PR authors may compromise the repository, for 
+example, by using build scripts controlled by the author of the PR or reading 
+token in memory. This check does not detect whether untrusted code checkouts are 
+used safely, for example, only on pull request that have been assigned a label.
```
laurentsimon, line 255: this check is only supported on GitHub. @olivekl do we need to use `merge request` instead of `pull request` in this sentence?

laurentsimon:
I fixed the pre-submit error 
```
--- FAIL: TestSARIFOutput (0.00s)
    --- FAIL: TestSARIFOutput/check-8 (0.00s)
        sarif_test.go:800: check-8: invalid result: 1
```
in https://github.com/ossf/scorecard/pull/1233
Once you push the error should go away

docs/checks.md
olivekl: Since it's only on GitHub I think `pull request` is fine.

asraa:
ping? anything else to do here? 
(assuming I'll update this to use the github workflow parser next)

olivekl (approved):

laurentsimon:
> ping? anything else to do here? (assuming I'll update this to use the github workflow parser next)

nop, go ahead update with workflow parser. Unless it's a trivial change, do it in a follow-up PR.

azeemshaikh38:
> > ping? anything else to do here? (assuming I'll update this to use the github workflow parser next)
> 
> nop, go ahead update with workflow parser. Unless it's a trivial change, do it in a follow-up PR. Let me know which option you prefer.

+1

dota17:
> 
It is recommended that the Readme  of  the project be updated synchronously. Currently, there are 17 check items, whereas the Readme  has only 16 check items.
https://github.com/ossf/scorecard/blob/main/README.md

naveensrinivasan:
Thanks, we would appreciate if you can do PR to fix the issue.

Issue #1184, opened 2021-10-28
Feature: Add supported repo interface in checks.yaml


laurentsimon:
As part of the local repo interface implementation, some checks are not supported.
We want to add this as part of the `checks.yaml` file, and use it to disable some test at runtime.

See this discussion thread https://github.com/ossf/scorecard/pull/1146#discussion_r736761403 for more information

PR #1195, opened 2021-10-29
✨ Use checks.yaml to store which repo types are supported by each check

laurentsimon:

This is a feature with no breaking changes. 

close https://github.com/ossf/scorecard/issues/1184

The general behavior for check selection is as follows:
1. `--repo=X` => uses all checks supported by the repo interface
2. `--repo=X --check=Y` => uses check Y and verifies it's supported by repo interface
3. `--repo=X --policy=Z` => uses all checks defined in the policy file AND supported by repo interface
4. `--repo=X --policy=Z --checks=Y` => uses Y AND verifies Y is defined in the policy and verifies Y checks are supported by the repo interface

The one thing I don't like is that, if a user passes a policy, we're silently discarding checks not supported by the interface. I wonder if this may be confusing for users.

### DIFF
--- a/docs/checks/internal/checks.yaml
+++ b/docs/checks/internal/checks.yaml
@@ -18,6 +18,7 @@ checks:
   Maintained:
     risk: High
     tags: supply-chain, security
+    repos: GitHub
     short: Determines if the project is "actively maintained".
     description: |
       Risk: `High` (possibly unpatched vulnerabilities)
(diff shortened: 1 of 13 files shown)


docs/checks/internal/checks.yaml
```diff
@@ -18,6 +18,7 @@ checks:
   Maintained:
     risk: High
     tags: supply-chain, security
+    repos: GitHub
```
azeemshaikh38, line 21: Maybe simpler to add a boolean like `run-on-local-files`? That way `checks.yaml` doesn't have to be updated with newer repo implementations.

azeemshaikh38 (approved):
Thanks!

docs/checks/internal/checks.yaml
laurentsimon: If you don't mind, I'll keep it as-is because it lends itself well to supporting other interfaces: gitlab, local-git, etc. I don't think we'll be able to support all of them with a binary "run on local file or not".

PR #1210, opened 2021-11-02
✨ Validate check.yaml's repo interface support

laurentsimon:

This PR adds validation to ensure the `repos` field in the checks.yaml file is consistent with the check's implementation. It currently uses regex to look for repo client APIs, but later we could use a callgraph if needed.

No breaking change in this PR.

### DIFF
--- a/docs/checks/internal/validate/main.go
+++ b/docs/checks/internal/validate/main.go
@@ -129,36 +129,75 @@ func isSubsetOf(a, b []string) bool {
 	return true
 }
 
-func validateRepoTypeAPIs(checkName string, repoTypes []string, checkFiles map[string]string) error {
-	// For now, we only list APIs in a check's implementation.
-	// Long-term, we should use the callgraph feature using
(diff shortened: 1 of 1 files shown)


azeemshaikh38 (approved):

docs/checks/internal/validate/main.go
```diff
@@ -15,67 +15,266 @@ package main
...
+		}
+
+		res := regex.FindStringSubmatch(string(content))
+		if len(res) != 2 {
+			continue
+		}
+
+		r := res[1]
+		if entry, exists := checkFiles[r]; exists {
+			//nolint:goerr113
```
naveensrinivasan, line 85: Can we avoid this? Not a blocker. Just wanted to call out.

naveensrinivasan (approved):
Thanks!

docs/checks/internal/validate/main.go
laurentsimon: need a pass to remove all of them, right. Fortunately this is not part of an API we expose.

PR #1211, opened 2021-11-02
✨ Add `--local` option to CLI

laurentsimon:

This add `--local` option as argument. No breaking changes.

### DIFF
--- a/clients/localdir/client.go
+++ b/clients/localdir/client.go
@@ -110,7 +110,7 @@ func listFiles(clientPath string) ([]string, error) {
 	return files, nil
 }
 
-func applyPredicate(clientPath string, clientFiles []string,
+func applyPredicate(clientFiles []string,
 	errFiles error, predicate func(string) (bool, error)) ([]string, error) {
 	if errFiles != nil {
 		return nil, errFiles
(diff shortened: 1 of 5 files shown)


clients/localdir/client.go
```diff
@@ -116,10 +136,9 @@ func listFiles(clientPath string, predicate func(string) (bool, error)) ([]strin
 // ListFiles implements RepoClient.ListFiles.
 func (client *localDirClient) ListFiles(predicate func(string) (bool, error)) ([]string, error) {
 	client.once.Do(func() {
-		client.files, client.errFiles = listFiles(client.path, predicate)
+		client.files, client.errFiles = listFiles(client.path)
 	})
-
-	return client.files, client.errFiles
+	return applyPredicate(client.path, client.files, client.errFiles, predicate)
```
azeemshaikh38, line 141: Consider making this a struct fn so you don't have to pass all the `client.*` objects.

clients/localdir/client.go
```diff
@@ -92,22 +98,36 @@ func listFiles(clientPath string, predicate func(string) (bool, error)) ([]strin
...
+func applyPredicate(clientPath string, clientFiles []string,
+	errFiles error, predicate func(string) (bool, error)) ([]string, error) {
+	if errFiles != nil {
+		return nil, errFiles
+	}
+
+	files := []string{}
+	for _, pathfn := range clientFiles {
+		// Remove prefix of the folder.
+		p := trimPrefix(pathfn, clientPath)
```
azeemshaikh38, line 122: Save the trimmed path to `client.files` to avoid running `trimPrefix` everytime?

azeemshaikh38 (approved):

clients/localdir/client.go
laurentsimon: just realized the call was only needed once in the `do.Once`. My bad. So I removed the call entirely here.

clients/localdir/client.go
laurentsimon: I removed `client.path` as it was not needed, so it takes only 3 arguments.

PR #1212, opened 2021-11-02
🐛  Use only olivekl@ in CODEOWNER

laurentsimon:
We recently changed CODEOWNER file, but it now automatically adds several people as reviewers for all PRs. This is overwhelming.
Let's only use this for the documentation and workflows.


naveensrinivasan (approved):
Thanks!

PR #1214, opened 2021-11-03
:seedling: Move docker build checks to ko

naveensrinivasan:

    Move the docker builds checks to ko. Part of moving to https://github.com/ossf/scorecard/issues/744

 None

* **Other information**:


.github/workflows/integration.yml
```diff
@@ -89,22 +89,6 @@ jobs:
...
-
-      - name: Test docker image
-        env:
-          GITHUB_AUTH_TOKEN: ${{ secrets.GH_AUTH_TOKEN }}
-        run: |
-              echo "Skipping e2e tests. Will be re-enabled soon."
-              #mkdir $OUTPUT_PATH
-              #docker run -e GITHUB_AUTH_TOKEN=$GITHUB_AUTH_TOKEN  $IMAGE_NAME  --show-details  --repo=https://github.com/ossf/scorecard --metadata=openssf --format json > ./$OUTPUT_PATH/results.json
-              #ginkgo -p -v --focus="E2E TEST:executable" ./e2e/...
-
```
azeemshaikh38, line 107: Why are we removing this? Is it because `make build` will run during the `Build` step of the GitHub workflow?

Makefile
```diff
@@ -10,6 +10,7 @@ PROTOC := $(shell which protoc)
 IMAGE_NAME = scorecard
 OUTPUT = output
 IGNORED_CI_TEST="E2E TEST:blob|E2E TEST:executable"
+PLATFORM="linux/amd64,linux/arm64,linux/386,linux/arm"
```
azeemshaikh38, line 13: What about `windows` and `darwin`?

Makefile
```diff
@@ -95,7 +96,7 @@ tree-status: ## Verify tree is clean and all changes are committed
 build-cron: build-pubsub build-bq-transfer build-github-server build-webhook build-add-script \
 	  build-validate-script build-update-script
 
-build-targets = generate-docs build-proto build-scorecard build-cron dockerbuild
+build-targets = generate-docs build-proto build-scorecard build-cron ko-build-scorecard ko-build-batch-controller ko-build-batch-worker ko-build-bq-transfer ko-build-webhook ko-build-github-server
```
azeemshaikh38, line 99: Create a single `ko-build` target that builds the `ko-build-*` targets.

Makefile
```diff
@@ -95,7 +96,7 @@ tree-status: ## Verify tree is clean and all changes are committed
 build-cron: build-pubsub build-bq-transfer build-github-server build-webhook build-add-script \
 	  build-validate-script build-update-script
 
-build-targets = generate-docs build-proto build-scorecard build-cron dockerbuild
```
azeemshaikh38, line 98: Why are we removing `dockerbuild`? Don't we want both `dockerbuild` and `ko-build`?

azeemshaikh38 (changes requested):

.github/workflows/integration.yml
naveensrinivasan: It is already there in the other action. So this is a duplicate.

Makefile
naveensrinivasan: There is no docker containers for windows/darwin it is only for Linux.

Makefile
naveensrinivasan: Why do we want dockerbuild and ko build? The progression is to replace dockerfiles with `ko`

Makefile
naveensrinivasan: With this approach, I can run individual builds and it makes it easier to build. Also, the plan is to have a few more jobs in  GH actions for each one of these build targets so that it finishes sooner.

.github/workflows/integration.yml
naveensrinivasan: https://github.com/ossf/scorecard/blob/6562cc1f4488c7a018b5c6b4e031f990058d95a2/.github/workflows/main.yml#L41 

The `make all` has the build step.

Makefile
azeemshaikh38: Before we remove `dockerbuild`, have we confirmed that the images generated by `ko` vs. `docker` don't diverge? If you have tested this locally somehow, do paste the results in the PR description.

Makefile
azeemshaikh38: What I mean is to create a phony target for all these individual targets for better readability. Basically:

```
ko-build: ko-build-scorecard ko-build-controller ...
```

And then use this phony target in `build-target`.

Makefile
azeemshaikh38: To clarify what I mean, `ko-build-batch-controller` for example, does not include the data files `cron/data/projects.*csv` which are required for it to function correctly. There might be other such nuances and subtleties which we might miss by completely doing away with `dockerbuild` without thorough testing.

We can submit this PR as is, but let's keep the `dockerbuild` target.

Makefile
azeemshaikh38: I'm updating this comment based on our recent offline conversation: having multiple `ko-build-*` targets specified as `build-target` is redundant and does not improve readability. To help us maintain quality of our code base, please consider updating this line of code to the suggestion above.

azeemshaikh38 (changes requested):

Makefile
naveensrinivasan: I am going to accept this. But this approach has a drawback. Our GitHub actions are slow because they all use a single Job. With all these in a single job, it goes sequential but if they were an independent target it would be running parallel which will speed up GH runs.

Makefile
naveensrinivasan: Agree, it is a valid argument. I will also include the docker build.

Makefile
```diff
@@ -95,7 +96,7 @@ tree-status: ## Verify tree is clean and all changes are committed
...
+build-targets = generate-docs build-proto build-scorecard build-cron ko-build-binaries dockerbuild
```
azeemshaikh38, line 99: Does this need to be `ko-build-everything`?

azeemshaikh38 (approved):
Thanks @naveensrinivasan !

Makefile
azeemshaikh38: Thanks for updating the code. We can still get parallelism using `make -j`. Using the example above:

```
ko-build: ko-build-scorecard ko-build-controller ...
```

in `.yml` workflow file:

```
make -j <num_parallel_jobs> ko-build
```
 
We can do this in a separate PR. Can submit this as is for now.

Makefile
naveensrinivasan: Thanks, @azeemshaikh38! 🤦

Issue #1072, opened 2021-09-27
Remove docker container from GitHub action


laurentsimon:
see https://github.com/ossf/scorecard/pull/1069#discussion_r716919804

https://docs.github.com/en/actions/creating-actions/creating-a-composite-action
https://dev.to/github/build-your-own-github-action-without-a-docker-container-1eic

PR #1222, opened 2021-11-05
✨  Pull request support for GitHub action

laurentsimon:
This adds support for pull request in the GH action.
No breaking changes.

The changes required updating how we generate our SARIF results.

This https://github.com/github/codeql-action/issues/811 is an improvement that would be nice, but non-blocking.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -154,9 +154,10 @@ func getEnabledChecks(sp *spol.ScorecardPolicy, argsChecks []string,
 		// Populate checks to run with policy file.
 		for checkName := range sp.GetPolicies() {
 			if !isSupportedCheck(supportedChecks, checkName) {
-				return enabledChecks,
-					sce.WithMessage(sce.ErrScorecardInternal,
-						fmt.Sprintf("repo type %s: unsupported check: %s", repoType, checkName))
+				// We silently ignore the check, like we do
+				// for the default case when no argsChecks
+				// are given by the user.
+				continue
 			}
 
 			if !enableCheck(checkName, &enabledChecks) {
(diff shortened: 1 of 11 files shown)


actions/Dockerfile
```diff
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Testing: docker run -e INPUT_SARIF_FILE=results.sarif -e GITHUB_WORKSPACE=/ -e INPUT_POLICY_FILE="/policy.yml" -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN -e GITHUB_REPOSITORY="ossf/scorecard" laurentsimon/scorecard-action:latest
+# Testing: docker run -e GITHUB_EVENT_NAME=pull_request -e INPUT_SARIF_FILE=results.sarif -e GITHUB_WORKSPACE=/ -e INPUT_POLICY_FILE="/policy.yml" -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN -e GITHUB_REPOSITORY="ossf/scorecard" laurentsimon/scorecard-action:latest
```
naveensrinivasan, line 15: This is not part of this PR. But could we also please change the base image. This reduces the attack vector.

laurentsimon: Distroless images don't support shell. So to use distroless, we need to remove the shell script and create another golang project to call as a binary. Doable, but needs some work.
We also have this tracking issue https://github.com/ossf/scorecard/issues/1072

azeemshaikh38 (approved):

chrismcgehee (approved):

PR #1226, opened 2021-11-08
🌱 Using library to parse github workflows

chrismcgehee:

Refactor

We currently parse Github workflow files manually. This is error prone and increase our maintenance burden.
Part of https://github.com/ossf/scorecard/issues/839

We'll use a 3rd-party library [rhysd/actionlint](https://github.com/rhysd/actionlint) to parse Github workflow files. This has several advantages:
1. Should help prevent bugs where our parsing logic wasn't handling edge cases correctly
2. Enables us to delete a decent amount of code we were using for parsing
2. This will make it much easier to add line numbers to results concerning workflow files.

As part of this PR, I also moved the code for dealing with workflow files into its own package. This was done in its own commit in case you'd like to review it separately.

No

* **Other information**:
Once https://github.com/ossf/scorecard/pull/1168 is merged, I'll convert that code to use actionlint for parsing.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -150,3 +150,6 @@ linters-settings:
       - ptrToRefParam
       - typeUnparen
       - unnecessaryBlock
+  wrapcheck:
+    ignorePackageGlobs:
+      - github.com/ossf/scorecard/v3/checks/fileparser
--- a/checks/testdata/github-workflow-multiple-unpinned-uses.yaml
+++ b/checks/testdata/github-workflow-multiple-unpinned-uses.yaml
@@ -20,7 +20,5 @@ jobs:
     steps:
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v1
-        run: echo "unpinned dependency"
       - name: some name
-        run: echo "unpinned dependency"
         uses: docker/build-push-action@1.2.3
--- a/checks/fileparser/errors.go
+++ b/checks/fileparser/errors.go
@@ -1,0 +1,21 @@
+// Copyright 2021 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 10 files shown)


checks/testdata/github-workflow-multiple-unpinned-uses.yaml
```diff
@@ -20,7 +20,5 @@ jobs:
     steps:
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v1
-        run: echo "unpinned dependency"
       - name: some name
-        run: echo "unpinned dependency"
```
chrismcgehee, line 25: Having both `uses` and `run` in a step in not valid for a workflow. I tested this in a workflow file, and the action failed with:
image

laurentsimon (approved):
Thanks Chris!

.golangci.yml
```diff
@@ -150,3 +150,6 @@ linters-settings:
       - ptrToRefParam
       - typeUnparen
       - unnecessaryBlock
+  wrapcheck:
+    ignorePackageGlobs:
```
laurentsimon, line 154: I'm not very familiar with golangci: does this avoid linter errors in error returned that are not wrapped? If so, why do we need to disable it? Why can we not use `sce.WithMessage`?

chrismcgehee: > does this avoid linter errors if the error returned is not wrapped? 

Yes

> Why can we not use `sce.WithMessage`?

The functions within `fileparser` already returns errors that use `sce.WithMessage`, so it seemed redundant to wrap it again. Essentially, the reason I added this was because `fileparser` is more an internal package and `wrapcheck` is meant to flag _external_ package errors that are not wrapped.

azeemshaikh38 (approved):
Woohoo, this is awesome @chrismcgehee! Thanks :)

naveensrinivasan:
@chrismcgehee  Thank you! Is this good to be merged?

chrismcgehee:
> @chrismcgehee Thank you! Is this good to be merged?

Yep!

laurentsimon:
@chrismcgehee late coment: `GetShellForStep` is the only function called outside of the fileparser package. Can we make the other functions non-exported? Or is the intention to export these functions?

chrismcgehee:
@laurentsimon I made them exported because I could see them being useful outside the package in the future. If you prefer though, I can open an PR to make them non-exported.

PR #1232, opened 2021-11-10
:sparkles:Add LGTM to the SAST check

evverx:
According to https://github.com/apps/lgtm-com
"LGTM is a code analysis platform for identifying vulnerabilities early and preventing
them from reaching production". It's used by `systemd`, `lxc` and a lot of other large
open source projects. The check is
still kind of broken in the sense that it fails to detect
projects where every PR is analyzed by LGTM before getting merged
but it's better than nothing I guess.

PR #1233, opened 2021-11-10
🐛  Make sarif generation deterministic

laurentsimon:
This makes the generation of sarif results deterministic: it will always use the same order for run categories and details.
No breaking changes

Found in https://github.com/ossf/scorecard/pull/1232#issuecomment-964754088


naveensrinivasan (approved):
Thanks!

PR #1235, opened 2021-11-10
🐛 Validate doc on pre-submit

laurentsimon:
We currently do not validate documentation in pre-submit checks. This is particularly important as the repo interfaces support only a specific number of checks that are defined via the `repos` field in the checks.yaml. The field is used to determined which checks are run, so a change in behavior needs to be detected.


naveensrinivasan (approved):
Thank you!

PR #1236, opened 2021-11-10
✨ Remove isScorecardRepo

laurentsimon:
This PR removes the `isScorecardRepo()` function from the repo interface.
it was used to ignore test files if the under-analysis repo was scorecard's.
I've removed this to ensure we run the same analysis on scorecard as we do for other repos.

I've kept the `testdata` folder as one we ignore in the analysis, to reduce the amount of false positives (see https://github.com/ossf/scorecard/pull/1235/checks?check_run_id=4168036778: when running on local folders, we always set `isScorecardRepo: false` which explains the number of errors reported). We should re-visit this implementation once we have separated data gathering from score calculation. The score calculation (which can be thought of as a policy) will be able to take an `ignore-path` field to ignore path; whilst keeping the underlying results canonical.

No breaking changes.


naveensrinivasan (approved):
Thanks!

azeemshaikh38 (commented):
This is awesome, thanks!

PR #1239, opened 2021-11-10
:bug:Security-Policy: really look for the security policy

evverx:
It was tested with [the systemd project](https://github.com/systemd/systemd/security) where the security policy
is kept in [docs/SECURITY.md](https://github.com/systemd/systemd/blob/main/docs/SECURITY.md). Without this patch `scorecard`
says that the security policy can't be found.


naveensrinivasan (approved):
Thank you!

Issue #1231, opened 2021-11-09
The "Fuzzing" check doesn't seem to always work correctly


evverx:
**Describe the bug**

According to `scorecards` some projects like [`util-linux`](https://github.com/google/oss-fuzz/tree/master/projects/util-linux) and [`selinux`](https://github.com/google/oss-fuzz/blob/master/projects/selinux/project.yaml) aren't fuzzed on OSS-Fuzz even though they have been fuzzed there since 2020 as far as I can remember.

**Reproduction steps**
```
$ ./scorecard-linux-amd64 version
GitVersion:	v3.1.1
GitCommit:	fd238d0e40ebc898e4aa592ad133b1f15687384f
GitTreeState:	clean
BuildDate:	1635283925
GoVersion:	go1.17.2
Compiler:	gc
Platform:	linux/amd64
```
```
$ GITHUB_AUTH_TOKEN=*** ./scorecard-linux-amd64 --show-details  --checks=Fuzzing --repo=https://github.com/util-linux/util-linux
Starting [Fuzzing]
Finished [Fuzzing]

RESULTS
-------
Aggregate score: 0.0 / 10

Check scores:
|--------|---------|--------------------------------|---------|--------------------------------------------------------------------------------------------------------|
| SCORE  |  NAME   |             REASON             | DETAILS |                                       DOCUMENTATION/REMEDIATION                                        |
|--------|---------|--------------------------------|---------|--------------------------------------------------------------------------------------------------------|
| 0 / 10 | Fuzzing | project is not fuzzed in       |         | https://github.com/ossf/scorecard/blob/fd238d0e40ebc898e4aa592ad133b1f15687384f/docs/checks.md#fuzzing |
|        |         | OSS-Fuzz                       |         |                                                                                                        |
|--------|---------|--------------------------------|---------|--------------------------------------------------------------------------------------------------------|
```
```
$ GITHUB_AUTH_TOKEN=*** ./scorecard-linux-amd64 --show-details  --checks=Fuzzing --repo=https://github.com/SELinuxProject/selinux
Starting [Fuzzing]
Finished [Fuzzing]

RESULTS
-------
Aggregate score: 0.0 / 10

Check scores:
|--------|---------|--------------------------------|---------|--------------------------------------------------------------------------------------------------------|
| SCORE  |  NAME   |             REASON             | DETAILS |                                       DOCUMENTATION/REMEDIATION                                        |
[...]

PR #1241, opened 2021-11-11
:bug: fix special character in search query to fix fuzzing check

asraa:

* Fixes a bug where special character / was causeing a false negative on the fuzzing query that queries for repositories
By github search API docs:
```
You can't use the following wildcard characters as part of your search query: . , : ; / \ ` ' " = * ! ? # $ & + ^ | ~ < > ( ) { } [ ] @. The search will simply ignore these symbols.
```

https://docs.github.com/en/search-github/searching-on-github/searching-code#considerations-for-code-search

https://github.com/ossf/scorecard/issues/1231

Successfully detects fuzzing for the repos

* **Other information**:


asraa:
I don't think any of the special characters are relevant for repository searches, but if you think it makes sense to replace all of those, let me know!

clients/githubrepo/search.go
```diff
@@ -59,7 +59,7 @@ func (handler *searchHandler) buildQuery(request clients.SearchRequest) (string,
 	}
 	var queryBuilder strings.Builder
 	if _, err := queryBuilder.WriteString(
-		fmt.Sprintf("%s repo:%s/%s", request.Query, handler.owner, handler.repo)); err != nil {
+		fmt.Sprintf("%s repo:%s/%s", strings.ReplaceAll(request.Query, "/", " "), handler.owner, handler.repo)); err != nil {
```
laurentsimon, line 62: why was there a `/` in the query that caused the issue? I thought org and repo did not have if they had been extracted properly from `owner/repo`?

azeemshaikh38, line 62: Nit: Provide the GitHub link in comment - https://docs.github.com/en/search-github/searching-on-github/searching-code#considerations-for-code-search for reference.

go.mod
```diff
@@ -40,71 +38,6 @@ require (
 	gotest.tools v2.2.0+incompatible
 )
 
-require (
-	cloud.google.com/go v0.94.1 // indirect
```
laurentsimon, line 44: is it normal all of the below dependencies are removed?

go.mod
```diff
@@ -4,9 +4,7 @@ go 1.17
 
 require (
 	cloud.google.com/go/bigquery v1.24.0
-	cloud.google.com/go/monitoring v0.1.0 // indirect
```
azeemshaikh38, line 7: Are these `go.mod` changes expected?

laurentsimon (approved):

clients/githubrepo/search.go
azeemshaikh38: This *may* not fully solve the problem. We should consider maintaining a list of special characters from https://docs.github.com/en/search-github/searching-on-github/searching-code#considerations-for-code-search to replace with space.

azeemshaikh38 (approved):
Thanks for the quick turnaround @asraa !

azeemshaikh38:
> I don't think any of the special characters are relevant for repository searches, but if you think it makes sense to replace all of those, let me know!

+1. I think its ok for this PR. But maybe just leave a comment so we know this could be a potential issue.

clients/githubrepo/search.go
asraa: Ah, for fuzzing we check if the URI is in a `project.yaml`.

e.g. we search for `github.com/util-linux/util-linux` in the main_repo field of the YAML https://github.com/google/oss-fuzz/blob/56b840aad7907d69ad0de08a9339ba7e18a03514/projects/util-linux/project.yaml#L17

go.mod
asraa: hmmm somehow that showed up in my `go mod tidy`.

evverx:
Thanks! I can confirm that now both `util-linux` and `selinux` are fuzzed according to `scorecard`.

(on a somewhat unrelated note It seems the "Fuzzing" check doesn't take disabled projects (where `disabled` is set to `true`) like https://github.com/google/oss-fuzz/blob/master/projects/influxdb/project.yaml into account. It has nothing to with this PR though)

clients/githubrepo/search.go
asraa: It solves the problem for github URL since they can only have `-` and alphanums, but I suppose it's possible scorecard will later to more complex queries.

asraa:
> (on a somewhat unrelated note It seems the "Fuzzing" check doesn't take disabled projects (where disabled is set to true) like https://github.com/google/oss-fuzz/blob/master/projects/influxdb/project.yaml into account. It has nothing to with this PR though)

Ooh right! Should we open an issue about this? I suppose we could search for the right project.yaml and then check for a disabled line.

evverx:
@asraa I'll try to open a new issue once I figure out why `scorecard` says that `influxdb` is fuzzed while `libra` isn't even though they're both disabled on OSS-Fuzz.

evverx:
Looks like `libra` triggered another corner case. It was renamed to "diem" but the "project.yaml" file wasn't changed accordingly so `scorecard` somehow looks for `diem/diem` instead of `libra/libra`. Basically it seems `scorecard` can't handle redirects so if for example `karelzak/util-linux` is looked for it redirects to `util-linux/util-linux` and because https://github.com/google/oss-fuzz/pull/6800 was merged it automagically works but if the `project.yaml` file hadn't been changed `scorecard` would say that the project isn't fuzzed. I'll try to open another issue then

asraa:
ping? anything else to do here? other issues with the fuzzing check have open issues now

azeemshaikh38:
> ping? anything else to do here? other issues with the fuzzing check have open issues now

Nothing from my side. Looks good to submit.

laurentsimon:
> Looks like `libra` triggered another corner case. It was renamed to "diem" but the "project.yaml" file wasn't changed accordingly so `scorecard` somehow looks for `diem/diem` instead of `libra/libra`. Basically it seems `scorecard` can't handle redirects so if for example `karelzak/util-linux` is looked for it redirects to `util-linux/util-linux` and because [google/oss-fuzz#6800](https://github.com/google/oss-fuzz/pull/6800) was merged it automagically works but if the `project.yaml` file hadn't been changed `scorecard` would say that the project isn't fuzzed. I'll try to open another issue then

yes, please!

laurentsimon:
> > ping? anything else to do here? other issues with the fuzzing check have open issues now
> 
> Nothing from my side. Looks good to submit.

LGTM too. Thanks @asraa !

Issue #1242, opened 2021-11-11
BUG: `RepoClient` data format


azeemshaikh38:
Currently, the `RepoClient` data format uses primitive Golang types. This means that when certain values are not returned, the result struct will have default values set. This can lead to incorrect results. 

We should update the `RepoClient` data to use `*bool`, `*string` etc. And the convention will be that if the value is `non-nil`, only then analyze the resut.

evverx:
Could it be that this causes https://github.com/ossf/scorecard/issues/1247?

azeemshaikh38:
Most likely yes. I have a PR out to fix this for `Branch-Protection`: #1243. With this PR, this is the result I get: 

```
scorecard  --format json --show-details --verbosity Debug --checks Branch-Protection  --repo=https://github.com/evverx/systemd | jq
{
  "date": "2021-11-11",
  "repo": {
    "name": "github.com/evverx/systemd",
    "commit": "1b1e6249a0987820cde646a11c1b559dcb02d62f"
  },
  "scorecard": {
    "version": "3.0.1",
    "commit": "6c1c789dc5b05cde492334f57b53807c786b038a"
  },
  "score": 2,
  "checks": [
    {
      "details": [
        "Info: 'force pushes' disabled on branch 'main'",
        "Info: 'allow deletion' disabled on branch 'main'",
        "Warn: linear history disabled on branch 'main'",
        "Warn: status checks for merging disabled on branch 'main'",
        "Info: number of required reviewers is 2 on branch 'main'",
        "Warn: Stale review dismissal disabled on branch 'main'",
        "Warn: Owner review not required on branch 'main'",
        "Warn: 'admininistrator' PRs are exempt from reviews on branch 'main'"
      ],
      "score": 2,
      "reason": "branch protection is not maximal on development and all release branches",
      "name": "Branch-Protection",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/6c1c789dc5b05cde492334f57b53807c786b038a/docs/checks.md#branch-protection",
        "short": "Determines if the default and release branches are protected with GitHub's branch protection settings."
      }
    }
  ],
  "metadata": null
}
```

Does that align with your expected Scorecard output?

evverx:
It does but with #1243 applied I got 
```
{
  "date": "2021-11-11",
  "repo": {
    "name": "github.com/evverx/systemd",
    "commit": "1b1e6249a0987820cde646a11c1b559dcb02d62f"
  },
[...]

PR #1243, opened 2021-11-11
🐛 Ignore nil values in Branch-Protection check

azeemshaikh38:

Part of #1242.

No.


laurentsimon (approved):
LGTM, thanks for the quick fix.
One question I have: when we ignore some setting, shall we add a 0 score to `scores` array? Otherwise the score won't see a difference when we run with or without admin token.

azeemshaikh38:
> LGTM, thanks for the quick fix. One question I have: when we ignore some setting, shall we add a 0 score to `scores` array? Otherwise the score won't see a difference when we run with or without admin token.

Not sure what you mean. The scores array collects scores *across* branches. But each branch only has a single score int. I have refactored the `branch_protection` check a bit more. Take a look and let me know.

laurentsimon:
> > LGTM, thanks for the quick fix. One question I have: when we ignore some setting, shall we add a 0 score to `scores` array? Otherwise the score won't see a difference when we run with or without admin token.
> 
> Not sure what you mean. The scores array collects scores _across_ branches. But each branch only has a single score int. I have refactored the `branch_protection` check a bit more. Take a look and let me know.

If run with admin token, there will be more settings we check for. I think this this means that when run with admin token, the overall score would drop (compared to a non-admin run) if all the settings that need admin settings are disabled

laurentsimon:
do you know if it fixes both https://github.com/ossf/scorecard/issues/1247 and https://github.com/ossf/scorecard/issues/1246?
I'm pretty sure it fixes the former.

evverx:
I'm not sure it works as expected. With this patch applied (with the same token) I got
```
{
  "date": "2021-11-11",
  "repo": {
    "name": "github.com/evverx/systemd",
    "commit": "33796123bccabdcc91ee9b87b5c042710bf50362"
  },
  "scorecard": {
    "version": "3.1.1-47-ga15225c",
    "commit": "a15225c59c0fb7d60f139795882990f8fde7badd"
  },
  "score": 3.0,
  "checks": [
    {
      "details": [
        "Warn: 'force pushes' enabled on branch 'main'",
        "Info: 'allow deletion' disabled on branch 'main'",
        "Warn: linear history disabled on branch 'main'",
        "Warn: 'administrator' PRs are exempt from reviews on branch 'main'",
        "Info: strict status check enabled on branch 'main'",
        "Warn: status checks for merging have no specific status to check on branch 'main'",
        "Warn: number of required reviewers is only 0 on branch 'main'",
        "Info: Stale review dismissal enabled on branch 'main'",
        "Warn: Owner review not required on branch 'main'"
      ],
      "score": 3,
      "reason": "branch protection is not maximal on development and all release branches",
      "name": "Branch-Protection",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/a15225c59c0fb7d60f139795882990f8fde7badd/docs/checks.md#branch-protection",
        "short": "Determines if the default and release branches are protected with GitHub's branch protection settings."
      }
    }
  ],
  "metadata": null
}
```
and here's what the rules look like
```
{
  "url": "https://api.github.com/repos/evverx/systemd/branches/main/protection",
  "required_pull_request_reviews": {
    "url": "https://api.github.com/repos/evverx/systemd/branches/main/protection/required_pull_request_reviews",
    "dismiss_stale_reviews": true,
    "require_code_owner_reviews": false,
    "required_approving_review_count": 2
  },
  "required_signatures": {
    "url": "https://api.github.com/repos/evverx/systemd/branches/main/protection/required_signatures",
    "enabled": false
  },
  "enforce_admins": {
    "url": "https://api.github.com/repos/evverx/systemd/branches/main/protection/enforce_admins",
    "enabled": false
  },
  "required_linear_history": {
    "enabled": false
  },
  "allow_force_pushes": {
    "enabled": true
  },
  "allow_deletions": {
    "enabled": false
  },
  "required_conversation_resolution": {
    "enabled": false
  }
}
```
It seems at least two subchecks aren't correct

azeemshaikh38:
> It seems at least two subchecks aren't correct

This is very surprising. I tested this PR both against your repo and a personal repo with an admin token and it is returning the expected results (matching with GitHub API). My only hunch here would be that somehow there is a delay between GitHub REST API and GraphQL APIs, so when you modify the settings and immediately test with Scorecard, the results seem off.

I'm going to submit this PR for now, since it is clearly an improvement over what we have today. Could you please try once more after the PR is submitted and let me know the results? If the issue persists, we can continue the debugging over #1246  and #1247.

Issue #1247, opened 2021-11-12
The "Branch-Protection" check seems to fail to detect the status of "status checks"


evverx:
I turned off the "Require status checks to pass before merging" setting and ran something like
```
curl  -u "***" -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/evverx/systemd/branches/main/protection/required_status_checks
```
to make sure they are off:
```
{
  "message": "Required status checks not enabled",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-status-checks-protection"
}
```

`scorecard` says that "Info: strict status check enabled on branch main" and increments the score:
```
scorecard  --format json --show-details --verbosity Debug --checks Branch-Protection  --repo=https://github.com/evverx/systemd
...
"Info: strict status check enabled on branch 'main'","Warn: status checks for merging have no specific status to check on branch 'main'"
```

laurentsimon:
Thanks for the report. This setting needs an admin token to be retrieved. We've prepared https://github.com/ossf/scorecard/pull/1243 to ignore the settings when it's not available.

evverx:
I use the same token with both curl and scorecard so I'm not sure why the results would be different. Does `scorecard` uses some other endpoints to get the protection rules?

laurentsimon:
we use graphQl APIs. @azeemsgoogle knows more about this part of the code

evverx:
Fixed in #1252. Closing.

Issue #1246, opened 2021-11-12
The "Branch-Protection" check seems to fail to get the number of required reviewers


evverx:
It seems my experimental branch fails to pass the Branch Protection check even though it requires at least 2 reviewers:
```
{
  "url": "https://api.github.com/repos/evverx/systemd/branches/main/protection/required_pull_request_reviews",
  "dismiss_stale_reviews": true,
  "require_code_owner_reviews": false,
  "required_approving_review_count": 2
}
```
`scorecard` says that "Warn: number of required reviewers is only 0 on branch 'main'"

I ran
```
scorecard  --show-details --verbosity Debug --checks Branch-Protection  --repo=https://github.com/evverx/systemd
```

evverx:
Fixed in #1252. Closing.

PR #1252, opened 2021-11-12
🐛 Fix branch protection results

laurentsimon:
BP results are wrong in certain cases.
See https://github.com/ossf/scorecard/issues/1247 and https://github.com/ossf/scorecard/issues/1246

I think this PR fixes it. There may be a cleaner way to re-arrange the code, but for now this should be enough for a quick fix.

During my tests, I found that a read-only non-admin token could retrieve all the results. I need someone else to confirm.

### DIFF
--- a/checks/branch_protection.go
+++ b/checks/branch_protection.go
@@ -45,13 +45,16 @@ var branchProtectionSettingScores = map[branchProtectionSetting]int{
 	allowForcePushes:     1,
 	allowDeletions:       1,
 	requireLinearHistory: 1,
-	enforceAdmins:        3,
+	// Need admin token.
+	enforceAdmins: 0,
 	// GitHub UI: "This setting will not take effect unless at least one status check is enabled".
+	// Need admin token.
 	requireStrictStatusChecks:   0,
 	requireStatusChecksContexts: 1,
 	requireApprovingReviewCount: 2,
 	// This is a big deal to enabled, so let's reward 3 points.
-	dismissStaleReviews:     3,
+	// Need admin token.
+	dismissStaleReviews:     0,
 	requireCodeOwnerReviews: 2,
 }
 
@@ -200,7 +203,8 @@ func checkReleaseAndDevBranchProtection(
 
 // isBranchProtected checks branch protection rules on a Git branch.
 func isBranchProtected(protection *clients.BranchProtectionRule, branch string, dl checker.DetailLogger) int {
-	totalScore := 15
+	// Temporary score calculation that does not account for admin settings.
+	totalScore := 8
 	score := 0
 
 	if protection.AllowForcePushes != nil {
--- a/checks/branch_protection.go
+++ b/checks/branch_protection.go
@@ -174,6 +174,7 @@ func checkReleaseAndDevBranchProtection(
 			return checker.CreateMinScoreResult(CheckBranchProtection,
 				fmt.Sprintf("branch protection not enabled on development/release branch: %s", b))
 		}
+		// TODO: if Protected==nil, we should not continue.
 		// The branch is protected. Check the protection.
 		score := isBranchProtected(&branch.BranchProtectionRule, b, dl)
 		scores = append(scores, score)
@@ -253,18 +254,22 @@ func isBranchProtected(protection *clients.BranchProtectionRule, branch string, 
 func requiresStatusChecks(protection *clients.BranchProtectionRule, branch string, dl checker.DetailLogger) int {
 	score := 0
 
-	switch protection.RequiredStatusChecks != nil {
+	switch protection.CheckRules.RequiresStatusChecks != nil {
 	case true:
-		dl.Info("status check enabled on branch '%s'", branch)
+		if *protection.CheckRules.RequiresStatusChecks {
+			dl.Info("status check enabled on branch '%s'", branch)
+		} else {
+			dl.Warn("status check disabled on branch '%s'", branch)
+		}
 
-		if protection.RequiredStatusChecks.UpToDate {
+		if protection.CheckRules.UpToDateBeforeMerge != nil &&
+			*protection.CheckRules.UpToDateBeforeMerge {
 			dl.Info("status checks require up-to-date branches for '%s'", branch)
-			score += branchProtectionSettingScores[requireStrictStatusChecks]
 		} else {
 			dl.Warn("status checks do not require up-to-date branches for '%s'", branch)
 		}
 
-		if len(protection.RequiredStatusChecks.Contexts) > 0 {
+		if protection.CheckRules.Contexts != nil && len(protection.CheckRules.Contexts) > 0 {
 			dl.Info("status checks have specific status enabled on branch '%s'", branch)
 			score += branchProtectionSettingScores[requireStatusChecksContexts]
 		} else {
@@ -272,7 +277,7 @@ func requiresStatusChecks(protection *clients.BranchProtectionRule, branch strin
 		}
 
 	case false:
-		dl.Warn("status check disabled on branch '%s'", branch)
+		dl.Debug("unable to retrieve status check for branch '%s'", branch)
 	}
 
 	return score
--- a/checks/branch_protection.go
+++ b/checks/branch_protection.go
@@ -179,7 +179,7 @@ func checkReleaseAndDevBranchProtection(
 			return checker.CreateMinScoreResult(CheckBranchProtection,
 				fmt.Sprintf("branch protection not enabled on development/release branch: %s", b))
 		}
-		// TODO: if Protected==nil, we should not continue.
+
 		// The branch is protected. Check the protection.
 		score := isBranchProtected(&branch.BranchProtectionRule, b, dl)
 		scores = append(scores, score)
--- a/checks/branch_protection.go
+++ b/checks/branch_protection.go
@@ -42,11 +42,12 @@ const (
 )
 
 var branchProtectionSettingScores = map[branchProtectionSetting]int{
-	allowForcePushes:            1,
-	allowDeletions:              1,
-	requireLinearHistory:        1,
-	enforceAdmins:               3,
-	requireStrictStatusChecks:   1,
+	allowForcePushes:     1,
+	allowDeletions:       1,
+	requireLinearHistory: 1,
+	enforceAdmins:        3,
+	// GitHub UI: "This setting will not take effect unless at least one status check is enabled".
+	requireStrictStatusChecks:   0,
 	requireStatusChecksContexts: 1,
 	requireApprovingReviewCount: 2,
 	// This is a big deal to enabled, so let's reward 3 points.
@@ -252,22 +253,26 @@ func isBranchProtected(protection *clients.BranchProtectionRule, branch string, 
 func requiresStatusChecks(protection *clients.BranchProtectionRule, branch string, dl checker.DetailLogger) int {
 	score := 0
 
-	if protection.RequiredStatusChecks.Strict != nil {
-		switch *protection.RequiredStatusChecks.Strict {
-		case false:
-			dl.Warn("status checks for merging disabled on branch '%s'", branch)
-			return score
-		case true:
-			dl.Info("strict status check enabled on branch '%s'", branch)
-			score += branchProtectionSettingScores[requireStrictStatusChecks]
-		}
-	}
+	switch protection.RequiredStatusChecks != nil {
+	case true:
+		dl.Info("status check enabled on branch '%s'", branch)
 
-	if len(protection.RequiredStatusChecks.Contexts) > 0 {
-		dl.Info("status checks for merging have specific status to check on branch '%s'", branch)
-		score += branchProtectionSettingScores[requireStatusChecksContexts]
-	} else {
-		dl.Warn("status checks for merging have no specific status to check on branch '%s'", branch)
+		if protection.RequiredStatusChecks.UpToDate {
+			dl.Info("status check require up-to-date branches for '%s'", branch)
+			score += branchProtectionSettingScores[requireStrictStatusChecks]
+		} else {
+			dl.Warn("status checks do not require up-to-date branches for '%s'", branch)
+		}
+
+		if len(protection.RequiredStatusChecks.Contexts) > 0 {
+			dl.Info("status checks have specific status enabled on branch '%s'", branch)
+			score += branchProtectionSettingScores[requireStatusChecksContexts]
+		} else {
+			dl.Warn("status checks have no specific status enabled on branch '%s'", branch)
+		}
+
+	case false:
+		dl.Info("status check disabled on branch '%s'", branch)
 	}
 
 	return score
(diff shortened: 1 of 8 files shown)


laurentsimon:
@evverx Can you try this PR and see if it fixes your problem?

evverx:
Thanks! Looks like https://github.com/ossf/scorecard/issues/1246 is gone but https://github.com/ossf/scorecard/issues/1247 isn't
```
{
  "date": "2021-11-11",
  "repo": {
    "name": "github.com/evverx/systemd",
    "commit": "e127b048e6a2e2b8ad195385b0620d42c9515cc5"
  },
  "scorecard": {
    "version": "3.1.1-53-g2cf6047",
    "commit": "2cf60476a04647160cb12f0f27fe9b6ddcba66f3"
  },
  "score": 4.0,
  "checks": [
    {
      "details": [
        "Warn: 'force pushes' enabled on branch 'main'",
        "Info: 'allow deletion' disabled on branch 'main'",
        "Warn: linear history disabled on branch 'main'",
        "Warn: 'administrator' PRs are exempt from reviews on branch 'main'",
        "Info: strict status check enabled on branch 'main'",
        "Warn: status checks for merging have no specific status to check on branch 'main'",
        "Info: number of required reviewers is 2 on branch 'main'",
        "Info: Stale review dismissal enabled on branch 'main'",
        "Warn: Owner review not required on branch 'main'"
      ],
      "score": 4,
      "reason": "branch protection is not maximal on development and all release branches",
      "name": "Branch-Protection",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/2cf60476a04647160cb12f0f27fe9b6ddcba66f3/docs/checks.md#branch-protection",
        "short": "Determines if the default and release branches are protected with GitHub's branch protection settings."
      }
    }
  ],
  "metadata": null
}
```
```
curl  -u "***" -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/evverx/systemd/branches/main/protection/required_status_checks
{
  "message": "Required status checks not enabled",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-status-checks-protection"
}
```

laurentsimon:
Please try the last commit I've pushed. It's fixed for me now. Tell me if it's fixed on your side.

evverx:
Judging by
```
...
  "score": 4.0,
  "checks": [
    {
      "details": [
        "Warn: 'force pushes' enabled on branch 'main'",
        "Info: 'allow deletion' disabled on branch 'main'",
        "Warn: linear history disabled on branch 'main'",
        "Warn: 'administrator' PRs are exempt from reviews on branch 'main'",
        "Info: status check disabled on branch 'main'",
        "Info: number of required reviewers is 2 on branch 'main'",
        "Info: Stale review dismissal enabled on branch 'main'",
        "Warn: Owner review not required on branch 'main'"
      ],
      "score": 4,
      ...
}
```
the issues are gone but I think it should be "Warn: status check disabled on branch 'main'". My understanding is that the "Info" messages are somewhat positive and increase the score while in this case it shouldn't I think.

laurentsimon:
I fixed the `Info` -> `Warn`. Let me know if it works on your end.
I will fix the score calculation once we have the warnings right. One thing at a time :-)

Thank you so much for your patience and help!

evverx:
Thanks! It seems to be working. I'll go ahead and close those issues

laurentsimon:
> Thanks! It seems to be working. I'll go ahead and close those issues

Great. I may ping you once I got the score fixed. In a few days :-)

laurentsimon:
@azeemsgoogle if you have time over the weekend, feel free to take over this PR. What's left to do is make the unit tests pass. T be honest, I don't think if it's worth the effort because I'm going to update the score calculation next week (discussion we had).

clients/githubrepo/branches.go
```diff
@@ -31,24 +31,70 @@ const (
...
-	DismissesStaleReviews      *bool
-	IsAdminEnforced            *bool
-	RequiresStrictStatusChecks *bool
+	// TODO: verify there is no conflicts.
+	// BranchProtectionRuleConflicts interface{}
 }
 
 type branch struct {
 	Name                 *string
-	RefUpdateRule        *refUpdateRule
```
azeemshaikh38, line 51: `RefUpdateRule` allows us to access settings viewable to non-admins. Won't removing this mean going back to our previous state where `Branch-Protection` check is mostly useful for admins?

image

clients/branch.go
```diff
@@ -28,13 +28,13 @@ type BranchProtectionRule struct {
...
 	EnforceAdmins              *bool
-	RequiredStatusChecks       StatusChecksRule
+	RequiredStatusChecks       *StatusChecksRule
 }
 
 // StatusChecksRule captures settings on status checks.
 type StatusChecksRule struct {
-	Strict   *bool
 	Contexts []string
+	UpToDate bool
```
azeemshaikh38, line 37: Could you explain the reasoning behind renaming this? This `bool` still captures `Strict` correct?

clients/branch.go
```diff
@@ -28,13 +28,13 @@ type BranchProtectionRule struct {
 	AllowForcePushes           *bool
 	RequireLinearHistory       *bool
 	EnforceAdmins              *bool
-	RequiredStatusChecks       StatusChecksRule
+	RequiredStatusChecks       *StatusChecksRule
```
azeemshaikh38, line 31: The data structure should be explicit about what information it captures. Using a `nil` `RequiredStatusChecks` struct to check for *status checks are not required* is confusing and is against the convention we are trying to build in #1242. Instead, how about something like:

```
type StatusChecksRule struct {
  RequiresStatusChecks *bool
  Strict/UpToDate *bool
  Contexts []string
}
``` 

Basically, we don't use pointers to `struct` and instead use pointers to primitive types.

azeemshaikh38 (commented):
Overall LGTM. My only concern is removing `RefUpdateRule`.

clients/githubrepo/branches.go
laurentsimon: good catch. I'm honestly confused. Not using it is what seemed to fix the problem. Can you try to reproduce the results I got with this PR? I actually tested in with a non-admin token all along. Maybe what they mean in their doc is `rules that are viewable *only* by non-admins`.
I will aslo try to verify the findings once more, and let's sync wafter that.

clients/branch.go
laurentsimon: took me 10mn to understand what `strict` meant in the GitHub UX settings. I can't find it in the API either. It may have been the old name they gave, I don't know. So I simply renamed it.. maybe `UpToDateBeforeMerge` would be more accurate. Let me know if you prefer `Strict` or if you're fine renaming it to something more meaningful.

clients/branch.go
laurentsimon: I did exactly that: use pointer to struct :-)
I'll update to using pointers to primitives. Thanks!

clients/githubrepo/branches.go
azeemshaikh38: I actually haven't been able to reproduce the problem from my side so far. But here's my preliminary suggestion - let's keep `RefUpdateRule` *and* the changes you have made as of now. We can fallback to the values returned by `RefUpdatedRule` if `BranchProtectionRule` returns `nil`. Do you think that would fix the problem while maintaining status quo?

clients/branch.go
azeemshaikh38: I'm ok with the rename, was just curious to understand the reasoning.

Nit: if `Strict` is confusing, how about `Enforced`?

clients/githubrepo/branches.go
laurentsimon: I suppose that'd work. I'm only worried that the code I added will have unwanted side effects. I'll do another round of  testing. In the graphQl API, they have `branchProtectionRules : A list of branch protection rules for this repository.` with no limitation whatsoever on differences between admin/non-admin, which is really confusing: I'll probably create a GH issue and ask them directly.

BTW, where did you find the list of branch protection settings that require admin tokens?

clients/branch.go
laurentsimon: enforced is confusing: enforcement of what? It could replace RequiresStatusChecks but not Strict, I think.
Note that it's even more complicated than that: in the UX, GH has a warning that `This setting will not take effect unless at least one status check is enabled`. We will check that in the check itself, though.

clients/githubrepo/branches.go
azeemshaikh38: My assumption was that since `RefUpdateRules` has all non-admin rules, `BranchProtectionRules` must be a super set of all rules available to both non-admins and admins.

clients/branch.go
azeemshaikh38: Ack.

laurentsimon:
I would like https://github.com/ossf/scorecard/issues/1274 to be resolved before merging this PR.

clients/githubrepo/branches.go
laurentsimon: I think that's correct. It just seems that when running as admin, `RefUpdateRules` is not available at all and everything appears in `BranchProtectionRules` instead.

laurentsimon:
@azeemsgoogle I've temporarily commented out the unit tests for this check because I'm going to change the score calculation in the next PR. Fixing them now and once again in next PR is unnecessary work I'd like to avoid.

laurentsimon:
friendly ping. I only commented out the unit tests, not the e2e ones.

docs/checks/internal/checks.yaml
```diff
@@ -139,12 +139,6 @@ checks:
       status checks before acceptance into a main branch, or preventing rewriting of
       public history.
 
-      Note: The following settings queried by the Branch-Protection check require an admin token: `DismissStaleReviews`, `EnforceAdmin`, and `StrictStatusCheck`. If
```
azeemshaikh38, line 142: Should we leave this as is since the status-quo does not change with this PR?

azeemshaikh38 (approved):

docs/checks/internal/checks.yaml
laurentsimon: absolutely. I forgot to put it back in.

clients/githubrepo/branches.go
```diff
@@ -114,35 +174,74 @@ func (handler *branchesHandler) listBranches() ([]*clients.BranchRef, error) {
 	return handler.branches, nil
 }
 
+func copyAdminSettings(src *branchProtectionRule, dst *clients.BranchProtectionRule) {
+	copyBoolPtr(src.IsAdminEnforced, &dst.EnforceAdmins)
+	copyBoolPtr(src.DismissesStaleReviews, &dst.RequiredPullRequestReviews.DismissStaleReviews)
+	if src.RequiresStatusChecks != nil && *src.RequiresStatusChecks {
```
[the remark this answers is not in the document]
laurentsimon: good catch. Done.

clients/githubrepo/branches.go
```diff
@@ -114,35 +174,74 @@ func (handler *branchesHandler) listBranches() ([]*clients.BranchRef, error) {
...
+func copyNonAdminSettings(src interface{}, dst *clients.BranchProtectionRule) {
+	// TODO: requiresConversationResolution, requiresSignatures, viewerAllowedToDismissReviews, viewerCanPush
+	switch v := src.(type) {
+	case *branchProtectionRule:
+		copyBoolPtr(v.AllowsDeletions, &dst.AllowDeletions)
+		copyBoolPtr(v.AllowsForcePushes, &dst.AllowForcePushes)
+		copyBoolPtr(v.RequiresLinearHistory, &dst.RequireLinearHistory)
+		copyInt32Ptr(v.RequiredApprovingReviewCount, &dst.RequiredPullRequestReviews.RequiredApprovingReviewCount)
+		copyBoolPtr(v.RequiresCodeOwnerReviews, &dst.RequiredPullRequestReviews.RequireCodeOwnerReviews)
+
```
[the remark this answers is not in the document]
laurentsimon: thanks. Done.

clients/githubrepo/branches.go
```diff
@@ -31,6 +31,55 @@ const (
 	refPrefix     = "refs/heads/"
 )
 
+// See https://github.community/t/graphql-api-protected-branch/14380
```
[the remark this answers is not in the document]
laurentsimon: done.

checks/branch_protection.go
```diff
@@ -249,22 +257,30 @@ func isBranchProtected(protection *clients.BranchProtectionRule, branch string,
...
+		}
+
+		if protection.CheckRules.UpToDateBeforeMerge != nil &&
+			*protection.CheckRules.UpToDateBeforeMerge {
+			dl.Info("status checks require up-to-date branches for '%s'", branch)
+		} else {
+			dl.Warn("status checks do not require up-to-date branches for '%s'", branch)
+		}
+
+		if protection.CheckRules.Contexts != nil && len(protection.CheckRules.Contexts) > 0 {
```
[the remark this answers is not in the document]
laurentsimon: thanks. Ive become paranoid :-)

checks/branch_protection.go
```diff
@@ -249,22 +257,30 @@ func isBranchProtected(protection *clients.BranchProtectionRule, branch string,
...
+	case true:
+		if *protection.CheckRules.RequiresStatusChecks {
+			dl.Info("status check enabled on branch '%s'", branch)
+		} else {
+			dl.Warn("status check disabled on branch '%s'", branch)
+		}
+
+		if protection.CheckRules.UpToDateBeforeMerge != nil &&
+			*protection.CheckRules.UpToDateBeforeMerge {
+			dl.Info("status checks require up-to-date branches for '%s'", branch)
```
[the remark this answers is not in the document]
laurentsimon: GitHub UX says `This setting will not take effect unless at least one status check is enabled`. So essentially it's a noop unless some contexts are defined. I can add the score and set it to 0. This is going to change in the next PR, but I'll add it anyway here.Thanks.

checks/branch_protection.go
```diff
@@ -249,22 +257,30 @@ func isBranchProtected(protection *clients.BranchProtectionRule, branch string,
...
-		case false:
-			dl.Warn("status checks for merging disabled on branch '%s'", branch)
-			return score
-		case true:
-			dl.Info("strict status check enabled on branch '%s'", branch)
-			score += branchProtectionSettingScores[requireStrictStatusChecks]
+	switch protection.CheckRules.RequiresStatusChecks != nil {
+	case true:
+		if *protection.CheckRules.RequiresStatusChecks {
+			dl.Info("status check enabled on branch '%s'", branch)
```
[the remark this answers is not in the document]
laurentsimon: same reason as above. Will add for now.

laurentsimon:
> > @azeemsgoogle I've temporarily commented out the unit tests for this check because I'm going to change the score calculation in the next PR. Fixing them now and once again in next PR is unnecessary work I'd like to avoid. if everything else is fine, please LGTM.
> 
> I have given LGTM. I'm hesitant about submitting this with all unit tests disabled though. Its not a practice we as a project should be encouraging. Please see if you can re-enable some of the basic test which check for things like nil-ptr dereference. If not, would be good if we have the follow up PR soon.

I intend to make it my next PR. Thanks!

Issue #1099, opened 2021-10-04
Make Token-Permission 'release'-aware


laurentsimon:
When permission `contents:write` is found in a workflow, we set the score to `0`.
We should check whether the workflow's job is a release job, i.e. one that creates a tag/release on the repo. This is a legitimate use case for the `contents:write` permission, and we should not reduce the score in this case.

PR #1253, opened 2021-11-13
✨ Reduce false positives in Token-Permissions for contents permission

laurentsimon:
This ignores, i.e. does not show a warning, if the `contents:write` is used in a packaging workflow.

close https://github.com/ossf/scorecard/issues/1099

### DIFF
--- a/checks/permissions.go
+++ b/checks/permissions.go
@@ -374,6 +374,9 @@ func createIgnoredPermissions(s, fp string, dl checker.DetailLogger) map[string]
 	if requiresPackagesPermissions(s, fp, dl) {
 		ignoredPermissions["packages"] = true
 	}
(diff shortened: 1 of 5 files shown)


oliverchang (approved):

azeemshaikh38:
This PR breaks e2e tests. I don't understand how a PR can get merged even though e2e tests are failing? This issue keeps re-occuring, we need to find a better solution.

laurentsimon:
> This PR breaks e2e tests. I don't understand how a PR can get merged even though e2e tests are failing? This issue keeps re-occuring, we need to find a better solution.

I looked into this. I don't think it's this PR, it's the dangerous workflow PR https://github.com/ossf/scorecard/pull/1168. Asra is sending a patch.

PR #1255, opened 2021-11-13
✨ [RAW] End-to-end support for raw results for Binary-Artifacts

laurentsimon:
This adds the plumbing to support arbitrary results via a `--raw` command. 

JSON output:
```
{
  "date": "2021-12-04",
  "repo": {
    "name": "...",
    "commit": "..."
  },
  "scorecard": {
    "version": "...",
    "commit": "..."
  },
  "metadata": null,
  "results": {
    "binaries": [
      {
        "path": "some/executable"
      },
      {
        "path": "executable2"
      }
    ]
  }
}

```


laurentsimon:
friendly ping

azeemshaikh38 (changes requested):
High-level feedback:

1. `RawResults` inside `CheckResult` limits us from sharing populated raw data across checks. Data populated by one check should be accessible across checks.
2. `RawResults` should be a well-defined struct and not an arbitrary format. This is also needed to ensure data sharing across checks.

So my suggestion would be to add a pointer to a struct, say `ScorecardData`, inside `ScorecardResult`. We can populate this struct by default and if `raw` bool is enabled, we print this field. Wdyt?

laurentsimon:
> High-level feedback:
> 
> 1. `RawResults` inside `CheckResult` limits us from sharing populated raw data across checks. Data populated by one check should be accessible across checks.

What do you mean? Checks run independently so they cannot access each others' data right now, right? 

> 2. `RawResults` should be a well-defined struct and not an arbitrary format. This is also needed to ensure data sharing across checks.

We can make this a structure instead of an `interface{}`. I have not done this because it requires adding as many structures as we have checks today so I started with the solution that minimized changes. But if we think it's better, I can do this. I think you're right that if we want to share this data with the `pkg.RunScorecard` API users, a structure is better to avoid having them casting it. Was this the main reason you had in mind or there were others?

> So my suggestion would be to add a pointer to a struct, say `ScorecardData`, inside `ScorecardResult`. We can populate this struct by default and if `raw` bool is enabled, we print this field. Wdyt?

That seems doable, yes.

This gives us the ability to change the internal structure without affecting the final output most users will rely upon. It will still affect API users (allstar) for whom we make no guarantees today.

So the solution would be: internal structures from checks -> copy to final structures `ScorecardData` -> set final structure within `ScorecardResult`. Correct?

laurentsimon:
> > So the solution would be: internal structures from checks -> copy to final structures `ScorecardData` -> set final structure within `ScorecardResult`. Correct?
> 
> I mean:
> 
> * A `ScorecardData` struct which lives in `pkg/` dir.
> * Update `CheckRequest` and `CheckResult` to hold a `*ScorecardData`.
>
> We don't need an internal struct except for convenience.

So you're suggesting we fix this (public) structure, pass it to checks and they will fill it. Right? Seems doable.

azeemshaikh38:
> > > So the solution would be: internal structures from checks -> copy to final structures `ScorecardData` -> set final structure within `ScorecardResult`. Correct?
> > 
> > 
> > I mean:
> > 
> > * A `ScorecardData` struct which lives in `pkg/` dir.
> > * Update `CheckRequest` and `CheckResult` to hold a `*ScorecardData`.
> > 
> > We don't need an internal struct except for convenience.
> 
> So you're suggesting we fix this (public) structure, pass it to checks and they will fill it. Right? Seems doable.

Yep, thanks!

laurentsimon:
> > > > So the solution would be: internal structures from checks -> copy to final structures `ScorecardData` -> set final structure within `ScorecardResult`. Correct?
> > > 
> > > 
> > > I mean:
> > > 
> > > * A `ScorecardData` struct which lives in `pkg/` dir.
> > > * Update `CheckRequest` and `CheckResult` to hold a `*ScorecardData`.
> > > 
> > > We don't need an internal struct except for convenience.
> > 
> > 
> > So you're suggesting we fix this (public) structure, pass it to checks and they will fill it. Right? Seems doable.
> 
> Yep, thanks!

Having a level of indirection is still useful, right?
Without it, I'm concerned that we might run into problems when filling the structure, e.g. if two checks are needed to fill. Note that I have not looked in details when this may happen, but there's a risk since we're moving things around. Wdut?

laurentsimon:
> > > > > So the solution would be: internal structures from checks -> copy to final structures `ScorecardData` -> set final structure within `ScorecardResult`. Correct?
> > > > 
> > > > 
> > > > I mean:
> > > > 
> > > > * A `ScorecardData` struct which lives in `pkg/` dir.
> > > > * Update `CheckRequest` and `CheckResult` to hold a `*ScorecardData`.
> > > > 
> > > > We don't need an internal struct except for convenience.
> > > 
> > > 
> > > So you're suggesting we fix this (public) structure, pass it to checks and they will fill it. Right? Seems doable.
> > 
> > 
> > Yep, thanks!
> 
> Having a level of indirection is still useful, right? Without it, I'm concerned that we might run into problems when filling the structure, e.g. if two checks are needed to fill the same part. Note that I have not looked in details when this may happen, but there's a risk since we're moving things around. Wdut?

additional thoughts to clarify. Here's where I'm at:
1. **Checks** evolve rapidly: As agreed, the checks can have their own internal structure instead of arbitrary `interface{}` format. 
2. **pkg.RunScorecard** outputs a format consumed by apps like allstar. We make no guarantees about breaking changes at the moment, but we should not break compatibility as often as checks change. So I envisage this function parsing the check results into an _independent_ structure
3. **pkg.AsJSON** outputs format consumed by most users and by policy makers like OPA. We want to change this structure as infrequently as possible. So I envisage this function parsing the results of point 2 into _another independent_ structure. As we refine the output for 2, we can, over time, incorporate them into this structure.

Hope this helps the discussion

laurentsimon:
> > > > > > So the solution would be: internal structures from checks -> copy to final structures `ScorecardData` -> set final structure within `ScorecardResult`. Correct?
> > > > > 
> > > > > 
> > > > > I mean:
> > > > > 
> > > > > * A `ScorecardData` struct which lives in `pkg/` dir.
> > > > > * Update `CheckRequest` and `CheckResult` to hold a `*ScorecardData`.
> > > > > 
> > > > > We don't need an internal struct except for convenience.
> > > > 
> > > > 
> > > > So you're suggesting we fix this (public) structure, pass it to checks and they will fill it. Right? Seems doable.
> > > 
> > > 
> > > Yep, thanks!
> > 
> > 
> > Having a level of indirection is still useful, right? Without it, I'm concerned that we might run into problems when filling the structure, e.g. if two checks are needed to fill the same part. Note that I have not looked in details when this may happen, but there's a risk since we're moving things around. Wdut?
> 
> additional thoughts to clarify. Here's where I'm at:

we talked offline. Below is what we decided:

> 1. **Checks** evolve rapidly: As agreed, the checks can have their own internal structure instead of arbitrary `interface{}` format.

should be replaced by a structure.

> 2. **pkg.RunScorecard** outputs a format consumed by apps like allstar. We make no guarantees about breaking changes at the moment, but we should not break compatibility as often as checks change. So I envisage this function parsing the check results into an _independent_ structure

optional for now. If allstar wants access to this results, we'll create a structure. Otherwise we'll wait till we make the code public.

> 3. **pkg.AsJSON** outputs format consumed by most users and by policy makers like OPA. We want to change this structure as infrequently as possible. So I envisage this function parsing the results of point 2 into _another independent_ structure. As we refine the output for 2, we can, over time, incorporate them into this structure.

optional for now. I've created this code so we can test things end-to-end anyway. Since we won't be making this code public, we can play around and update it for now.

> Hope this helps the discussion

laurentsimon:
@azeemsgoogle I've made the changes we discussed, PTAL

checker/check_request.go
```diff
@@ -28,4 +28,6 @@ type CheckRequest struct {
 	OssFuzzRepo clients.RepoClient
 	Dlogger     DetailLogger
 	Repo        clients.Repo
+	// UPGRADEv6: return raw results instead of scores.
```
azeemshaikh38, line 31: There are a lots of `// UPGRADEvX:` comments spread across the code. Can we instead create an issue and update the comment here to `// TODO(#issue): ...`?

checker/check_result.go
```diff
@@ -118,6 +118,35 @@ type CheckResult struct {
 	Details2 []CheckDetail `json:"-"` // Details of tests and sub-checks
 	Score    int           `json:"-"` // {[-1,0...10], -1 = Inconclusive}
 	Reason   string        `json:"-"` // A sentence describing the check result (score, etc)
+
+	// UPGRADEv6
+	// Stores raw data results.
+	// We re-use Error2 and Version.
+	RawResults *RawResults `json:"-"`
```
azeemshaikh38, line 125: Why do we need to copy over the `RawResults` ptr to each `CheckResult`? The ptr passed in `CheckRequest` is what gets updated right?

pkg/scorecard.go
```diff
@@ -101,7 +102,47 @@ func RunScorecards(ctx context.Context,
...
+	go runEnabledChecks(ctx, repo, nil, checksToRun, repoClient, ossFuzzRepoClient, ciiClient, resultsCh)
+	for result := range resultsCh {
+		ret.Checks = append(ret.Checks, result)
+	}
+	return ret, nil
+}
+
+// RunScorecardsRaw runs enabled Scorecard checks on a Repo
+// and returns raw results.
+func RunScorecardsRaw(ctx context.Context,
```
azeemshaikh38, line 114: Do we need a separate fn here? If `raw` bool is true, we simply need to pass a non-nil `RawResults` correct?

pkg/scorecard_raw_results.go
```diff
@@ -0,0 +1,30 @@
...
+package pkg
+
+import (
+	"time"
+
+	"github.com/ossf/scorecard/v3/checker"
+)
+
+// ScorecardRawResult struct is returned on a successful Scorecard run.
+type ScorecardRawResult struct {
```
azeemshaikh38, line 24: This struct is the same as `ScorecardResult` struct. Can we use the old struct itself? Or maybe simply add a `*RawResult` ptr to the old struct?

azeemshaikh38 (changes requested):
LGTM overall. Left some comments.

checker/check_result.go
laurentsimon: I think you're right, may not be needed. It's there in the code because of the original PR using `Raw interface{}`. I'll update, thanks!

pkg/scorecard.go
laurentsimon: you're right. I was hesitant to change the original function while we transition, to keep things as separate as possible. Shall I use a single function?

pkg/scorecard_raw_results.go
laurentsimon: I tried to keep everything separate while we transition. Shall I try to merge them now or later?

pkg/scorecard.go
azeemshaikh38: Yeah I think single fn will be better.

pkg/scorecard_raw_results.go
azeemshaikh38: IMO we can merge them now since the diff is minimal.

checker/check_result.go
laurentsimon: done.

pkg/scorecard_raw_results.go
laurentsimon: done.

pkg/scorecard.go
laurentsimon: done.

checker/check_request.go
laurentsimon: so there are just 3 `// UPGRADEv6`: I'm keeping them jut because there's no issue or things to fix, it's just indicating as comment that it will be released for v6. I think the ones mentioning `v2` can be cleaned up as you suggested.

laurentsimon:
need an actual LGTM now :-)

pkg/scorecard_raw_results.go
```diff
@@ -0,0 +1,31 @@
...
+	Repo       RepoInfo
+	Date       time.Time
+	Scorecard  ScorecardInfo
+	Checks     []checker.CheckResult
+	RawResults checker.RawResults
+	Metadata   []string
+}
```
azeemshaikh38, line 31: Remove this file?

azeemshaikh38 (approved):
Thanks!

checker/check_result.go
```diff
@@ -250,3 +275,14 @@ func CreateRuntimeErrorResult(name string, e error) CheckResult {
...
+// CreateRawResults creates raw results.
+// UPGRADEv6.
+func CreateRawResults(name string) CheckResult {
+	return CheckResult{
+		Name:    name,
+		Version: 6,
+		Error2:  nil,
+		// Note: RawResults are set by checks.
+	}
+}
```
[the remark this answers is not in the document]
laurentsimon: Either we keep it or we return `checker.CheckResult{}` in the functions. I don't mind keeping it to convey the fact that we're returning a raw results.. even if the function does not do much. You're right the `name` is not needed either.
Let me know what you think.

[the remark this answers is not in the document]
azeemshaikh38: Not sure what you mean. This is dead code right? No one is calling this fn AFAICT. Hence the suggestion to remove it.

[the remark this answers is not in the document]
laurentsimon: it was called until 5mn ago in `checks.BinaryArtifacts()` function. Your call whether we remove it or not :)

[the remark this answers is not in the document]
azeemshaikh38: Oh lol, ok. I prefer removing it. The code LG as is to me.

Issue #1256, opened 2021-11-13
Binary-Artifacts false positives


evverx:
As far as I can tell, the Binary-Artifacts check was supposed to catch executables people can run unintentionally after cloning repositories but in reality it seems to just look for extensions and, for example, flags projects using binary files for testing purposes. (The check would be even more noisy if it ignored extensions and searched for magic numbers by analogy with `file` because it would effectively penalize projects for keeping, for example, regression tests generated by fuzz targets (a lot of which look like executables or, maybe, even kind of executables if ELF files are in seed corpora). As an example, below is what the Binary-Artifacts says about `systemd`:
```
{
  "date": "2021-11-11",
  "repo": {
    "name": "github.com/systemd/systemd",
    "commit": "9cc615460830afdb51ad23e594906bbe60a3b25a"
  },
  "scorecard": {
    "version": "3.1.1-57-g8da30e6",
    "commit": "8da30e63afbc62e25c2c1252003aed9d34c3d04d"
  },
  "score": 0.0,
  "checks": [
    {
      "details": [
        "Warn: binary detected: test/dmidecode-dumps/HP-Z600.bin",
        "Warn: binary detected: test/dmidecode-dumps/Lenovo-ThinkPad-X280.bin",
        "Warn: binary detected: test/dmidecode-dumps/Lenovo-Thinkcentre-m720s.bin"
      ],
      "score": 0,
      "reason": "binaries present in source code",
      "name": "Binary-Artifacts",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/8da30e63afbc62e25c2c1252003aed9d34c3d04d/docs/checks.md#binary-artifacts",
        "short": "Determines if the project has generated executable (binary) artifacts in the source repository."
      }
    }
  ],
  "metadata": null
}
```

laurentsimon:
Thanks for the feedback. Scorecard looks for both magic number and extension. 
We check for binaries because the code cannot be reviewed and it may contain arbitrary code run by the user, as you say. Users may not directly run the code.

We discard binaries in folders named `testdata`, which you may use for unit tests, regression tests, etc. This is not documented so I'll add a note in the documentation. Do you think this is acceptable?

evverx:
> Scorecard looks for both magic number and extension

Judging by https://github.com/h2non/filetype#supported-types `filetype.Get` isn't exactly useful when it comes to detecting executables that can be run by accident so effectively `scorecard` almost always fall backs to file extensions.

[...]

Issue #1074, opened 2021-09-27
Try out scorecard action


laurentsimon:
once https://github.com/ossf/scorecard/pull/1073 is landed.
For each push, scorecard will run and the results will be available in https://github.com/ossf/scorecard/security/code-scanning 

I would like some feedback about the results. I've already noticed some improvements needed:
1. Short description should be shorter
2. Long description: shall we keep it long os just give users a URL?
3. Line numbers needed when referring to policy file violation
4. Successful results still appear in the GH security dashboard, because we mark them as `note`. Do we want to keep as-is or remove the notes entirely? Note that it may be useful to keep it in SARIF format in general, but remove it for GitHub actions only.

laurentsimon:
FYI @azeemsgoogle

asraa:
> Successful results still appear in the GH security dashboard, because we mark them as note. Do we want to keep as-is or remove the notes entirely? 

I would personally remove the notes entirely -- I don't think they have much value to the user (who could maybe go see a list of checks anyway on Scorecard), and I think it makes the important stuff less visible.

> Long description: shall we keep it long os just give users a URL?

I like it long, don't like link chasing.

> Short description should be shorter

Is this the title? If so is it canonical to have the title be the action of the check, or something indicating it's wrong, e.g. difference between "Determines if..." and "Check failed: Project workflow does not follow"

azeemshaikh38:
This looks great. Some comments:

1. The titles `Determines if ...` should be shorter and easier to understand. Something like `Dependencies are unpinned`.
2. The short description should be shorter. Also, it has some HTML elements which don't get rendered well.
3. Remove alerts which pop up as `Notes`.
4. Can we also make use of the `Severity` level (Low, Medium etc.)?

laurentsimon:
Thanks both of you for the feedback!

> This looks great. Some comments:
> 
> 1. The titles `Determines if ...` should be shorter and easier to understand. Something like `Dependencies are unpinned`.
> 2. The short description should be shorter. Also, it has some HTML elements which don't get rendered well.

the description is set to the long description atm. It does not support links. Maybe we should remove it? Note that the description is also used by GH next to the line with the rule ID https://github.com/ossf/scorecard/security/code-scanning/2865?query=ref%3Arefs%2Fheads%2Fmain so it's redundant. I have not tested what happens when there are multiple file locations reported.
> 3. Remove alerts which pop up as `Notes`.

Looks like nobody likes it :-)
[...]

PR #1263, opened 2021-11-14
:bug:Binary-Artifacts: skip files that look like tests

evverx:
Should address https://github.com/ossf/scorecard/issues/1256

I'm not sure how to figure out whether files are tests or not
by just looking at their names but it seems almost all the tools
analyzing source code I've seen so far either skip or demote files
containing "test" in their names. Some go even farther and admit that
they can't detect tests and allow users to specify where tests are one
way or another

### DIFF
--- a/checks/binary_artifact.go
+++ b/checks/binary_artifact.go
@@ -98,6 +98,13 @@ func checkBinaryFileContent(path string, content []byte,
 		return false, sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("filetype.Get:%v", err))
 	}
 
+	if strings.Contains(strings.ToLower(path), "test") {
+		dl.Debug3(&checker.LogMessage{
+			Text: fmt.Sprintf("%s looks like a test, skipping", path),
+		})
+		return true, nil
+	}
+
 	if _, ok := binaryFileTypes[t.Extension]; ok {
 		dl.Warn3(&checker.LogMessage{
 			Path: path, Type: checker.FileTypeBinary,


naveensrinivasan (approved):
Thanks!

checks/binary_artifact.go
```diff
@@ -98,6 +98,13 @@ func checkBinaryFileContent(path string, content []byte,
 		return false, sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("filetype.Get:%v", err))
 	}
 
+	if strings.Contains(strings.ToLower(path), "test") {
```
azeemshaikh38, line 101: We have `isTestData` fn already which looks for test files - https://github.com/ossf/scorecard/blob/main/checks/file_utils.go#L51. Maybe use that here instead?

azeemshaikh38 (approved):

checks/binary_artifact.go
evverx: It can't be used directly because it looks for "testdata" files only. It's possible to relax it by letting it skip "test" files but I'm not sure it's a good idea in the sense that I wouldn't want to start filtering out more files than I should in other places just because of the Binary-Artifacts check. I'd characterize this PR as a stop gap that should probably even look like a stop gap :-)

azeemshaikh38: Makes sense. We can submit this as is for now.

laurentsimon:
We discard binaries present in folders named `testdata` folder for this.

laurentsimon (commented):
We discard binaries present in folders named `testdata` folder for this. There is also a question around whether discarding binaries is a feature for result producers (repo owners) or result consumers (anyone who want more info about a repo before using it). I agree it's useful for repo owners and we can add this as an ignore list option to scorecard (we can add expose it in the GitHub action's policy file). For consumers, the best way to solve this is for the community to agree on naming conventions: testdata is already used by certain (several ?) languages.

evverx:
> There is also a question around whether discarding binaries is a feature for result producers (repo owners) or result consumers (anyone who want more info about a repo before using it).

Given that consumers usually report "issues" discovered by various code analyzers to maintainers I think they are all involved one way or another.

> For consumers, the best way to solve this is for the community to agree on naming conventions: testdata is already used by certain (several ?) languages.

I replied in https://github.com/ossf/scorecard/issues/1256#issuecomment-969232897

laurentsimon:
is it fair to say this PR may be closed since we've merged https://github.com/ossf/scorecard/pull/1300?
I'm sure someone else will raise the same issue for tests. Shall we re-visit when this happens and/or when we have a better understanding of what's a test?

cc @david-a-wheeler would love to add `where test files are` in the yaml config you mentioned during last meeting!

evverx:
> is it fair to say this PR may be closed since we've merged #1300?

I don't think it is. There are projects like elfutils that keep ELF files for testing purposes. Plus, basically almost all projects keeping their seed corpora close to their source code are affected I think

evverx:
I think it can be closed because I don't think that just looking for `test` is enough though. I asked how CodeQL and LGTM do it in https://github.com/ossf/scorecard/issues/1074#issuecomment-972337872 so in the meantime it would probably make sense to just wait for a reply.

@laurentsimon on an unrelated note I'm not sure who attends those meetings and whether they can affect any roadmaps but if would be great if for example GitHub fixed Dependabot along with their security tab: https://github.com/ossf/scorecard/pull/1336. I understand that the idea is to probably discuss high-level issues but it would be great if they could also lead to visible improvements in the tools scorecard itself recommends.

laurentsimon:
> I think it can be closed because I don't think that just looking for `test` is enough though. I asked how CodeQL and LGTM do it in [#1074 (comment)](https://github.com/ossf/scorecard/issues/1074#issuecomment-972337872) so in the meantime it would probably make sense to just wait for a reply.
> 
> @laurentsimon on an unrelated note I'm not sure who attends those meetings and whether they can affect any roadmaps but if would be great if for example GitHub fixed Dependabot along with their security tab: #1336. I understand that the idea is to probably discuss high-level issues but it would be great if they could also lead to visible improvements in the tools scorecard itself recommends.

let's create an issue to track other tools improvements, thanks for suggesting

PR #1279, opened 2021-11-16
:bug:Binary-Artifacts: no longer fall back to file extensions

evverx:
Looks like isArchive is supposed to cover all the filetypes
the Binary-Artifacts check is interested in:
https://github.com/h2non/filetype#archive. If types like dll
aren't supported they probably should be added to the matchers
library (by the way there is an in-flight PR trying to
achieve exactly that: https://github.com/h2non/filetype/pull/103)

With this patch applied scorecard no longer complains about
systemd:
```
    {
      "details": null,
      "score": 10,
      "reason": "no binaries found in the repo",
      "name": "Binary-Artifacts",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/63e3b92466f0403159c02f1ccedd43f9400e8b26/docs/checks.md#binary-artifacts",
        "short": "Determines if the project has generated executable (binary) artifacts in the source repository."
      }
```
and still can successfuly flag repositories with suspicious
elf files for example
```
./scorecard --verbosity Debug  --show-details --format json  --repo=https://github.com/evverx/binary-artifacts-test --checks Binary-Artifacts | jq | tee OUT
{
  "repo": {
    "name": "github.com/evverx/binary-artifacts-test",
    "commit": "11cbc0fddac398f346639b0552061f859e506b02"
  },
...
  "score": 0.0,
  "checks": [
    {
      "details": [
        "Warn: binary detected: date",
        "Warn: binary detected: nothing-to-see-here-hehe.txt"
      ],
      "score": 0,
      "reason": "binaries present in source code",
      "name": "Binary-Artifacts",
...
```

### DIFF
--- a/checks/binary_artifact.go
+++ b/checks/binary_artifact.go
@@ -15,12 +15,7 @@
 package checks
 
 import (
-	"fmt"
-	"path/filepath"
-	"strings"
-
 	"github.com/h2non/filetype"
-	"github.com/h2non/filetype/types"
 
 	"github.com/ossf/scorecard/v3/checker"
 	sce "github.com/ossf/scorecard/v3/errors"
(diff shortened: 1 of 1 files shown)


checks/binary_artifact.go
```diff
@@ -52,61 +47,11 @@ func BinaryArtifacts(c *checker.CheckRequest) checker.CheckResult {
...
-	if _, ok := binaryFileTypes[t.Extension]; ok {
-		dl.Warn3(&checker.LogMessage{
-			Path: path, Type: checker.FileTypeBinary,
-			Text: "binary detected",
-		})
-		*pfound = true
-		return true, nil
-	} else if _, ok := binaryFileTypes[strings.ReplaceAll(filepath.Ext(path), ".", "")]; ok {
-		// Falling back to file based extension.
+	if filetype.IsArchive(content) {
```
evverx, line 54: If the check shouldn't flag pdf files for example I think it would make sense to replace this part with something like
```
kind, _ := filetype.Archive(content)
if !stopList[kind.Extension] {
...
}
```
where `stopList` contains `pdf`

evverx: To get this to work more or less reliably I think some kind of https://github.com/h2non/filetype/pull/73 should be implemented first. I guess I'm back where I started :-)

PR #1287, opened 2021-11-16
✨ Update score for branch protection with levels

laurentsimon:
Update score to work like 'stages':
- first stage: force push, force deletion. if admin, enforce admins. Those give 3/10
- second stage: review >= 1. for admins, up-to-date branch. Those gives 6/10. Maybe up-to-date branch could go to first stage?
- third stage: status checks defined. Gives 8/10. 
- fourth stage : reviews >= 2. Gives 9/10
- fifth stage: for admins, dismissal (very hard to use). Gives 10/10

If a stage is not satisfied, the score won't go up regardless of the settings of the next stages. Warn/Info messages are shown for all stage, regardless of the score.
I scorecard is run without admin access, the admin settings are ignored.

@olivekl how shall I word this in the checks.yaml?

FYI, I'm working on the unit tests right now but that should not hamper the review.

### DIFF
--- a/checks/branch_protection.go
+++ b/checks/branch_protection.go
@@ -17,7 +17,6 @@ package checks
 import (
 	"errors"
 	"fmt"
-	"math"
 	"regexp"
 
 	"github.com/ossf/scorecard/v3/checker"
@@ -31,40 +30,53 @@ const (
 	// CheckBranchProtection is the exported name for Branch-Protected check.
 	CheckBranchProtection = "Branch-Protection"
 	minReviews            = 2
+	level1                = 3
+	level2                = 3
+	level3                = 2
+	level4                = 1
+	level5                = 1
 	// First level.
 	allowForcePushes branchProtectionSetting = iota
 	allowDeletions
-	requiresPullRequests
 	// Second level.
 	requireStatusChecksContexts
-	// Third level.
+	// Third and fourth level.
 	requireApprovingReviewCount
 	// Admin settings.
 	// First level.
 	enforceAdmins
 	// Second level.
 	requireUpToDateBeforeMerge
-	// Third level.
+	// Fourth level.
 	dismissStaleReviews
 	// requireCodeOwnerReviews no longer used.
 	// requireLinearHistory no longer used, see https://github.com/ossf/scorecard/issues/1027.
 )
 
 var branchProtectionSettingScores = map[branchProtectionSetting]int{
-	// The basics.
-	// We would like to have enforceAdmins,
-	// but it's only available to admins.
 	allowForcePushes:            1,
 	allowDeletions:              1,
-	requiresPullRequests:        1,
 	requireStatusChecksContexts: 1, // Gated by requireStatusChecks=true.
-
-	requireApprovingReviewCount: 2,
-
+	requireApprovingReviewCount: 1,
 	// Need admin token, so we cannot rely on them in general.
-	requireUpToDateBeforeMerge: 1, // Gated by requireStatusChecks=true.
 	enforceAdmins:              1,
+	requireUpToDateBeforeMerge: 1, // Gated by requireStatusChecks=true.
 	dismissStaleReviews:        1,
+}
+
+type scoresInfo struct {
+	basic               int
+	adminBasic          int
+	review              int
+	adminReview         int
+	context             int
+	thoroughReview      int
+	adminThoroughReview int
+}
+
+type levelScore struct {
+	scores scoresInfo
+	maxes  scoresInfo
 }
 
 //nolint:gochecknoinits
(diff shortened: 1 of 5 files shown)


laurentsimon:
friendly ping. ISE would like this check in the cron job's data in a couple weeks.

checks/branch_protection.go
```diff
@@ -28,36 +28,55 @@ type branchProtectionSetting int
 
 const (
 	// CheckBranchProtection is the exported name for Branch-Protected check.
-	CheckBranchProtection                         = "Branch-Protection"
-	minReviews                                    = 2
-	allowForcePushes      branchProtectionSetting = iota
+	CheckBranchProtection = "Branch-Protection"
+	minReviews            = 2
+	level1                = 3
```
asraa, line 33: maybe add a comment
`// Points incremented at each level`

checks/branch_protection.go
```diff
@@ -203,11 +429,28 @@ func checkReleaseAndDevBranchProtection(
...
+	score.scores.review, score.maxes.review = nonAdminReviewProtection(protection)
+	score.scores.adminReview, score.maxes.adminReview = adminReviewProtection(protection, branch, dl)
+	score.scores.context, score.maxes.context = nonAdminContextProtection(protection, branch, dl)
+	score.scores.thoroughReview, score.maxes.thoroughReview =
+		nonAdminThoroughReviewProtection(protection, branch, dl)
+	score.scores.adminThoroughReview, score.maxes.adminThoroughReview =
+		adminThoroughReviewProtection(protection, branch, dl)
+
+	return computeScore([]levelScore{score})
+}
```
azeemshaikh38, line 446: Move this fn to `_test.go` file where it is used.

checks/branch_protection.go
```diff
@@ -109,6 +128,198 @@ func BranchProtection(c *checker.CheckRequest) checker.CheckResult {
 	return checkReleaseAndDevBranchProtection(c.RepoClient, c.Dlogger)
 }
 
+func updateMaxScore(s1, s2 int) (int, error) {
+	if s2 != 0 && s1 != 0 && s1 != s2 {
```
asraa, line 132: I think I'm misunderstanding this logic :/ why is it the case that across all branches the settings are the same or one is 0? 

Is it because the scores are binary (at least for now, given the values in `branchProtectionSettingScores`)?

asraa (commented):
going to also comment offline as well..

checks/branch_protection.go
laurentsimon: good call! Done. Thanks!

checks/branch_protection.go
laurentsimon: done, thanks!

checks/branch_protection.go
laurentsimon: the name of the function may be confusing. This is not the result score: it's the maximum score possible for a level, calculated by adding `+1` for each settings enabled. The maximum score depends on whether you have admin access or not. When you don't have admin access, the max score for a level may be 0 because the settings are not available. I don''t want to say "we've removed points because we cannot access your admin-only settings".
So in this function, I update the maximum possible score but check that it's not 0, ie that is it available.
As yoou say, the maximum score across all branches MUST be the same: I added this code to also verify that the maximum possible score is the same across all branches: if this is violated it means there's a bug in our implementation.
Does this make sense?

laurentsimon: I think you're right there is unecessary logic I forgot to remove. let me update

laurentsimon: done, PTAL

olivekl:
Suggestions for wording this in checks.yaml (use `|` before to preserve formatting):

This test has tiered scoring. Each tier must be fully satisfied to achieve points at the next tier. For example, if you fulfill the Tier 3 checks but do not fulfill all the Tier 2 checks, you will not receive any points for Tier 3. 

Note: If Scorecard is run without an administrative access token, the requirements that specify “For administrators” are ignored. 

Tier 1 Requirements (3/10 points):

- Force push
- Force deletion
- For administrators: Include administrators

Tier 2 Requirements (6/10 points):

-  Required reviewers >=1 ​
- For administrators: Strict status checks (require branches to be up-to-date before merging) 

Tier 3 Requirements (8/10 points):

- Status checks defined

Tier 4 Requirements (9/10 points):

- Required reviewers >= 2

Tier 5 Requirements (10/10 points):

- For administrators: Dismiss stale reviews

checks/branch_protection.go
```diff
@@ -28,36 +28,56 @@ type branchProtectionSetting int
...
-	// Need admin token.
-	dismissStaleReviews:     3,
-	requireCodeOwnerReviews: 2,
+	allowForcePushes:            1,
+	allowDeletions:              1,
+	requireStatusChecksContexts: 1, // Gated by requireStatusChecks=true.
+	requireApprovingReviewCount: 1,
+	// Need admin token, so we cannot rely on them in general.
+	enforceAdmins:              1,
+	requireUpToDateBeforeMerge: 1, // Gated by requireStatusChecks=true.
```
asraa, line 64: I don't think this one is gated, is it? the scorecard check will use the settings I think because of this comment ("Even though it technically is not enforced by GitHub if Context=nil, we still show a positive outcome for users.")

checks/branch_protection.go
```diff
@@ -28,36 +28,56 @@ type branchProtectionSetting int
...
-	requireStatusChecks:         0,
-	requireStatusChecksContexts: 2,
-	requireApprovingReviewCount: 2,
-	// This is a big deal to enabled, so let's reward 3 points.
-	// Need admin token.
-	dismissStaleReviews:     3,
-	requireCodeOwnerReviews: 2,
+	allowForcePushes:            1,
+	allowDeletions:              1,
+	requireStatusChecksContexts: 1, // Gated by requireStatusChecks=true.
```
asraa, line 60: suggest revising this comment to "Used to check if required status checks before merging is non-empty, rather than checking if requireStatusChecks is set." or something like that

without the history that there used to be `requireStatusChecks` I find it confusing.

checks/branch_protection.go
```diff
@@ -28,36 +28,56 @@ type branchProtectionSetting int
...
+	review              int
+	adminReview         int
+	context             int
+	thoroughReview      int
+	adminThoroughReview int
+}
+
+type levelScore struct {
+	scores scoresInfo
+	maxes  scoresInfo
```
asraa, line 80: suggest comment
`// maximum score depending on whether admin token is used`

checks/branch_protection.go
```diff
@@ -28,36 +28,56 @@ type branchProtectionSetting int
...
-	allowForcePushes      branchProtectionSetting = iota
+	CheckBranchProtection = "Branch-Protection"
+	minReviews            = 2
+	// Points incremented at each level.
+	level1 = 3
+	level2 = 3
+	level3 = 2
+	level4 = 1
+	level5 = 1
+	// First level.
```
asraa, line 39: could you map the levels to the names you've given them? 
1: (admin/non-admin) basic score
2. (admin/non-admin) review score
3. (non-admin) context
4. (non-admin) thorough review
5. (admin) thorough review

checks/branch_protection.go
asraa: Ah! Gotcha. Was just a verification /debug check. Got it with the removed logic. Thank you!

checks/branch_protection.go
```diff
@@ -176,20 +386,33 @@ func checkReleaseAndDevBranchProtection(
...
+		score.scores.basic, score.maxes.basic = basicNonAdminProtection(&branch.BranchProtectionRule, b, dl)
+		score.scores.adminBasic, score.maxes.adminBasic = adminBasicProtection(&branch.BranchProtectionRule, b, dl)
+		score.scores.review, score.maxes.review = nonAdminReviewProtection(&branch.BranchProtectionRule)
+		score.scores.adminReview, score.maxes.adminReview = adminReviewProtection(&branch.BranchProtectionRule, b, dl)
+		score.scores.context, score.maxes.context = nonAdminContextProtection(&branch.BranchProtectionRule, b, dl)
+		score.scores.thoroughReview, score.maxes.thoroughReview =
+			nonAdminThoroughReviewProtection(&branch.BranchProtectionRule, b, dl)
+		score.scores.adminThoroughReview, score.maxes.adminThoroughReview =
+			adminThoroughReviewProtection(&branch.BranchProtectionRule, b, dl) // Do we want this?
+
```
azeemshaikh38, line 403: How about we write it this way:

```
tier1Score, tier1Passing := calculateTier1Score(branch.BranchProtectionRule)

if tier1Passing {
  tier2Score, tier2Passing := calculateTier2Score(...)
}
if tier2Passing {
  ...
}
score := tier1Score + tier2Score + ...
```

wherein, the `calculateTierXScore` returns true either if:
(i) all requirements for the tier are met, or
(ii) some fields are `nil`, indicating no data, so we don't penalize the repo.

This way we avoid the complex structs and also avoid leaking `admin token info` here in the check. Wdyt?

azeemshaikh38 (changes requested):

checks/branch_protection.go
laurentsimon: We can do the tier calculation in this part, because it's iterating over all branches: we need the data from all branches to calculate the tiers. This function is collecting the result. The tier calculation is done the `computeScore()` function. Can you check it?

asraa: > (ii) some fields are nil, indicating no data, so we don't penalize the repo.

I think without overhauling and separating out score/check (maybe not in this PR, i'm happy to give this a tackle on monday), you will need to propogate a max score since nil will either mean nil or no admin token. so i'm not sure this will work unless you get that info, in which case i think it's pretty much equivalent to what's currently implemented

checks/branch_protection.go
laurentsimon: sure

checks/branch_protection.go
azeemshaikh38: > We cannot do the tier calculation in this part, because it's iterating over all branches: we need the data from all branches to calculate the tiers. This function is collecting the result. The tier calculation is done in the computeScore() function and essentially does what you describe here. Can you check it?

Ah, I see. Then, let's calculate per-branch tier scores here and aggregate them later? So each branch will have:

```
{
  tier1: <score, passing>,
  tier2: <score, passing>,
 ...
}
```

which will be later aggregated across all branches. Would that work?

azeemshaikh38: > nil will either mean nil or no admin token

Right. And there is no way for us differentiate between the 2. Even with the knowledge that some fields are `admin-only`, a `nil` does not mean that it is because a lack of admin privileges. It simply means there was no data. So I see no reason to make speculations about why certain fields were returned `nil`.

checks/branch_protection.go
laurentsimon: right. When I tested, I think this setting was always nil when Context=nil. So I think my comment no longer applies. Let me remove it.

checks/branch_protection.go
laurentsimon: which comment?

checks/branch_protection.go
laurentsimon: > Ah, I see. Then, let's calculate per-branch tier scores here and aggregate them later? So each branch will have:

{
  tier1: <score, passing>,
  tier2: <score, passing>,
 ...
}
which will be later aggregated across all branches. Would that work?

I think this would work. However, it goes against the idea of separating the check from score calculation because we now would be doing score calculation at the same time that we retrieve settings. The current code just retrieves data, and score calculation is done after, which is what we want in the long term. Btw, we also need the max score for each branch for sanity checking, I think.

checks/branch_protection.go
laurentsimon: sg, will do. consider it done.

azeemshaikh38 (approved):

checks/branch_protection.go
azeemshaikh38: Sg, lgtm

checks/branch_protection.go
asraa: `Gated by requireStatusChecks=true` -- unless I'm misunderstanding!

docs/checks.md
```diff
@@ -97,6 +97,28 @@ Additionally, in some cases these rules will need to be suspended. For example,
 if a past commit includes illegal content such as child pornography, it may be
 necessary to use a force push to rewrite the history rather than simply hide the
 commit. 
+
+This test has tiered scoring. Each tier must be fully satisfied to achieve points at the next tier. For example, if you fulfill the Tier 3 checks but do not fulfill all the Tier 2 checks, you will not receive any points for Tier 3.
+
+Note: If Scorecard is run without an administrative access token, the requirements that specify “For administrators” are ignored.
+
+Tier 1 Requirements (3/10 points):
+  - Force push
```
asraa, line 106: suggest:
- Prevent force push
- Prevent branch deletion
- Include administrator for review

asraa (approved):
LGTM besides the small comment nits

laurentsimon:
Thanks @asraa and @azeemsgoogle . Did not mean to be too pushy on this one... kinda came across this way though. @asraa please feel free to start an issue/PR to separate this check, if you're interested in tackling it.

docs/checks.md
laurentsimon: done, thanks.

checks/branch_protection.go
laurentsimon: gotcha, will do. Thanks

laurentsimon:
FYI ```Makefile:261: *** GITHUB_AUTH_TOKEN is undefined.  Stop.
Error: Process completed with exit code 2.```

laurentsimon:
@naveensrinivasan @azeemshaikh38 please approve my integration tests. Thanks

laurentsimon:
Note: the PR has 2 LGTMs but this does not trigger the automatic approval of the integration test :/

laurentsimon:
Looks like I'm able to approve my own integration tests!

PR #1288, opened 2021-11-17
:bug:Binary-Artifacts: no longer complain about ".bin" files

evverx:
Those files most likely contain binary data used by tests for
example. It should be safe to remove this because executables
disguised as ".bin" files will still be caught and flagged by scorecard
before it even has a chance to look at extensions.

It should address https://github.com/ossf/scorecard/issues/1256


laurentsimon (approved):
Thanks

laurentsimon:
You also suggested using `IsArchive()` which seemed like a reasonable approach to me. Could you dump here your arguments in favor of it from a previous issue/PR?

@naveensrinivasan write this part of the code, let see what he thinks.

evverx:
I think `scorecard` can switch to `IsArchive` once the library starts to recognize archives deterministically. As far as tell, if for example a file is a `.deb` package and an `.ar` archive at the same time `filetype` can return either `.deb` or `.ar` depending on the implementation details. Effectively that means that to catch deb packages `scorecard` still have to fall back to extensions.

PR #1297, opened 2021-11-18
:sparkles:Pinned-Dependencies: show where exactly parsing fails

evverx:
Looks like scorecard can't parse comments quoted with backticks like
```
cmd -a \
    -b `# withouth backticks -c below would be a separate command` \
    -c
```
and fails with something like
```
error parsing shell code: 82:26: reached EOF without closing quote `
```

This PR turns that message into
```
Debug: error parsing shell code: vagrant/bootstrap_scripts/arch-sanitizers-clang.sh: 82:26: reached EOF without closing quote `
```
which is a bit more useful.


checks/shell_download_validate.go
```diff
@@ -768,7 +768,7 @@ func validateShellFileAndRecord(pathfn string, content []byte, files map[string]
 		// Note: this is caught by internal caller and only printed
 		// to avoid failing on shell scripts that our parser does not understand.
 		// Example: https://github.com/openssl/openssl/blob/master/util/shlib_wrap.sh.in
-		return false, sce.WithMessage(sce.ErrorShellParsing, err.Error())
+		return false, sce.WithMessage(sce.ErrorShellParsing, fmt.Sprintf("%s: %s", pathfn, err.Error()))
```
evverx, line 771: Looking at https://pkg.go.dev/mvdan.cc/sh/v3/syntax#Parser.Parse, I think I should pass `pathfn` to `Parse` instead

laurentsimon (approved):
Thanks!

checks/shell_download_validate.go
laurentsimon: LGTM if you've verified it works.

evverx: I certainly did :-) I'll rebase it once the integration test is fixed and https://github.com/ossf/scorecard/pull/1298 is merged.

FWIW the relevant "mvdan/sh" issue is https://github.com/mvdan/sh/issues/636. As far as I can tell it will never be fixed.

laurentsimon:
please commit something to restart the integration test, then I can approve them to run. Thanks

evverx:
I force pushed the branch to trigger the tests

evverx:
I'm not sure why the integration test failed. I force-pushed the branch once again to retrigger the tests.

laurentsimon:
I think we're hitting API rate limits. Errors `repo unreachable`, `Rate limit exceeded`
@naveensrinivasan @azeemsgoogle

azeemshaikh38:
> I think we're hitting API rate limits. Errors `repo unreachable`, `Rate limit exceeded` @naveensrinivasan @azeemsgoogle

Yep, that's the API rate limits for sure. Sadly I think we share the PATs between e2e tests and the cron job :D Need to wait and retry.

PR #1300, opened 2021-11-18
✨ Remove multimedia formats from Binary Artifacts

laurentsimon:
see title. Binary-Artifacts check mostly cares about compiled files that may be run or dynamically loaded and cannot be reviewed thru pull requests.


laurentsimon:
@evverx any comments on this change?

azeemshaikh38 (changes requested):
My concern with this change is we lose info about the presence of all these file formats that we used to detect before. We should be growing this list over time not trimming it :) 

Maybe we can modify scores depending on filetype? Similar to what @jeffmendoza  mentioned in his AllStar doc. We can give really low scores for executables but be more generous towards non-exec files. Wdyt?

laurentsimon:
> My concern with this change is we lose info about the presence of all these file formats that we used to detect before. We should be growing this list over time not trimming it :)

I've mostly removed audio files and things like torrent, which we look for internally. They are a risk that someone may run them and be compromised, but are not binary artifacts of a project that end up in code that may affect a dependent project

> Maybe we can modify scores depending on filetype? Similar to what @jeffmendoza mentioned in his AllStar doc. We can give really low scores for executables but be more generous towards non-exec files. Wdyt?

good suggestion. Note that we're currently not looking for zip/tar or other compressed format, which was in the AllStar doc.

@naveensrinivasan please also provide your thoughts.

naveensrinivasan:
> > My concern with this change is we lose info about the presence of all these file formats that we used to detect before. We should be growing this list over time not trimming it :)
> 
> I've mostly removed audio files and things like torrent, which we look for internally. They are a risk that someone may run them and be compromised, but are not binary artifacts of a project that end up in code that may affect a dependent project
> 
> > Maybe we can modify scores depending on filetype? Similar to what @jeffmendoza mentioned in his AllStar doc. We can give really low scores for executables but be more generous towards non-exec files. Wdyt?
> 
> good suggestion. Note that we're currently not looking for zip/tar or other compressed format, which was in the AllStar doc.
> 
> @naveensrinivasan please also provide your thoughts.

I agree with @azeemshaikh38 on increasing the more binary formats. But that said I don't see an issue with removing `audio files` which this PR is removing. So my vote is YES for this PR.

@azeemshaikh38 What are your concerns in specifically removing audio files?  Probably I am missing your thought process and concerns. I would like to understand that.

azeemshaikh38:
> I agree with @azeemshaikh38 on increasing the more binary formats. But that said I don't see an issue with removing `audio files` which this PR is removing. So my vote is YES for this PR.
> 
> @azeemshaikh38 What are your concerns in specifically removing audio files? Probably I am missing your thought process and concerns. I would like to understand that.

In the long-term, Scorecard plans to provide facts about a repo. So it's useful if Scorecard can *detect* presence even of  audio files. The policy evaluators can decide whether the repo should be penalized for it or not. So I don't want to lose the functionality where we are able to detect these files today.

For the short-term, I'm suggesting we use this PR to update the code for `Binary-Artifacts` scoring instead. If Scorecard *detects* audio files, we don't penalize the repo. That way, we don't lose any functionality and also improve our scoring. Bonus, we also generalize `Binary-Artifact` scoring from a boolean 0 or 10 to a more granular version.

Hope that makes sense.

naveensrinivasan:
> > I agree with @azeemshaikh38 on increasing the more binary formats. But that said I don't see an issue with removing `audio files` which this PR is removing. So my vote is YES for this PR.
> > @azeemshaikh38 What are your concerns in specifically removing audio files? Probably I am missing your thought process and concerns. I would like to understand that.
> 
> In the long-term, Scorecard plans to provide facts about a repo. So it's useful if Scorecard can _detect_ presence even of audio files. The policy evaluators can decide whether the repo should be penalized for it or not. So I don't want to lose the functionality where we are able to detect these files today.
> 
> For the short-term, I'm suggesting we use this PR to update the code for `Binary-Artifacts` scoring instead. If Scorecard _detects_ audio files, we don't penalize the repo. That way, we don't lose any functionality and also improve our scoring. Bonus, we also generalize `Binary-Artifact` scoring from a boolean 0 or 10 to a more granular version.
> 
> Hope that makes sense.

Thanks, @azeemshaikh38!  We probably can have a discussion on this if we want in our next meeting. 
@laurentsimon Thoughts?

laurentsimon:
> > > I agree with @azeemshaikh38 on increasing the more binary formats. But that said I don't see an issue with removing `audio files` which this PR is removing. So my vote is YES for this PR.
> > > @azeemshaikh38 What are your concerns in specifically removing audio files? Probably I am missing your thought process and concerns. I would like to understand that.
> > 
> > 
> > In the long-term, Scorecard plans to provide facts about a repo. So it's useful if Scorecard can _detect_ presence even of audio files. The policy evaluators can decide whether the repo should be penalized for it or not. So I don't want to lose the functionality where we are able to detect these files today.

Good point. One limitation is that it'd probably force users to add more post-processing (basically remove all irrelevant files they don't care about). Note that scorecard wants to detect stuff about a repo, but we should limit ourselves to things that are security related for the check. It's not clear that Pdf, audio files are relevant when ingesting dependencies.

> > For the short-term, I'm suggesting we use this PR to update the code for `Binary-Artifacts` scoring instead. If Scorecard _detects_ audio files, we don't penalize the repo. That way, we don't lose any functionality and also improve our scoring. Bonus, we also generalize `Binary-Artifact` scoring from a boolean 0 or 10 to a more granular version.
> > Hope that makes sense.

would you log the files in details but not increase the score? Would it be Warn or Info or Debug? We're trying to make Info a positive thing, and Warn a negative things that can be fixed.

> Thanks, @azeemshaikh38! We probably can have a discussion on this if we want in our next meeting. @laurentsimon Thoughts?

sounds good!

evverx:
@laurentsimon this PR looks good to me.

> So it's useful if Scorecard can detect presence even of audio files. The policy evaluators can decide whether the repo should be penalized for it or not. So I don't want to lose the functionality where we are able to detect these files today.

I don't think the `filetype` library is suitable for this. In https://github.com/ossf/scorecard/issues/1256#issuecomment-969442525 I compared it to `libmagic` and, well, in my opinion it's somewhat limited.

> Note that we're currently not looking for zip/tar or other compressed format, which was in the AllStar doc

I don't think it should be implemented at least until `scorecard` can figure out where tests are. Experimenting with `filetype` I unleashed `isArchive` on the util-linux repository and it failed the "binary-artifacts" check spectacularly because to test the code working with various archives/images it has to keep files like that there

evverx:
> One limitation is that it'd probably force users to add more post-processing (basically remove all irrelevant files they don't care about).

I don't think it should. As discussed elsewhere I think scorecard findings should be meaningful by default without any changes to it. If AllStar cares about file types and so on it should do it itself somewhere without affecting regular users of scorecard.

laurentsimon:
> > One limitation is that it'd probably force users to add more post-processing (basically remove all irrelevant files they don't care about).
> 
> I don't think it should. As discussed elsewhere I think scorecard findings should be meaningful by default without any changes to it. If AllStar cares about file types and so on it should do it itself somewhere without affecting regular users of scorecard.

so you're in favor of removing those audio/pdf files?

evverx:
> so you're in favor of removing those audio/pdf files?

Yes, I'm. In my opinion this check should focus on files that can be left in repositories by build pipelines (or people) and that can be accidentally run and certainly it shouldn't base its decisions on file extensions

laurentsimon:
@oliverchang give us your opinion as well, thanks

laurentsimon:
@naveensrinivasan @azeemsgoogle please approve the integration tests. Thanks

inferno-chromium:
i think goal should be removing the false positives from a check. multimedia files are very common in a repo, i wonder why we added them in the first place. i think goal for binary artifact check was not have source compiled binary formats, so protect against submitting compromised build artifacts. rest is all ok.

naveensrinivasan (approved):
Now that we have some consensus with @azeemshaikh38 and I. I am approving this.

oliverchang:
> @oliverchang give us your opinion as well, thanks

Also +1 to removing these since they don't really have security implications.

azeemshaikh38:
> i think goal should be removing the false positives from a check. multimedia files are very common in a repo, i wonder why we added them in the first place. i think goal for binary artifact check was not have source compiled binary formats, so protect against submitting compromised build artifacts. rest is all ok.

SG, lgtm.

azeemshaikh38 (approved):

laurentsimon:
Thanks everyone for chiming in!

azeemshaikh38:
@laurentsimon integration tests are failing, can you please check?

PR #1301, opened 2021-11-18
🌱 Run `Dangerous-Workflow` in release tests

azeemshaikh38:

Start running the `Dangerous-Workflow` check in the release tests to build confidence before we enable it in production.

No.

### DIFF
--- a/cron/config/config.yaml
+++ b/cron/config/config.yaml
@@ -22,6 +22,10 @@ completion-threshold: 0.99
 shard-size: 10
 webhook-url: 
 cii-data-bucket-url: gs://ossf-scorecard-cii-data
+# TODO: Temporarily remove SAST and CI-Tests which require lot of GitHub API tokens.
+# TODO(#859): Re-add Contributors after fixing inconsistencies.
+# TODO: Add Dangerous-Workflow in v4
+blacklisted-checks: SAST,CI-Tests,Contributors,Dangerous-Workflow
 metric-exporter: stackdriver
 # UPGRADEv2: to remove.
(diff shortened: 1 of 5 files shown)


asraa (approved):

cron/config/config.yaml
```diff
@@ -22,6 +22,10 @@ completion-threshold: 0.99
 shard-size: 10
 webhook-url: 
 cii-data-bucket-url: gs://ossf-scorecard-cii-data
+# TODO: Temporarily remove SAST and CI-Tests which require lot of GitHub API tokens.
```
laurentsimon, line 25: note: ISE wants to use CI-Tests it in their dashboard

laurentsimon (changes requested):
See comment on ISE above.

cron/config/config.yaml
azeemshaikh38: To be clear, `CI-Tests` have been disabled for a quite a while now. This PR is not disabling any new checks. 

If we want to re-enable `CI-Tests`, need to look into making the check more performant before adding in back in. Don't have the cycles to do that this quarter. Can you create an issue and maybe we can attempt this in Q1?

laurentsimon: Good to know! Which other checks are currently disabled?

laurentsimon (approved):
CI-Tests was already disabled. So approving. Can we add this info to the README under cron section?

cron/config/config.yaml
azeemshaikh38: SAST, CI-Test and Contributors. We had even updated the doc with this info, IIRC. Can't seem to find it now.

PR #1308, opened 2021-11-19
✨ Use `pull_request_target` + protected env for e2e

azeemshaikh38:

* **Other information**:

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,70 +12,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Run secret-dependent integration tests only after /ok-to-test approval
-on:
-  pull_request:
-  repository_dispatch:
-    types: [ok-to-test-command]
-
+# Run secret-dependent integration tests only after approval
 name: Integration tests
-env:
-  IMAGE_NAME: scorecard
-  OUTPUT_PATH: output
+on: pull_request_target
 
 permissions:
   contents: read
   pull-requests: write
 
 jobs:
-  integration-trusted:
+  approve:
     runs-on: ubuntu-latest
     steps:
-      - id: debug-outputs
-        run: |
-          echo ${{ github.event_name }}
-          echo ${{ github.event.client_payload }}
+      - name: approve
+        run: echo For security reasons, all pull requests need to be approved before running integration tests.
 
-      - id: verify-sha-input
-        if: github.event_name == 'repository_dispatch'
-        run: |
-          echo \"${{ github.event.client_payload.pull_request.head.sha }}\"
-          echo \"${{ github.event.client_payload.slash_command.args.named.sha }}\"
-          SHAINPUT=$(echo ${{github.event.client_payload.slash_command.args.named.sha}} | cut -c1-7)
-          if [ ${#SHAINPUT} -le 6 ]; then echo "error::input sha not at least 7 characters long" ; exit 1
-          else echo "done"
-          fi
-          SHAHEAD=$(echo ${{github.event.client_payload.pull_request.head.sha}} | cut -c1-7)
-          echo ${#SHAINPUT}
-          echo ${#SHAHEAD}
-          if [ "${SHAHEAD}" != "${SHAINPUT}" ]; then echo "sha input from slash command does not equal the head sha" ; exit 1
-          else echo "shas are equal"
-          fi
-
+  integration-trusted:
+    runs-on: ubuntu-latest
+    environment: integration-test
+    needs: [approve]
+    steps:
       - name: pull_request actions/checkout
-        if: github.event_name == 'pull_request'
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
-
-      - name: pull_request actions/checkout
-        if: github.event_name == 'repository_dispatch'
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
-        with:
-          ref: 'refs/pull/${{ github.event.client_payload.pull_request.number }}/merge'
 
       - name: setup-go
         uses: actions/setup-go@331ce1d993939866bb63c32c6cbbfd48fa76fc57 # v2.1.3
         with:
           go-version: '1.17'
-
-      - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@daadedc81d5f9d3c06d2c92f49202a3cc2b919ba # v0.2.1
-        with:
-          project_id: ${{ secrets.GCP_PROJECT_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -30,53 +28,17 @@ permissions:
...
-          fi
-          SHAHEAD=$(echo ${{github.event.client_payload.pull_request.head.sha}} | cut -c1-7)
-          echo ${#SHAINPUT}
-          echo ${#SHAHEAD}
-          if [ "${SHAHEAD}" != "${SHAINPUT}" ]; then echo "sha input from slash command does not equal the head sha" ; exit 1
-          else echo "shas are equal"
-          fi
-
       - name: pull_request actions/checkout
         if: github.event_name == 'pull_request'
```
laurentsimon, line 38: why do we need this if the `pull_request` trigger is not defined for this workflow?

laurentsimon (commented):
did you try using `workflow_run` as suggested by someone? Why can we not use it?

.github/workflows/integration.yml
azeemshaikh38: Done.

azeemshaikh38:
I can't get the integration test workflow to run in this PR. How do we test changes to such workflows?

laurentsimon:
> I can't get the integration test workflow to run in this PR. How do we test changes to such workflows?

the code used by GitHub is the one in the target/base branch. You'll have to merge into main to test it out. Maybe try it on a cloned repo?

.github/workflows/integration.yml
```diff
@@ -12,71 +12,34 @@
...
-      - name: pull_request actions/checkout
-        if: github.event_name == 'pull_request'
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+      - name: approve
+        run: echo For security reasons, all pull requests need to be approved before running integration tests.
 
+  integration-trusted:
+    runs-on: ubuntu-latest
+    environment: integration-test
+    needs: [approve]
```
asraa, line 33: doesn't the approve job always succeed? confused what approval is given

azeemshaikh38: I should probably name that job better. `approve` job here just prints a message (I was testing whether `integration` workflow actually runs). The real approval comes from the `environment: integration-test` line. That is a protected environment which can only be accessed by PRs that have approval from one of the required approvers.

azeemshaikh38:
> > I can't get the integration test workflow to run in this PR. How do we test changes to such workflows?
> 
> the code used by GitHub is the one in the target/base branch, IFAIK. You'll have to merge into main to test it out. Maybe try it on a cloned repo?

So I setup my fork to hav this change - https://github.com/azeemshaikh38/scorecard/blob/main/.github/workflows/integration.yml. Could you send my fork a PR from your fork to test this?

laurentsimon:
> > > I can't get the integration test workflow to run in this PR. How do we test changes to such workflows?
> > 
> > 
> > the code used by GitHub is the one in the target/base branch, IFAIK. You'll have to merge into main to test it out. Maybe try it on a cloned repo?
> 
> So I setup my fork to have this change - https://github.com/azeemshaikh38/scorecard/blob/main/.github/workflows/integration.yml. Could you send my fork a PR from your fork to test this?

you can create PRs to you fork, right? Does it make a difference if I send the PR or you do?
Want to be sure you can iterate on sending PRs for testing, not be blocked by me to resend, etc

azeemshaikh38:
> you can create PRs to you fork, right? Does it make a difference if I send the PR or you do? I want to be sure you can iterate on sending more PRs for testing, and not be blocked by me to resend, etc

You are right I was able to test it out - 
image

I think this PR is good to submit.

.github/workflows/integration.yml
```diff
@@ -12,71 +12,34 @@
...
-  OUTPUT_PATH: output
+on: pull_request_target
 
 permissions:
   contents: read
   pull-requests: write
 
 jobs:
-  integration-trusted:
+  approve:
```
laurentsimon, line 24: how does this check a reviewer approved? I see only a printed message.

laurentsimon (approved):
LGTM overall. I don't think this is the right approach because it affects productivity so much. Every time we submit a PR, we'll have to wait for someone to LGTM to run the integration test. 

As an interim solution, it's better than having PR blocked for ever so let's merge in.

I think the right approach is to use `workflow_run`.

wdut?

laurentsimon (changes requested):
see comment about printed message. I don't understand how we know it's been reviewed.

azeemshaikh38:
> see comment about printed message. I don't understand how we know it's been reviewed.

See https://github.com/ossf/scorecard/pull/1308#discussion_r753348139

.github/workflows/integration.yml
azeemshaikh38: It doesn't it only prints a message as you mentioned. The approve happens because of `environment` field.

laurentsimon: Thanks for clarifying.

laurentsimon (approved):
Thanks Azeem. Let's merge this one in and merge all the blocked PRs. Do you agree that we should move towards `workflow_run` to avoid affecting productivity? (In a follow-up PR)

azeemshaikh38:
> Thanks Azeem. Let's merge this one in and merge all the blocked PRs. Do you agree that we should move towards `workflow_run` to avoid affecting productivity? (In a follow-up PR)

No `workflow_run` will not work for us. At least, AFAICT. We can build the binaries in `pull_request` and pass it to `workflow_run` so that it can use the secret token, but then the binary itself will be a security threat.

azeemshaikh38:
This is still stuck :( I'm going to try and force submit it. Hopefully it should fix the issue for the other PRs.

laurentsimon:
> > Thanks Azeem. Let's merge this one in and merge all the blocked PRs. Do you agree that we should move towards `workflow_run` to avoid affecting productivity? (In a follow-up PR)
> 
> No `workflow_run` will not work for us. At least, AFAICT. We can build the binaries in `pull_request` and pass it to `workflow_run` so that it can use the secret token, but then the binary itself will be a security threat.

I see, that makes sense. Let's see if it's usable or not. Otherwise let's move to a model where we run the e2e tests without admin token. My understanding is that the token requires admin for certain check like branch protection: is my understanding correct? If that's the case, maybe we could at least separate the branch protection from the rest, to avoid blocking PR authors.

@naveensrinivasan has allowed/approved many PRs to unblock us, but it was not actually a code review: there was no LGTM. What I'm concerned with is that we're going to approve in a rush to allow the tests to run, without doing a proper review. @naveensrinivasan how is the user experience to approve these test ti run?
I'll try to go thru it on my end as well.

naveensrinivasan:
TBH this is painful!

Issue #1307, opened 2021-11-19
Pinned_Dependencies check stops on faulty shell files when it is meant to continue


lehors:
**Describe the bug**
When the shell parser returns an error the check stops when it should continue to the next file.

**Reproduction steps**
```
./scorecard --repo=https://github.com/openssl/openssl --checks=Pinned-Dependencies

RESULTS
-------
Aggregate score: ?

Check scores:
|-------|---------------------|--------------------------------|--------------------------------------------------------------------------------------------------------------------|
| SCORE |        NAME         |             REASON             |                                             DOCUMENTATION/REMEDIATION                                              |
|-------|---------------------|--------------------------------|--------------------------------------------------------------------------------------------------------------------|
| ?     | Pinned-Dependencies | error parsing shell code:      | https://github.com/ossf/scorecard/blob/71e8698617d25b006cf0935da5e1fb5487bff292/docs/checks.md#pinned-dependencies |
|       |                     | 5:19: a command can only       |                                                                                                                    |
|       |                     | contain words and redirects;   |                                                                                                                    |
|       |                     | encountered (                  |                                                                                                                    |
|-------|---------------------|--------------------------------|--------------------------------------------------------------------------------------------------------------------|

```

**Expected behavior**
Rather than quit on the first faulty shell file the check should continue, process the rest of the files, and still return some score.

**Additional context**

A simple fix (PR upcoming) gives the following more useful result:
```
RESULTS
-------
Aggregate score: 6.0 / 10

Check scores:
|--------|---------------------|--------------------------------|--------------------------------------------------------------------------------------------------------------------|
| SCORE  |        NAME         |             REASON             |                                             DOCUMENTATION/REMEDIATION                                              |
|--------|---------------------|--------------------------------|--------------------------------------------------------------------------------------------------------------------|
| 6 / 10 | Pinned-Dependencies | dependency not pinned by hash  | https://github.com/ossf/scorecard/blob/71e8698617d25b006cf0935da5e1fb5487bff292/docs/checks.md#pinned-dependencies |
|        |                     | detected -- score normalized   |                                                                                                                    |
|        |                     | to 6                           |                                                                                                                    |
|--------|---------------------|--------------------------------|--------------------------------------------------------------------------------------------------------------------|

```

azeemshaikh38:
Hi @lehors, thanks for starting this issue. I was about to respond to your email on ossf-scorecard-dev@ group, but glad that you started this issue as this is easier to track. 

[...]

PR #1312, opened 2021-11-19
:bug: Pinned-Dependencies: Fix invalid shell file handling

lehors:
Parsing errors are meant to be discarded but aren't. This patch
changes the code so that the error is indeed discarded and checking
continues as intended.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

- [ x] Tests for the changes have been added (for bug fixes / features)
- [ x] PR title follows the guidelines defined in  https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process

Bug fix

See Issue #1307 

The parsing error is discarded as intended and a score is given as a result.
The bottom of the description of Issue #1307 shows the new output.

No


checks/shell_download_validate.go
```diff
@@ -914,6 +914,7 @@ func validateShellFile(pathfn string, content []byte, dl checker.DetailLogger) (
 	if err != nil && errors.Is(err, sce.ErrorShellParsing) {
 		// Discard and print this particular error for now.
 		dl.Debug(err.Error())
+		err = nil
```
azeemshaikh38, line 917: Explicit `return false, nil` here since value of `r` is undefined.

azeemshaikh38 (approved):
Thanks @lehors !

checks/shell_download_validate.go
lehors: `r` is not undefined actually. It is set to false. See: https://github.com/ossf/scorecard/blob/23b0ddb8aa96356321cf31a2709723e29b15a951/checks/shell_download_validate.go#L771

checks/shell_download_validate_test.go
```diff
@@ -86,3 +88,18 @@ func TestIsSupportedShellScriptFile(t *testing.T) {
 		})
 	}
 }
+func TestValidateShellFile(t *testing.T) {
```
chrismcgehee, line 91: If you just add a newline before this method and add
```
	t.Parallel()
```
at the beginning of this method, the linter should pass.

Other than that, looks great! Thanks for fixing this!

chrismcgehee (changes requested):

checks/shell_download_validate_test.go
lehors: My bad. I fixed it. Thanks.

chrismcgehee (approved):

Issue #1306, opened 2021-11-19
pip install shouldn't be banned


evverx:
From https://github.com/ossf/scorecard/pull/611#discussion_r752910507

> I think we should just recommend pipenv instead, and consider all direct pip install commands as unpinned

I think it's overkill though when all I need is to install a package in a test script. I'm planning to switch to -r in systemd/systemd#21438 and I can theoretically add --require-hashes there since Dependabot supports it but switching to pipenv or any other tool creating virtual python environments is absolutely unwarranted in that particular case I think

laurentsimon:
Thanks for sharing. Tests are typically run on pull requests, in which case getting pwned is less serious. So this is again an instance of test false positive, it seems?

We need to start a doc around best security practice for pipy in general, because there's no general doc or agreement so it's hard for scorecard to implement something that does not exists. We've started one for npm but it's WIP.

evverx:
> Tests are typically run on pull requests, in which case getting pwned is less serious

Scripts like that can be run on push events or be "scheduled" as well and in some repositories they can have access to GITHUB_TOKEN with the "write" permissions granted to GHActions by default: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token so in theory they can pwn repositories.

evverx:
Just to clarify, I think in this case `scorecard` rightfully complains and pip should be changed there one way or another but it isn't obvious how it can be done without turning it into unmaintainable mess

laurentsimon:
cc @di who may know what pipy are planning to improve tooling

di:
There's not really anything functionally different between using pipenv's lockfile and using `requirements.txt` with `--require-hashes`, a `pip install` with `--require-hashes` should be allowed IMO.

laurentsimon:
thanks @di. One limitation of `reauirements.txt` is that it does not include transitive dependencies. Is this correct? If so, any plans to change this behavior?

di:
A `requirements.txt` file without hashes does not need to include transitive dependencies, but using `--require-hashes` enables strict hash checking mode, in which hashes must be included for all packages to be installed, including transitive dependencies, so it has the effect of requiring transitive dependencies to be included.

Example:
```
$ cat requirements.txt
sampleproject==2.0.0 \
    --hash=sha256:2b0c55537193b792098977fdb62f0acbaeb2c3cfc56d0e24ccab775201462e04 \
[...]

PR #1313, opened 2021-11-20
✨  Allow pip install with --require-hashes only

laurentsimon:
See discussion in https://github.com/ossf/scorecard/issues/1306#issuecomment-974539197
We've had a TODO in https://github.com/ossf/scorecard/blob/main/checks/shell_download_validate.go#L484 for a while.

This means that any `pip install`, `pip --upgrade`, `pip -U` are flagged, as is already the case.
We only allow: `pip install -r <> --require-hashes` and `pip install *.whl`

### DIFF
--- a/checks/shell_download_validate.go
+++ b/checks/shell_download_validate.go
@@ -469,7 +469,11 @@ func isUnpinnedPipInstall(cmd []string) bool {
 
 	isInstall := false
 	hasWhl := false
+	hashHashes := false
+	hasAddArgs := false
+	fmt.Println("testing", cmd, len(cmd))
 	for i := 1; i < len(cmd); i++ {
+		fmt.Println("", cmd[i])
 		// Search for install commands.
 		if strings.EqualFold(cmd[i], "install") {
 			isInstall = true
@@ -483,19 +487,29 @@ func isUnpinnedPipInstall(cmd []string) bool {
 		// TODO(laurent): https://github.com/ossf/scorecard/pull/611#discussion_r660203476.
 		// Support -r <> --require-hashes.
 
+		if strings.EqualFold(cmd[i], "--require-hashes") {
+			hashHashes = true
+			continue
+		}
+
 		// Exclude *.whl as they're mostly used
 		// for tests. See https://github.com/ossf/scorecard/pull/611.
 		if strings.HasSuffix(cmd[i], ".whl") {
 			hasWhl = true
+			// We continue because a command may contain
+			// multiple packages to install, not just `.whl` files.
 			continue
 		}
 
-		// Any other arguments are considered unpinned.
-		return true
+		// Any other arguments are considered unpinned,
+		// unless hashHashes is true.
+		hasAddArgs = true
 	}
 
-	// We get here only for `pip install [bla.whl ...]`.
-	return isInstall && !hasWhl
+	// We get here only for
+	// `pip install [bla.whl ...]`, `pip install <> --require-hashes`
+	fmt.Println(cmd, !((hasWhl && !hasAddArgs) || hashHashes))
+	return isInstall && !((hasWhl && !hasAddArgs) || hashHashes)
 }
 
 func isPythonCommand(cmd []string) bool {
--- a/checks/shell_download_validate.go
+++ b/checks/shell_download_validate.go
@@ -468,7 +468,9 @@ func isUnpinnedPipInstall(cmd []string) bool {
 	}
 
 	isInstall := false
-	hasUnresolvedDeps := false
+	hasRequireHashes := false
+	hasAdditionalArgs := false
+	hasWheel := false
 	for i := 1; i < len(cmd); i++ {
 		// Search for install commands.
 		if strings.EqualFold(cmd[i], "install") {
@@ -482,7 +484,7 @@ func isUnpinnedPipInstall(cmd []string) bool {
 
 		// https://github.com/ossf/scorecard/issues/1306#issuecomment-974539197.
 		if strings.EqualFold(cmd[i], "--require-hashes") {
-			hasUnresolvedDeps = false
+			hasRequireHashes = true
 			break
 		}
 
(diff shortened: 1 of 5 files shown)


laurentsimon:
@rsprabery @evverx

laurentsimon:
@di

evverx:
@laurentsimon thanks! With this patch applied `evverx/systemd` containing https://github.com/evverx/systemd/pull/56 kind of passed the check
```
      "details": [
        "Warn: no lock files detected for a package manager",
        "Warn: dependency not pinned by hash (job 'fuzzing'): .github/workflows/cifuzz.yml:37",
        "Warn: dependency not pinned by hash (job 'fuzzing'): .github/workflows/cifuzz.yml:44",
        "Info: GitHub-owned actions are pinned",
        "Info: Dockerfile dependencies are pinned",
        "Info: no insecure (not pinned by hash) dependency downloads found in Dockerfiles",
        "Info: no insecure (not pinned by hash) dependency downloads found in shell scripts",
        "Info: no insecure (not pinned by hash) dependency downloads found in GitHub workflows"
      ],
```
I'm not sure what "Warn: no lock files detected for a package manager" is but it doesn't seem to be relevant to this PR

laurentsimon:
Ignore `no lock files detected for a package manager`, it's being commented out in https://github.com/ossf/scorecard/pull/1315

checks/shell_download_validate.go
```diff
@@ -480,22 +482,28 @@ func isUnpinnedPipInstall(cmd []string) bool {
...
-		return true
+		// Other arguments are present.
+		hasAddArgs = true
 	}
 
-	// We get here only for `pip install [bla.whl ...]`.
-	return isInstall && !hasWhl
+	// We get here only for
+	// `pip install [bla.whl ...]`, `pip install <> --require-hashes`
+	return isInstall && !((hasWhl && !hasAddArgs) || hashHashes)
```
oliverchang, line 506: This logic is getting a bit hard to reason about.

Maybe expand out some of these inversions into:

```
return isInstall &&  !hasHashes && (!hasWhl || hasAddArgs)
```

Or break these up into smaller conditions to make them easier to understand.

oliverchang (approved):

checks/shell_download_validate.go
```diff
@@ -469,6 +469,8 @@ func isUnpinnedPipInstall(cmd []string) bool {
 
 	isInstall := false
 	hasWhl := false
+	hashHashes := false
+	hasAddArgs := false
```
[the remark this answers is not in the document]
di: IIUC, this is to capture a case like `pip install some_project_from_pypi`? If so I'd call this something like `hasUnresolvedDeps`.

checks/shell_download_validate.go
```diff
@@ -480,22 +482,28 @@ func isUnpinnedPipInstall(cmd []string) bool {
 			continue
 		}
 
-		// TODO(laurent): https://github.com/ossf/scorecard/pull/611#discussion_r660203476.
-		// Support -r <> --require-hashes.
+		// https://github.com/ossf/scorecard/issues/1306#issuecomment-974539197.
+		if strings.EqualFold(cmd[i], "--require-hashes") {
+			hashHashes = true
+			continue
```
di, line 488: If at any point we find `--require-hashes`, we can set `hasUnresolvedDeps = false` and break out of this loop. 

```suggestion
			hasUnresolvedDeps = false
			break
```

checks/shell_download_validate.go
```diff
@@ -480,22 +482,28 @@ func isUnpinnedPipInstall(cmd []string) bool {
...
 			hasWhl = true
+			// We continue because a command may contain
+			// multiple packages to install, not just `.whl` files.
 			continue
 		}
 
-		// Any other arguments are considered unpinned.
-		return true
+		// Other arguments are present.
+		hasAddArgs = true
```
di, line 501:
```suggestion
		// Unresolved dependencies are present.
		hasUnresolvedDeps = true
```

checks/shell_download_validate.go
di: I think this can just be:
```suggestion
	return isInstall && hasUnresolvedDeps
```

checks/shell_download_validate.go
di, line 473:
```suggestion
	hasUnresolvedDeps := false
```

checks/shell_download_validate.go
```diff
@@ -480,22 +482,28 @@ func isUnpinnedPipInstall(cmd []string) bool {
...
+		// https://github.com/ossf/scorecard/issues/1306#issuecomment-974539197.
+		if strings.EqualFold(cmd[i], "--require-hashes") {
+			hashHashes = true
+			continue
+		}
 
 		// Exclude *.whl as they're mostly used
 		// for tests. See https://github.com/ossf/scorecard/pull/611.
 		if strings.HasSuffix(cmd[i], ".whl") {
 			hasWhl = true
```
di, line 494: We don't actually care whether `.whl` files are present or not, just that they don't trigger `hasUnresolvedDeps`.
```suggestion
```

checks/shell_download_validate.go
```diff
@@ -480,22 +482,28 @@ func isUnpinnedPipInstall(cmd []string) bool {
 			continue
```
di, line 482: I think we need to assert that `install` can only be the first value here, otherwise this would trigger on an unlikely but technically valid command like `pip download install`. If it's not the first value, we can just break out of this loop:

```suggestion
			break
```

laurentsimon:
@di I have had to re-introduce some variables because some patterns were not captured by your simplification, e.g. `pip install bla.whl pkg1`

checks/shell_download_validate.go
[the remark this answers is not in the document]
laurentsimon: It could also be `--upgrade`

laurentsimon:
@naveensrinivasan @azeemsgoogle any idea why the license check is failing? I see `tree is dirty, please commit all changes`, but not sure what this means. The PR branch seems up to date with the main branch AFAICT.

PR #1314, opened 2021-11-20
✨ npm only allows `npm ci`

laurentsimon:
`npm ci` is the only command that honors hashes and verifies them. Other commands like `npm install`, `npm update` should never base used in scripts. Updates should be performed via a tool like dependabot/renovabot. 
even in the presence of a lock files, `npm install` may pull in un-pinned dependencies if some dependencies are not present in the lock file.

close https://github.com/ossf/scorecard/issues/1175

Issue #689, opened 2021-07-13
Frozen check: differentiate libs and programs


laurentsimon:
A library's dependency should not be pinned. We have a [TODO](https://github.com/ossf/scorecard/blob/main/checks/frozen_deps.go#L319) in the code to account for this already, but we have not implemented it yet.

It may be possible to detect if the package/code is a library or a program thru the manifest file.

Other ideas around pinning https://jbeckwith.com/2019/12/18/package-lock/

laurentsimon:
cursory read of of package.json:
`bin` entry declares a CLI/program, `main` declares an entry point for a library, aka module. Example [bin](https://github.com/tclindner/npm-package-json-lint/blob/master/package.json#L22) and [module](https://github.com/tclindner/npm-package-json-lint/blob/master/package.json#L29).
It is possible to declare a package that contains both a module and a CLI: in this case I think we should flag the use a lock files for the module.

nanikjava:
> A library's dependency should not be pinned. We have a [TODO](https://github.com/ossf/scorecard/blob/main/checks/frozen_deps.go#L319) in the code to account for this already, but we have not implemented it yet.

@laurentsimon went through the `main` branch there isn't any `frozen_deps.go` file ?

laurentsimon:
It's been renamed to `pinned-dependencies.go`.
Do you want me to assign this bug to you?

nanikjava:
@laurentsimon yes please. Thank you :)

laurentsimon:
Thanks @nanikjava !

nanikjava:
Hi @laurentsimon 

After reading through the comments and looking at https://docs.npmjs.com/cli/v7/configuring-npm/package-json following is my understanding for the frozen check:

* This check is for `node.js` project where the code will check for existence of `package.json` in root directory of the project  

* If the `package.json` has `bin` information  (eg: https://github.com/tclindner/npm-package-json-lint/blob/master/package.json#L23) it is considered as `bin`

* If the `package.json` has `main` information (eg: https://github.com/tclindner/npm-package-json-lint/blob/master/package.json#L30) it is considered as `module`

Now to the question:

* What score do we want to have for the above ?

* Are there any other check besides `bin` and `module` ?

Let me know if I'm on the right track.

laurentsimon:
You're definitely on the right track. That's how I understand `package.json` as well.
In terms of score:
1. If it's a module and it pins dependencies, score = 0
[...]

PR #1315, opened 2021-11-20
🐛 Disable pinning lock file search in repo

laurentsimon:
The code looking for lock files in package managers is incorrect:

- libraries vs cli https://github.com/ossf/scorecard/issues/689 => false positives
- we only look for lock files in the top folder => false positives

This PR disables this part of the code until we have a better solution. Looking for a lock files next to a manifest file is also error prone: users may copy lock files into dockerfiles from different locations, etc.

I think the right approach is to ensure any scripted installation command requires all dependencies and transitive dependencies to be pinned by hash:
- https://github.com/ossf/scorecard/pull/1314 - `npm ci`
- https://github.com/ossf/scorecard/pull/1313 - `pip install -r <> --require-hashes`

Updates should be performed by tools (dependabot/renovabot), never in scripts.

@meder @inferno-chromium  FYI

### DIFF
--- a/checks/pinned_dependencies.go
+++ b/checks/pinned_dependencies.go
@@ -651,7 +651,7 @@ func isPackageManagerLockFilePresent(c *checker.CheckRequest) (int, error) {
 // validatePackageManagerFile will validate the if frozen dependecies file name exists.
 // TODO(laurent): need to differentiate between libraries and programs.
 // TODO(laurent): handle multi-language repos.
-//nolint:unused,deadcode
+//nolint:unused
 func validatePackageManagerFile(name string, dl checker.DetailLogger, data fileparser.FileCbData) (bool, error) {
 	switch strings.ToLower(name) {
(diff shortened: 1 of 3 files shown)


checks/pinned_dependencies.go
```diff
@@ -45,10 +45,15 @@ func init() {
 // PinnedDependencies will check the repository if it contains frozen dependecies.
 func PinnedDependencies(c *checker.CheckRequest) checker.CheckResult {
 	// Lock file.
+	/* WARNING: this code is inherently incorrect:
+	- does not differenetiate between libs and main
+	- only looks at root fodler.
```
oliverchang, line 50: nit: s/fodler/folder/

oliverchang (approved):

laurentsimon:
@naveensrinivasan @azeemsgoogle FYI, the e2e2 tests are getting errors `repo unreachable`. Could this be because of GitHub's rate limiting? Or something else?

Issue #1317, opened 2021-11-20
BUG: integration tests are not testing the PR code


laurentsimon:
See this line https://github.com/ossf/scorecard/blob/main/.github/workflows/integration.yml#L36
We're not checking out the PR using `ref`, so we're effectively testing the main branch.

I unfortunately pushed a breaking change https://github.com/ossf/scorecard/pull/1316

PR #1318, opened 2021-11-20
🐛 Fix integration test

laurentsimon:
See this line https://github.com/ossf/scorecard/blob/main/.github/workflows/integration.yml#L36
We're not checking out the PR using `ref`, so we're effectively testing the main branch.

I unfortunately pushed a breaking change https://github.com/ossf/scorecard/pull/1316

To merge it it, we will need to disable integration tests temporarily in the GitHub UX.

close https://github.com/ossf/scorecard/issues/1317


azeemshaikh38 (approved):
Thanks!

PR #1323, opened 2021-11-22
🐛 Get OSes from matrix.include if present

chrismcgehee:

Bug fix

This [workflow file](https://github.com/mahmoud/boltons/blob/270e974975984f662f998c8f6eb0ebebd964de82/.github/workflows/tests.yaml) was causing an error for the Pinned-Dependencies check. It was because the only OSes listed in strategy.matrix were in the `include` section. Previously, `fileparser.JobAlwaysRunsOnWindows()` may have returned incorrect results if `strategy.matrix.os` and 'strategy.matrix.include` had different operating systems listed. That will be fixed here.
Part of https://github.com/ossf/scorecard/issues/839#issuecomment-974667652.

No

### DIFF
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -87,6 +87,14 @@ func getJobStrategyMatrixRows(job *actionlint.Job) map[string]*actionlint.Matrix
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


azeemshaikh38 (approved):
Thanks for the quick fix!

PR #1324, opened 2021-11-22
🐛 Skip pinned dependencies check for template Dockerfiles

chrismcgehee:

Bug fix

This [Dockerfile](https://github.com/traefik/mesh/blob/957da2d9ca4d4cdd4e00d0bdfbb9f5ae4f2685ff/tmpl.Dockerfile) was causing an error for the pinned dependencies check. It is a template Dockerfile, so it was unable to be parsed.
Part of https://github.com/ossf/scorecard/issues/839#issuecomment-974667652

Template Dockerfiles will be skipped for this check. I did some searching of Github for different formats that filenames for template Dockerfiles take, and so this should cover most of them.

No

### DIFF
--- a/checks/fileparser/listing.go
+++ b/checks/fileparser/listing.go
@@ -222,3 +222,22 @@ func CheckFileContainsCommands(content []byte, comment string) bool {
 	}
 	return false
 }
+
+// IsTemplateFile returns true if the file name contains a string commonly used in template files.
+func IsTemplateFile(pathfn string) bool {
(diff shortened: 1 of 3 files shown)


azeemshaikh38 (approved):
Thanks!

laurentsimon:
have we considered logging with `Debug` when an error is found in `validateDockerfileIsPinned`, instead of skipping the whole sub-check? Relying on dockerfile name seems fragile, no? 
Maybe we should create an `Error()` function to log errors in sub-checks without. Not sure what's best. wdut?

PR #1333, opened 2021-11-23
✨ Only run license check and not everything

laurentsimon:
The main workflow currently uses `make all` even though it only needs the licensecheck binary. This makes the check really slow.

In addition, I've pinned the dependency


naveensrinivasan (approved):

laurentsimon:
I. think we're hitting API rate limits, `Rate limit exceeded` happening on integration test.
FYI @azeemsgoogle @naveensrinivasan

Issue #1285, opened 2021-11-16
"internal error: cannot list check runs by ref" with the bcc repository


evverx:
Looks like https://github.com/iovisor/bcc/pull/3626 confuses `scorecard` somehow
```
./scorecard --verbosity Debug  --show-details --format json  --repo=https://github.com/iovisor/bcc --checks CI-Tests
...
{
  "date": "2021-11-15",
  "repo": {
    "name": "github.com/iovisor/bcc",
    "commit": "67f59ee80fcf5deedaacba1436d9fa09d32a16a0"
  },
  "scorecard": {
    "version": "3.1.1-64-gcc49494-dirty",
    "commit": "cc4949465b6730ee398e49a096e0132f02078372"
  },
  "score": -1.0,
  "checks": [
    {
      "details": [
        "Debug: checking PR: 3626"
      ],
      "score": -1,
      "reason": "internal error: cannot list check runs by ref",
      "name": "CI-Tests",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/cc4949465b6730ee398e49a096e0132f02078372/docs/checks.md#ci-tests",
        "short": "Determines if the project runs tests before pull requests are merged."
      }
    }
  ],
  "metadata": null
```
To get that "Debug: checking PR: 3626" message I applied the following patch:
```diff
diff --git a/checks/ci_tests.go b/checks/ci_tests.go
index 515cbb1..a84d984 100644
--- a/checks/ci_tests.go
+++ b/checks/ci_tests.go
@@ -53,6 +53,9 @@ func CITests(c *checker.CheckRequest) checker.CheckResult {

                var foundCI bool

+               c.Dlogger.Debug3(&checker.LogMessage{
+                       Text: fmt.Sprintf("checking PR: %d", pr.Number),
+               })
                // Github Statuses.
                prSuccessStatus, err := prHasSuccessStatus(pr, c)
                if err != nil {
```

evverx:
https://github.com/systemd/casync triggers it too
```
    {
      "details": null,
      "score": -1,
      "reason": "internal error: cannot list check runs by ref",
      "name": "CI-Tests",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/bd9b6df2d372718aebe7f1fb4c2a6c32149e5868/docs/checks.md#ci-tests",
        "short": "Determines if the project runs tests before pull requests are merged."
      }
    },
```

Issue #1322, opened 2021-11-22
scorecard seems to miss some status checks


evverx:
Looking at the results of the SAST check I noticed that sometimes `scorecard` says that some status checks (namely `lgtm-com`) weren't run even though they were. GitHub API described at https://docs.github.com/en/rest/reference/checks#get-a-check-suite has no problem finding them:
https://github.com/libbpf/libbpf/pull/353
```
$ curl --silent  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/libbpf/libbpf/commits/aafe84ae9b8e7b837c6469e48e85f7c3769c79b6/check-suites | jq '.check_suites[] | .app.slug, .latest_check_runs_count'
"lgtm-com"
1
"github-actions"
13
"github-actions"
13
"github-actions"
13
```
https://github.com/mesonbuild/meson/pull/9576
```
$ curl --silent  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/mesonbuild/meson/commits/53120d13e56f8536dfb093a35fb90559c171295a/check-suites | jq '.check_suites[] | .app.slug, .latest_check_runs_count'
"lgtm-com"
2
"github-actions"
1
"github-actions"
3
"github-actions"
1
"github-actions"
1
"github-actions"
2
"github-actions"
1
"github-actions"
3
"github-actions"
12
"github-actions"
4
"azure-pipelines"
7
"codecov"
2
```
I'm not sure how exactly `ListCheckRunsForRef` is implemented but it seems it's different from what the "check-suites" returns.

evverx:
It's kind of related to https://github.com/ossf/scorecard/issues/1322

laurentsimon:
It's using https://github.com/google/go-github/blob/master/github/checks.go#L250
They also expose check suites API, see https://github.com/google/go-github/blob/master/github/checks.go#L274

if you give it a try, let us know.

cc @azeemsgoogle

evverx:
Apparently https://github.com/google/go-github/blob/master/github/checks.go#L250 is where `lgtm-com` gets lost because with `check-runs` instead of `check-suites` I can no longer see it anywhere:
```
curl --silent  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/mesonbuild/meson/commits/53120d13e56f8536dfb093a35fb90559c171295a/check-runs |
grep -i lgtm
```

laurentsimon:
[...]

PR #1335, opened 2021-11-23
🐛 CI-Tests: no longer fail if there are no check suites

evverx:
In PRs like https://github.com/iovisor/bcc/pull/3626 no check suites
are triggered:
```
$ curl --silent  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/iovisor/bcc/commits/3fcf0f1b5862ff2c5908723b38b57c5d1c97c1cf/check-runs
{
  "total_count": 0,
  "check_runs": [

  ]
}
```
```
 curl --silent  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/iovisor/bcc/commits/3fcf0f1b5862ff2c5908723b38b57c5d1c97c1cf/check-suites
{
  "total_count": 0,
  "check_suites": [

  ]
}
```
The check should just keep going because "statuses" still can be
triggered so it should use them instead

Closes https://github.com/ossf/scorecard/issues/1285


evverx:
FWIW I think this check should switch from `check-runs` to `check-suites` (as discussed in https://github.com/ossf/scorecard/issues/1322) but until then it shouldn't fail when there're no check runs (it shouldn't fail when there're no check suites either though :-))

laurentsimon (approved):
LGTM, thanks. cc @azeemsgoogle: can you confirm the `crs == nil` was not intentional ?

evverx:
FWIW I don't think it can be intentional because because of that check projects with no tests get `-1` instead of `0` and I don't think that's correct.

laurentsimon:
@naveensrinivasan @azeemsgoogle the e2e tests are failing because of the OSV check. Has anything changed recently? This PR does not touch the OSV/Vulnerabilities check.

naveensrinivasan:
It is failing on 
```
• Failure [12.665 seconds]
323
E2E TEST:Fuzzing
324
/home/runner/work/scorecard/scorecard/e2e/fuzzing_test.go:29
325
  E2E TEST:Validating use of fuzzing tools
326
  /home/runner/work/scorecard/scorecard/e2e/fuzzing_test.go:30
327
    Should return use of OSS-Fuzz [It]
328
    /home/runner/work/scorecard/scorecard/e2e/fuzzing_test.go:31
329

330
    Expected
331
        <bool>: false
332
    to be true
333

334
    /home/runner/work/scorecard/scorecard/e2e/fuzzing_test.go:55
335
------------------------------
336
E2E TEST:Contributors E2E TEST:Validating project contributors 
337
  Should return valid project contributors
338
  /home/runner/work/scorecard/scorecard/e2e/contributors_test.go:31
339
•
340

341
Summarizing 1 Failure:
342

343
[Fail] E2E TEST:Fuzzing E2E TEST:Validating use of fuzzing tools [It] Should return use of OSS-Fuzz
```

azeemshaikh38:
> 

+1. I think its an overall improvement to return 0 score over -1, so this PR LGTM.

Issue #1257, opened 2021-11-13
The "Token-Permissions" check seems to complain about codeql and labeler


evverx:
Both `codeql` and `labeler` are widely used GH Actions that require non read-only tokens. The links to the "official" documentation where the permissions they have to have are mentioned:
https://github.com/github/codeql-action#usage
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input

```
./scorecard --verbosity Debug --checks Token-Permissions --show-details --format json  --repo=https://github.com/systemd/systemd | jq
...
      "details": [
        "Info: permissions set to 'read-all': .github/workflows/build_test.yml:1",
        "Debug: not a publishing workflow: .github/workflows/build_test.yml:1",
        "Debug: not a codeql workflow: .github/workflows/build_test.yml:1",
        "Debug: not a codeql upload SARIF workflow: .github/workflows/build_test.yml:1",
        "Debug: no permission defined: .github/workflows/build_test.yml:1",
        "Info: permissions set to 'read-all': .github/workflows/cifuzz.yml:1",
        "Debug: not a publishing workflow: .github/workflows/cifuzz.yml:1",
        "Debug: not a codeql workflow: .github/workflows/cifuzz.yml:1",
        "Debug: not a codeql upload SARIF workflow: .github/workflows/cifuzz.yml:1",
        "Debug: no permission defined: .github/workflows/cifuzz.yml:1",
        "Warn: no permission defined: .github/workflows/codeql-analysis.yml:1",
        "Debug: not a publishing workflow: .github/workflows/codeql-analysis.yml:1",
        "Debug: codeql workflow detected: .github/workflows/codeql-analysis.yml:1",
        "Info: 'actions' permission set to 'read': .github/workflows/codeql-analysis.yml:1",
        "Info: 'contents' permission set to 'read': .github/workflows/codeql-analysis.yml:1",
        "Debug: 'security-events' permission set to 'write': .github/workflows/codeql-analysis.yml:1",
        "Info: permissions set to 'read-all': .github/workflows/coverity.yml:1",
        "Debug: not a publishing workflow: .github/workflows/coverity.yml:1",
        "Debug: not a codeql workflow: .github/workflows/coverity.yml:1",
        "Debug: not a codeql upload SARIF workflow: .github/workflows/coverity.yml:1",
        "Debug: no permission defined: .github/workflows/coverity.yml:1",
[...]

PR #1336, opened 2021-11-23
📖 docs: be more specific about what Dependabot brings with it

evverx:
It would have helped me to decide whether I needed it or not

https://github.com/dependabot/dependabot-core/issues/2804 has been open for almost a year (and https://github.com/dependabot/dependabot-core/issues/2198 since 2019) so it doesn't seem to make sense to wait for it to be fixed anytime soon.


evverx:
The documentation should probably mention that RenovateBot doesn't support dependencies pinned to SHAs in a way that would be human-friendly: https://github.com/evverx/systemd/pull/30 but since I'm not sure whether it's configurable or not I decided not to include it in this PR.

More generally, it would be great if before recommending anything in its documentation the scorecard project could somehow make sure that whatever it recommends is more or less usable.

evverx:
I'd personally probably stop recommending CodeQL as well due its hidden security tab with alerts. It was discussed in https://github.com/ossf/scorecard/issues/1257#issuecomment-970717868 and https://github.com/ossf/scorecard/issues/1074#issuecomment-971096713 though

evverx:
There seems to be another missing feature: https://github.com/dependabot/dependabot-core/issues/2835#issuecomment-902136798

azeemshaikh38 (approved):
Thanks!

azeemshaikh38:
Thanks @evverx. Feel free to open issues about the comments you have added here and we can continue the discussion there.

PR #1344, opened 2021-12-02
✨ Cleanup Branch Protection and add e2e tests

laurentsimon:
This cleans up the branch protection check and adds additional e2e tests.
It also fixes a bug whereby score would be 10 if none of the branches had branch protection enabled.


asraa (approved):
Simplification makes sense to me!

Non-blocking because I think this will need to be refactored anyway but I think using `protected` as the `doLogging` bool might be better named or commented (`warn_protected`, but it's awkward to use in both scenarios)

PR #1345, opened 2021-12-02
🌱  [UNDER TESTING] Increase parallelism

azeemshaikh38:

* **Other information**:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
      - name: Run presubmit tests
        run: |
             go env -w GOFLAGS=-mod=mod
-            make -j 5 all
+            make -j $(nproc) all
   license-check:
     name: license boilerplate check
     runs-on: ubuntu-latest


.github/workflows/main.yml
```diff
@@ -38,7 +38,7 @@ jobs:
      - name: Run presubmit tests
        run: |
             go env -w GOFLAGS=-mod=mod
-            make -j 5 all
+            make -j 12 all
```
laurentsimon, line 41: have you tried using "make -j`nproc`" to dynamically retrieve the number of processors?

laurentsimon (approved):

laurentsimon:
error `release failed after 1800.01s error=context deadline exceeded`. The Validate does many checks: shall we split it into smaller ones? The linter does not need to build the code, for example...

inferno-chromium (approved):

.github/workflows/main.yml
```diff
@@ -38,7 +38,7 @@ jobs:
      - name: Run presubmit tests
        run: |
             go env -w GOFLAGS=-mod=mod
-            make -j 5 all
+            make -j $(nproc) all
```
naveensrinivasan, line 41: Instead of `make -j all`, split them into individual jobs with the actions and they run parallel.

.github/workflows/main.yml
azeemshaikh38: Thanks for the suggestion. `nproc` worked great!

.github/workflows/main.yml
azeemshaikh38: IIUC, we need to maintain order in our build process. For example, `update-dependencies` step should always run before any other step while `tree-status` should be the last step. Similarly, some steps which generate files like proto and mocks should run before we build. Explicitly specifying this ordering in the Makefile is the best thing I could come up with. It's not ideal, but if folks have better ideas, please do let me know.

q464pp:
可以拉取消合併嗎 謝謝

.github/workflows/main.yml
justaugustus: > split them into individual jobs with the actions and they run parallel.

+1, example from Kubernetes RelEng: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/release-utils/release-utils-presubmits.yaml

There we usually prefer to run separate Prow jobs to split them up along the following boundaries:
- unit/integration tests
- e2e tests
- verifies (linting, `go mod tidy` diffs, boilerplate)

naveensrinivasan:
@azeemshaikh38  Can we close this? Now that we have some parallelism.

azeemshaikh38:
> @azeemshaikh38 Can we close this? Now that we have some parallelism.

Done, thanks!

PR #1346, opened 2021-12-02
✨ Update action names

laurentsimon:
Update action names to satisfy the market place requirements.
Also added an option to share results - currently does nothing, though.
No breaking changes

### DIFF
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -4,6 +4,7 @@ on:
     # Only the default branch is supported.
     branches: [main, master]
   schedule:
+    # Weekly on Saturdays.
     - cron:  '30 1 * * 6'
   #pull_request:
     # All branches are supported.
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -3,6 +3,8 @@ on:
   push:
     # Only the default branch is supported.
     branches: [main, master]
+  schedule:
+    - cron:  '30 1 * * 6'
   #pull_request:
     # All branches are supported.
   #  branches: [main]
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -3,9 +3,9 @@ on:
   push:
     # Only the default branch is supported.
     branches: [main, master]
-  pull_request:
+  #pull_request:
     # All branches are supported.
-    branches: [main]
+  #  branches: [main]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/scorecard-analysis.yml
```diff
@@ -21,13 +21,18 @@ jobs:
...
       - name: "Run analysis"
-        # TODO: pin our action.
-        uses: ossf/scorecard-actions/analyze@test/dogfood
+        uses: ossf/scorecard-action@5fc8ff3ee41559cbd1079b561414c8fe3272afab
         with:
           policy_file: .github/scorecard-policy.yml
           results_file: results.sarif
           results_format: sarif
           repo_token: ${{ secrets.GITHUB_TOKEN }}
+          # Scorecard team runs a weekly scan of public GitHub repos,
```
laurentsimon, line 30: @olivekl I'd like to have you input on the text and command name. Basically, this option will change the behavior of the action and upload the results to our BigTable https://github.com/ossf/scorecard#public-data
We already have the data. But enabling sharing allows to offload the cost to GitHub infra instead of ours. It also reduces API rate limiting. In a nutshell, it helps us scale.
I don't want to scare users: we *already* have the data, so there's nothing bad about sharing it.

naveensrinivasan (approved):

.github/workflows/scorecard-analysis.yml
olivekl: How about: `streamline_results_sharing`? Is that too long?

For the comment text:

“The Scorecard team runs a weekly scan of public GitHub repositories in order to track the overall security health of the open source ecosystem. Setting `streamline_results_sharing: true` simplifies this process by sharing your results directly from your workflow, so we do not have to run the scans again. **No additional data will be shared** by setting `streamline_results_sharing: true`. It simply helps us scale by cutting down on repeated workflows and API requests.

See  https://github.com/ossf/scorecard#public-data for more information.”

azeemshaikh38: How about something like `generate_attestation`? 

Uploading to the weekly cron is *one* of the many things we'll be using it for. This boolean really generates a claim by Scorecard so that others can consume.

laurentsimon: thanks both. Another idea is to call it `badge_support` or `enable_badges`: if users want to display badges, they would ave to enable this option and we can have a disclaimer similar to @olivekl 's above to say that it implies sharing the results. Any more thoughts?

Issue #1353, opened 2021-12-03
BUG: --local option is broken


laurentsimon:
at this line https://github.com/ossf/scorecard/blob/main/cmd/root.go#L322, we're creating clients that use GitHub APIs. When run on a local repo, this forces users to define a GitHub token which is not needed.

These clients should only be created if they are requested by a user who runs scorecard, i.e. if the checks are requested.

I think the fix is as simple as leaving these variables `nil` if the checks are not requested by users.

laurentsimon:
@azeemsgoogle can you tackle this? 

fyi, @naveensrinivasan this is another case that requires unit test to prevent it from happening again.

laurentsimon:
I'll take it, since I needed to fix it for testing anyway.

azeemshaikh38:
Ouch. I had also created a commit for it. We can go with your PR itself. Just added comment to your PR.

laurentsimon:
closed by https://github.com/ossf/scorecard/pull/1355

PR #1354, opened 2021-12-03
✨ Fix local repo asking for GitHub token

laurentsimon:
See https://github.com/ossf/scorecard/issues/1353
no breaking changes

close https://github.com/ossf/scorecard/issues/1353

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -319,13 +319,6 @@ var rootCmd = &cobra.Command{
 		}
 		defer repoClient.Close()
 
-		ciiClient := clients.DefaultCIIBestPracticesClient()
-		ossFuzzRepoClient, err := githubrepo.CreateOssFuzzRepoClient(ctx, logger)
-		if err != nil {
-			log.Fatal(err)
-		}
-		defer ossFuzzRepoClient.Close()
-
 		// Read docs.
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -342,6 +335,21 @@ var rootCmd = &cobra.Command{
...
+			ciiClient = clients.DefaultCIIBestPracticesClient()
+		}
+
+		if _, exists := enabledChecks[checks.CheckCIIBestPractices]; exists {
+			ossFuzzRepoClient, err = githubrepo.CreateOssFuzzRepoClient(ctx, logger)
+			if err != nil {
+				log.Fatal(err)
+			}
+			defer ossFuzzRepoClient.Close()
+		}
```
azeemshaikh38, line 351: Instead of basing this on whats enabled, we can use repoType. Basically, update `getRepoAccessors` to return both CII and OssFuzz clients.

azeemshaikh38: So in case of `local` these will be `nil` irrespective of which check is enabled.

laurentsimon: does it help to do that? These are actually needed unless they're enabled, right?

laurentsimon:
superseded by another PR. Closing.

Issue #1340, opened 2021-11-26
scorecard shouldn't complain about actions in the same repository


evverx:
It's possible to reference actions that are kept in the same repository using `./path/to/dir`: https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#referencing-an-action-in-the-same-repository-where-a-workflow-file-uses-the-action. I don't think scorecard should complain about them.

laurentsimon:
Great catch, thanks! Feel free to send a PR and cc me on it for approval.

laurentsimon:
see https://github.com/ossf/scorecard/pull/1357

PR #1357, opened 2021-12-03
✨ Ignore local actions that are not pinned

laurentsimon:
It makes no sense to warn users if they use locally-defined actions
closes https://github.com/ossf/scorecard/issues/1340
no breaking changes


checks/pinned_dependencies.go
```diff
@@ -602,6 +602,14 @@ func validateGitHubActionWorkflow(pathfn string, content []byte,
 				// Cannot check further, continue.
 				continue
 			}
+
+			// nolint:lll
+			// Check whether this is an action defined in the same repo,
+			// https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#referencing-an-action-in-the-same-repository-where-a-workflow-file-uses-the-action.
+			if strings.HasPrefix(execAction.Uses.Value, "./") {
```
evverx, line 609: Thanks! As far as I understand it's possible to reference local actions using "{owner}/{repo}" but I don't think I have ever seen anyone really use it (probably because "./" is shorter :-)

laurentsimon: good point. Hopefully this will encourage users to use the right idiom:-)

naveensrinivasan (approved):

oliverchang (approved):

PR #1370, opened 2021-12-06
✨ update log msg for non-pinned actions

laurentsimon:
Add whether the non-pinned action is GitHub-owned or not.
No breaking changes.


naveensrinivasan (approved):
Thanks

PR #1372, opened 2021-12-07
✨ [DRAFT: RAW]: Security policy support

laurentsimon:
PR separates checks from policy for the Security-Policy check.
No breaking changes on output.

### DIFF
--- a/checks/raw/binary_artifact.go
+++ b/checks/raw/binary_artifact.go
@@ -86,10 +86,17 @@ func checkBinaryFileContent(path string, content []byte,
 	exists1 := binaryFileTypes[t.Extension]
 	exists2 := binaryFileTypes[strings.ReplaceAll(filepath.Ext(path), ".", "")]
 	if exists1 || exists2 {
+<<<<<<< HEAD
 		*pfiles = append(*pfiles, checker.File{
 			Path:   path,
 			Type:   checker.FileTypeBinary,
 			Offset: 1,
+=======
+		*pfiles = append(*pfiles, File{
+			Path:   path,
+			Type:   checker.FileTypeBinary,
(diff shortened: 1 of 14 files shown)


checks/raw/binary_artifact.go
```diff
@@ -30,7 +30,7 @@ import (
 
 // BinaryArtifacts retrieves the raw data for the Binary-Artifacts check.
 func BinaryArtifacts(c clients.RepoClient) (checker.BinaryArtifactData, error) {
-	var files []checker.File
+	files := []checker.File{}
```
azeemshaikh38, line 33: If you want to create an empty slice prefer: `make([]checker.File, 0)` to make the intent explicit.

docs/checks/internal/validate/main.go
```diff
@@ -70,6 +71,10 @@ func listCheckFiles() (map[string]string, error) {
 			continue
 		}
 
+		// WARNING: assume the filename is the same during migration
```
azeemshaikh38, line 74: Unrelated to this PR - can we do away with these validations where we string parse the checks and the API usage?

azeemshaikh38 (approved):
Thanks!

checks/raw/binary_artifact.go
laurentsimon: thanks.

docs/checks/internal/validate/main.go
laurentsimon: Using a CFG is better but requires more work. If you have something specific in mind that's easy to implement, sure. The string based approach currently works, so I don't see this as a priority unless it keeps failing.
What do you have in mind?

PR #1373, opened 2021-12-07
✨ Don't use the policy filename in SARIF results

laurentsimon:
This uses a default "no-source-file" for results that don't have explicit files to display.
Since we're removing policy files from the GH action, that makes it less confusing for users.


azeemshaikh38 (changes requested):
As discussed offline, let's hold on this until our sync.

laurentsimon:
friendly ping to LGTM, as discussed offline

azeemshaikh38 (approved):

Issue #1260, opened 2021-11-14
The "Code Review" check seems to reject "reviewed" PRs


evverx:
Looking at https://github.com/ossf/scorecard/blob/51de6b6e5d9b025561b15c30cbb498bf31101427/checks/code_review.go#L118-L128

and https://github.com/systemd/systemd/pull/21305 (which was rejected by the check because due to the "rebase" button it was presumably created, reviewed and merged by the same person as far as I can tell) I don't think the check should always rely on whatever `pr.MergeCommit.Committer.Login` is. It seems `.merged_by.login` could have helped in this case:
```
curl  --silent  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/systemd/systemd/pulls/21305 |
jq .merged_by.login,.user.login
```
```
"poettering"
"ml-"
```

azeemshaikh38:
Is this now fixed?

evverx:
I don't think it's fixed. https://github.com/ossf/scorecard/pull/1272 is still open because as far as I understand it needs more tests

azeemshaikh38:
Thanks @evverx. Is #1272 the only pending thing here?

evverx:
Yes, I think it is.

As far as I can remember I ran into another issue with the "code review" check but I haven't figured out what it was yet so I think it's safe to assume that it has nothing to do with this issue :-)

I'm not sure if it helps but with #1272 applied the systemd code-review score went from
```
      "score": 7,
      "reason": "GitHub code reviews found for 23 commits out of the last 30 -- score normalized to 7",
```
to
```
      "score": 9,
      "reason": "GitHub code reviews found for 29 commits out of the last 30 -- score normalized to 9",
```

evverx:
As far as I understand scorecard switched to commits and associated PRs (where it should be relatively easy to figure out who commiters and authors are). I'm not sure whether it works as expected in its current form but I think this issue is kind of out of date and can be closed.

PR #1375, opened 2021-12-07
✨ Code-Review: show PRs merged without code review

evverx:
to make it easier to figure out whether those PRs are really merged
without code review or whether there is a bug in scorecard like
https://github.com/ossf/scorecard/issues/1260 that prevents it
from finding reviewed PRs. Other than that, the "CI-Tests" check
already shows "untested" PRs so it seems the "Code-Review" check
should follow suit.


laurentsimon (approved):
Awesome, thanks!

PR #1379, opened 2021-12-09
✨ Comply with GH-specific rules for SARIF

laurentsimon:
This makes a few changes to comply with GH-specific rules:
1. rule ID should be different from rule name
2. rule ID should be pascal case

In addition, I've create a category for the branch protection check, to allow us to run it alone without running other checks - see https://github.com/ossf/scorecard-action/issues/19


naveensrinivasan (approved):
Didn't test it. Thanks!

azeemshaikh38 (approved):

Issue #886, opened 2021-08-23
Feature: add line number to Token-Permissions detail reporting


laurentsimon:
see the title. It's identical to https://github.com/ossf/scorecard/issues/725 but for `Token-Permissions` check.

laurentsimon:
FYI @chrismcgehee

laurentsimon:
@chrismcgehee I think you're working on it, right? Assigning to you. If you're not, please un-assign.

laurentsimon:
Does https://github.com/ossf/scorecard/pull/1363 close this issue?

chrismcgehee:
> Does #1363 close this issue?

It does not. These functions (and the functions they call) still need the line number to be set correctly:
- [createIgnoredPermissions](https://github.com/ossf/scorecard/blob/86fd966dc41886edec168e071ee11227ea262c71/checks/permissions.go#L400)
- [isSARIFUploadWorkflow](https://github.com/ossf/scorecard/blob/86fd966dc41886edec168e071ee11227ea262c71/checks/permissions.go#L416)
- 
The fix here will be more involved because we'll have to go several steps up in the call stack to pass in a structured object instead of a string.

Issue #1100, opened 2021-10-04
Make Packaging check 'job'-aware


laurentsimon:
currently the Token-Permission workflow works at the workflow level to decide whether certain dangerous permissions are acceptable (`requiresPackagesPermissions` and `isSARIFUploadWorkflow`). We should make the check 'job'-aware, since jobs run isolated from each other.

laurentsimon:
FYI @chrismcgehee

PR #1381, opened 2021-12-12
✨ Adding line numbers for rest of Token-Permissions (and by extension, Packaging)

chrismcgehee:
Packaging)

Feature

When the Token-Permission check makes logs for packaging workflow, it does not set the line number. Same thing for the Packaging check.
Closes https://github.com/ossf/scorecard/issues/886

All logs from the Token-Permission and Packaging checks will include line numbers. This PR also makes the check for packaging workflow more precise. Instead of doing a string search for an action name, it will parse the workflow file and check that a `uses` field on a step matches the action name.

No

* **Other information**:

### DIFF
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/rhysd/actionlint"
 
+	"github.com/ossf/scorecard/v3/checker"
 	sce "github.com/ossf/scorecard/v3/errors"
 )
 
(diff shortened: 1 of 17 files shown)


checks/testdata/github-workflow-permissions-run-package-workflow-write.yaml
```diff
@@ -22,5 +22,4 @@ jobs:
       packages: write
     steps:
       - name: some name
-        run: echo "write-and-read workflow"
```
chrismcgehee, line 25: Removed this line because parsing this file caused an error. A step is not allowed to have both a `run` and a `uses` field.

azeemshaikh38: Nit suggestion: this is a good corner case to test for. We should probably have a test which tests that we handle these parsing issues gracefully.

azeemshaikh38 (approved):
Thanks @chrismcgehee! Great PR as always.

laurentsimon:
Thanks @chrismcgehee , this is really cool!

I think this PR closes https://github.com/ossf/scorecard/issues/1100, is that right?

closes https://github.com/ossf/scorecard/issues/1100

chrismcgehee:
> I think this PR closes #1100, is that right?

Maybe it does; I'll comment on that issue.

chrismcgehee:
> Nit suggestion: this is a good corner case to test for. We should probably have a test which tests that we handle these parsing issues gracefully.

I'm all for testing corner cases, but I don't think this case is a good fit for our existing unit tests. If a workflow file does not conform to Github's specifications for a workflow file, I think it should be indeterminate what our code does; in this case it will still run the token-permissions check on this file, but it would treat is as a `run` step and ignore the `uses`. I think that behavior is fine, and I also think it would be fine if our code reported an error on this file. What we need to do is create a larger test that checks multiple files and continues checking if 1 of the files has some sort of error.

PR #1385, opened 2021-12-13
:seedling: Fix the stale configuration.

naveensrinivasan:
The number of issues and PR aren't getting attention and this will help
us with this.

* **Other information**:

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -17,7 +17,7 @@ on:
   schedule:
   - cron: "30 1 * * *"
 
-permissions: read-all
+permissions: 
 
 jobs:
   stale:
@@ -36,3 +36,7 @@ jobs:
         stale-pr-label: 'no-pr-activity'
         exempt-pr-labels: 'awaiting-approval,work-in-progress'
         only-labels: 'awaiting-feedback,awaiting-answers'
+        days-before-pr-stale: '10'
+        days-before-pr-close: '20'
+        days-before-issue-stale: '60'
+


azeemshaikh38 (approved):
Great, thanks! Will wait for @laurentsimon also to chime in here.

.github/workflows/stale.yml
```diff
@@ -36,3 +36,7 @@ jobs:
         stale-pr-label: 'no-pr-activity'
         exempt-pr-labels: 'awaiting-approval,work-in-progress'
         only-labels: 'awaiting-feedback,awaiting-answers'
+        days-before-pr-stale: '10'
+        days-before-pr-close: '20'
+        days-before-issue-stale: '60'
```
laurentsimon, line 41: what happens when something becomes stale? Is the issue labeled?

Most issues have no updates in practice, right? I'm worried that issues labeled as stale may deter external contributors from working on them.

On the other hand, having a stale label would help us decide whether we should close them manually or not, which seems like a good idea. How do we remove the stale issue? Manually? Or we're supposed to add a comment to restart the 60 day period?

.github/workflows/stale.yml
```diff
@@ -36,3 +36,7 @@ jobs:
         stale-pr-label: 'no-pr-activity'
         exempt-pr-labels: 'awaiting-approval,work-in-progress'
         only-labels: 'awaiting-feedback,awaiting-answers'
+        days-before-pr-stale: '10'
+        days-before-pr-close: '20'
```
laurentsimon, line 40: I have PRs that are probably more than 20 days that I don't want be closed automatically (those with `RAW` in their title). Will I have to re-open them? Maybe we should only label the PRs as stale, but not automatically close them (since there aren't a lot of PRs anyway, it should not add to much work for us). wdyt?

Do PRs labeled "Draft" also get closed automatically?

.github/workflows/stale.yml
```diff
@@ -17,7 +17,7 @@ on:
   schedule:
   - cron: "30 1 * * *"
 
-permissions: read-all
+permissions:
```
laurentsimon, line 20: Since no permissions are declared, there's nothing the action can do, e.g., label the issue, etc. So what does the action do?

naveensrinivasan: The specific job has permissions.
```
permissions:
      issues: write
      pull-requests: write
```

.github/workflows/stale.yml
naveensrinivasan: We can add a label that will not mark the PR as stale or close it. For example, this label will avoid it as stale`work-in-progress`

.github/workflows/stale.yml
naveensrinivasan: `If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart` also we can have exempt-labels to avoid marking an issue as stale.

.github/workflows/stale.yml
laurentsimon: sgtm. Let's pilot it and see how well it works then :-)

Issue #1366, opened 2021-12-06
Feature: generate SBOMs based on SPDX format supported by ko


developer-guy:
**Is your feature request related to a problem? Please describe.**
No, this is not related to a problem.

**Describe the solution you'd like**

We've worked on building container images by making use of the `google/ko` project recently[^1]. Then I realized that `google/ko` has the support of generating SBOMs based on SPDX format, and it'll be released soon[^2]. Maybe we might want to use this feature to generate SBOMs for the scorecard container images.

**Describe alternatives you've considered**

Alternatively, there are various tooling around this[^3][^4][^5][^6] concept to generate it There are several methods designed specifically for delivering SBOMs, including [SPDX (Software Package Data Exchange)](https://spdx.dev), [Software Identification (SWID) Tags](https://nvd.nist.gov/products/swid), and [Cyclone DX](https://cyclonedx.org).

**Additional context**

[^1]: https://github.com/ossf/scorecard/pull/1127
[^2]: https://github.com/google/ko/pull/511
[^3]: https://github.com/anchore/syft
[^4]:  https://github.com/spdx/spdx-sbom-generator
[^5]: https://github.com/tern-tools/tern
[^6]: https://github.com/kubernetes/release/tree/master/cmd/bom

cc: @naveensrinivasan @dentrax

laurentsimon:
what info does `ko` support for the output? What I mean is, it seems to support listing the goland project's dependencies using `go version -m`. What else? Does it capture dependencies installed via `apt-get`, etc?

I have not played around with `google/ko`, so forgive me if those questions make no sense :-)

developer-guy:
it also supports SPDX format which you see [here](https://github.com/spdx/spdx-examples) by using a flag named "--sbom=spdx", IMHO, it should be the default value btw. Here is the result from my binary by running the following commands:

```shell
$ KO_DOCKER_REPO=docker.io/devopps VERSION=123 ko build --sbom spdx github.com/GoTurkiye/hello-world-cli
$ crane ls docker.io/devopps/hello-world-cli-41ae0754a7c4a818c3a0d29b10e48ca5
latest
sha256-1257a5feeb1ac7e14a56203f015086ca47342d0b05f75dccd94624cc8cfccb7f.sbom # will create this file
[...]

PR #1392, opened 2021-12-15
🐛 Fix ko build workflows in Makefile

imjasonh:
Progress toward https://github.com/ossf/scorecard/issues/1366

Bugfixes

`make ko-build-everything` fails for me.

`make ko-build-everything` builds images, but does not push them.

Images are built for all binaries, not just `scorecard`, and (if `--push=false` is removed) are pushed to, e.g.:

```
gcr.io/imjasonh/scorecard-github-server/server@sha256:a142bd9ad872d94fc55e3b6b92b24d862f0ebfd044fec1154d1e6872544e0a1a
```

SBOMs are also attached to each platform-specific image:

```
gcr.io/imjasonh/scorecard-github-server/server:sha256-3bf12ae036f73fa4d93f3dd16b5be8ee6f9ba2a8796ada98e1d8c9a545b41304.sbom
```

Binaries inside images are built using Go 1.17 (as it's the base for the `ko` image).

It's possible, but I doubt it, especially since images aren't pushed in this PR.

* **Other information**:

cc @developer-guy @naveensrinivasan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,36 +197,43 @@ ko-build-everything: $(ko-targets)
 scorecard-ko:
 	KO_DATA_DATE_EPOCH=$(SOURCE_DATE_EPOCH) KO_DOCKER_REPO=${KO_PREFIX}/scorecard LDFLAGS="$(LDFLAGS)" \
 	ko publish -B \
+			   --push=false \
 			   --platform=$(PLATFORM)\
 			   --tags latest,$(GIT_VERSION),$(GIT_HASH) github.com/ossf/scorecard/v3
 cron-controller-ko:
 	KO_DATA_DATE_EPOCH=$(SOURCE_DATE_EPOCH) KO_DOCKER_REPO=${KO_PREFIX}/$(IMAGE_NAME)-batch-controller LDFLAGS="$(LDFLAGS)" \
 	ko publish -B \
+			   --push=false \
 			   --platform=$(PLATFORM)\
 			   --tags latest,$(GIT_VERSION),$(GIT_HASH) github.com/ossf/scorecard/v3/cron/controller
 cron-worker-ko:
 	KO_DATA_DATE_EPOCH=$(SOURCE_DATE_EPOCH) KO_DOCKER_REPO=${KO_PREFIX}/$(IMAGE_NAME)-batch-worker LDFLAGS="$(LDFLAGS)" \
 	ko publish -B \
+			   --push=false \
 			   --platform=$(PLATFORM)\
 			   --tags latest,$(GIT_VERSION),$(GIT_HASH) github.com/ossf/scorecard/v3/cron/worker
 cron-cii-worker-ko:
 	KO_DATA_DATE_EPOCH=$(SOURCE_DATE_EPOCH) KO_DOCKER_REPO=${KO_PREFIX}/$(IMAGE_NAME)-cii-worker LDFLAGS="$(LDFLAGS)" \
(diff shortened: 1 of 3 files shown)


naveensrinivasan (commented):
>This relies on a new image ghcr.io/google/ko, which is built from ko at HEAD. This might introduce unintended breakages until this behavior is released and tagged with a stable tag. If you'd rather wait until then, that's totally fine, I just wanted to get the conversation started.

Great! Thanks, When is it likely to be merged and tagged? 

So we could essentially remove all other cloudbuild.yaml files once this one is merged?

Makefile
```diff
@@ -195,46 +195,39 @@ ko-targets = scorecard-ko cron-controller-ko cron-worker-ko cron-cii-worker-ko c
 ko-build-everything: $(ko-targets)
```
naveensrinivasan, line 195: We need something to test in CI so that it doesn't push to the registry. Would this change work with that? We don't have access from the GH to GCP so all the container builds happen only in GCP with cloudbuild

imjasonh: If this is intended just for testing that `ko` builds work, then `--push=false` is probably what you want (and not `--local`, unless you want to make the resulting image available in the Docker daemon, which it looks like you don't)

I've added `--push=false` to all of these for now, and reverted `cloudbuild/scorecard.yaml` since it won't actually push your image now. We can figure out how we want to integrate `ko publish` into your actual releases in another PR.

imjasonh:
> Great! Thanks, When is it likely to be merged and tagged?
> 
> So we could essentially remove all other cloudbuild.yaml files once this one is merged?

Unclear, there's [some work we'd like to do before ko v0.10](https://github.com/google/ko/milestone/2), at which point this will be tagged as `ghcr.io/google/ko:v0.10`).

You can either have one cloudbuild.yaml that builds and pushes everything, or individual cloudbuild.yamls that each build and push one component, or both if you'd prefer. Really whatever you want.

naveensrinivasan:
>You can either have one cloudbuild.yaml that builds and pushes everything, or individual cloudbuild.yamls that each build and push one component, or both if you'd prefer. Really whatever you want.

We could choose either IMO. I would let others chime in on this. 

My preference would be a single cloubuild.yaml

azeemshaikh38:
Thanks @imjasonh for starting this PR, this is great! My feedback:

- As long as this PR does not push images (which looks to be true AFAICT), we can go ahead with submitting this PR. 
- Before we start pushing images using `ko`, we need to figure out how to handle cases like https://github.com/ossf/scorecard/blob/main/cron/controller/Dockerfile which generate and attach data along with the image. Maybe you have some ideas there @imjasonh?
- W.r.t single `cloudbuild.yaml` vs. multiple. I'm ok with either. Would probably lean more towards the multiple files model to maintain status quo, but honestly don't think it matters.

imjasonh:
> * Before we start pushing images using `ko`, we need to figure out how to handle cases like https://github.com/ossf/scorecard/blob/main/cron/controller/Dockerfile which generate and attach data along with the image. Maybe you have some ideas there @imjasonh?

That seems like a good fit for using [`kodata`](https://github.com/google/ko#static-assets). When you `ko publish`, if there's anything in a directory named `kodata`, `ko` will pack it up into a separate layer, and inject an env var `KO_DATA_PATH` where that data can be found in the running container.

This might take some work to get fully sorted out, but I don't think there's any reason this shouldn't work. I'll try to tackle that next.

PR #1395, opened 2021-12-15
:seedling: Included cron data when using KO

naveensrinivasan:
@imjasonh Does this look good for including the data files?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,7 @@ scorecard-ko:
 			   --platform=$(PLATFORM)\
 			   --tags latest,$(GIT_VERSION),$(GIT_HASH) github.com/ossf/scorecard/v3
 cron-controller-ko:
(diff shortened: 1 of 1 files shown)


imjasonh:
> @imjasonh Does this look good for including the data files?

Unfortunately this won't do what it looks like.

When `ko publish` is run, it looks for a directory under the importpath being build called `kodata/` (there's no renaming this directory). `ko` will bundle up everything in that directory into a new layer in your image, at some path, and expose that path to your code as `KO_DATA_PATH` -- that env var is available at runtime, not at build-time.

So the change you'd need is to:
1. create a directory `cron/controller/kodata`, which either contains the data currently in `cron/data`, or is a symlink to that directory (`ko` will chase symlinks when creating the `kodata` layer).
2. have your code load that data from `filepath.Join(os.Getenv("KO_DATA_PATH"), projects.release.csv)`, instead of taking the location as arg[0] ([as here](https://github.com/ossf/scorecard/blob/a13b63eae27192342429a423ebb5f4afb4a7bd2f/cron/controller/main.go#L87)).

This won't shuffle the data like you seem to be doing, I'm not sure why you do that, or if not doing it would cause problems.

PR #1396, opened 2021-12-15
✨ [RAW] Branch Protection support

laurentsimon:
Add support for raw results for Branch-Protection check.
No breaking changes.
Output:
For a repo with admin access:
```
{
  "results": {
    ...
    "branch-protections": [
      {
        "protection": {
          "required-reviewer-count": 1,
          "allows-deletions": false,
          "allows-force-pushes": false,
          "requires-code-owner-review": false,
          "required-linear-history": true,
          "dismisses-stale-reviews": false,
          "enforces-admin": true,
          "requires-status-checks": null,
          "requires-updated-branches-to-merge": true,
          "status-checks-contexts": [
            "Validate",
            "license boilerplate check",
            "integration-trusted"
          ]
        },
        "name": "main"
      }
    ]
  }
}

```

For a repo *without* admin access:
```
    ...
    "branch-protections": [
      {
        "protection": {
          "required-reviewer-count": 0,
          "allows-deletions": false,
          "allows-force-pushes": false,
          "requires-code-owner-review": false,
          "required-linear-history": false,
          "dismisses-stale-reviews": null,
          "enforces-admin": null,
          "requires-status-checks": null,
          "requires-updated-branches-to-merge": null,
          "status-checks-contexts": []
        },
        "name": "main"
      }
    ]
  }
}
```


laurentsimon:
cc @naveensrinivasan @azeemsgoogle  e2e test are failing:
```
with:
    max_attempts: 3
    retry_on: error
    command: make ci-e2e
    retry_wait_seconds: 10
    polling_interval_seconds: 1
    warning_on_retry: true
    continue_on_error: false
  env:
    GOROOT: /opt/hostedtoolcache/go/1.17.4/x64
    GITHUB_AUTH_TOKEN: ***
```

Sorry, I have not been following some of the recent updates to the workflow.

checker/raw_result.go
```diff
@@ -71,6 +32,15 @@ type MergeRequest struct {
 	// TODO: add fields, e.g., State=["merged"|"closed"]
 }
 
+// File represents a file.
+type File struct {
+	Path    string
```
laurentsimon, line 37: for some reasons that had disappeared after merging frm main. This code was added in previous PRs.

checker/raw_result.go
```diff
@@ -84,3 +54,60 @@ type Tool struct {
 	Desc          string
 	ConfigFiles   []File
 }
+
+// SecurityPolicyData contains the raw results
```
laurentsimon, line 58: A lot of this code was added in previous PRs and somewhat disappeared too. Not really part of this PR.

checks/branch_protection_test.go
```diff
@@ -432,286 +423,15 @@ func TestReleaseAndDevBranchProtected(t *testing.T) {
...
+			r := BranchProtection(&req)
 			if !scut.ValidateTestReturn(t, tt.name, &tt.expected, &r, &dl) {
 				t.Fail()
 			}
 			ctrl.Finish()
 		})
 	}
 }
-
-func TestIsBranchProtected(t *testing.T) {
```
laurentsimon, line 444: this test was moved to `raw/branch_protection_test.go`

checks/raw/branch_protection.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/ossf/scorecard/v3/checker"
+	"github.com/ossf/scorecard/v3/clients"
+	sce "github.com/ossf/scorecard/v3/errors"
+)
+
+type branchMap map[string]*clients.BranchRef
+
+// BranchProtection retrieves the raw data for the Branch-Protection check.
+func BranchProtection(c clients.RepoClient) (checker.BranchProtectionsData, error) {
```
laurentsimon, line 30: same code as before but storing information in the raw structure. Nothing new really.

pkg/json_raw_results.go
```diff
@@ -46,6 +46,24 @@ type jsonTool struct {
...
+	RequiresLinearHistory               *bool    `json:"required-linear-history"`
+	DismissesStaleReviews               *bool    `json:"dismisses-stale-reviews"`
+	EnforcesAdmins                      *bool    `json:"enforces-admin"`
+	RequiresStatusChecks                *bool    `json:"requires-status-checks"`
+	RequiresUpToDateBranchBeforeMerging *bool    `json:"requires-updated-branches-to-merge"`
+	StatusCheckContexts                 []string `json:"status-checks-contexts"`
+}
+
+type jsonBranchProtection struct {
+	Protection *jsonBranchProtectionSettings `json:"protection"`
```
laurentsimon, line 63: we use a pointer: if the branch is not protected, nothing is populated and `Protection` is set to `nil`

checker/raw_result.go
```diff
@@ -84,3 +54,60 @@ type Tool struct {
...
+	Name                string
+}
+
+// BranchProtectionsData contains the raw results
+// for the Branch-Protection check.
+type BranchProtectionsData struct {
+	Branches []BranchProtectionData
+}
+
+// RawResults contains results before a policy
```
azeemshaikh38, line 106: Nit: put the RawResults struct as the first struct in the file. Readability is better when the main struct is first and all the nested structs follow.

checker/raw_result.go
```diff
@@ -84,3 +54,60 @@ type Tool struct {
...
+	// Tools contains a list of tools.
+	// Note: we only populate one entry at most.
+	Tools []Tool
+}
+
+// BranchProtectionData contains the raw results
+// for one branch.
+//nolint:govet
+type BranchProtectionData struct {
+	Protected                           *bool
```
azeemshaikh38, line 84: This makes me wonder should we follow the same pattern for other raw structs too? That is ptr to primitive types. Can do it in a separate PR if you agree.

azeemshaikh38 (approved):
Thanks!

checker/raw_result.go
laurentsimon: the pointer `nil` means we could not retrieve the information. Other checks don't have this problem, right?
I've not thought this through, but I'm not sure it would help for the other checks I have migrated, like Security-Policy, Dependency-Tool and Binary-Artifacts. Do you have another idea? Which checks do you you think would benefit from it?

checker/raw_result.go
laurentsimon: makes sense. I think I inadvertently un-did what you did in the previous PR.. the RawResults was top of the file. My bad!

laurentsimon:
interesting error from integration tests in the latest run:
```
make: *** [Makefile:137: build-scorecard] Terminated
Error: Timeout of 30000ms hit
```

cc @naveensrinivasan @azeemshaikh38

azeemshaikh38:
> interesting error from integration tests in the latest run:
> 
> ```
> make: *** [Makefile:137: build-scorecard] Terminated
> Error: Timeout of 30000ms hit
> ```
> 
> cc @naveensrinivasan @azeemshaikh38

Thi is now fixed.

checker/raw_result.go
azeemshaikh38: > the pointer `nil` means we could not retrieve the information. Other checks don't have this problem, right?

Well, it should be `nil` even if we don't run the check. Which can happen with any check.

IMO it'll be useful to have a consistent convention across checks. Anyways, this is not a blocking issue. We can migrate all checks to the raw/evaluation format and then reconsider this.

Issue #1400, opened 2021-12-16
BUG: Token-Permission check outputs wrong score


laurentsimon:
The Token-Permission check stores all wokflows' results in a single map. This mean we cannot differentiate between workflow files. Running scorecard on our repo returns 0. That's because we originally only looked for top-level permission being defined, but this recently changed.

PR #1401, opened 2021-12-16
🐛  Fix score calculation for multiple files

laurentsimon:
Token permission check does not handle multiple files for score calculation, due to the fact that all workflows shared the same permission map. This is because the check used to only look for undefined top-level permissions, globally. Now we've switched to looking for both run-level and top-level permissions. We want to only flag when some files don't declare both of them; or if only the run-level permissions are declared.

closes https://github.com/ossf/scorecard/issues/1400

### DIFF
--- a/checks/permissions.go
+++ b/checks/permissions.go
@@ -65,6 +65,7 @@ type permissions struct {
 }
 
 type permissionCbData struct {
+	// map of filename to write permissions used.
 	workflows map[string]permissions
 }
 
--- a/checks/permissions.go
+++ b/checks/permissions.go
@@ -32,6 +32,25 @@ const (
 	topLevelPermission    = "top level"
 )
 
+type permission string
+
+const (
+	permissionAll            = permission("all")
+	permissionStatuses       = permission("statuses")
+	permissionChecks         = permission("checks")
+	permissionSecurityEvents = permission("security-events")
+	permissionDeployments    = permission("deployments")
+	permissionContents       = permission("contents")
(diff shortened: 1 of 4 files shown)


laurentsimon:
error 
```
make: *** [Makefile:137: build-scorecard] Terminated
Error: Timeout of 30000ms hit
```

seems to be happening for all PRs now
cc @naveensrinivasan @azeemsgoogle

checks/permissions.go
```diff
@@ -244,66 +261,73 @@ func calculateScore(result permissionCbData) int {
...
-	// status: https://docs.github.com/en/rest/reference/repos#statuses.
-	// May allow an attacker to change the result of pre-submit and get a PR merged.
-	// Low risk: -0.5.
-	if permissionIsPresent(result, "statuses") {
-		score -= 0.5
-	}
+		// status: https://docs.github.com/en/rest/reference/repos#statuses.
+		// May allow an attacker to change the result of pre-submit and get a PR merged.
+		// Low risk: -0.5.
+		if permissionIsPresent(perms, "statuses") {
```
azeemshaikh38, line 280: Nit: make these strings - "all", "statuses", "checks" etc. constants or better yet define a new type for these strings. It helps avoid bugs due to typos or definition mismatches.

azeemshaikh38 (approved):
Thanks. Added a nit comment to fix.

checks/permissions.go
laurentsimon: done, thanks.

PR #1405, opened 2021-12-20
✨ Do not expose sarif and policy command

laurentsimon:
In order to simplify maintenance and support, we won't publicly advertise support for sarif and the score-based policy file. 
1. sarif will be used internally for the action. Currently, we run the scorecard CLI; but long-term we may have a dedicated CLI for the action. 
2. The policy file will be replaced by a setting-based policy file once we have finished separating checks from policy score-based policy calculation.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -147,10 +147,6 @@ func scorecardCmd(cmd *cobra.Command, args []string) {
 		log.Panic("policy not supported yet")
 	}
 
-	if local != "" && !sarifEnabled {
-		log.Panic("--local option not supported yet")
-	}
-
 	var v6 bool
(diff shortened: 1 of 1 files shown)


laurentsimon:
same error as for my other PR:
```
> [internal] load metadata for docker.io/library/golang@sha256:3c4de86eec9cbc619cdd72424abd88326ffcf5d813a8338a7743c55e5898734f:
------
failed to solve with frontend dockerfile.v0: failed to create LLB definition: failed to authorize: rpc error: code = Unknown desc = failed to fetch oauth token: unexpected status: 503 Service Unavailable
```

Re-starting but I suppose it will fail again.
cc @azeemsgoogle @naveensrinivasan

azeemshaikh38 (approved):

Issue #1245, opened 2021-11-12
Feature: Separate check for policy/score evaluation


laurentsimon:
We need to separate check and its policy evaluation. This will allow us to return 'raw' results for users to apply arbitrary policies.
The first step is to do the separation within the `checks` package. Then all checks are migrated, we will create a `pkg.RunRawScorecards()` and create our default policy (currently the scores) thru an additional call.
We will later expose this in the CLI

naveensrinivasan:
> We need to separate check and its policy evaluation. This will allow us to return 'raw' results for users to apply arbitrary policies.
> The first step is to do the separation within the `checks` package. Then all checks are migrated, we will create a `pkg.RunRawScorecards()` and create our default policy (currently the scores) thru an additional call.
> We will later expose this in the CLI

Cool! Is this part of v4 release?

laurentsimon:
> > We need to separate check and its policy evaluation. This will allow us to return 'raw' results for users to apply arbitrary policies.
> > The first step is to do the separation within the `checks` package. Then all checks are migrated, we will create a `pkg.RunRawScorecards()` and create our default policy (currently the scores) thru an additional call.
> > We will later expose this in the CLI
> 
> Cool! Is this part of v4 release?

no. Just too many checks to migrate :-) v5 or v6.

azeemshaikh38:
@laurentsimon assigning to you for now.

laurentsimon:
cc @richsalz this is the  large blocker that will help us provide more granularity in the results and give is the ability to provide remediation steps for each warning. (something that came up in https://github.com/ossf/scorecard/issues/1500)

richsalz:
Thanks for CC'ing me, @laurentsimon!  Have you thought of assigning each check a unique ID, and then allowing command-line flags like `--ignore filename:id` as in `--ignore .github/workflows/ci.yml:unpinned`

laurentsimon:
We thought about something along these lines early on. I think we'll achieve what you describe with policies. Once we have the separation of checks vs policy, we'll be able to apply `--ignore .github/workflows/ci.yml:unpinned`, and even more expressive policies such as `ignore unpinned action org/name`. The file approach works well on simple examples, but is hard to generalize to other uses case like, say "warn me if a reviewers is not in list X,Y,Z".

We feel the policies will allow flexibility for advanced users, without bloating the CLI arguments. We will have built-in policies, of course.

Wdut?

richsalz:
[...]

PR #1408, opened 2021-12-22
:seedling: [W.I.P] feat(listing): add ignorelist for scanning

### LINKED ISSUE #1406: add ability to pass ignore list
**What would you like to be added**:

We are skipping the binary-check for the `testdata` folders as implemented in [`isTestdataFile()`](https://github.com/ossf/scorecard/blob/090ae4f0bbc3b6956971bec83530c86696e1e75d/checks/fileparser/listing.go#L134) function.

Not sure but passing a new `ignoreDirs []string` field to struct [CheckRequest](https://github.com/ossf/scorecard/blob/090ae4f0bbc3b6956971bec83530c86696e1e75d/checker/check_request.go#L24) will solve the issue. (It should apply `*foo*` logic for every element that we [traversed](https://github.com/ossf/scorecard/blob/090ae4f0bbc3b6956971bec83530c86696e1e75d/checks/fileparser/listing.go#L151)?) Which we call in [BinaryArtifacts](https://github.com/ossf/scorecard/blob/090ae4f0bbc3b6956971bec83530c86696e1e75d/checks/binary_artifact.go#L33) function afterward.

**Why is this needed**:

To prevent [Binary-Artifacts false positives](https://github.com/ossf/scorecard/issues/1256).

From the discussion #593 started by @developer-guy; @westonsteimel [said](https://github.com/anchore/syft/issues/593#issuecomment-990745927) we probably need some enhancements to consider other names for test directories, as @laurentsimon already [proposed an idea](https://github.com/ossf/scorecard/pull/1263#pullrequestreview-806274849) to ability to skip _given_ folders in the config.

**Additional context**:

Out of context:  In the pull https://github.com/ossf/scorecard/pull/1263/files I'm not sure whether it's that _secure_ to check a single line of `if` command to trigger skip logic by force asserting `*test*` matching. Can someone please enlighten me about the following logic; wouldn't it be dangerous from the security perspective? If I put some vulnerable executables in random `*test*` folders?

```golang
if strings.Contains(strings.ToLower(path), "test")
```

Dentrax:
Signed-off-by: Furkan <furkan.turkal@trendyol.com> @Dentrax 
Co-authored-by: Batuhan <batuhan.apaydin@trendyol.com> @developer-guy 

Fixes #1406

See #1406 and https://github.com/anchore/syft/issues/593

We have the ability to pass ignorelist.

No

* **Other information**:

### DIFF
--- a/checker/check_request.go
+++ b/checker/check_request.go
@@ -28,6 +28,7 @@ type CheckRequest struct {
 	OssFuzzRepo clients.RepoClient
 	Dlogger     DetailLogger
 	Repo        clients.Repo
+	IgnoreDirs  []string
 	// UPGRADEv6: return raw results instead of scores.
 	RawResults *RawResults
 }
--- a/checks/fileparser/listing.go
+++ b/checks/fileparser/listing.go
@@ -49,10 +49,24 @@ func isMatchingPath(pattern, fullpath string, caseSensitive bool) (bool, error) 
 	return match, nil
 }
 
-func isTestdataFile(fullpath string) bool {
-	// testdata/ or /some/dir/testdata/some/other
-	return strings.HasPrefix(fullpath, "testdata/") ||
-		strings.Contains(fullpath, "/testdata/")
+func containsTestdataFile(fullpath string, ignoreDirs []string) bool {
+	check := func(name string) bool {
+		// name/ or /some/dir/name/some/other
+		return strings.HasPrefix(fullpath, name+"/") ||
+			strings.Contains(fullpath, "/"+name+"/")
+	}
+
+	if check("testdata") {
+		return false
+	}
+
+	for _, dir := range ignoreDirs {
+		if check(dir) {
+			return false
+		}
+	}
+
+	return true
 }
 
 // FileCbData is any data the caller can act upon
@@ -79,7 +93,7 @@ func CheckFilesContent(shellPathFnPattern string,
 ) error {
 	predicate := func(filepath string) (bool, error) {
 		// Filter out test files.
-		if isTestdataFile(filepath) {
+		if containsTestdataFile(filepath, c.IgnoreDirs) {
 			return false, nil
 		}
 		// Filter out files based on path/names using the pattern.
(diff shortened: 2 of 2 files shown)


Dentrax:
This is currently WIP. Passing the `ignoreDirs` field is not implemented yet, since we couldn't way out to pass project-wide ignore array.

We add a new `IgnoreDirs  []string` field in the `CheckRequest` struct. We should fill this from the same field/type that we need to add in the `config.go/config` struct?

@laurentsimon Can you please enlighten us on this?

laurentsimon:
The current idea we have is to build the "ignore list" as a policy applied on the *canonical* scorecard results. Here, I call *canonical* results the scorecard results without filtering. 

Consumers of the canonical results can then decide to apply a policy of their choice, e.g., "I don't care about results referencing path /some/path for check CheckName". The policy file may be different for every consumer, and we aim to provide a policy language to express more complicated policies, e.g. using https://cuelang.org/ or https://www.openpolicyagent.org/docs/latest/policy-language/ or anything users may want.

If we were to provide a directory list that changes scorecard's behavior at the time it gathers information (this PR), it becomes harder to tell what the *ground truth* is. 

So we envisage a world where:
1. scorecard results are canonical and do not depend on filtering
2. consumers apply policies to filter out results they don't care about, possibly using a hint from the consumer. A consumer may be the producer as well: for example, in the GitHub action, the repo owner will be able to add a config file. We envisage this policy being applied on the canonical results.

In order to build expressive policies, we aim to add some features in scorecard to output the results in a more structure format. We're tracking it in https://github.com/ossf/scorecard/issues/1245. The first step is to split the checks into 2 parts:
1. The part that gathers information
2. The part that applies a policy. Today, you can think of the score as a simple policy. Long-term, users will be able to express more complicated policies using a policy language. You have already split some of our checks, see this folder https://github.com/ossf/scorecard/tree/main/checks/raw (data gathering) and https://github.com/ossf/scorecard/tree/main/checks/evaluation ()score policy calculation. You can try it out using `SCORECARD_V6=1 scorecard ... --raw`.

Let me know if you have questions.

Dentrax:
A bit confused and lost @laurentsimon on this PR. Thanks for your thoughts. How should we proceed here? 🙏

laurentsimon:
We're going to address the `ignore-list` using policies. Please see the discussion starting on this https://github.com/ossf/scorecard/issues/1245#issuecomment-1018882971
Feel free to comment on the issue.

PR #1410, opened 2021-12-22
:seedling: Included examples of using BigQuery

naveensrinivasan:

Instead of keeping this in https://github.com/naveensrinivasan/scorecarddata it makes sense to move it within scorecard as an example which help
teams find ways to solve their problems

* **Other information**:


naveensrinivasan:
@azeemshaikh38 @laurentsimon  thoughts on this.

After we have a consensus I will fix the go.mod and license header. I wanted to throw the idea out to get some feedback.

README.md
```diff
@@ -362,6 +363,9 @@ between **0 and 10**. A confidence of 0 indicates that the check was unable to
 achieve any real signal, and that the result should be ignored. A confidence of 10
 indicates the check was completely sure of the result. 
 
+### Using public data
+[example of using scorecard bigquery](examples/bigquery/scorecarddata)
```
azeemshaikh38, line 367: How about we link to `naveensrinivasan/scorecarddata` repo here? Since this is a repo of its own, IMO we should maintain this repo separately rather than including it in scorecard repo.

azeemshaikh38 (changes requested):
I like the idea of exposing this tool to our users, but would prefer exposing it as a separate repo.

README.md
naveensrinivasan: Sounds good. I will do a PR for this.

justaugustus:
> I like the idea of exposing this tool to our users, but would prefer exposing it as a separate repo.

+1 to a separate repo for this.

justaugustus:
Closing based on the previous feedback here.

Issue #710, opened 2021-07-19
BUG - Frozen Deps handle args


naveensrinivasan:
**Describe the bug**
The frozen-deps check fails on ARGS being used in the FROM
```
ARG BASEIMAGE
ARG BINARY


FROM "${BASEIMAGE}"
COPY ${BINARY} /usr/local/bin/${BINARY}
```
https://raw.githubusercontent.com/kubernetes/kubernetes/master/build/server-image/Dockerfile

naveensrinivasan:
cc @laurentsimon

laurentsimon:
`ARG` is given via a command line. The `ARG` may be pinned, but dependabot is enable to update a pinned script command.
So it seems like intended result. wdyt?

laurentsimon:
as we roll this check thru allstar, I'll try to see how we can better avoid these (possibly) false positives. May require code annotations for developers to configure some opt-out (scorecard could then have an option to use or ignore the annotations). We could go one step further and require annotations to be LGTM'ed by a security team. Suggestions welcome.

naveensrinivasan:
> as we roll this check thru allstar, I'll try to see how we can better avoid these (possibly) false positives. May require code annotations for developers to configure some opt-out (scorecard could then have an option to use or ignore the annotations). We could go one step further and require annotations to be LGTM'ed by a security team. Suggestions welcome.

I like the idea of added annotations. Let’s discuss this in our next sync up.

laurentsimon:
I think to ensure that that dependency is pinned in `FROM`, the line should look like
```
FROM ${ARG}:@sha256:${HASH}
```

We do this in https://github.com/ossf/scorecard-actions/blob/main/analyze/Dockerfile#L17

wdut?

laurentsimon:
FYI @nanikjava

Issue #701, opened 2021-07-15
BUG: docker pinning using AS


laurentsimon:
If the dockerfile uses
```
FROM pk@sha256:hash as base

FROM base
```

the check incorrectly reports `base` as not pinned.

There's already a test file we can re-use to test our fix`testdata/Dockerfile-not-pinned-as`

PR #1413, opened 2021-12-23
✨ Add line numbers for insecure downloads

laurentsimon:
We don't have support for lines for secure download yet. This PR adds support. I happened to have time to look into it...
no breaking changes.

closes https://github.com/ossf/scorecard/issues/710
closes https://github.com/ossf/scorecard/issues/725
closes https://github.com/ossf/scorecard/issues/701

### DIFF
--- a/checks/dangerous_workflow.go
+++ b/checks/dangerous_workflow.go
@@ -167,9 +167,9 @@ func checkJobForUntrustedCodeCheckout(job *actionlint.Job, path string,
 			continue
 		}
 		if strings.Contains(ref.Value.Value, "github.event.pull_request") {
-			line := 1
+			line := checker.OffsetDefault
 			if step.Pos != nil {
-				line = step.Pos.Line
+				line = uint(step.Pos.Line)
 			}
 			dl.Warn3(&checker.LogMessage{
 				Path:   path,
@@ -225,9 +225,9 @@ func checkVariablesInScript(script string, pos *actionlint.Pos, path string,
 		// Check if the variable may be untrustworthy.
 		variable := script[s+3 : s+e]
 		if containsUntrustedContextPattern(variable) {
-			line := 1
+			line := checker.OffsetDefault
 			if pos != nil {
-				line = pos.Line
+				line = uint(pos.Line)
 			}
 			dl.Warn3(&checker.LogMessage{
 				Path:   path,
--- a/checks/dangerous_workflow.go
+++ b/checks/dangerous_workflow.go
@@ -222,10 +222,7 @@ func checkVariablesInScript(script string, pos *actionlint.Pos, path string,
 		// Check if the variable may be untrustworthy.
 		variable := script[s+3 : s+e]
 		if containsUntrustedContextPattern(variable) {
-			line := checker.OffsetDefault
-			if pos != nil {
-				line = uint(pos.Line)
-			}
+			line := fileparser.GetLineNumber(pos)
 			dl.Warn3(&checker.LogMessage{
 				Path:   path,
 				Type:   checker.FileTypeSource,
--- a/checks/dangerous_workflow.go
+++ b/checks/dangerous_workflow.go
@@ -167,10 +167,7 @@ func checkJobForUntrustedCodeCheckout(job *actionlint.Job, path string,
 			continue
 		}
 		if strings.Contains(ref.Value.Value, "github.event.pull_request") {
-			line := checker.OffsetDefault
-			if step.Pos != nil {
-				line = uint(step.Pos.Line)
-			}
+			line := fileparser.GetLineNumber(step.Pos)
 			dl.Warn3(&checker.LogMessage{
 				Path:   path,
 				Type:   checker.FileTypeSource,
--- a/checks/pinned_dependencies.go
+++ b/checks/pinned_dependencies.go
@@ -229,7 +229,7 @@ func validateShellScriptIsFreeOfInsecureDownloads(pathfn string, content []byte,
 		addPinnedResult(pdata, true)
 		return true, nil
 	}
-	r, err := validateShellFile(pathfn, 0, 0, content, dl)
+	r, err := validateShellFile(pathfn, 0, 0 /*unknown*/, content, dl)
 	if err != nil {
 		return false, err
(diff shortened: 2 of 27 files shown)


checks/dangerous_workflow.go
```diff
@@ -167,9 +167,9 @@ func checkJobForUntrustedCodeCheckout(job *actionlint.Job, path string,
 			continue
 		}
 		if strings.Contains(ref.Value.Value, "github.event.pull_request") {
-			line := 1
+			line := checker.OffsetDefault
 			if step.Pos != nil {
-				line = step.Pos.Line
+				line = uint(step.Pos.Line)
 			}
```
chrismcgehee, line 173: These 4 lines can be replaced with `fileparser.GetLineNumber(step.Pos)`. Similar on line 228 in this file.

checks/pinned_dependencies.go
```diff
@@ -523,19 +522,14 @@ func validateGitHubWorkflowIsFreeOfInsecureDownloads(pathfn string, content []by
...
-			scriptContent = fmt.Sprintf("%v\n%v", scriptContent, string(script))
-		}
-	}
-
-	if scriptContent != "" {
-		var err error
-		validated, err = validateShellFile(pathfn, []byte(scriptContent), dl)
-		if err != nil {
-			return false, err
+			validated, err := validateShellFile(pathfn, uint(execRun.Run.Pos.Line), 0 /*unknown*/, []byte(script), dl)
```
chrismcgehee, line 525: I believe the reason the `TestGitHubWorflowRunDownload/wget_across_steps` test is failing is because we're now calling `validateShellFile` for each step instead of waiting to call it after we've gotten the `run` actions for all steps. A similar problem probably also would occur for Dockerfiles when something is downloaded and executed in different steps, although it doesn't look like we have a test to catch that.

checks/pinned_dependencies_test.go
```diff
@@ -1038,10 +1198,80 @@ func TestGitHubWorkflowUsesLineNumber(t *testing.T) {
...
+			filename: "testdata/github-workflow-download-lines.yaml",
+			expected: []struct {
+				snippet   string
+				startLine uint
+				endLine   uint
+			}{
+				{
+					snippet:   "bash /tmp/file",
+					startLine: 27,
+					endLine:   2,
```
chrismcgehee, line 1238: That's weird to me that endLine is less than startLine. Why is that?

checks/shell_download_validate.go
```diff
@@ -428,6 +429,12 @@ func isGoUnpinnedDownload(cmd []string) bool {
 		}
 
 		pkg := cmd[i+1]
+		// Consider strings that do not start with github.com
+		// as referring to a local path or an golang package
+		// from golang.org.
+		if !strings.HasPrefix(pkg, "github.com/") {
+			return false
+		}
```
chrismcgehee, line 437: Why add this? If someone had the command `go install google.golang.org/protobuf/cmd/protoc-gen-go@latest`, then `isGoUnpinnedDownload` would return false when I think it should return true.

checks/shell_download_validate.go
```diff
@@ -570,35 +577,38 @@ func isUnpinnedPakageManagerDownload(node syntax.Node, cmd, pathfn string,
...
-			Type:    checker.FileTypeSource,
-			Offset:  0, // TODO: add line numbers
-			Snippet: cmd,
-			Text:    "insecure (not pinned by hash) download detected",
+			Path:      pathfn,
+			Type:      checker.FileTypeSource,
+			Offset:    startLine + node.Pos().Line(),
+			EndOffset: endLine + node.Pos().Line(),
+			Snippet:   cmd,
+			Text:      "go installation not pinned by hash",
```
chrismcgehee, line 585: I like that you've made log messages more specific. :+1:

chrismcgehee (commented):
Thanks for taking this one on, Laurent. You got to it a lot sooner than I would have.

checks/dangerous_workflow.go
laurentsimon: thanks, will do!

checks/pinned_dependencies.go
laurentsimon: correct. I will fix it. I'll let you know once I have.

checks/pinned_dependencies_test.go
laurentsimon: will fix, slipped thru the cracks :-)

checks/shell_download_validate.go
laurentsimon: thanks for catching it. I'd like to catch instances like https://github.com/envoyproxy/envoy/blob/main/examples/load-reporting-service/Dockerfile-lrs#L9, where the installation is not pinned but it's using a local folder. Any idea how I coould do this?

chrismcgehee: You'd like scorecard to flag `go install /local/folder` as an unpinned dependency? If so, then I think you could just remove lines 435-437.

laurentsimon: no, the opposite: I'd like to *not* flag these as unpinned, since they are either:
1. committed to the repo, so they are pinned
2. downloaded from some random location, in which case other parts of our code will catch them (curl followed by install, etc). Note this is not implemented yet

chrismcgehee: Ok, that sounds like a good thing to do. I'd probably recommend using a regex to detect if the arg passed to `go install` is in the format of a url something like `\w+\.\w+/\w+`. If it doesn't match that regex, assume it's a local download and return false.

checks/pinned_dependencies.go
laurentsimon: Done, PTAL. I've simply added a `taintedFiles = map[string]bool` as input argument to keep track f downloaded files across invocations.

laurentsimon:
I have addressed all comments from @chrismcgehee . PTAL

laurentsimon:
friendly ping

laurentsimon:
tests are passing. can someone take and look and LGTM?

naveensrinivasan (approved):

PR #1414, opened 2021-12-23
✨ Add details to message for default location in SARIF

laurentsimon:
we add details for additional information to SARIF when there is no file path to point to: branch protection, code reviews, etc.
Example, users will see
```
branch protection is not maximal on development and all release branches:
Warn: settings do not apply to administrators on branch 'main'
Warn: no status checks found to merge onto branch 'main'
Warn: number of required reviewers is 0 on branch 'main'
Warn: Stale review dismissal disabled on branch 'main'
```

instead of only:
```
branch protection is not maximal on development and all release branches
```


naveensrinivasan (approved):
Thank you!

Issue #1418, opened 2021-12-23
BUG: CodeQl wrongly detected


laurentsimon:
We currently look for the string `github/codeql-action` in a workflow to detect CodeQl.
However, strings like `github/codeql-action/upload-sarif` don't indicate codeQl is enabled, only thaht a tool downloads a SARIF file to GitHub. Scorecard's own action does use it, for example.

We should be checking for the presence of `github/codeql-action/analyze` instead, I think. See https://github.com/github/codeql-action

PR #1419, opened 2021-12-23
🐛 Fix incorrect detection of CodeQl

laurentsimon:
See https://github.com/ossf/scorecard/issues/1418
no breaking changes.
closes https://github.com/ossf/scorecard/issues/1418


naveensrinivasan (approved):
Thanks!

PR #1429, opened 2021-12-28
🐛  Documentation out of sync with Pinned-Dependencies check

laurentsimon:
Update the doc to reflect the changes in the check, in particular around lock files.
No breaking changes.
Note that once we have proper documentation for package managers, we'll point to it.


naveensrinivasan (approved):
Thanks!

PR #1431, opened 2021-12-29
✨ Update doc for scorecard container

laurentsimon:
Add information about running a scorecard container for a specific scorecard version.
no breaking changes.


naveensrinivasan (approved):
Thanks!

PR #1432, opened 2022-01-03
:seedling: Refactor vulnerabilities 

naveensrinivasan:

  Refactor the osv check into a interface so that it can be tested 

* **Other information**:


naveensrinivasan:
I will include test in the next PR.

clients/vulnerabilities.go
```diff
@@ -0,0 +1,80 @@
...
+	Commit string `json:"commit"`
+}
+
+// VulnerabilitiesClient cheks for vulnerabilities in api.osv.dev.
+type VulnerabilitiesClient interface {
+	HasUnfixedVulnerabilities(context context.Context, commit string) (OSVResponse, error)
+}
+
+// OSVResponse is the response from the OSV API.
+type OSVResponse struct {
```
azeemshaikh38, line 38: Since this is a response from an interface and not from a specific implementation of the interface, rename to something like `VulnerabilitiesResponse`.

clients/vulnerabilities.go
```diff
@@ -0,0 +1,80 @@
...
+	Vulns []struct {
+		ID string `json:"id"`
+	} `json:"vulns"`
+}
+
+type vulns struct{}
+
+// DefaultVulnerabilities is a new Vulnerabilities client.
+func DefaultVulnerabilities() VulnerabilitiesClient {
```
azeemshaikh38, line 47: `DefaultVulnerabilities` sounds like it returns the vulnerabilities themselves and not the client. Maybe rename to `DefaultVulnerabilitiesClient` to be more explicit.

azeemshaikh38 (commented):
Thanks @naveensrinivasan!

azeemshaikh38 (approved):

clients/vulnerabilities.go
naveensrinivasan: Agreed, Thanks. Fixed it.

clients/vulnerabilities.go
naveensrinivasan: Agreed. Thanks!

laurentsimon (approved):
Thanks @naveensrinivasan !

PR #1440, opened 2022-01-06
✨ Update messages for pinning warning

laurentsimon:
Update messages for pinning warning `"insecure (not pinned by hash) download"` -> `"insecure (not pinned by hash) download-then-run"`. This is better in cases where the highlighted line only executes a file, and the download is many lines before and not shown to user.
no braking changes


naveensrinivasan (approved):
Thanks!

PR #1443, opened 2022-01-06
🐛 Inconclusive results should not be reported

laurentsimon:
I broke some invariant in a previous commit thru refactoring. Inconclusive results are reported in SARIF, but should not be. Thanks you @azeemsgoogle for reporting!

This PR fixes it.
no breaking changes


azeemshaikh38 (approved):
Thanks!

PR #1445, opened 2022-01-06
✨ Change filename when no file is available

laurentsimon:
We currently display "on file available" when no file in the repo can be pointed to in the SARIF report. We use this, for example, for branch-Protection, Code-Review, Security-Policy. It's been reported to be confusing for users.
This PR changes the filename to be empty as " ". "" is not a valid name, which is why I provide a single space.

List of alerts:

Screen Shot 2022-01-06 at 1 42 42 PM

When clicking on the alert, no name is shown as:

Screen Shot 2022-01-06 at 1 45 24 PM

### DIFF
--- a/pkg/sarif.go
+++ b/pkg/sarif.go
@@ -591,7 +591,7 @@ func (r *ScorecardResult) AsSARIF(showDetails bool, logLevel zapcore.Level,
 		RuleIndex := len(run.Tool.Driver.Rules) - 1
 		if len(locs) == 0 {
 			// Use an "empty" filename to avoid confusing users.
-			locs = addDefaultLocation(locs, "no file associated with this results")
+			locs = addDefaultLocation(locs, "no file associated with this alert")
 			msg := createDefaultLocationMessage(&check)
 			cr := createSARIFCheckResult(RuleIndex, sarifCheckID, msg, &locs[0])
 			run.Results = append(run.Results, cr)
--- a/pkg/sarif.go
+++ b/pkg/sarif.go
@@ -590,7 +590,8 @@ func (r *ScorecardResult) AsSARIF(showDetails bool, logLevel zapcore.Level,
 		// so it's the last position for us.
 		RuleIndex := len(run.Tool.Driver.Rules) - 1
 		if len(locs) == 0 {
-			// Use an "empty" filename to avoid confusing users.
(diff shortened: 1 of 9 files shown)


laurentsimon:
cc @olivekl

azeemshaikh38 (commented):
Do you mind adding a screenshot of what the updated UI would look like with this PR?

olivekl:
Agreed that "no file available" is confusing. I agree with using either a blank filename, or something like "not associated with any file."

laurentsimon:
> Do you mind adding a screenshot of what the updated UI would look like with this PR?

added the screenshot in the description. PTAL

laurentsimon:
we could also use as file the checks.md#check-name location... random idea

laurentsimon:
here's an example if I link to the documentation of checks:

List:
Screen Shot 2022-01-06 at 6 43 13 PM

After clicking:

Screen Shot 2022-01-06 at 6 44 07 PM

laurentsimon:
friendly ping. 
Forget about the link to the `checks.md` suggestion, it does not work because the file does not live in the repo under analysis.
I've changed the file name to: `no file associated with this results`.
I've also added a `Click Remediation section below to solve this issue` below the highlighted text fr all the checks, and updated the text `Remediation:` to `Remediation (click "Show more" below):` (the GitHub UX requires an additional click that is not obvious). Thanks @olivekl for brainstorming this.

azeemshaikh38:
I apologize, I'm caught up in getting the Scorecard site working and some other issues. Will take a look at this by EOD tomo.

pkg/sarif.go
```diff
@@ -592,7 +590,8 @@ func (r *ScorecardResult) AsSARIF(showDetails bool, logLevel zapcore.Level,
 		// so it's the last position for us.
 		RuleIndex := len(run.Tool.Driver.Rules) - 1
 		if len(locs) == 0 {
-			locs = addDefaultLocation(locs, "no file available")
+			// Use an "empty" filename to avoid confusing users.
```
azeemshaikh38, line 593: Remove comment?

azeemshaikh38 (approved):
LGTM

pkg/sarif.go
laurentsimon: good catch. Will do!

Issue #1420, opened 2021-12-24
BUG: SAST and CI-Tests wrongly succeed when action is enabled


laurentsimon:
SAST and CI-Test use the `PR.App.Slug` field to determine if a SAST/CI tool is used. When running scorecard's GitHub action, we detect our own action as a SAST/CI tool.

We should exclude it. The URL can be used to determine the name of the tool run, e.g., https://api.github.com/repos/laurentsimon/scorecard-action-test-3/check-runs/4143313445

laurentsimon:
@azeemsgoogle is this something you'd like to add to the repo interface?

rmjspaz:
> @azeemsgoogle is thi
 something you'd like to add to the repo interface, i.e. a way to retrieve the tool name from the run/Slug?

U

rmjspaz:
> > @azeemsgoogle is thi
>  something you'd like to add to the repo interface, i.e. a way to retrieve the tool name from the run/Slug?
> 
> U

azeemshaikh38:
It seems non-intuitive to me that users can't improve their SAST/CI scores by enabling Scorecard actions. We shouldn't ignore Scorecard action in these checks IMO. 

If these checks seem to "wrongly succeed", then I would argue that the way we dole out points in these checks is where the problem lies. Maybe we should re-think the scoring system in these checks instead?

laurentsimon:
> It seems non-intuitive to me that users can't improve their SAST/CI scores by enabling Scorecard actions. We shouldn't ignore Scorecard action in these checks IMO.

good idea. I initially thought scorecard should be running silently in the background and not affect the results. But you make a good point.

> If these checks seem to "wrongly succeed", then I would argue that the way we dole out points in these checks is where the problem lies. Maybe we should re-think the scoring system in these checks instead?

I agree the scoring system is not great for those checks. I think it requires more information about SAST tools. Originally I wanted to classify what features or vulnerability classes each tool can find, but this seems to ambitious at this point. A more accessible scoring could be:
1. presence of linter: give 2-3 points
2. presence of supply-chain tool: give 2-3 points
3. language-specific tool: give remaining points, possibly using https://github.com/ossf/scorecard/issues/1427

Until we get there, we could maybe start by implementing point 2, ie give 3 points for scorecard as supply-chain tool. wdut? 

For CI-Tests check, I'm not sure what we should do. The check is currently coarse . Maybe we can keep it as-is for the time being?

azeemshaikh38:
[...]

Issue #1380, opened 2021-12-10
Feature: add linter check


laurentsimon:
Linters are a great way to enforce certain coding practices and avoid mistakes, as well as improving the code quality to help reviewers.

I think it'd be useful to support this in scorecard. Maybe this could live under the existing SAST check. I think it's be enough to start with check for the presence of this action https://github.com/github/super-linter, which wraps all linters.

naveensrinivasan:
I think this needs some discussion.

justaugustus:
Related to the CLOMonitor requests here, but requires more discussion: https://github.com/cncf/clomonitor/issues/1466, https://github.com/ossf/scorecard/issues/2614

PR #1487, opened 2022-01-18
✨ Support more SAST tools

laurentsimon:
This PR addresses https://github.com/ossf/scorecard/issues/1420, where we came to the conclusion that we want more granularity in the SAST check.
This PR does not contain unit tests yet, because there's a lot to discuss and I want to see if we're all OK with this changes.

**WARNING:  please ignore the comments and the calls to Println()**: I will clean up once we have agreed everything else is fine.

This PR does look for:
- Linter tools run on each PR, and reward 3 points if it's the case
- Code analysis tools run on each PR, and reward 4 points if its the case
- Supply-chain analysis tools run on each PR, and reward 3 points if it's the case. Since scorecard is the only supply-chain tool at the moment and we don't advertise PR support, we give 3 points if scorecard is enabled, even if not run on PRs.

This PR gives no additional points if a tool is declared in a workflow but not run on PRs - with the exception of scorecard above.

closes https://github.com/ossf/scorecard/issues/1380
closes https://github.com/ossf/scorecard/issues/1420
closes https://github.com/ossf/scorecard/issues/1580

### DIFF
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -299,21 +299,10 @@ func IsWorkflowFile(pathfn string) bool {
 	}
 }
 
-<<<<<<< HEAD
-<<<<<<< HEAD
-=======
->>>>>>> 3cd962d (e2e tests)
 // IsGithubWorkflowFileCb determines if a file is a workflow
 // as a callback to use for repo client's ListFiles() API.
 func IsGithubWorkflowFileCb(pathfn string) (bool, error) {
 	return IsWorkflowFile(pathfn), nil
-<<<<<<< HEAD
-=======
-func IsWorkflowFileCb(filename string) (bool, error) {
-	return strings.HasPrefix(strings.ToLower(filename), ".github/workflows"), nil
->>>>>>> 8d875e4 (draft)
-=======
->>>>>>> 3cd962d (e2e tests)
 }
 
 // IsGitHubOwnedAction checks if this is a github specific action.
@@ -347,7 +336,7 @@ func AnyJobsMatch(workflow *actionlint.Workflow, jobMatchers []JobMatcher, fp st
 ) bool {
 	for _, job := range workflow.Jobs {
 		for _, matcher := range jobMatchers {
-			if !matcher.matches(job) {
+			if !matcher.Matches(job) {
 				continue
 			}
 
@@ -370,8 +359,8 @@ func AnyJobsMatch(workflow *actionlint.Workflow, jobMatchers []JobMatcher, fp st
 	return false
 }
 
-// matches returns true if the job matches the job matcher.
-func (m *JobMatcher) matches(job *actionlint.Job) bool {
+// Matches returns true if the job matches the job matcher.
+func (m *JobMatcher) Matches(job *actionlint.Job) bool {
 	for _, stepToMatch := range m.Steps {
 		hasMatch := false
 		for _, step := range job.Steps {
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -299,10 +299,15 @@ func IsWorkflowFile(pathfn string) bool {
 	}
 }
 
+<<<<<<< HEAD
 // IsGithubWorkflowFileCb determines if a file is a workflow
 // as a callback to use for repo client's ListFiles() API.
 func IsGithubWorkflowFileCb(pathfn string) (bool, error) {
 	return IsWorkflowFile(pathfn), nil
+=======
+func IsWorkflowFileCb(filename string) (bool, error) {
+	return strings.HasPrefix(strings.ToLower(filename), ".github/workflows"), nil
+>>>>>>> 8d875e4 (draft)
 }
 
 // IsGitHubOwnedAction checks if this is a github specific action.
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -300,14 +300,20 @@ func IsWorkflowFile(pathfn string) bool {
 }
 
 <<<<<<< HEAD
+<<<<<<< HEAD
+=======
+>>>>>>> 3cd962d (e2e tests)
 // IsGithubWorkflowFileCb determines if a file is a workflow
 // as a callback to use for repo client's ListFiles() API.
 func IsGithubWorkflowFileCb(pathfn string) (bool, error) {
 	return IsWorkflowFile(pathfn), nil
+<<<<<<< HEAD
 =======
 func IsWorkflowFileCb(filename string) (bool, error) {
 	return strings.HasPrefix(strings.ToLower(filename), ".github/workflows"), nil
 >>>>>>> 8d875e4 (draft)
+=======
+>>>>>>> 3cd962d (e2e tests)
 }
 
 // IsGitHubOwnedAction checks if this is a github specific action.
--- a/checks/packaging.go
+++ b/checks/packaging.go
@@ -48,8 +48,12 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
 		return checker.CreateRuntimeErrorResult(CheckPackaging, e)
 	}
 
-	for _, fp := range matchedFiles {
-		fc, err := c.RepoClient.GetFileContent(fp)
+	for _, fn := range matchedFiles {
+		if !fileparser.IsWorkflowFile(fn) {
+			continue
+		}
+
+		fc, err := c.RepoClient.GetFileContent(fn)
 		if err != nil {
 			e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("RepoClient.GetFileContent: %v", err))
 			return checker.CreateRuntimeErrorResult(CheckPackaging, e)
@@ -60,18 +64,23 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
 			e := fileparser.FormatActionlintError(errs)
 			return checker.CreateRuntimeErrorResult(CheckPackaging, e)
 		}
-		if !isPackagingWorkflow(workflow, fp, c.Dlogger) {
+		if !isPackagingWorkflow(workflow, fn, c.Dlogger) {
 			continue
 		}
 
-		runs, err := c.RepoClient.ListSuccessfulWorkflowRuns(filepath.Base(fp))
+		runs, err := c.RepoClient.ListSuccessfulWorkflowRuns(filepath.Base(fn))
 		if err != nil {
 			e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("Client.Actions.ListWorkflowRunsByFileName: %v", err))
 			return checker.CreateRuntimeErrorResult(CheckPackaging, e)
 		}
 		if len(runs) > 0 {
+<<<<<<< HEAD
 			c.Dlogger.Info(&checker.LogMessage{
 				Path:   fp,
+=======
+			c.Dlogger.Info3(&checker.LogMessage{
+				Path:   fn,
+>>>>>>> d9ff5a7 (draft)
 				Type:   checker.FileTypeSource,
 				Offset: checker.OffsetDefault,
 				Text:   fmt.Sprintf("GitHub publishing workflow used in run %s", runs[0].URL),
@@ -79,8 +88,13 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
 			return checker.CreateMaxScoreResult(CheckPackaging,
 				"publishing workflow detected")
 		}
+<<<<<<< HEAD
 		c.Dlogger.Debug(&checker.LogMessage{
 			Path:   fp,
+=======
+		c.Dlogger.Debug3(&checker.LogMessage{
+			Path:   fn,
+>>>>>>> d9ff5a7 (draft)
 			Type:   checker.FileTypeSource,
 			Offset: checker.OffsetDefault,
 			Text:   "GitHub publishing workflow not used in runs",
--- a/checks/packaging.go
+++ b/checks/packaging.go
@@ -38,20 +38,13 @@ func init() {
 
 // Packaging runs Packaging check.
 func Packaging(c *checker.CheckRequest) checker.CheckResult {
-<<<<<<< HEAD
 	matchedFiles, err := c.RepoClient.ListFiles(fileparser.IsGithubWorkflowFileCb)
-=======
-	matchedFiles, err := c.RepoClient.ListFiles(fileparser.IsWorkflowFileCb)
->>>>>>> 8d875e4 (draft)
 	if err != nil {
 		e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("RepoClient.ListFiles: %v", err))
 		return checker.CreateRuntimeErrorResult(CheckPackaging, e)
 	}
 
 	for _, fn := range matchedFiles {
-		if !fileparser.IsWorkflowFile(fn) {
-			continue
-		}
 
 		fc, err := c.RepoClient.GetFileContent(fn)
 		if err != nil {
@@ -74,13 +67,8 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
 			return checker.CreateRuntimeErrorResult(CheckPackaging, e)
 		}
 		if len(runs) > 0 {
-<<<<<<< HEAD
-			c.Dlogger.Info(&checker.LogMessage{
-				Path:   fp,
-=======
 			c.Dlogger.Info3(&checker.LogMessage{
 				Path:   fn,
->>>>>>> d9ff5a7 (draft)
 				Type:   checker.FileTypeSource,
 				Offset: checker.OffsetDefault,
 				Text:   fmt.Sprintf("GitHub publishing workflow used in run %s", runs[0].URL),
@@ -88,13 +76,8 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
 			return checker.CreateMaxScoreResult(CheckPackaging,
 				"publishing workflow detected")
 		}
-<<<<<<< HEAD
-		c.Dlogger.Debug(&checker.LogMessage{
-			Path:   fp,
-=======
 		c.Dlogger.Debug3(&checker.LogMessage{
 			Path:   fn,
->>>>>>> d9ff5a7 (draft)
 			Type:   checker.FileTypeSource,
 			Offset: checker.OffsetDefault,
 			Text:   "GitHub publishing workflow not used in runs",
--- a/checks/packaging.go
+++ b/checks/packaging.go
@@ -67,7 +67,7 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
 			return checker.CreateRuntimeErrorResult(CheckPackaging, e)
 		}
 		if len(runs) > 0 {
-			c.Dlogger.Info3(&checker.LogMessage{
+			c.Dlogger.Info(&checker.LogMessage{
 				Path:   fn,
 				Type:   checker.FileTypeSource,
 				Offset: checker.OffsetDefault,
@@ -76,7 +76,7 @@ func Packaging(c *checker.CheckRequest) checker.CheckResult {
 			return checker.CreateMaxScoreResult(CheckPackaging,
 				"publishing workflow detected")
 		}
-		c.Dlogger.Debug3(&checker.LogMessage{
+		c.Dlogger.Debug(&checker.LogMessage{
 			Path:   fn,
 			Type:   checker.FileTypeSource,
 			Offset: checker.OffsetDefault,
--- a/checks/packaging.go
+++ b/checks/packaging.go
@@ -38,7 +38,11 @@ func init() {
 
 // Packaging runs Packaging check.
 func Packaging(c *checker.CheckRequest) checker.CheckResult {
+<<<<<<< HEAD
 	matchedFiles, err := c.RepoClient.ListFiles(fileparser.IsGithubWorkflowFileCb)
+=======
+	matchedFiles, err := c.RepoClient.ListFiles(fileparser.IsWorkflowFileCb)
+>>>>>>> 8d875e4 (draft)
 	if err != nil {
 		e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("RepoClient.ListFiles: %v", err))
 		return checker.CreateRuntimeErrorResult(CheckPackaging, e)
--- a/checks/sast.go
+++ b/checks/sast.go
@@ -205,7 +205,7 @@ func isToolUsed(cats workflowCategories, c sastCategory, mrScore float64) bool {
 
 func isToolUsedInWorkflows(cats workflowCategories, c sastCategory) bool {
 	for _, cats := range cats {
-		if _, exists := cats[sastCategoryCodeAnalysis]; exists {
+		if _, exists := cats[c]; exists {
 			return true
 		}
 	}
--- a/checks/sast.go
+++ b/checks/sast.go
@@ -16,10 +16,13 @@ package checks
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/ossf/scorecard/v4/checker"
+	"github.com/ossf/scorecard/v4/checks/fileparser"
 	"github.com/ossf/scorecard/v4/clients"
 	sce "github.com/ossf/scorecard/v4/errors"
+	"github.com/rhysd/actionlint"
 )
 
 // CheckSAST is the registered name for SAST.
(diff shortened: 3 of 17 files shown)


checks/packaging.go
```diff
@@ -17,7 +17,6 @@ package checks
 import (
```
laurentsimon, line 17: ignore this file, it's mostly typo changes.

clients/githubrepo/workflows.go
```diff
@@ -31,15 +35,85 @@ type workflowsHandler struct {
...
+}
+
+func (handler *workflowsHandler) listWorkflowRuns(opts *clients.ListWorkflowRunOptions) ([]clients.WorkflowRun, error) {
+	endpoint := fmt.Sprintf("repos/%s/%s/actions/runs", handler.owner, handler.repo)
+	u, err := addOptions(endpoint, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := handler.client.NewRequest("GET", u, nil)
```
laurentsimon, line 100: go-github does not support `check_suite_id` as input https://github.com/google/go-github/blob/178169fc04bc4a05ca48f98709cdd02ab5b556e3/github/actions_workflow_runs.go#L56

So I've simply copied their code from https://github.com/google/go-github/blob/178169fc04bc4a05ca48f98709cdd02ab5b556e3/github/actions_workflow_runs.go#L91 until they add support. We should file an issue for it.

clients/workflows.go
```diff
@@ -16,5 +16,19 @@ package clients
...
 type WorkflowRun struct {
-	URL string
+	URL          string
+	PullRequests []PullRequest
+	WorkflowID   *int64
+}
+
+// Workflow represents a workflow.
+type Workflow struct {
+	ID   int64
```
laurentsimon, line 26: need discussion: pointer or not?

naveensrinivasan (commented):
Great, But we need tests!

laurentsimon:
> Great, But we need tests!

so much to discuss before I add the test. Ill add them at the end. Test-man :p

laurentsimon:
cc @evverx

evverx:
> Linter tools run on each PR, and reward 3 points if it's the case
> Supply-chain analysis tools run on each PR, and reward 3 points if it's the case

I don't think that linters and supply-chain tools (combined) should outweigh SAST tools like LGTM, CodeQL or Coverity so it seems to me that they should be downgraded to 1 or something like that. I mean, systemd for example uses superlinter and it finds issues from time to time but I'm not sure it would be fair to say that in terms of security it's almost as important as LGTM or Coverity. It wouldn't be fair if projects using only LGTM received 4 while projects using only superlinter received 3 either.

I'll try to take a closer look later. Thanks!

evverx:
> This PR gives no additional points if a tool is declared in a workflow but not run on PRs

It's not always feasible to run SAST tools on every PR but if they are run daily or weekly it's still useful and should be rewarded I think. For example it takes CodeQL about 40 minutes to analyze systemd so it doesn't make much sense to clog its GHActions pool on every PR but it's still run daily

evverx:
systemd runs LGTM and superlinter on every PR, CodeQL daily and Coverity Scan daily and with this PR applied its score went from 10 to 6. I'm not sure it's what it should get :-)

evverx:
FWIW The curl score went from 7 to 0 with this PR applied.

Anyway, on the whole I think I like the idea of putting various tools into separate groups and assigning different "scores" to them. Though the way those "scores" are assigned depending on the group, frequency and so on needs tweaking I think. Thanks!

azeemshaikh38:
Will take a closer look at this tomo. Overall LGTM.

laurentsimon:
Thanks for the feedback @evverx I totally agree we need tweak the score; and also that tools that run on commits/schedules should be rewarded. Maybe give 70% of the points if run on commits vs PRs. Feel free to suggest.

laurentsimon:
> systemd runs LGTM and superlinter on every PR, CodeQL daily and Coverity Scan daily and with this PR applied its score went from 10 to 6. I'm not sure it's what it should get :-)

you'd actually get a 10 if you install scorecard as an action. It suffices for one linter and one static analysis code (LGTM in your case) to be run on all PRs + scorecard installed to get 10.

evverx:
> Maybe give 70% of the points if run on commits vs PRs

I think it depends on how frequent those runs are. If projects run SAST tools daily (or a few times a day) I think they should be scored higher than projects with weekly or monthly runs.

> you'd actually get a 10 if you install scorecard as an action.

I get that :-) but it's complicated because on the one hand I think scorecard is helpful but on the other hand I don't want to promote OSSF. Other than that looking at a couple of checks that have already been introduced and issues where new checks are outlined it seems to me that scorecard is moving towards "consumers" instead of "producers" (which is understandable) but I'm not sure it will make sense for "producers" to keep it upstream at some point (time will tell I think). Just to clarify, I have absolutely no problems with this direction but somehow tools catered to "consumers" somehow shift things that should be done by "consumers" to "producers"  (the latest example would be https://github.com/google/oss-fuzz/issues/7146) and generally I wouldn't want to annoy maintainers.

clients/githubrepo/workflows.go
azeemshaikh38: Could you file an issue and add a TODO comment here to point to it?

azeemshaikh38 (approved):

azeemshaikh38:
LGTM. 2 open issues would be - unit tests and addressing @evverx's comment.

clients/checkruns.go
```diff
@@ -16,10 +16,13 @@ package clients
...
 type CheckRun struct {
-	Status     string
-	Conclusion string
-	URL        string
-	App        CheckRunApp
+	Status       string
+	Conclusion   string
+	URL          string
+	App          CheckRunApp
+	Name         string
```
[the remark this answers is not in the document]
naveensrinivasan: Can you please add the reason as to why would like to change to a pointer? 

Every change requested should have a reason as to why it is requested. It helps others understand the thought process and the reasoning. This is applicable across all these comments. 

Pointer sematic has drawbacks we have to be clear on why we need it?  `Go` doesn't encourage pointer semantics. Most of the standard library has value semantics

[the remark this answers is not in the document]
azeemshaikh38: Sure. As part of https://github.com/ossf/scorecard/issues/1242, we are in the process of migrating the current data structure to a more consistent format. The consistent formatting lets us use a convention that implies `nil` == `no data`.

clients/githubrepo/workflows.go
```diff
@@ -31,15 +35,85 @@ type workflowsHandler struct {
...
+	}
+
+	return clients.Workflow{
+		ID:   *workflow.ID,
+		Path: *workflow.Path,
+		Name: *workflow.Name,
+	}, nil
+}
+
+func addOptions(s string, opts interface{}) (string, error) {
```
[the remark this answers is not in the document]
naveensrinivasan: Why pointer semantics `*ListWorkflowRunOptions`? I would recommend it being value sematics.

checks/fileparser/github_workflow.go
```diff
@@ -298,6 +298,10 @@ func IsWorkflowFile(pathfn string) bool {
 	}
 }
 
+func IsWorkflowFileCb(filename string) (bool, error) {
```
naveensrinivasan, line 301: Why have a `func` that never returns error as part of the return?

naveensrinivasan: Could we remove the `error`?

One line `go` funcs are discouraged. Is there a specific reason to have this in a separate func?

clients/githubrepo/checkruns.go
```diff
@@ -49,14 +49,33 @@ func (handler *checkrunsHandler) listCheckRunsForRef(ref string) ([]clients.Chec
...
-		})
+		}
+		if checkRun.Name != nil {
+			cr.Name = *checkRun.Name
+		}
+		if checkRun.CheckSuite != nil &&
+			checkRun.CheckSuite.ID != nil {
+			cr.CheckSuiteID = checkRun.CheckSuite.ID
+		}
+		/*
```
naveensrinivasan, line 67: Could we please remove this commented code?

clients/githubrepo/workflows.go
```diff
@@ -51,9 +125,22 @@ func (handler *workflowsHandler) listSuccessfulWorkflowRuns(filename string) ([]
...
 	for _, workflowRun := range data.WorkflowRuns {
-		workflowRuns = append(workflowRuns, clients.WorkflowRun{
-			URL: workflowRun.GetURL(),
-		})
+		r := clients.WorkflowRun{
+			URL:        workflowRun.GetURL(),
+			WorkflowID: workflowRun.WorkflowID,
+		}
+
+		/*prs := workflowRun.PullRequests
```
naveensrinivasan, line 133: Could we please remove this commented code?

clients/repo_client.go
```diff
@@ -35,6 +35,8 @@ type RepoClient interface {
 	ListReleases() ([]Release, error)
 	ListContributors() ([]Contributor, error)
 	ListSuccessfulWorkflowRuns(filename string) ([]WorkflowRun, error)
+	ListWorkflowRuns(opts *ListWorkflowRunOptions) ([]WorkflowRun, error)
```
naveensrinivasan, line 38: Could this not be pointer `*ListWorkflowRunOption`? Makes the API cleaner.

naveensrinivasan (approved):
Thanks. Few comments.

checks/sast.go
```diff
@@ -16,186 +16,510 @@ package checks
...
+		{
+			display: "CodeQL",
+			slug:    "github-actions",
+			action:  "github/codeql-action/analyze",
+		},
+		{
+			display: "CodeQL",
+			slug:    "github-actions",
+			action:  "github/codeql-action/analyze",
+		},
```
mjpieters, line 102: Did you mean to list CodeQL twice here?

laurentsimon: nope. Thanks for catching!

laurentsimon:
update: I've not tweaked the logic for score computation as follows:
- linter are run on all PRs. Linters are cheap so this seems acceptable. 1 points is awarded if it's the case
- supply-chain tool are used. We don't check if it's run on all PRs. So long as it's defined in a workflow and/or run on at least one commit, we award 2 points for it
- static analysis. Same as for supply-chain, and we award 5 points. If one tool is run on all PRs, we give an additional 2 points.

Wdut?

evverx:
> supply-chain tool are used. We don't check if it's run on all PRs. So long as it's defined in a workflow and/or run on at least one commit, we award 2 points for it

I think supply chain tools should follow the same rules as static analyzers in the sense that if they can't be run on PRs they should be downgraded to 1. It should help to prompt the developers of those tools to make them compatible with the PR workflow (which is supposed to catch issues as early as possible).

laurentsimon:
> > supply-chain tool are used. We don't check if it's run on all PRs. So long as it's defined in a workflow and/or run on at least one commit, we award 2 points for it
> 
> I think supply chain tools should follow the same rules as static analyzers in the sense that if they can't be run on PRs they should be downgraded to 1. It should help to prompt the developers of those tools to make them compatible with the PR workflow (which is supposed to catch issues as early as possible).

good point. I'll do that then. Thank you for your feedback!

laurentsimon:
thoughts on rate limiting 

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions 
```
When using GITHUB_TOKEN, the rate limit is 1,000 requests per hour per repository. For requests to resources that belong to an enterprise account on GitHub.com, GitHub Enterprise Cloud's rate limit applies, and the limit is 15,000 requests per hour per repository.
```

That's fine since each commit should have a different token.

For PAT, https://docs.github.com/en/developers/apps/building-github-apps/rate-limits-for-github-apps
```
GitHub Apps making server-to-server requests use the installation's minimum rate limit of 5,000 requests per hour. If an application is installed on an organization with more than 20 users, the application receives another 50 requests per hour for each user. Installations that have more than 20 repositories receive another 50 requests per hour for each repository. The maximum rate limit for an installation is 12,500 requests per hour.
```

For a repository like `systemd/systemd`, we'd need 30*6 = ~200 API requests, so let's say 300 for each push. 10-15 commits per hours should be fine.

FYI, we hope to be able to support GITHUB_TOKEN in the future. For cron job, highly unlikely we'll ever be able to run SAST check if we apply the changes proposed here.

@azeemshaikh38 any further thought?

evverx:
> For cron job, highly unlikely we'll ever be able to run SAST check if we apply the changes proposed here.

FWIW Looking at https://metrics.openssf.org/grafana/d/default/metric-dashboard?orgId=1&var-PackageURL=pkg%3Agithub/systemd/systemd it seems it doesn't work even without this PR (assuming the cron job is responsible for updating that dashboard)

laurentsimon:
> > For cron job, highly unlikely we'll ever be able to run SAST check if we apply the changes proposed here.
> 
> FWIW Looking at https://metrics.openssf.org/grafana/d/default/metric-dashboard?orgId=1&var-PackageURL=pkg%3Agithub/systemd/systemd it seems it doesn't work even without this PR (assuming the cron job is responsible for updating that dashboard)

you're correct we currently don't run it. Fyi, the data in the dashboard is out of date.
cc @scovetta @inferno-chromium

evverx:
FWIW @laurentsimon with https://github.com/systemd/systemd/commit/74b781de2545c80b988981aace9e04148cf3dfe3 merged it seems scorecard can no longer detect super-linter. As far as I understand the check isn't aware of the slim image: https://github.com/github/super-linter#slim-image

evverx:
Though looking at
```
    {
      "details": [
        "Info: code-analysis tool 'CodeQL' detected in workflow: .github/workflows/codeql-analysis.yml",
        "Warn: linter tool run on 0 commits out of 15 commits",
        "Warn: supply-chain tool run on 0 commits out of 15 commits",
        "Info: code-analysis tool run on the last 15 commits"
      ],
      "reason": "SAST category of tools are not used%!(EXTRA string=code-analysis, string=supply-chain)",
      "name": "SAST",
```
it seems either the commit based analysis doesn't work as expected or it just doesn't look at PRs there. Anyway I'm not sure why it stopped working.

azeemshaikh38:
> Though looking at
> 
> ```
>     {
>       "details": [
>         "Info: code-analysis tool 'CodeQL' detected in workflow: .github/workflows/codeql-analysis.yml",
>         "Warn: linter tool run on 0 commits out of 15 commits",
>         "Warn: supply-chain tool run on 0 commits out of 15 commits",
>         "Info: code-analysis tool run on the last 15 commits"
>       ],
>       "reason": "SAST category of tools are not used%!(EXTRA string=code-analysis, string=supply-chain)",
>       "name": "SAST",
> ```
> 
> it seems either the commit based analysis doesn't work as expected or it just doesn't look at PRs there. Anyway I'm not sure why it stopped working.


This might be a similar issue to https://github.com/ossf/scorecard/issues/1580#issuecomment-1027946049. We use GitHub Search APIs for detecting SAST tools and GitHub's Search APIs likely unreliable. I tried running Scorecard locally on `systemd` and the results varied. We should probably consider updating the `Search` API in `RepoClient` to use locally available code instead. @laurentsimon fyi.

evverx:
> We use GitHub Search APIs for detecting SAST tools and GitHub's Search APIs likely unreliable.

Could it be that because of that the "code review" check is kind of broken as well?
```
  "repo": {
    "name": "github.com/systemd/systemd",
    "commit": "d74da762a3b1edbb7935c3b3a229db601f734125"
  },
  "scorecard": {
    "version": "4.0.1-86-geb0730a",
    "commit": "eb0730ae797377d24676cdedfa27216f0d649091"
  },
...
    {
      "details": null,
      "score": 4,
      "reason": "GitHub code reviews found for 13 commits out of the last 30 -- score normalized to 4",
      "name": "Code-Review",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/eb0730ae797377d24676cdedfa27216f0d649091/docs/checks.md#code-review",
        "short": "Determines if the project requires code review before pull requests (aka merge requests) are merged."
      }
```

azeemshaikh38:
> re: search API in repoClient @azeemshaikh38 do we need this API at all or can we just let checks use the file listing API https://github.com/ossf/scorecard/blob/main/checks/fileparser/listing.go?

It would be nice to support a generic `Search` API, especially for cases which require searching large repos like oss-fuzz like we do [today](https://github.com/ossf/scorecard/blob/eb0730ae797377d24676cdedfa27216f0d649091/checks/fuzzing.go#L55). Having said that, I'm open to trying out `ListFiles` and `GetFileContent` APIs for solving SAST checks if that gets us more consistency.

evverx:
Also it seems scorecard (with `--verbosity Debug --show-details`) no longer shows useful logs that can be used to somewhat dispute its findings. I think it would be great if they could be brought back given the number of false positives it produces mainly due to the GH API as far as I can tell.

azeemshaikh38:
> Could it be that because of that the "code review" check is kind of broken as well?

Code-Review does not use GitHub search APIs. It looks like #1505 changed the behavior here. Commits before that, give `systemd/systemd` a score of 10 for Code-Review. @laurentsimon could you look into this?

evverx:
> Commits before that, give systemd/systemd a score of 10 for Code-Review

I think commits before that are affected by https://github.com/ossf/scorecard/issues/1260 so it doesn't always work as expected either. Anyway, without "debug" logs it's hard to tell what scorecard does and why.

evverx:
Regarding logs it appears something like that was reported in https://github.com/ossf/scorecard/issues/95 and then it got a bit better and now those logs seem to be gone again.

azeemshaikh38:
> > Commits before that, give systemd/systemd a score of 10 for Code-Review
> 
> I think commits before that are affected by #1260 so it doesn't always work as expected either. Anyway, without "debug" logs it's hard to tell what scorecard does and why.

Ok, I think I know whats going on. Some PRs are repeated for `systemd` commits. Its as if the same PR number gets re-used and latest PR data (including reviews) overwrites old data. Here's an example:

 

```
{
              "oid": "d74da762a3b1edbb7935c3b3a229db601f734125",
              "associatedPullRequests": {
                "nodes": [
                  {
                    "repository": {
                      "owner": {
                        "login": "systemd"
                      },
                      "name": "systemd"
                    },
                    "number": 22506,
                    "title": "gpt-auto: some (primarily cosmetic) fixes to backing block device detection in gpt-auto-generator/sd-device",
                    "mergeCommit": {
                      "oid": "d74da762a3b1edbb7935c3b3a229db601f734125"
                    }
                  }
                ]
              }
            },
{
              "oid": "d5cb053cd93d516f516e0b748271b55f9dfb3a29",
              "associatedPullRequests": {
                "nodes": [
                  {
                    "repository": {
                      "owner": {
                        "login": "systemd"
                      },
                      "name": "systemd"
                    },
                    "number": 22506,
                    "title": "gpt-auto: some (primarily cosmetic) fixes to backing block device detection in gpt-auto-generator/sd-device",
                    "mergeCommit": {
                      "oid": "d74da762a3b1edbb7935c3b3a229db601f734125"
                    }
                  }
                ]
              }
            },
```

2 different commits have the same `associatedPR`, while the PR data itself reflects the latest change. This is why Scorecard finds unreviewed commits, because there is no review data associated with these commits. 

The real question is why is this behavior happening on `systemd` at all? @evverx do you have any clues?

evverx:
My guess would be that that's because that PR was merged so the two commits and the "merge" commit ended up in the repository. If it had been squashed or rebased only that commit would have been mapped to that PR. I need to take a closer look.

azeemshaikh38:
> My guess would be that that's because that PR was merged so the two commits and the "merge" commit ended up in the repository. If it had been squashed or rebased only that commit would have been mapped to that PR. I need to take a closer look.

Ah, I think you are right. https://github.com/systemd/systemd/pull/22506 has 2 commits - [d5cb053cd93d516f516e0b748271b55f9dfb3a29](https://github.com/systemd/systemd/pull/22506/commits/d5cb053cd93d516f516e0b748271b55f9dfb3a29) and [f1ad2c9238f03198a05e36167397d24983ebc527](https://github.com/systemd/systemd/pull/22506/commits/f1ad2c9238f03198a05e36167397d24983ebc527). So it shows up as 3 commits - [d5cb053cd93d516f516e0b748271b55f9dfb3a29](https://github.com/systemd/systemd/pull/22506/commits/d5cb053cd93d516f516e0b748271b55f9dfb3a29), [f1ad2c9238f03198a05e36167397d24983ebc527](https://github.com/systemd/systemd/pull/22506/commits/f1ad2c9238f03198a05e36167397d24983ebc527) and a [merge commit](https://github.com/systemd/systemd/commit/d74da762a3b1edbb7935c3b3a229db601f734125).

This is a very helpful find @evverx, thanks! Let me look into figuring out a fix for this.

evverx:
I think I guessed it because the logs where that PR was mentioned were right in front of me :-) It would be great if they could be kept. Thanks!

evverx:
FWIW looking at the "code review" check it seems it can't cover projects like https://github.com/libbpf/libbpf where patches are reviewed on the kernel mailing list and then get synced with the kernel on GitHub. I think it's a corner case probably but maybe scorecard could look for "Signed-Off", "Reviewed-by" or something like that. It can easily be forged though so I'm not sure what to do about that. Anyway, it has nothing to do with the SAST check so I'll show myself out. Thanks!

azeemshaikh38:
> I think I guessed it because the logs where that PR was mentioned were right in front of me :-) It would be great if they could be kept. Thanks!

Sent #1637 for this. Also added more logging, hopefully that helps.

evverx:
> Regarding logs it appears something like that was reported in https://github.com/ossf/scorecard/issues/95 and then it got a bit better and now those logs seem to be gone again.

Looks like it was an overstatement because the logs weren't actually gone. It's just `--verbosity Debug` no longer works: https://github.com/ossf/scorecard/pull/1637#issuecomment-1039813280. Looking at a few PRs that have been opened recently it's a bit sad that unlike Allstar (against which changes are actually tested and which was the reason the logger was changed in the first place) open source maintainers (or generally people rolling out their own scripts) don't seem to be considered "dowstream consumers" scorecard should care about.

laurentsimon:
> > Regarding logs it appears something like that was reported in #95 and then it got a bit better and now those logs seem to be gone again.
> 
> Looks like it was an overstatement because the logs weren't actually gone. It's just `--verbosity Debug` no longer works: [#1637 (comment)](https://github.com/ossf/scorecard/pull/1637#issuecomment-1039813280). Looking at a few PRs that have been opened recently it's a bit sad that unlike Allstar (against which changes are actually tested and which was the reason the logger was changed in the first place) open source maintainers (or generally people rolling out their own scripts) don't seem to be considered "dowstream consumers" scorecard should care about.

we consider them downstream users. @naveensrinivasan is working on adding unit tests and this one will be caught once he's added them. Thanks for reminding us, though. Please create a tracking issue if there's none specifically for the CLI.

evverx:
> @naveensrinivasan is working on adding unit tests and this one will be caught once he's added them.

FWIW looking at the number of tests that have been added relatively recently I have to say it's impressive and I don't doubt that eventually corner cases like this will be covered as well (and I usually completely agree with @naveensrinivasan  when it comes to testing (including fuzzing).

>  Please create a tracking issue if there's none specifically for the CLI.

@laurentsimon I came here to report a bug in the SAST check and then I updated scorecard and ran into another bug in the Code-Review check and while I was testing that PR I ran into yet another bug/feature related to "debug" mode. I'm not sure I can spend so much time reporting scorecard issues. Anyway, I didn't mean to say that scorecard shouldn't be used in general. My takeaway is that it can be used indirectly via the action or Allstar but it's just not stable enough to rely on outside of the OSSF ecosystem (and there is nothing wrong with that). I think I can wait until it's more or less stabilized.

azeemshaikh38:
> Looks like it was an overstatement because the logs weren't actually gone. It's just `--verbosity Debug` no longer works: [#1637 (comment)](https://github.com/ossf/scorecard/pull/1637#issuecomment-1039813280). Looking at a few PRs that have been opened recently it's a bit sad that unlike Allstar (against which changes are actually tested and which was the reason the logger was changed in the first place) open source maintainers (or generally people rolling out their own scripts) don't seem to be considered "dowstream consumers" scorecard should care about.

I hear you @evverx. It's an unfortunate result of us not having tests like Laurent mentioned. #1650 should fix the problem you are having. Let us know if we can help in any other way.

clients/workflows.go
```diff
@@ -16,5 +16,19 @@ package clients
 
 // WorkflowRun represents VCS WorkflowRun.
 type WorkflowRun struct {
-	URL string
+	URL          string
```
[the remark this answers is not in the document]
laurentsimon: will keep as-is because the github-go API use a string. I don't disagree with the suggestion though

clients/githubrepo/workflows.go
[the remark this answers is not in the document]
laurentsimon: I made it a pointer. Pointers are better for large structure, right? Otherwise it's copying around data for calls, etc, I think

clients/checkruns.go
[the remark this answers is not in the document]
laurentsimon: If I make `Name` a pointer, shall I make everything else one too then?

laurentsimon:
@azeemshaikh38 take a final look before I merge. Ignore the comments, I'm going to remove them before the merge.

evverx:
FWIW Looking at https://github.com/ossf/scorecard/pull/1816 I think by that logic the SAST check should be inconclusive as well when scorecard can't find static analyzers it's aware of. Coverity is often hidden in bash scripts sending data to the scanner and it's unlikely that `scorecard` can ever detect that reliably.

azeemshaikh38:
@laurentsimon are we blocked on anything to get this merged?

laurentsimon:
no real blocker, mostly delayed to fix the e2e tests.

laurentsimon:
Please don't close, Im going to get to this PR sometimes after I've migrated other checks to raw results.

naveensrinivasan:
> Please don't close, Im going to get to this PR sometimes after I've migrated other checks to raw results.

I apologize it was a mistake.

laurentsimon:
> > Please don't close, Im going to get to this PR sometimes after I've migrated other checks to raw results.
> 
> I apologize it was a mistake.

np, happens to be all the time :)

naveensrinivasan:
@laurentsimon  What do we want to do about this PR?

laurentsimon:
Let's keep it open if you don't mind. The code uses too many API calls, but I think if we relax some of checks, a large part of the code can be re-used. I'm not working on it atm, though.

naveensrinivasan:
> Let's keep it open if you don't mind. The code uses too many API calls, but I think if we relax some of checks, a large part of the code can be re-used. I'm not working on it atm, though.

OK, sounds good! Thanks

naveensrinivasan:
Hi @laurentsimon,

 I wondered if keeping this Pull Request open would be beneficial, considering it is over a year old.

naveensrinivasan:
> Hi @laurentsimon,
> 
> I wondered if keeping this Pull Request open would be beneficial, considering it is over a year old.

A friendly ping @laurentsimon

laurentsimon:
> > Hi @laurentsimon,
> > I wondered if keeping this Pull Request open would be beneficial, considering it is over a year old.
> 
> A friendly ping @laurentsimon

Up to you. There's a lot of useful code someone may be able to re-use to implement the SAST. But we can close it a search for it if need be.

naveensrinivasan:
> > > Hi @laurentsimon,
> > > I wondered if keeping this Pull Request open would be beneficial, considering it is over a year old.
> > 
> > 
> > A friendly ping @laurentsimon
> 
> Up to you. There's a lot of useful code someone may be able to re-use to implement the SAST. But we can close it a search for it if need be.

Thanks for keeping it clean. I will close it.

Issue #1273, opened 2021-11-15
Should scorecard be forcing users to use a zap.Logger?


naveensrinivasan:
External packages like Allstar would like to use as a scorecard as a Library. 
https://github.com/ossf/allstar/blob/913bb71913d373d0daaf2a59957c3610d2522e50/pkg/policies/binary/binary.go#L120
https://github.com/ossf/scorecard/blob/1050b1cd607b3686641c96ac071ae6774e60588f/clients/githubrepo/client.go#L183

This API call is forcing the consumers to use `zap.Logger`. This should not be forced to use a specific logger. It could be interface for logging, not a specific implementation.

laurentsimon:
I agree it should not.

naveensrinivasan:
@azeemshaikh38 FYI...

jeffmendoza:
Ideally CreateGitHubRepoClient() can be refactored to take a [`DetailLogger`](https://github.com/ossf/scorecard/blob/63e3b92466f0403159c02f1ccedd43f9400e8b26/checker/check_result.go#L81) instead of a zap.Logger.

justaugustus:
I _might_ be able to pick this up after https://github.com/ossf/allstar/pull/98.
(Just commenting to shake up the stale bot...)

azeemshaikh38:
Thanks @justaugustus!

justaugustus:
Working on this in https://github.com/ossf/scorecard/pull/1502.

justaugustus:
The first bit is merged in https://github.com/ossf/scorecard/pull/1502.
Before I continue, I want check in on something...

What's the motivation behind https://github.com/ossf/scorecard/blob/0d76deace2b3edd9216a15ec5a8387130acd8db4/checker/detail_logger_impl.go#L42-L50?

More specifically, the function argument, `(msg *LogMessage)`:

```go
func (l *logger) Info3(msg *LogMessage) { ... }
```

`*LogMessage` makes it hard to do `logger.Info3()` without first doing something like:

```go
log := &LoggerMessage{}
log.Text = "the thing I want to log"

logger.Info3(log)
```

I just want to make sure I'm not missing something about the intended ergonomics, before trying to implement.

Perhaps something variadic could work nicely here?

```go
// Adds a struct of fields to the log entry. All it does is call `WithField` for
// each `Field`.
func (logger *Logger) WithFields(fields Fields) *Entry {
	entry := logger.newEntry()
	defer logger.releaseEntry(entry)
[...]

PR #1502, opened 2022-01-20
:warning: Create a dedicated logging package to encapsulate calls to `zap`

justaugustus:

    We're roughly mimicking existing functionality and the interface may change in future. Logging usage that has changed exists in packages that already have test coverage.

    tl;dr -- If existing tests pass, we should be fine here.

    Introduces a logging package to eventually replace go.uber.org/zap.
    Here we're effectively quarantining usage of the go.uber.org/zap and go.uber.org/zap/zapcore to a single package, `log`.

    See https://github.com/ossf/scorecard/issues/1273.

    Functions that create instances of `zap.Logger` now create instances of `log.Logger`:

    ```go
    type Logger struct {
        Zap *zap.Logger
    }
    ```

    ...which means usage of zap will now look something like:

    ```go
    logger.Zap.Info()
    ```

    As @jeffmendoza mentions in https://github.com/ossf/scorecard/issues/1273#issuecomment-969485674:

    > Ideally CreateGitHubRepoClient() can be refactored to take a [`DetailLogger`](https://github.com/ossf/scorecard/blob/63e3b92466f0403159c02f1ccedd43f9400e8b26/checker/check_result.go#L81) instead of a zap.Logger.

    the goal would to support logging without zap, opting to use `DetailLogger` instead.

    Consumers of functions like:

    ```go
    func CreateGithubRepoClient(ctx context.Context, logger *zap.Logger) clients.RepoClient { ... }
    ```

    will need to use `log.Logger` as an argument:

    ```go
    func CreateGithubRepoClient(ctx context.Context, logger *log.Logger) clients.RepoClient { ... }
    ```

    as well as update log calls from:

    ```go
    ghClient := CreateGithubRepoClient(ctx, logger)
    logger.Info(...)
    ```

    to:

    ```go
    ghClient := CreateGithubRepoClient(ctx, logger)
    logger.Zap.Info(...)
    ```

* **Other information**:

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -21,12 +21,17 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+// Logger exposes logging capabilities.
+// The struct name was chosen to closely mimic other logging facilities within
+// to project to make them easier to search/replace.
+// Initial implementation was designed to encapsulate calls to `zap`, but
+// future iterations will seek to directly expose logging methods.
 type Logger struct {
 	Zap *zap.Logger
 }
 
 // NewLogger creates an instance of *zap.Logger.
-// Copied from clients/githubrepo/client.go
+// Copied from clients/githubrepo/client.go.
 func NewLogger(logLevel Level) (*Logger, error) {
 	zapCfg := zap.NewProductionConfig()
 	zapLevel := parseLogLevelZap(string(logLevel))
--- a/log/log.go
+++ b/log/log.go
@@ -52,14 +52,19 @@ type Level string
 // TODO(log): Revisit if all levels are required. The current list mimics zap
 //            log levels.
 const (
-	DebugLevel  Level = "debug"
-	InfoLevel   Level = "info"
-	WarnLevel   Level = "warn"
-	ErrorLevel  Level = "error"
-	DPanicLevel Level = "dpanic"
-	PanicLevel  Level = "panic"
-	FatalLevel  Level = "fatal"
+	DefaultLevel       = InfoLevel
+	DebugLevel   Level = "debug"
+	InfoLevel    Level = "info"
+	WarnLevel    Level = "warn"
+	ErrorLevel   Level = "error"
+	DPanicLevel  Level = "dpanic"
+	PanicLevel   Level = "panic"
+	FatalLevel   Level = "fatal"
 )
+
+func (l Level) String() string {
+	return string(l)
+}
 
 // parseLogLevelZap parses a log level string and returning a zapcore.Level,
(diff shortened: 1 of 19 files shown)


justaugustus:
@inferno-chromium @naveensrinivasan @azeemshaikh38 @laurentsimon @olivekl @jeffmendoza -- This is in a good place for initial review. I don't want run to far ahead in case there's something design-wise that needs fixing.

naveensrinivasan:
I had a quick look and 👍. I will let @azeemshaikh38  and @laurentsimon chime in

justaugustus:
> I had a quick look and 👍. I will let @azeemshaikh38 and @laurentsimon chime in

Thanks so much, Naveen!

log/log.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"fmt"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+type Logger struct {
+	Zap *zap.Logger
```
azeemshaikh38, line 25: Maybe add a `// TODO` here to eventually stop exposing the `Zap` object and instead expose `Info()`, `Warn()`, `Sync()` methods directly.

azeemshaikh38 (approved):
This looks awesome @justaugustus, thanks!

log/log.go
justaugustus: @azeemshaikh38 -- Would you prefer I add that in this PR or can I take it as a follow-up, post-merge?

justaugustus: (nvm, tests failed, so I'll add it now 🙃)

azeemshaikh38: Whichever you prefer. You can simply leave a comment and address in a separate PR if you'd like.

justaugustus: Added a comment in https://github.com/ossf/scorecard/pull/1502/commits/ac1d50f11a146f768fc1cb93777c061b40fb2d0b.

justaugustus:
@ naveensrinivasan @azeemshaikh38 -- Fixed up the reviews. Should be good to retrigger CI on :)

naveensrinivasan:
> @ naveensrinivasan @azeemshaikh38 -- Fixed up the reviews. Should be good to retrigger CI on :)

Thanks!

azeemshaikh38 (approved):

azeemshaikh38:
Thanks @justaugustus, looks good to merge to me.

laurentsimon (commented):
Just wanted to join and say thanks! This is a massive PR!

PR #1505, opened 2022-01-21
✨ Raw data for code review check

laurentsimon:
Add support for raw results for Branch-Protection check.
No breaking changes.
The hope is that we will be able t re-use the "commit" structure for SAST.

Output:
```
"commits": [
      {
        "committer": {
          "login": "dependabot[bot]"
        },
        "review": {
          "platform": {
            "name": "GitHub"
          },
          "authors": [
            {
              "login": "azeemshaikh38"
            }
          ],
          "merge-request": {
            "number": 1459,
            "author": {
              "login": "dependabot"
            }
          }
        }
      },
      {
        "committer": {
          "login": "justaugustus"
        },
        "review": {
          "platform": {
            "name": "GitHub"
          },
          "authors": [
            {
              "login": "azeemshaikh38"
            }
          ],
          "merge-request": {
            "number": 1502,
            "author": {
              "login": "justaugustus"
            }
          }
        }
      }
...
```

cc @jeffmendoza @asraa to give you some idea of what the raw results look like at the moment.


checker/raw_result.go
```diff
@@ -97,12 +110,43 @@ type Issue struct {
...
+	Committer User
+	// Note: not all commits are reviewed.
+	// `nil` indicate no reviews.
+	Review *Review
+}
+
 // MergeRequest represents a merge request.
 type MergeRequest struct {
-	URL string
+	Number int
```
azeemshaikh38, line 125: Maybe `ID` instead of `Number`?

checker/raw_result.go
```diff
@@ -97,12 +110,43 @@ type Issue struct {
...
 	// TODO: add fields, e.g., State=["merged"|"closed"]
 }
 
+// User represent a user.
+type User struct {
+	Login string
+}
+
+// ReviewPlatform represents a review platform.
+type ReviewPlatform struct {
```
azeemshaikh38, line 136: If as of now there are no other fields, maybe just use a string instead of a struct. We can create one later when needed.

checker/raw_result.go
```diff
@@ -97,12 +110,43 @@ type Issue struct {
...
 }
 
+// Commit represents a commit.
+type Commit struct {
+	// Note: SHA is not directly accessible from a pull request.
+	// TODO:SHA          string
+	Committer User
+	// Note: not all commits are reviewed.
+	// `nil` indicate no reviews.
+	Review *Review
```
azeemshaikh38, line 120: I find it a bit non-intuitive to have `Review` inside `Commit`. A single PR can have multiple "commits" but the "review" is for a PR and not the commits individually.

I think it should be the other way around, i.e a `MergeCommit` (of type `Commit`) inside a `Review`. Wdut?

azeemshaikh38: Another suggestion here would be to update `CodeReviewData` to return both `[]Commits` and `[]PullRequests` (MergedPRs)

checker/raw_result.go
laurentsimon: PRs have both an `ID` and a `Number`,and those are different things. The `Number` is what we use in the UX, so I think that's what we want. Agreed?

checker/raw_result.go
laurentsimon: that's true. Bit if we need to add, we'll always break backward compatibility. I see what you mean: it feels overkill. But would it be nice to not have backward compatibility, especially if users have to write policies?

checker/raw_result.go
laurentsimon: re: counter-intuitive. A commit here is a commit to main branch. There's only one merge commit for a PR. So essentially, a commit to main may (or not) have an associated pull request.  Maybe I should rename to `MergeCommit` to be clearer? 

Is there a use case where we look a commits to non-main branch? Just thinking out lout whether the `Commit` field may be re-usable in the future. (we'll find out as we iterate :-))

re: to return both `[]Commits` and `[]PullRequests`. Do we gain something by doing this? I see a point for commits in general (if we have other use cases for it, maybe another check); but I don't immediately see that it helps for the Code-Review check. Can you clarify?

Thanks!

azeemshaikh38:
@laurentsimon will look at the PR over the weekend or maybe Monday. If this is blocking you in any way, let me know and I can prioritize this.

laurentsimon:
> @laurentsimon will look at the PR over the weekend or maybe Monday. If this is blocking you in any way, let me know and I can prioritize this.

Monday's fine. I don't have work blocked on this PR, np.

checker/raw_result.go
azeemshaikh38: Makes sense. In that case how about the below suggestion: 

```
Commit {
  MergeRequest *MergeRequest
  Committer *User
}

MergeRequest {
  Reviews []Review
  Author
  Number
}

Review {
  Author User
  Platform ReviewPlatform
  Status <APPROVED|DECLINED|...>
}
```

Here's my reasoning to suggest the update:

- There can be multiple reviews for a given commit which the current structure does not imply. 
- Every `Review` can have more info like `approved`, `declined` etc. which cannot be capture in the current structure.
- Reviews are associated with MergeRequest, not the other way round. i.e you can have a `MergeRequest` without a `Review`, but no `Review` without `MergeRequest`.

Let me know what you think.

asraa: +1 on review status

checks/code_review.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Security Scorecard Authors
+// Copyright 2021 Security Scorecard Authors
```
asraa, line 1: if this is changing, maybe 2022?

checker/raw_result.go
laurentsimon: > Makes sense. In that case how about the below suggestion:

we can tweak it. Below a few more points

> ```
> Commit {
>   MergeRequest *MergeRequest
>   Committer *User
> }

a review is not always done thru a pull request, so MergeRequest must be inside a Review, not the other way around. Gerrit is an example: it reviews commit but does not use merge requests.

> MergeRequest {
>   Reviews []Review
>   Author
>   Number
> }
> 
> Review {
>   Author User
>   Platform ReviewPlatform
>   Status <APPROVED|DECLINED|...>
> }
> ```

I don't think we care about capturing multiple review systems for a single review. The project either uses GitHub reviews or Gerrit or Prow, we don't need Platform to be associated with each reviewer's review. 

> Here's my reasoning to suggest the update:
> 
> * There can be multiple reviews for a given commit which the current structure does not imply.

it is captured by the current design. A review can have multiple "authors". Authors can be changed to a more generic structure `ReviewData` that contains more than Authors, such as State or anything else we want to add in the future. 

> * Every `Review` can have more info like `approved`, `declined` etc. which cannot be capture in the current structure.

we can add a status, that's fair. I don't see a strong use case for it since the review system takes care of deciding when merging is acceptable, but I don't have a strong opinion on having it. It complicates the policy evaluation a little but not much. 

> * Reviews are associated with MergeRequest, not the other way round. i.e you can have a `MergeRequest` without a `Review`, but no `Review` without `MergeRequest`.

that's not the case, see comment above for Gerrit.

> Let me know what you think.

laurentsimon: let's finalize in our sync. Meanwhile, I'll add the State to the structure instead of just a list of Authors

laurentsimon: update: it's hard to have a review status for different review systems, because we detect the review system based on certain fields that indicate reviews. So we actually don't know what system a non-reviewed commit is using...
I suppose Gerrit and Prow have their ways of indicating a "need changes" or other status, but we don't support these atm. For context, to detect GitHub reviews we look at the merge request's `Status` field; to detect Gerrit reviews we look as labels' content; and so on.

1. I think most users will want to define policies based on what scorecard supports, like "use GitHub review" or "use GitHub review and the the bot named X is acceptable".

2. There is another category of policies for which scorecard has no support like "merge request has status X and comment message contains Y", which essentially tries to use the results to implement support for a review system we do not support. I am not against these uses cases, but I feel a review system should be detected within scorecard rather than being implemented as a policy. 

My original goal was to address case 1. If we don't make a judgement call on what a review means at all, this means we're aiming to address point 2 and in this case we cannot really expose a review field. Policy writers will need to write complicated policies (including us for the default one) that re-implement what a review means. I kinda feel this should be implemented in scorecard rather than a policy.

Maybe there's a way to address both by providing digested data for case 1; whilst still allowing case 2. I need to think a bit more about it.

Feel free to chime in :-)

laurentsimon: additional thoughts. Checks like Token-Permission, Dangerous-Workflows, etc may require the presence of case 2 policies, like "no secret in workflow unless action is one of X,Y,Z ". So maybe there's no way around supporting cases in category 2) above. This leads to other questions:
1. We'll have to build policy API/building blocks to help users write simple policies for case 1 policies. Essentially, this means writing all the check's evaluation using a policy language. 
2. The original idea was to be policy-language agnostic and leave it to users to write their policy with a policy language of their choice. That no longer seems to be the case if we have to support building blocks for developers to use.
3. What does a policy return? a Fail/Pass or a list of things that made the policy failed. I suppose the latter. What is the format of these results? Is it defined by the JSON input the policy is applied to, or can it be arbitrarily defined by us?
5. Our first customer is allstar and binary authorization. How will allstar evaluation a policy? Will it use its own policy language on top of the JSON raw output we provide? Does it expect scorecard to accept a policy file that scorecard evaluates itself (which goes against point 2 above)? 

Another approach we could consider is for the raw results to expose both complete raw data, and some data about what scorecard supports. In the context of this commit, it would be:
```
commits{ // raw results
  SHA
  MergeRequest{
   Number, Reviews{ Status, Authors}, MergeCommit, labels[]
  }
}

reviews[]{ // what scorecard supports
  review{
    commitSHA
    system (GitHub, Prow, etc)
    reviewers[]{}
  }
}
```

We could merge this PR this way to see where this takes us. I think we also we need a design doc.

laurentsimon: cc @asraa @jeffmendoza @oliverchang @chrismcgehee

naveensrinivasan (commented):
Tests are required for such a large feature change. We shouldn't merge up until that.

laurentsimon:
> Tests are required for such a large feature change. We shouldn't merge up until that.

we won't. Btw, the existing unit tests still pass. This PR does not changing the logics, it just re-factors the code.

checker/raw_result.go
azeemshaikh38: Setup a sync to discuss this over gvc.

checks/code_review.go
justaugustus: Copyright dates do not need to be updated.
They only need to reflect current date if it's a net-new file.

checker/raw_result.go
laurentsimon: so I'll go ahead and do this:
```
commits{ 
  // Raw data
  SHA
  CommitMessage
  MergeRequest{
   Number, Reviews{ Status, Authors}, MergeCommit, labels[]
  }
  // "Opinionated" results
  approved-reviews{ // what scorecard supports
  system (GitHub, Prow, etc)
  []review{
    reviewers[]{ Login, }
  }
}
}
```

checks/code_review.go
justaugustus: > Copyright dates do not need to be updated.
> They only need to reflect current date if it's a net-new file.

ref: https://github.com/ossf/scorecard/issues/1534 for discussion

checker/raw_result.go
justaugustus: > Setup a sync to discuss this over gvc.

We should push discussions publicly as much as possible.

checker/raw_result.go
laurentsimon: this is no longer exposed as we've decided to use repo-interface raw results.

checker/raw_result.go
laurentsimon: we've decided to use 
```
commits{ 
  SHA
  CommitMessage
  MergeRequest{
   Number, Reviews{ Status, Authors}, MergeCommit, labels[]
  }
```

and push the massaged version (easier to write policies for) in a different part of scorecard.

laurentsimon:
this is finally ready for review. Thanks for all the discussions and input!

azeemshaikh38 (approved):

naveensrinivasan:
> # [Codecov](https://codecov.io/gh/ossf/scorecard/pull/1505?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ossf) Report
> > Merging [#1505](https://codecov.io/gh/ossf/scorecard/pull/1505?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ossf) ([00d1917](https://github.com/ossf/scorecard/commit/00d1917bd1cd65888ccb7f554328aea4e92aafe3)) into [main](https://codecov.io/gh/ossf/scorecard/commit/009aa85e3f649048ad9e4c377afc45b8bc48f6b7?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ossf) ([009aa85](https://github.com/ossf/scorecard/commit/009aa85e3f649048ad9e4c377afc45b8bc48f6b7)) will **decrease** coverage by `2.59%`.
> > The diff coverage is `1.83%`.
> 
> ```diff
> @@            Coverage Diff             @@
> ##             main    #1505      +/-   ##
> ==========================================
> - Coverage   54.42%   51.82%   -2.60%     
> ==========================================
>   Files          70       72       +2     
>   Lines        6175     6313     +138     
> ==========================================
> - Hits         3361     3272      -89     
> - Misses       2603     2836     +233     
> + Partials      211      205       -6     
> ```

@laurentsimon Disabled auto merge. How do we address this?

laurentsimon:
weird. 
Tho things:

1. the file `pkg/json_raw_results.go` just copies the raw results into a new structure before display. I agree we should add unit tests for this one in the long-term. But maybe not in this PR because the aim is just to split the check and policies. Let me know what you think.

2. I'm puzzled why file `checks/raw/code_review.go` shows with no hits. I've re-factored the code but not changed the logic. Why does it show no coverage? It's literally telling us none of the file was tested. Do you know what's going on?

naveensrinivasan:
> The bulk of the coverage drops appears to be from file `checks/raw/code_review.go` which shows with no hits. I've re-factored the code but not changed the logic. Why does it show no coverage? It's literally telling us none of the lines were tested. Do you know what's going on?

TBH I don't know. If you think it is good, please merge it 👍

laurentsimon:
> > The bulk of the coverage drops appears to be from file `checks/raw/code_review.go` which shows with no hits. I've re-factored the code but not changed the logic. Why does it show no coverage? It's literally telling us none of the lines were tested. Do you know what's going on?
> 
> TBH I don't know. If you think it is good, please merge it 👍

something is off but I don't know why codecov does not see it :/ Let's merge and please remind me to see what the following PRs show about these lines.

Issue #986, opened 2021-09-09
BUG: unit tests


laurentsimon:
See comment in https://github.com/ossf/scorecard/pull/980

azeemshaikh38:
@naveensrinivasan thanks for the awesome job here. Testing new code is a breeze!

naveensrinivasan:
Thanks

naveensrinivasan:
Closing this issue as it has been addressed.

PR #1512, opened 2022-01-21
:seedling: Unit tests for binary_artifact

naveensrinivasan:

    https://github.com/ossf/scorecard/issues/986

* **Other information**:


checks/raw/binary_artifact_test.go
```diff
@@ -0,0 +1,92 @@
...
+			logger, err := githubrepo.NewLogger(log.DebugLevel)
+			if err != nil {
+				t.Errorf("githubrepo.NewLogger: %v", err)
+			}
+
+			// nolint
+			defer logger.Zap.Sync()
+
+			ctrl := gomock.NewController(t)
+			repo, err := localdir.MakeLocalDirRepo(tt.inputFolder)
```
azeemshaikh38, line 72: Why are we creating LocalDir client here? Shouldn't we be using the MockClient?

azeemshaikh38 (approved):

checks/raw/binary_artifact_test.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+
+			ctx := context.Background()
+
+			client := localdir.CreateLocalDirClient(ctx, logger)
+			if err := client.InitRepo(repo); err != nil {
+				t.Errorf("InitRepo: %v", err)
+			}
+			_, err = BinaryArtifacts(client)
+			if !errors.Is(err, tt.err) {
```
laurentsimon, line 85: should you be testing for the content of the results https://github.com/ossf/scorecard/blob/main/checker/raw_result.go#L35?

azeemshaikh38 (commented):
Approved by mistake

checks/raw/binary_artifact_test.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// TestBinaryArtifact tests the BinaryArtifact checker.
+func TestBinaryArtifacts(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name        string
+		inputFolder string
+		err         error
+		expected    checker.CheckResult
```
laurentsimon, line 37: raw part of check does not return`checker.CheckResult` but the results defined in https://github.com/ossf/scorecard/blob/main/checker/raw_result.go#L35. I think we should be testing for this instead, right?

checks/raw/binary_artifact_test.go
naveensrinivasan: Thanks!

checks/raw/binary_artifact_test.go
naveensrinivasan: Yes, agreed took care of it. Now the func checks for the results count.

azeemshaikh38 (approved):
Awesome, thanks!

Issue #1488, opened 2022-01-18
BUG: Maintained check list issue update from non-maintainers


laurentsimon:
In the Maintained check, we look for issue thru `c.RepoClient.ListIssues()`. I think anyone can create issues, but that does not indicate that the maintainers are active. We should either look for closed issues or comments by maintainers (users who have committed to the main branch, for example)

I have not looked in details, though.

cc @azeemsgoogle

chrismcgehee:
I could work on this one. The [Issue](https://docs.github.com/en/graphql/reference/objects#issue) object has `closedAt` so it will be easy to check if the issue has been closed in the past 90 days. It also has a `comments` field which has an `authorAssociation` field. We can check if any comments from the past 90 days are from one of these [personas](https://docs.github.com/en/graphql/reference/enums#commentauthorassociation):
- OWNER
- MEMBER
- COLLABORATOR

azeemshaikh38:
That's a pretty neat solution @chrismcgehee, I like it! If it's possible to figure out the persona who `opened`, `closed` or `updated` the issue, that'd be awesome. If not, let's proceed with the `comment` and `authorAssociation` part of it for now and we can iterate on it further. Wdut?

laurentsimon:
very neat! Are we going to change the signature of `ListIssues()` to have input parameters?

chrismcgehee:
@azeemshaikh38 I can use the `authorAssociation` field on the issue itself to figure out the persona who opened the issue. Figuring out updates on the issue would be trickier. I think I'd have to use `timelineItems` which is a union of like 30 different types, so the code would be very complex if we wanted to find the persona of the updater. I won't do that now; it probably won't ever make sense to do it, but we can always revisit later.

@laurentsimon I don't think I'll need to give input parameters to `ListIssues()`. I'll add additional fields to the `Issue` struct, but other than that, the `RepoClient` interface should stay the same.

azeemshaikh38:
SGTM @chrismcgehee, thanks!

laurentsimon:
fyi, I don't think https://github.com/ossf/scorecard/blob/main/checks/maintained.go#L93 is true. I think the creator of an issue can close it even if they are external to a repo and not a collaborator. I think we need to verify who closed the issue, if feasible.

PR #1515, opened 2022-01-22
🐛 Issue activity only counts if done by a maintainer

chrismcgehee:

Bug fix

For the Maintained check, any issues activity by any Github user increased the score.
https://github.com/ossf/scorecard/issues/1488

Now only if an issue is created, closed, or commented on by a user who is a collaborator, owner, or member will the score be increased.

No

### DIFF
--- a/clients/githubrepo/graphql.go
+++ b/clients/githubrepo/graphql.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"sync"
-	"time"
 
 	"github.com/shurcooL/githubv4"
 
@@ -27,11 +26,12 @@ import (
 )
 
 const (
-	pullRequestsToAnalyze = 30
-	issuesToAnalyze       = 30
-	reviewsToAnalyze      = 30
-	labelsToAnalyze       = 30
-	commitsToAnalyze      = 30
+	pullRequestsToAnalyze  = 30
+	issuesToAnalyze        = 30
+	issueCommentsToAnalyze = 30
+	reviewsToAnalyze       = 30
+	labelsToAnalyze        = 30
+	commitsToAnalyze       = 30
 )
 
 // nolint: govet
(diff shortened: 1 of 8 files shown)


clients/issue.go
```diff
@@ -18,6 +18,15 @@ import "time"
 
 // Issue represents a thread like GitHub issue comment thread.
 type Issue struct {
-	URI       *string
-	UpdatedAt *time.Time
+	URI               *string
+	CreatedAt         time.Time
+	ClosedAt          time.Time
+	AuthorAssociation string
```
azeemshaikh38, line 24: Update `time.Time` -> `*time.Time` and `string` -> `*string`. We are in the process of moving to a convention where we can use `nil` values to represent no data instead of relying on default values. See #1242 for context.

clients/issue.go
```diff
@@ -18,6 +18,15 @@ import "time"
...
+	Comments          []IssueComment
+}
+
+// IssueComment represents a comment on an issue.
+type IssueComment struct {
+	CreatedAt         time.Time
+	AuthorAssociation string
```
azeemshaikh38, line 31: Similar comment here as above.

azeemshaikh38, line 31: Nit: maybe update `AuthorAssociation` to be of enum type so that different implementations of the `RepoClient` interface can return valid author associations.

clients/githubrepo/graphql.go
```diff
@@ -86,8 +86,16 @@ type graphqlData struct {
...
-				UpdatedAt *time.Time
+				Url               *string
+				AuthorAssociation githubv4.String
+				CreatedAt         githubv4.DateTime
+				ClosedAt          githubv4.DateTime
+				Comments          struct {
+					Nodes []struct {
+						AuthorAssociation githubv4.String
+						CreatedAt         githubv4.DateTime
+					}
```
azeemshaikh38, line 97: Using pointer values here will easily let you know when the API returns no data. E.g `AuthorAssociation` would be `nil` instead of "" when an association could not be made.

azeemshaikh38 (commented):
Thanks @chrismcgehee. LGTM overall, left some minor comments. PTAL.

clients/issue.go
chrismcgehee: Yeah, this makes the code nicer. I'm glad you suggested it.

azeemshaikh38 (approved):
Nice, thanks!

PR #1516, opened 2022-01-23
⚠️ log: Initial `logr`/`logrusr` implementation

justaugustus:

Continues https://github.com/ossf/scorecard/pull/1502.
ref: https://github.com/ossf/scorecard/issues/1273

https://github.com/ossf/scorecard/pull/1502 implemented a generic logger that wrapped `zap`. Create a new logger is done by calling `log.Logger(log.Level)`.

`log.Logger` now uses `logr`, which provides a sane interface and has multiple popular implementations.

From @thockin in https://github.com/ossf/scorecard/issues/1273#issuecomment-1018822830:

> Stephen pinged me for a consult, so here I am :)
> 
> Opinion: About 1% of apps can actually justify writing logging abstractions, but 90% of us end up doing it because there isn't a (literal or de facto) standard.
> 
> I offer you [http://github.com/go-logr/logr](https://github.com/go-logr/logr) as that standard. logr defines an API, but not an impl. There are many implementations available (including glog, klog, zap, logrus, and zerolog). Your main package can have full knowledge of the impl, set it up as it likes, and then pass it down to the rest of the app.
> 
> EVERYTHING ELSE only needs to know the logr API. The logr package has no external dependencies (and will not grow them, scout's honor), so it is a "safe" thing to depend on. And it's TINY (500 LOC, 3/4 of that is comments). And it has testing implementations and trivial "give me a func() to call" implementations.
> 
> The more of us that use it (klog already uses it internally, for example :) the more inertia we build.
> 
> Writing log wrappers is a nerd-snipe.

While [`zapr`](https://github.com/go-logr/zapr) exists, I opted to choose [`logrusr`](https://github.com/bombsimon/logrusr) as [`logrus`](https://github.com/sirupsen/logrus) just happens to be used by a few more projects in the ecosystem that I'm familiar with (including [Kubernetes Release Engineering logging](https://github.com/kubernetes-sigs/release-utils/blob/main/log/log.go))

Yes!

  Consumers of functions like:

  ```go
  func CreateGithubRepoClient(ctx context.Context, logger *zap.Logger) clients.RepoClient { ... }
  ```

  will need to use `log.Logger` as an argument:

  ```go
  func CreateGithubRepoClient(ctx context.Context, logger *log.Logger) clients.RepoClient { ... }
  ```

  as well as update log calls from:

  ```go
  ghClient := CreateGithubRepoClient(ctx, logger)
  logger.Info(...)
  ```

  to:

  ```go
  ghClient := CreateGithubRepoClient(ctx, logger)
  logger.Zap.Info(...)
  ```

* **Other information**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,8 +8,10 @@ require (
 	cloud.google.com/go/pubsub v1.17.0
 	cloud.google.com/go/trace v0.1.0 // indirect
 	contrib.go.opencensus.io/exporter/stackdriver v0.13.8
+	github.com/bombsimon/logrusr/v2 v2.0.1
 	github.com/bradleyfalzon/ghinstallation/v2 v2.0.4
 	github.com/go-git/go-git/v5 v5.4.2
+	github.com/go-logr/logr v1.2.2
 	github.com/golang/mock v1.6.0
 	github.com/google/go-cmp v0.5.7
 	github.com/google/go-containerregistry v0.8.0
@@ -22,6 +24,7 @@ require (
 	github.com/onsi/gomega v1.18.0
 	github.com/shurcooL/githubv4 v0.0.0-20201206200315-234843c633fa
 	github.com/shurcooL/graphql v0.0.0-20200928012149-18c5c3165e3a // indirect
+	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.3.0
 	github.com/xeipuuv/gojsonschema v0.0.0-20180618132009-1d523034197f
 	go.opencensus.io v0.23.0
@@ -34,11 +37,10 @@ require (
 	mvdan.cc/sh/v3 v3.4.2
 )
 
+// TODO(go.mod): Is there a reason these deps are kept separately from the
+//               other `require`s?
 require (
-	github.com/bombsimon/logrusr/v2 v2.0.1
-	github.com/go-logr/logr v1.0.0
 	github.com/rhysd/actionlint v1.6.8
-	github.com/sirupsen/logrus v1.8.1
 	gotest.tools v2.2.0+incompatible
 )
 
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/spf13/cobra v1.3.0
 	github.com/xeipuuv/gojsonschema v0.0.0-20180618132009-1d523034197f
 	go.opencensus.io v0.23.0
-	go.uber.org/zap v1.20.0
 	gocloud.dev v0.24.0
 	golang.org/x/tools v0.1.8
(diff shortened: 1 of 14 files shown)


log/log.go
```diff
@@ -15,35 +15,54 @@
...
 // future iterations will seek to directly expose logging methods.
 type Logger struct {
-	Zap *zap.Logger
+	*logr.Logger
+	// Zap *zap.Logger
 }
 
 // NewLogger creates an instance of *zap.Logger.
 // Copied from clients/githubrepo/client.go.
 func NewLogger(logLevel Level) (*Logger, error) {
```
naveensrinivasan, line 37: Probably you are still working on it. But part of refactoring this shouldn't return `error`

justaugustus: Yep, yep, I'll drop that in the next push to this PR.
If only all of our `err`s could always be `nil`! 😜

naveensrinivasan:
If we merge this change then the next release would have to be `v5.0.0`. This is breaking API.

justaugustus:
@naveensrinivasan @azeemshaikh38 -- I think this is ready for review!

go.mod
```diff
@@ -35,6 +37,8 @@ require (
 	mvdan.cc/sh/v3 v3.4.2
 )
 
+// TODO(go.mod): Is there a reason these deps are kept separately from the
```
azeemshaikh38, line 40: @naveensrinivasan might know why?

azeemshaikh38 (approved):
This is so much neater, thanks @justaugustus!

azeemshaikh38:
> If we merge this change then the next release would have to be `v5.0.0`. This is breaking API.

True. For now, I think we can merge this in and discuss in this week's meet about how to handle the release version.

justaugustus:
> This is so much neater, thanks @justaugustus!

@azeemshaikh38 -- Happy to help! I just rebased to deconflict w/ `main`, so once tests pass, this should be good to merge.

naveensrinivasan (approved):
Thanks!

log/log.go
```diff
@@ -15,54 +15,76 @@
...
+	}
+
+	return logger
+}
+
+// ParseLevel takes a string level and returns the Logrus log level constant.
+// If the level is not recognized, it defaults to `logrus.InfoLevel` to swallow
+// potential configuration errors/typos when specifying log levels.
+// https://pkg.go.dev/github.com/sirupsen/logrus#ParseLevel
+func ParseLevel(lvl string) logrus.Level {
```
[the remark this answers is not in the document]
justaugustus: > Could we add some unit tests for this?

Created https://github.com/ossf/scorecard/issues/1528 for tracking!

PR #1525, opened 2022-01-24
:seedling: Unit tests for Fuzzing

naveensrinivasan:

    Unit tests checks for fuzzing.

    https://github.com/ossf/scorecard/issues/986

* **Other information**:

### DIFF
--- a/checks/fuzzing_test.go
+++ b/checks/fuzzing_test.go
@@ -1,0 +1,308 @@
+// Copyright 2020 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checks
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+
+	"github.com/ossf/scorecard/v4/checker"
+	"github.com/ossf/scorecard/v4/clients"
+	mockrepo "github.com/ossf/scorecard/v4/clients/mockclients"
+	sce "github.com/ossf/scorecard/v4/errors"
+	scut "github.com/ossf/scorecard/v4/utests"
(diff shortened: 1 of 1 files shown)


checks/fuzzing_test.go
```diff
@@ -0,0 +1,241 @@
...
+				req.OssFuzzRepo = nil
+			}
+
+			got, err := checkOSSFuzz(&req)
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("checkOSSFuzz() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
```
laurentsimon, line 97: can you add a verification of Warn/Info messages? (Same comment for other functions)

naveensrinivasan: Can you point me to an example of this?

laurentsimon: sure, see https://github.com/ossf/scorecard/blob/main/checks/dangerous_workflow_test.go#L160

naveensrinivasan: This would involve refactoring the original implementation for including the logging changes. Am I right? I don't see it in the `Fuzzing` check.

laurentsimon: I think in this case it would just be `NumberOfWarn: 0, NumberOfInfo: 0, NumberOfDebug: 0`, right?
Let me know if I'm mis-understanding the question.

naveensrinivasan: Sorry, what value does that add? I don't get it. Probably we should discuss this.

naveensrinivasan: Also the `dangerous_workflow.go` has reference to `checker.DetailLogger` but whereas `fuzzing.go` doesn't use that. So I don't see the reason. 

Probably I am missing the point.

laurentsimon: I mean that we want to know that no messages are being printed, and it should not change unexpectedly because it would affect the SARIF output for GitHub scanning dashboard. So if we ever add calls to Warn(), that's something we want to catch.

laurentsimon: re: no logger in Fuzzing(). logger is passed as a field in the checker request https://github.com/ossf/scorecard/blob/main/checker/check_request.go#L29. You just nee dto set the logger in the request instead of passing it as argument.

naveensrinivasan: Got it! Thanks for the details, it helped to understand it. But I am worried that this is going to make it harder for new devs to contribute to the project.

laurentsimon (approved):
Thanks!

PR #1558, opened 2022-01-28
🌱 Use committer name if login isn't available

### LINKED ISSUE #1543: BUG: ListCommits() has empty committer name for "verified" commits
The repo client's code https://github.com/ossf/scorecard/blob/main/clients/githubrepo/graphql.go#L218 contains empty string for the committer name for certain commits.

Here are example commits that have an empty committer name:
d50788f6381d1fea21e4e50caa2de41012a38982  
5f9fff3b20ce7eb933978c7a4f9391cb2c9b3d89  
7a6eb2812a85ed047582b86a3e77e16302506558  
16c0d375d64391c617df8f3994f15167b6798dbb  

and more.

Looking at https://github.com/ossf/scorecard/commits/main, all these PRs have one thing in common: they are marked "verified" `This commit was created on GitHub.com`. Not sure why this changes the behavior though.

This blocks https://github.com/ossf/scorecard/issues/1524

@azeemsgoogle assigning to you since you've played with graphql and you may know an easy fix.

azeemshaikh38:

If committer login is not available use another field from GH. Closes #1543 

No.

### DIFF
--- a/clients/githubrepo/graphql.go
+++ b/clients/githubrepo/graphql.go
@@ -17,6 +17,7 @@ package githubrepo
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -33,6 +34,7 @@ const (
 	reviewsToAnalyze       = 30
 	labelsToAnalyze        = 30
 	commitsToAnalyze       = 30
+	allowedCommitterName   = "github"
 )
 
 // nolint: govet
@@ -48,8 +50,9 @@ type graphqlData struct {
 							Message       githubv4.String
 							Oid           githubv4.GitObjectID
 							Committer     struct {
+								Name *string
 								User struct {
-									Login githubv4.String
+									Login *string
 								}
 							}
 						}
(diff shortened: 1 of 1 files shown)


clients/githubrepo/graphql.go
```diff
@@ -210,12 +211,18 @@ func pullRequestsFrom(data *graphqlData) []clients.PullRequest {
 func commitsFrom(data *graphqlData) []clients.Commit {
 	ret := make([]clients.Commit, 0)
 	for _, commit := range data.Repository.DefaultBranchRef.Target.Commit.History.Nodes {
+		var committer string
+		if commit.Committer.User.Login != nil {
+			committer = *commit.Committer.User.Login
+		} else if commit.Committer.Name != nil {
+			committer = *commit.Committer.Name
```
laurentsimon, line 218: this can be spoofed so we should not use it as committer, right?
Should it be 
```
} else{
  committer = "github"
}
```

as you suggested in https://github.com/ossf/scorecard/issues/1543#issuecomment-1023816104?

azeemshaikh38: > this can be spoofed so we should not use it as committer, right?

I didn't realize this can this be spoofed. How would someone do that? 

> as you suggested in [#1543 (comment)](https://github.com/ossf/scorecard/issues/1543#issuecomment-1023816104)?

Yes, this will be `github`. The `committer.name` is set to that value for commits which have `nil` User.

laurentsimon: I see. Maybe add a comment. Maybe we should also verify this is `github` in the code to validate our understanding, and catch any violating of our understanding.

laurentsimon (approved):
Awesome!

Issue #575, opened 2021-06-14
Commit based scorecard


KPostOffice:
**Is your feature request related to a problem? Please describe.**
I would like to be able to create a score card for either specific releases or commits on GitHub so that I can meaningfully show how a project's health has changed on a release by release basis.

I understand that some scorecard indicators might not work because they interact directly with the GitHub api and are not dependent on the current head commit.  Is there a subset of checks that this could work with though?

laurentsimon:
That's an interesting feature to have, thanks for the suggestion. Any checks that are looking at file content (frozen dependencies, binary artifacts checked-in, etc) would work in this model. In the future, we would like to support other type of code source: not only online repos, but also files/directories on disk, e.g. `scorecard --repo=file://`. This would make commit-based checks feasible.

We are also working on a BigQuery end point to expose the results of our cron run of scorecard. The projects we crawl are in [cron/data/projects.csv](https://github.com/ossf/scorecard/blob/main/cron/data/projects.csv).

@azeemshaikh38 who's leading the above and will have more comments.

laurentsimon:
as part of the work to create a GitHub action that works on pull requests (content-based checks), I'm going to work on this soon. What I plan to do is create a new [repoClient](https://github.com/ossf/scorecard/blob/main/clients/repo_client.go) that takes as input a folder containing code that is already downloaded. The new repo client implementation will only support content-based checks and will use the folder passed as argument by users. Scorecard will then be run as:
```
$git clone some/repo
$scorecard ... --source-code=./repo
```
I will re-use this functionality for the GitHub action on pull request events.

Any further comment?

goern:
I'd love to see a Tekton task for this too ;) Let me know if you need some help with it or if Project Thoth shall go ahead and publish on on the tekton catalog.

laurentsimon:
Hi @goern , thanks for the interest in the project. Having a tekton task sounds like an awesome thing to have!

Would you need all the checks for the task (like we do for our GitHub actions on cron and push triggers https://github.com/ossf/scorecard/issues/1074), or a subset of checks that run on source-code only (basically, this PR)?

What output would you need? (we have JSON and SARIF). SARIF is based on a policy, whereas JSON is the raw output with scores for each check and an aggregate score for the repo.

If you have cycles to build work on it, we'd be more than happy for you to own it.

azeemshaikh38:
Will give this a shot. Might be something worth pursuing for v5.

laurentsimon:
[...]

PR #1562, opened 2022-01-29
🌱 Only return PRs associated with recent commits

### LINKED ISSUE #1524: BUG: Checks using ListMergedPRs() need hardening
`ListMergedPRs()` lists the last 30 or so PRs in a repo. Imagine the following scenario:
1. Repo owners using PRs and review their code, etc
2. They stop using PRs and only commit directly to head

IIUC, Scorecard will be looking at old PRs, e.g. in the Code-Review check, and conclude that reviews are done on the repo, even though it's not the case.

We need to harden all the checks that list PRs and compare their `MergeCommit` to the ones returned by `ListCommits()`.

Could we do this within the repo interface implementation, in order to avoid making mistakes in check implementations. For example, shall we make `ListMergedPRs()` return only the PRs that are in the `ListCommits()`? Something along these lines. Note: we'd need to be able to differentiate between "n PRs" only are available for this repo and "n PRs out of last m commits"

azeemshaikh38:

Only returns PRs used to merge recent commits if any. None otherwise.

Fixes #1524. Part of #575.

No.

### DIFF
--- a/clients/githubrepo/graphql.go
+++ b/clients/githubrepo/graphql.go
@@ -49,44 +49,53 @@ type graphqlData struct {
 							CommittedDate githubv4.DateTime
 							Message       githubv4.String
 							Oid           githubv4.GitObjectID
-							Committer     struct {
+							Author        struct {
+								User struct {
+									Login githubv4.String
+								}
+							}
+							Committer struct {
 								Name *string
 								User struct {
 									Login *string
 								}
 							}
+							AssociatedPullRequests struct {
+								Nodes []struct {
+									Repository struct {
+										Name  githubv4.String
+										Owner struct {
+											Login githubv4.String
+										}
+									}
+									Author struct {
+										Login githubv4.String
+									}
+									Number      githubv4.Int
+									HeadRefOid  githubv4.String
+									MergedAt    githubv4.DateTime
+									MergeCommit struct {
+										// NOTE: only used for sanity check.
+										// Use original commit oid instead.
+										Oid githubv4.GitObjectID
+									}
+									Labels struct {
+										Nodes []struct {
+											Name githubv4.String
+										}
+									} `graphql:"labels(last: $labelsToAnalyze)"`
+									Reviews struct {
+										Nodes []struct {
+											State githubv4.String
+										}
+									} `graphql:"reviews(last: $reviewsToAnalyze)"`
+								}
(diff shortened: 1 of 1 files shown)


naveensrinivasan (approved):

azeemshaikh38:
Ok I need some inputs here @laurentsimon @naveensrinivasan @justaugustus - if we get the PRs from `commit.associatedPR` instead of the repository, the behavior is different on some repos like forks. Basically, forks do not have any PRs, so GitHub APIs return no PRs on forks. But with `commit.associatedPR`, GitHub recognizes that the commit has a PR on the original repo so it returns that PR as part of `commit.associatedPR`. Now, the question is, is this a behavior we want?

I personally think this is the right behavior, since we want to analyze "commits" not PRs. Commits might come from Gerrit, Prow or a GitHub PR of a different repo and we should be able to recognize that. What do you folks think?

azeemshaikh38:
> Ok I need some inputs here @laurentsimon @naveensrinivasan @justaugustus - if we get the PRs from `commit.associatedPR` instead of the repository, the behavior is different on some repos like forks. Basically, forks do not have any PRs, so GitHub APIs return no PRs on forks. But with `commit.associatedPR`, GitHub recognizes that the commit has a PR on the original repo so it returns that PR as part of `commit.associatedPR`. Now, the question is, is this a behavior we want?
> 
> I personally think this is the right behavior, since we want to analyze "commits" not PRs. Commits might come from Gerrit, Prow or a GitHub PR of a different repo and we should be able to recognize that. What do you folks think?

Let's discuss further on #575. I was able to figure out how to keep the current behavior unchanged, but this is something we should consider.

laurentsimon:
> Ok I need some inputs here @laurentsimon @naveensrinivasan @justaugustus - if we get the PRs from `commit.associatedPR` instead of the repository, the behavior is different on some repos like forks. Basically, forks do not have any PRs, so GitHub APIs return no PRs on forks. But with `commit.associatedPR`, GitHub recognizes that the commit has a PR on the original repo so it returns that PR as part of `commit.associatedPR`. Now, the question is, is this a behavior we want?
> 
> I personally think this is the right behavior, since we want to analyze "commits" not PRs. Commits might come from Gerrit, Prow or a GitHub PR of a different repo and we should be able to recognize that. What do you folks think?

LGTM at first sight.

PR #1566, opened 2022-01-30
⚠️ Remove ListMergedPRs API

azeemshaikh38:

Instead of relying on 2 potentially divergent APIs - `ListCommits` and `ListMergedPRs`, reduce the API to only `ListCommits`. Any PRs associated with this commit will be return in the `associatedPR` field. It makes Scorecard more tightly tied to analyzing a repo's commits (part of #575) and helps avoid confusion in future implementations of `RepoClient` (long-term fix of #1524).

Yes.

### DIFF
--- a/checks/ci_tests.go
+++ b/checks/ci_tests.go
@@ -39,16 +39,18 @@ func init() {
 
 // CITests runs CI-Tests check.
 func CITests(c *checker.CheckRequest) checker.CheckResult {
-	prs, err := c.RepoClient.ListMergedPRs()
+	commits, err := c.RepoClient.ListCommits()
 	if err != nil {
-		e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("RepoClient.ListMergedPRs: %v", err))
+		e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("RepoClient.ListCommits: %v", err))
 		return checker.CreateRuntimeErrorResult(CheckCITests, e)
 	}
 
 	totalMerged := 0
 	totalTested := 0
-	for index := range prs {
-		pr := &prs[index]
+	for i := range commits {
+		pr := &commits[i].AssociatedMergeRequest
+		// TODO(#575): We ignore associated PRs if Scorecard is being run on a fork
+		// but the PR was created in the original repo.
 		if pr.MergedAt.IsZero() {
 			continue
 		}
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -14,6 +14,8 @@
 
 package checker
 
+import "time"
+
 // RawResults contains results before a policy
 // is applied.
 type RawResults struct {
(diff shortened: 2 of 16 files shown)


checks/ci_tests.go
```diff
@@ -39,17 +39,17 @@ func init() {
...
 	}
 
 	totalMerged := 0
 	totalTested := 0
-	for index := range prs {
-		pr := &prs[index]
-		if pr.MergedAt.IsZero() {
+	for i := range commits {
+		pr := &commits[i].AssociatedMergeRequest
+		if pr.MergedAt.IsZero() || pr.Repository != c.Repo.String() {
```
laurentsimon, line 52: I'm curious: does ListCommits() list all commits to branches used during PRs? If not, why is `pr.Repository != c.Repo.String()` needed?

azeemshaikh38: `ListCommits` is basically `git log` on a repository. However, the associated PR with that commit might come from a different repo (say in case of a fork). So, I'm ignoring such PRs to maintain our current behavior. 

There is a bigger discussion to be had on whether we should honor commits whose PRs originated on a different repo. i.e, should such commits be considered code reviewed etc. But I don't want that discussion to affect this PR. The aim of this PR is to remove the API but keep behavior the same.

laurentsimon: > `ListCommits` is basically `git log` on a repository. However, the associated PR with that commit might come from a different repo (say in case of a fork). So, I'm ignoring such PRs to maintain our current behavior.

sorry, what does a different repo mean? When we submit PR to scorecard repo, we use a different repo, which is a fork of the original one. But it is technically a different repo, right? Does it mean these PRs are ignored? Or does "different repo" mean something different?

> There is a bigger discussion to be had on whether we should honor commits whose PRs originated on a different repo. i.e, should such commits be considered code reviewed etc. But I don't want that discussion to affect this PR. The aim of this PR is to remove the API but keep behavior the same.

azeemshaikh38: So let's say you run Scorecard on `laurentsimon/scorecard`. `ListCommits()` will return [58865e959e3a782d3f3cd5a5ae952ac308c11a46](https://github.com/laurentsimon/scorecard/commit/58865e959e3a782d3f3cd5a5ae952ac308c11a46)  as one of the commits in this repository. But the PR which created this commit is https://github.com/ossf/scorecard/pull/1562.  

So `commit.associatedPR` will point to #1562 which belongs to a different repository. Does that make sense?

laurentsimon: ho I see. So this only affects forks. Let's add a comment to the code to clarify things. I suppose it's OK to consider these as reviews, but we will discuss this in another issue. Thanks for the explaination

laurentsimon (approved):

checks/ci_tests.go
azeemshaikh38: Done.

laurentsimon (approved):
LGTM overall. Thanks

azeemshaikh38:
Friendly ping for a review.

PR #1567, opened 2022-01-30
🌱 Refactor some code

azeemshaikh38:

Refactors code. Minor behavior change - instead of `--raw` option, use `--format=raw`.

No.

### DIFF
--- a/clients/localdir/repo.go
+++ b/clients/localdir/repo.go
@@ -21,17 +21,11 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"strings"
 
 	clients "github.com/ossf/scorecard/v4/clients"
 )
 
-var (
-	errNotDirectory = errors.New("not a directory")
-	errInvalidURI   = errors.New("invalid URI")
-)
-
-var filePrefix = "file://"
+var errNotDirectory = errors.New("not a directory")
 
 type repoLocal struct {
 	path     string
@@ -78,10 +72,7 @@ func (r *repoLocal) AppendMetadata(m ...string) {
 
 // MakeLocalDirRepo returns an implementation of clients.Repo interface.
 func MakeLocalDirRepo(pathfn string) (clients.Repo, error) {
-	if !strings.HasPrefix(pathfn, filePrefix) {
-		return nil, fmt.Errorf("%w", errInvalidURI)
-	}
-	p := path.Clean(pathfn[len(filePrefix):])
+	p := path.Clean(pathfn)
 	repo := &repoLocal{
 		path: p,
 	}
--- a/cmd/package_managers.go
+++ b/cmd/package_managers.go
@@ -1,0 +1,156 @@
+// Copyright 2020 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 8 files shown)


cmd/package_managers.go
```diff
@@ -0,0 +1,157 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	sce "github.com/ossf/scorecard/v4/errors"
+)
+
+var errInvalidInput = errors.New("only one package name must be non-empty")
+
+// TODO(cmd): Add unit tests for this.
```
naveensrinivasan, line 30: Can you please create an issue for this?

cmd/package_managers.go
```diff
@@ -0,0 +1,157 @@
...
+func fetchGitRepositoryFromPackageManagers(npm, pypi, rubygems string) (bool, string, error) {
```
naveensrinivasan, line 31: IMO tuple is good to read in returns, anything more than `2` should be a struct. It makes it easier to maintain and read. This is with respect to  `bool, string, error`

cmd/root.go
```diff
@@ -400,154 +387,43 @@ func getEnabledChecks(sp *spol.ScorecardPolicy, argsChecks []string,
...
-	err = json.NewDecoder(resp.Body).Decode(v)
-	if err != nil {
-		return "", sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("failed to parse ruby gem json: %v", err))
-	}
-	if v.SourceCodeURI == "" {
-		return "", sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("could not find source repo for ruby gem: %v", err))
-	}
-	return v.SourceCodeURI, nil
+func getRepoAccessors(ctx context.Context, repoURI, localURI string, logger *sclog.Logger) (
+	clients.Repo, // repo
```
naveensrinivasan, line 391: I like the comments on these returns. Thanks 👍

naveensrinivasan (approved):
Thank you! Few suggestions on the code. But I am OK if you think these aren't required from your perspective. Feel free to merge 👍

cmd/package_managers.go
azeemshaikh38: Done - #1568

cmd/package_managers.go
azeemshaikh38: Done.

clients/localdir/repo.go
```diff
@@ -78,10 +72,7 @@ func (r *repoLocal) AppendMetadata(m ...string) {
 
 // MakeLocalDirRepo returns an implementation of clients.Repo interface.
 func MakeLocalDirRepo(pathfn string) (clients.Repo, error) {
-	if !strings.HasPrefix(pathfn, filePrefix) {
-		return nil, fmt.Errorf("%w", errInvalidURI)
-	}
-	p := path.Clean(pathfn[len(filePrefix):])
```
laurentsimon, line 84: make sure to test nothin breaks after this change.

cmd/package_managers.go
```diff
@@ -0,0 +1,156 @@
+// Copyright 2020 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
```
laurentsimon, line 5: is this something anyone uses in practice? Is it worth maintaining?

laurentsimon (approved):

clients/localdir/repo.go
azeemshaikh38: Tested with `--local` on my workstation and it worked fine.

cmd/package_managers.go
azeemshaikh38: No clue. Worth discussing this though. Maybe start a new issue for this?

I don't want to add breaking changes in this PR though, so will leave it as is for now.

clients/localdir/repo.go
azeemshaikh38: Let me know if there is any other tests I can run for this.

laurentsimon: should be fine. Can't think of anything else atm

PR #1575, opened 2022-01-31
✨ Mention renovatebot's settings

laurentsimon:
Explicitly say that renovatebot can help migrate from version pinning to hash pinning via a setting.
See https://github.com/dependabot/dependabot-core/issues/3699#issuecomment-1025715911

### DIFF
--- a/docs/checks/internal/checks.yaml
+++ b/docs/checks/internal/checks.yaml
@@ -78,6 +78,10 @@ checks:
         be enabled for forks where security updates have ever been turned on so projects
         maintaining stable forks should evaluate whether this behavior is satisfactory
         before turning it on.
+      - >-
+        Renovatebot has support to migrate dockerfiles' dependencies from version pinning to hash pinning 
+        via the [pinDigests setting](https://docs.renovatebot.com/configuration-options/#pindigests) without
+        aditional manual effort.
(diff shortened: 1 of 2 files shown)


naveensrinivasan (approved):
Thank you!

docs/checks/internal/checks.yaml
```diff
@@ -78,6 +78,10 @@ checks:
         be enabled for forks where security updates have ever been turned on so projects
         maintaining stable forks should evaluate whether this behavior is satisfactory
         before turning it on.
+      - >-
+        Unlike dependabot, renovatebot has support to migrate dockerfiles' dependencies from version pinning to hash pinning
```
azeemshaikh38, line 82: Nit: might come across as if we are putting down dependabot. Consider changing to:

```
Renovatebot supports migrating Dockerfile dependencies from ...
```

azeemshaikh38 (approved):

docs/checks/internal/checks.yaml
azeemshaikh38: I thought `conversation resolution before merging` should stop this PR from being merged, lol.

laurentsimon: it should have!

Issue #435, opened 2021-05-11
Feature: improve unit tests and e2e test for better coverage


laurentsimon:
there are many tests, so this should be done progressively

azeemshaikh38:
+1. We really need this. I also think we should make existing tests to be hermetic and not flaky.

spencerschrock:
Given the age I'm going to mark this as done, especially given the unit test work Naveen did last year.

PR #1579, opened 2022-02-01
🌱 Only run allowed checks in different modes

azeemshaikh38:

Replace file parsing logic for recognizing supported/unsupported checks. 

No.

### DIFF
--- a/checks/binary_artifact.go
+++ b/checks/binary_artifact.go
@@ -26,7 +26,10 @@ const CheckBinaryArtifacts string = "Binary-Artifacts"
 
 //nolint
 func init() {
-	if err := registerCheck(CheckBinaryArtifacts, BinaryArtifacts); err != nil {
+	var supportedRequestTypes = []checker.RequestType{
+		checker.FileBased,
+	}
+	if err := registerCheck(CheckBinaryArtifacts, BinaryArtifacts, supportedRequestTypes); err != nil {
 		// this should never happen
 		panic(err)
 	}
--- a/docs/checks/internal/validate/main.go
+++ b/docs/checks/internal/validate/main.go
@@ -14,233 +14,19 @@
 package main
 
 import (
-	"errors"
 	"fmt"
-	"os"
-	"path"
-	"regexp"
 	"strings"
-
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"github.com/ossf/scorecard/v4/checks"
 	docs "github.com/ossf/scorecard/v4/docs/checks"
 )
 
-var (
-	allowedRisks     = map[string]bool{"Critical": true, "High": true, "Medium": true, "Low": true}
-	allowedRepoTypes = map[string]bool{"GitHub": true, "local": true}
-	supportedAPIs    = map[string][]string{
-		// InitRepo is supported for local repos in general. However, in the context of checks,
-		// this is only used to look up remote data, e.g. in Fuzzing check.
-		// So we only have "GitHub" supported.
-		"InitRepo":                   {"GitHub"},
-		"URI":                        {"GitHub", "local"},
-		"IsArchived":                 {"GitHub"},
-		"ListFiles":                  {"GitHub", "local"},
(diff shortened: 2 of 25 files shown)


docs/checks/internal/validate/main.go
```diff
@@ -14,235 +14,21 @@
...
 	docs "github.com/ossf/scorecard/v4/docs/checks"
 )
 
-var (
-	allowedRisks     = map[string]bool{"Critical": true, "High": true, "Medium": true, "Low": true}
-	allowedRepoTypes = map[string]bool{"GitHub": true, "local": true}
-	supportedAPIs    = map[string][]string{
-		// InitRepo is supported for local repos in general. However, in the context of checks,
-		// this is only used to look up remote data, e.g. in Fuzzing check.
-		// So we only have "GitHub" supported.
```
laurentsimon, line 37: I think it's dangerous to remove some sanity checks. This code already caught one problem when someone updated a check and it would have changed the supported checks. I'm pretty sure we will update some checks; forget to update the `registerCheck()` and it will break the `--local` feature, and later the actions running on pull requests (not quite there yet though :-)) Today we don't have unit tests for the `--repo` option, which will make it hard to catch.

What can we do to add sanity checks?

azeemshaikh38: #435 would add e2e tests for `--local` option?

laurentsimon: in the pipeline, yes.

azeemshaikh38: We can consider prioritizing it if that addresses your concern here. I need this to unblock myself for `commit-based` Scorecard work. Would prefer getting this PR in and in parallel working to get the e2e tests for this.

laurentsimon: I think this should work. Let's try to avoid releasing a scorecard action before these checks are back in.

checks/binary_artifact.go
```diff
@@ -26,7 +26,10 @@ const CheckBinaryArtifacts string = "Binary-Artifacts"
 
 //nolint
 func init() {
-	if err := registerCheck(CheckBinaryArtifacts, BinaryArtifacts); err != nil {
+	var supportedRequestTypes = []checker.RequestType{
+		checker.FileBased,
+	}
+	if err := registerCheck(CheckBinaryArtifacts, BinaryArtifacts, supportedRequestTypes); err != nil {
```
laurentsimon, line 32: fyi, when we added the `--local` repo, we decided to have the metadata in the checks.yaml rather than hardcoded in the implementation. What's the motivation for changing it?

laurentsimon (approved):
LGTm overall. Added a question why the type is moved from checks.yaml to the check's implementation

checks/binary_artifact.go
azeemshaikh38: > we decided to have the metadata in the checks.yaml rather than hardcoded in the implementation.

True. After our current implementation I have found it confusing that the `docs` package validates the correctness of `checks`. My main motivation is changing this is so that each check is complete in itself and can self-validate, if that makes sense. 

Let me know. Will wait on your response to submit this PR.

docs/checks/internal/validate/main.go
azeemshaikh38: SGTM.

checks/binary_artifact.go
oliverchang: Moving here seems OK to me. Assuming I'm understanding this correctly, this level of granularity like an implementation detail that users don't really have to understand. 

`local` on the other hand would be useful for users of scorecards to know and be documented in `checks.yaml`/`checks.md`.

oliverchang: Or actually maybe I've misunderstood this -- the plan is to move the source of truth for "local" completely to being in the implementation. That seems fine to me, and while our output `checks.md` doesn't show this today, in the future we might want to show this there as well. 

Either way, this seems fine to me.

azeemshaikh38: Thanks! I'm merging this in that case.

PR #1601, opened 2022-02-03
:bug: Fix Dependabot check to accept .yaml file extension

### LINKED ISSUE #1600: BUG: dependency update tool check does not accept .yaml extension
The Dependabot check does not consider`.github/dependabot.yaml` file, however both `.yml` and `.yaml` extensions should be accepted according to the [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).

behnazh-w:
The Dependabot check does not consider `.github/dependabot.yaml` file, however both `.yml` and `.yaml` extensions should be accepted according to the [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).

This PR fixes this bug by adding `.github/dependabot.yaml` to the files checked in the dependency update tool check, and provides a test to check for `.github/dependabot.yaml` file.

Closes #1600

### DIFF
--- a/checks/raw/dependency_update_tool.go
+++ b/checks/raw/dependency_update_tool.go
@@ -43,7 +43,7 @@ func checkDependencyFileExists(name string, data fileparser.FileCbData) (bool, e
 	}
 
 	switch strings.ToLower(name) {
-	case ".github/dependabot.yml":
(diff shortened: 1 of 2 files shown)


laurentsimon (approved):
Good catch, Thank you!

PR #1615, opened 2022-02-08
✨ Check for secrets in workflows run on pull requests

laurentsimon:
See point 5 in https://github.com/ossf/scorecard/issues/426#issuecomment-856939499

This PR looks for secrets used in job's env, jobs' `with` argument, job's `run` scripts and top-level `env` of the workflow.

We add a constraint that the workflow must run on `pull_request` and jobs must have a call to `action/checkout` of the PR.


checks/dangerous_workflow.go
```diff
@@ -120,23 +128,70 @@ func validateGitHubActionWorkflowPatterns(path string, content []byte, dl checke
...
 	}
+
 	return nil
 }
 
-func checkPullRequestTrigger(workflow *actionlint.Workflow) bool {
+func usesPullRequestTrigger(workflow *actionlint.Workflow) bool {
+	// Check if the webhook event trigger is a pull_request_target
+	for _, event := range workflow.On {
+		e, ok := event.(*actionlint.WebhookEvent)
```
azeemshaikh38, line 185: Is there a reason to validate this interface's underlying implementation? Can we not simply do `event.EventName()`?

checks/dangerous_workflow.go
```diff
@@ -120,23 +128,70 @@ func validateGitHubActionWorkflowPatterns(path string, content []byte, dl checke
...
+		}
+	}
+
+	return false
+}
+
+func usesPullRequestTargetTrigger(workflow *actionlint.Workflow) bool {
 	// Check if the webhook event trigger is a pull_request_target
 	for _, event := range workflow.On {
 		e, ok := event.(*actionlint.WebhookEvent)
```
azeemshaikh38, line 197: Same comment as above - can we replace with `event.EventName()`?

checks/dangerous_workflow.go
```diff
@@ -120,23 +128,70 @@ func validateGitHubActionWorkflowPatterns(path string, content []byte, dl checke
...
+		if err := checkJobForUntrustedCodeCheckout(job, path, dl, pdata); err != nil {
+			return err
 		}
 	}
+
 	return nil
 }
 
-func checkPullRequestTrigger(workflow *actionlint.Workflow) bool {
+func usesPullRequestTrigger(workflow *actionlint.Workflow) bool {
```
azeemshaikh38, line 182: Consider generalizing these event lookup fns like these:

```
func usesEventTrigger(workflow *actionlint.Workflow, eventName string) bool {}
```

checks/dangerous_workflow.go
```diff
@@ -148,6 +203,69 @@ func checkPullRequestTrigger(workflow *actionlint.Workflow) bool {
...
+		return nil
+	}
+
+	// https://docs.github.com/en/actions/security-guides/encrypted-secrets#naming-your-secrets
+	for _, step := range job.Steps {
+		if step == nil {
+			continue
+		}
+
+		if err := checkSecretInActionArgrs(step, path, dl, pdata); err != nil {
```
azeemshaikh38, line 218: Typo in `Args`

azeemshaikh38 (approved):

checks/dangerous_workflow.go
```diff
@@ -120,23 +128,70 @@ func validateGitHubActionWorkflowPatterns(path string, content []byte, dl checke
...
+		return nil
+	}
+
+	// Secrets used in env at the top of the wokflow.
+	if err := checkSecretInEnv(workflow.Env, path, dl, pdata); err != nil {
+		return err
+	}
+
+	// Secrets used on jobs.
+	for _, job := range workflow.Jobs {
```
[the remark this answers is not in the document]
laurentsimon: correct, the check is in `checkJobForUsedSecrets()`

Note: I'm going to update the code today to also verify whether an `environment` is declared or not, since this indicates a env secret and is typically gated by an approval, so we can discard such jobs. We can add further logic later to read the actual environment rules if we want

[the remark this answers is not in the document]
laurentsimon: updated the PR description to reflect my last comment.

checks/dangerous_workflow.go
laurentsimon: I've replaced.

checks/dangerous_workflow.go
laurentsimon: replaced.

Issue #1393, opened 2021-12-15
Feature: deprecate Pass/Fail model


azeemshaikh38:
Scorecard will deprecate Pass/Fail model along with the confidence score by EO'Q1 2022. This issue tracks the progress of it.

PR #1628, opened 2022-02-11
🌱 Simplify DetailLogger

azeemshaikh38:

Part of #1393 

No.


checks/ci_tests.go
```diff
@@ -83,8 +83,9 @@ func CITests(c *checker.CheckRequest) checker.CheckResult {
 		}
 
 		if !foundCI {
-			c.Dlogger.Debug3(&checker.LogMessage{
-				Text: fmt.Sprintf("merged PR without CI test: %d", pr.Number),
+			c.Dlogger.Debug(&checker.LogMessage{
+				Text:    fmt.Sprintf("merged PR without CI test: %d", pr.Number),
+				Version: 3,
```
laurentsimon, line 88: do you want to set version to 3 in the Dlogger's functions to avoid mistakes? Setting version 3 by default is equivalent to `Warn3(Text:...)`, ie if there's no `Path` it will be ignored.

The version is only used at https://github.com/ossf/scorecard/blob/main/pkg/common.go#L45, and it's the same as https://github.com/ossf/scorecard/blob/main/pkg/common.go#L48. If `Path==nil`, it prints only the text message.

laurentsimon (approved):
Thanks!

checks/ci_tests.go
azeemshaikh38: I personally don't mind removing `Version` altogether. My only reasoning to keep it was to ensure overall behavior stayed the same. If you agree, can send a PR to remove the `Version` field and simplify this further.

laurentsimon: makes sense. Can do in follow-up PR.

Issue #1618, opened 2022-02-09
Improve scorecard score for scorecard repo


inferno-chromium:
**Describe the bug**

**Reproduction steps**
https://deps.dev/go/github.com%2Fossf%2Fscorecard
Are these false positives? If not, please fix. Also, for false positives, is there any easy way to add exceptions/ignore-lists [this is important as these workflows issues probably happen on other criticial repos as well]

**Expected behavior**
10/10 score for scorecard repo.

**Additional context**

---

- [ ] Branch protection: https://github.com/ossf/scorecard/pull/1630
- [ ] OpenSSF Best Practices badge: https://github.com/ossf/scorecard/pull/1629
- [ ] Dangerous workflow: 
- [ ] Fuzzing: https://github.com/google/oss-fuzz/pull/7268
- [ ] Token permissions:

naveensrinivasan:
Here is the latest release https://deps.dev/go/github.com%2Fossf%2Fscorecard%2Fv4

justaugustus:
Sounds fun!
/assign

justaugustus:
Local run against https://github.com/ossf/scorecard/pull/1629:

```console
./scorecard --repo=https://github.com/ossf/scorecard --show-details
Starting [Maintained]
Starting [Dependency-Update-Tool]
Starting [Fuzzing]
Starting [Contributors]
Starting [CI-Tests]
Starting [Signed-Releases]
Starting [SAST]
Starting [License]
Starting [Token-Permissions]
Starting [Vulnerabilities]
Starting [CII-Best-Practices]
Starting [Code-Review]
Starting [Pinned-Dependencies]
Starting [Binary-Artifacts]
Starting [Branch-Protection]
Starting [Security-Policy]
Starting [Dangerous-Workflow]
Starting [Packaging]
Finished [Maintained]
Finished [Fuzzing]
Finished [Contributors]
Finished [CI-Tests]
Finished [Signed-Releases]
Finished [SAST]
Finished [Dependency-Update-Tool]
Finished [Token-Permissions]
Finished [Vulnerabilities]
Finished [CII-Best-Practices]
Finished [Code-Review]
Finished [Pinned-Dependencies]
Finished [License]
Finished [Branch-Protection]
Finished [Security-Policy]
Finished [Dangerous-Workflow]
Finished [Packaging]
Finished [Binary-Artifacts]
```

RESULTS
-------
Aggregate score: 8.0 / 10

Check scores:

[...]

PR #1630, opened 2022-02-12
🌱 .github: Add settings.yml to enforce repo configuration

justaugustus:
- [x] (N/A) Tests for the changes have been added (for bug fixes / features)

Part of https://github.com/ossf/scorecard/issues/1618.
**REQUIRES ENABLING [`probot/settings`](https://github.com/apps/settings) ON ORG/REPO-LEVEL**

- Updates team/collaborator permissions
- Enforces branch protection settings

* **Other information**:

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -1,0 +1,87 @@
+repository:
+  # See https://developer.github.com/v3/repos/#edit for all available settings.
+
+  # The name of the repository. Changing this will rename the repository
+  name: scorecard
+
+  # A short description of the repository that will show up on GitHub
+  description: Security Scorecards - Security health metrics for Open Source
+
+  # A URL with more information about the repository
+  homepage: ''
+
+  # Updates the default branch for this repository.
+  default_branch: main
+
+  # Either `true` to enable automated security fixes, or `false` to disable
+  # automated security fixes.
+  enable_automated_security_fixes: true
+
+  # Either `true` to enable vulnerability alerts, or `false` to disable
+  # vulnerability alerts.
+  enable_vulnerability_alerts: true
+
+# See https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions for available options
+teams: []
+    # The permission to grant the team. Can be one of:
+    # * `pull` - can pull, but not push to or administer this repository.
+    # * `push` - can pull and push, but not administer this repository.
+    # * `admin` - can pull, push and administer this repository.
+    # * `maintain` - Recommended for project managers who need to manage the repository without access to sensitive or destructive actions.
+
+# Collaborators: give specific users access to this repository.
+# See https://docs.github.com/en/rest/reference/collaborators for available options
+
+# THIS LIST SHOULD ALWAYS MATCH TOP-LEVEL CODEOWNERS
+# https://github.com/ossf/scorecard/blob/main/.github/CODEOWNERS
+
+# TODO(owners): For ease of management, this should eventually shift to a
+#               defined GitHub team instead of individual usernames
+collaborators:
+  - name: azeemshaikh38
+    permission: admin
+
+  - name: inferno-chromium
+    permission: admin
+
+  - name: justaugustus
+    permission: admin
+
+  - name: laurentsimon
+    permission: admin
+
+  - name: naveensrinivasan
+    permission: admin
+
+branches:
+  - name: main
+    # https://docs.github.com/en/rest/reference/repos#update-branch-protection
+    # Branch Protection settings. Set to null to disable
+    protection:
+      # Required. Require at least one approving review on a pull request, before merging. Set to null to disable.
+      required_pull_request_reviews:
+        # The number of approvals required. (1-6)
+        required_approving_review_count: 1
+        # Dismiss approved reviews automatically when a new commit is pushed.
+        dismiss_stale_reviews: true
+        # Blocks merge until code owners have reviewed.
+        require_code_owner_reviews: true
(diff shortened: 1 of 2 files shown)


.github/settings.yml
```diff
@@ -0,0 +1,87 @@
...
+#               defined GitHub team instead of individual usernames
+collaborators:
+  - name: azeemshaikh38
+    permission: admin
+
+  - name: inferno-chromium
+    permission: admin
+
+  - name: justaugustus
+    permission: admin
```
naveensrinivasan, line 48: Write now @justaugustus has `write permissions`.

naveensrinivasan (commented):
Thanks! I approve. @azeemshaikh38 @inferno-chromium @laurentsimon 👀

.github/settings.yml
```diff
@@ -0,0 +1,87 @@
...
+  - name: main
+    # https://docs.github.com/en/rest/reference/repos#update-branch-protection
+    # Branch Protection settings. Set to null to disable
+    protection:
+      # Required. Require at least one approving review on a pull request, before merging. Set to null to disable.
+      required_pull_request_reviews:
+        # The number of approvals required. (1-6)
+        required_approving_review_count: 1
+        # Dismiss approved reviews automatically when a new commit is pushed.
+        dismiss_stale_reviews: true
```
inferno-chromium, line 66: we have this false right now as it makes it very annoying to do back and forth.

.github/settings.yml
inferno-chromium: Should be ok, @justaugustus has enough work on scorecards to justify this. you all should reeval later if we need that many admins or switch to 2 or something.

inferno-chromium (approved):

.github/settings.yml
azeemshaikh38: +1. We tried `dismiss_stale_reviews` once and it was difficult to work with. Would prefer to keep this as `false`.

.github/settings.yml
```diff
@@ -0,0 +1,87 @@
...
+        # Specify which users and teams can dismiss pull request reviews. Pass an empty dismissal_restrictions object to disable. User and team dismissal_restrictions are only available for organization-owned repositories. Omit this parameter for personal repositories.
+        dismissal_restrictions:
+          users: []
+          teams: []
+      # Required. Require status checks to pass before merging. Set to null to disable
+      required_status_checks:
+        # Required. Require branches to be up to date before merging.
+        strict: true
+        # Required. The list of status checks to require in order to merge into this branch
+        contexts: []
```
azeemshaikh38, line 78: Shouldn't this be populated to include the current status checks?

azeemshaikh38 (approved):

laurentsimon:
would the OSSF run the app or would we be OK giving a 3P the ability to change our settings (branch protection, admin, users, etc)?

.github/settings.yml
naveensrinivasan: I agree this should be false.

justaugustus (commented):
Left some comments to the comments :)

.github/settings.yml
justaugustus: Fixed in b31c8e3ad793e1f25cfcaf03ddd6c64c8f29ea8e.

justaugustus:
> would the OSSF run the app or would we be OK giving a 3P the ability to change our settings (branch protection, admin, users, etc)?

It's a GitHub App; are we fine with that? https://github.com/apps/settings
We could theoretically also run our own: https://github.com/probot/settings/blob/master/docs/deploy.md

naveensrinivasan:
> > would the OSSF run the app or would we be OK giving a 3P the ability to change our settings (branch protection, admin, users, etc)?
> 
> It's a GitHub App; are we fine with that? https://github.com/apps/settings We could theoretically also run our own: https://github.com/probot/settings/blob/master/docs/deploy.md

I am skeptical about giving an app to manage the admin settings.

.github/settings.yml
azeemshaikh38: That's fair, but I think we've had such a bad experience with this, I think it would be preferable to keep it off for now. 

I'm open to trying a different version of this (like the one you mentioned being used in Kubernetes). But I would prefer to keep that out of scope for this PR atleast and take it up as a separate issue/discussion.

azeemshaikh38:
> It's a GitHub App; are we fine with that? https://github.com/apps/settings
> We could theoretically also run our own: https://github.com/probot/settings/blob/master/docs/deploy.md

Oh I did not realize this was a third party GitHub app and not something natively supported by GitHub. I agree with @naveensrinivasan, I'm a bit skeptical about this too. Especially having read - https://github.com/probot/settings#security-implications (yikes!)

.github/settings.yml
laurentsimon: I also like the k8's way of lgtm :-) feature request for GitHub!

laurentsimon:
> > It's a GitHub App; are we fine with that? https://github.com/apps/settings
> > We could theoretically also run our own: https://github.com/probot/settings/blob/master/docs/deploy.md
> 
> Oh I did not realize this was a third party GitHub app and not something natively supported by GitHub. I agree with @naveensrinivasan, I'm a bit skeptical about this too. Especially having read - https://github.com/probot/settings#security-implications (yikes!)

an option is to run it ourselves, but it still creates a central point of failure. Who has access to the cloud project where the app runs? Essentially these folks now have branch protection bypass, and I don't know if configuration of a cloud project is that easy; and the config is not publicly available (@justaugustus maybe terraform,/k8 config can help?). 

To be fair, today all of us can change the settings anyway, so we *can* bypass it. 

GH apps, in general, make it harder to pin dependencies or know what code is running.

Not saying we should not use it, just adding additional thoughts to the conversation.

Issue #1389, opened 2021-12-13
Feature: add Scorecard to OSS-Fuzz and CIFuzz


azeemshaikh38:
Improve Scorecard's Fuzzing score by adding it to OSS-Fuzz and enabling CIFuzz GH Action.

@ jonathanmetzman fyi.

naveensrinivasan:
Which API do you want to fuzz within scorecard? 

If it is just to get the checkmark for fuzzing then it doesn't add value.

Probably good for discussion.

evverx:
FWIW Looking at https://github.com/ossf/scorecard/issues/1256#issuecomment-969848255, I think it would make sense to cover h2non/filetype at least (assuming scorecard will keep using it for parsing untrusted data).

evverx:
As far as I know scorecard also parses random bash scripts with https://github.com/mvdan/sh so ideally it should probably be covered as well (though I agree technically it doesn't improve the fuzzing score of scorecard itself)

naveensrinivasan:
> As far as I know scorecard also parses random bash scripts with https://github.com/mvdan/sh so ideally it should probably be covered as well (though I agree technically it doesn't improve the fuzzing score of scorecard itself)

AFAIK the https://github.com/mvdan/sh is fuzzed probably randomly not using oss-fuzz.

naveensrinivasan:
> FWIW Looking at https://github.com/ossf/scorecard/issues/1256#issuecomment-969848255, I think it would make sense to cover h2non/filetype at least (assuming scorecard will keep using it for parsing untrusted data).

I agree,I think it would be better to include the fuzzing that repository instead of Scorecard. 

Thanks for bringing it up. 

It makes sense.

evverx:
scorecard has its own file parser located in https://github.com/ossf/scorecard/tree/main/checks/fileparser though (which I think is a good candidate for fuzzing).

> AFAIK the https://github.com/mvdan/sh is fuzzed probably randomly not using oss-fuzz

Good to know. Thanks! But given that finding and fixing bugs found by fuzz targets is like playing whack-a-mole if it isn't fuzzed continuous it's probably safe to say it isn't fuzzed.

naveensrinivasan:
> scorecard has its own file parser located in https://github.com/ossf/scorecard/tree/main/checks/fileparser though (which I think is a good candidate for fuzzing).
> 
> 
> 
[...]

PR #1631, opened 2022-02-12
🌱 Add policy fuzzer

AdamKorcz:
Adds a fuzzer for `ParseFromYAML` as discussed in https://github.com/google/oss-fuzz/pull/7269

Adds a fuzz test.

No.

* **Other information**:

### DIFF
--- a/policy/policy_fuzz_test.go
+++ b/policy/policy_fuzz_test.go
@@ -1,0 +1,28 @@
+//go:build gofuzz
+// +build gofuzz
+
+// Copyright 2022 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package policy
+
+import (
+	"testing"
+)
+
+func FuzzParseFromYAML(f *testing.F) {
+	f.Fuzz(func(t *testing.T, b []byte) {
+		ParseFromYAML(b)
+	})
+}


evverx:
I think it would be great to turn on CIFuzz/CFLite as well. Hopefully scorecard complaining about unpinned dependencies in scorecard :-) could help to fix issues like https://github.com/google/oss-fuzz/issues/7206 and https://github.com/google/oss-fuzz/issues/6836 a bit faster.

justaugustus (approved):
Thanks @AdamKorcz!

naveensrinivasan (commented):
Thanks, we also need to update the build scripts to validate. 

If not when the API changes it isn't going to fail.

naveensrinivasan:
Please don't merge this till we have build integrated to validate the fuzz code.

evverx:
> Please don't merge this till we have build integrated to validate the fuzz code.

@naveensrinivasan https://google.github.io/oss-fuzz/getting-started/continuous-integration/ is supposed to do exactly that. I should have posted that link probably. Sorry!

naveensrinivasan:
The scorecard has to compile anytime the code changes locally to avoid breaking this. 


Here is an example of fuzzing compile target 
https://github.com/sigstore/sigstore/blob/main/Makefile

evverx:
> The scorecard has to compile anytime the code changes locally to avoid breaking this.

I agree. Ideally it should be integrated into the build process but in general I wouldn't expect extIernal contributors to always build and run all the tests/fuzzers before PRs are submitted so CIFuzz is supposed to help there.

naveensrinivasan:
The external contributors need not but the CI on GitHub should validate this.

evverx:
> The external contributors need not but the CI on GitHub should validate this.

@naveensrinivasan I completely agree with you. What I was trying to say is that in addition to that it usually makes sense to turn on CIFuzz because the build and runtime environments provided by OSS-Fuzz are different from what fuzz target normally face locally or on GitHub. For example, CIFuzz caught https://github.com/systemd/systemd/actions/runs/1833518567 today even though the fuzz targets were built and run locally and on GitHub Actions.

naveensrinivasan:
Also we need documentation as to how build this and test it. 

Does it need go-tip, Go-fuzz?

We need to refer to Go-Fuzz in the go.mod. This makes adding additional dependency. 

Probably the fuzz should be a separate module to avoid these additional dependencies. 

So my recommendation would be we discuss these options and then merge this in. 

The fuzz need not be for just a func. The goal would be to add to many targets.

naveensrinivasan:
> > The external contributors need not but the CI on GitHub should validate this.
> 
> 
> 
> @naveensrinivasan I completely agree with you. What I was trying to say is that in addition to that it usually makes sense to turn on CIFuzz because the build and runtime environments provided by OSS-Fuzz are different from what fuzz targets normally face locally or on GitHub. For example, CIFuzz caught https://github.com/systemd/systemd/actions/runs/1833518567 today even though the fuzz targets were built and run locally and on GitHub Actions.

@evverx Agreed! Thank you for the great insights 👍.

naveensrinivasan (changes requested):
I am blocking this PR , till we have consensus. I agree with the code and it is just that we don't have the other infrastructure things to make this work , which I have discussed in detail https://github.com/ossf/scorecard/issues/1389#issuecomment-1038522409

Thank you! for the contribution

azeemshaikh38 (approved):
Thanks!

PR #1639, opened 2022-02-15
:seedling: Parallelize the builds

naveensrinivasan:

    The builds are slow on CI.

    - Created a workflow with multiple jobs for each of the docker builds
    - Created a workflow with multiple jobs for each of the ko builds
    - Removed the reference to dockerbuild and kobuild in the build-targets
      make target
    - This should reduce the time required to finish the CI builds as it
      makes it parallel.

* **Other information**:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,276 @@
+# Copyright 2021 Security Scorecard Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: docker-build
+on: [push, pull_request]
+jobs:
+  scorecard:
+    name: scorecard-docker
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+     - name: Install Protoc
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,276 @@
...
+     - name: Clone the code
+       uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+       with:
+          fetch-depth: 0
+     - name: Setup Go
+       uses: actions/setup-go@bfdd3570ce990073878bf10f6b2d79082de49492 # v2.2.0
+       with:
+         go-version: '^1.17'
+     - name: docker build
+       run: make scorecard-docker
```
azeemshaikh38, line 54: Could we parallelize this by saying `make -j X dockerbuild`? That way, a new docker build steps needs to be added only to the Makefile and not in 2 places (Makefile and here).

.github/workflows/ko.yml
```diff
@@ -0,0 +1,315 @@
...
+     - name: Clone the code
+       uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+       with:
+          fetch-depth: 0
+     - name: Setup Go
+       uses: actions/setup-go@bfdd3570ce990073878bf10f6b2d79082de49492 # v2.2.0
+       with:
+         go-version: '^1.17'
+     - name: ko build
+       run: make scorecard-ko
```
azeemshaikh38, line 87: Similar comment here.

azeemshaikh38 (commented):
Thanks, LGTM overall. Left a comment, PTAL.

.github/workflows/ko.yml
naveensrinivasan: The same applies here https://github.com/ossf/scorecard/pull/1639#discussion_r807079525

.github/workflows/docker.yml
naveensrinivasan: Let me know if you disagree.

azeemshaikh38 (approved):

PR #1643, opened 2022-02-15
✨ Add tests for graphQL costs

### LINKED ISSUE #1623: BUG: Fixer high GitHub API usage
#1566 increases the cost/complexity of the GraphQL queries. The query cost goes up ~20x due to that PR. We should fix this before the next release. 2-part solution here:

1. Decrease the number of `associatedPullRequests` to 1, since for every commit we are only interested in a single PR (the one used to submit the commit). Doing this brings back the cost to original, so ~20x savings.
2. Add some tests to verify the cost of our GraphQL queries. Any future issues can be caught using these tests. 

I would say, we block the next release on this issue since IMO it's important to fix this.

azeemshaikh38:

Closes #1623 

No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ test: $(test-targets)
 unit-test: ## Runs unit test without e2e
 	# Run unit tests, ignoring e2e tests
 	# run the go tests and gen the file coverage-all used to do the integration with codecov
-	go test -race -covermode=atomic  -coverprofile=unit-coverage.out `go list ./... | grep -v e2e`
+	SKIP_GINKGO=1 go test -race -covermode=atomic  -coverprofile=unit-coverage.out `go list ./...`
 
 $(GINKGO): install
 
--- a/clients/githubrepo/branches.go
+++ b/clients/githubrepo/branches.go
@@ -98,6 +98,7 @@ type branch struct {
 	BranchProtectionRule *branchProtectionRule
 }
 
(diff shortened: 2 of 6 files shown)


naveensrinivasan (approved):
Awesome! Thanks

laurentsimon (approved):

Makefile
```diff
@@ -261,7 +261,7 @@ test: $(test-targets)
 unit-test: ## Runs unit test without e2e
 	# Run unit tests, ignoring e2e tests
 	# run the go tests and gen the file coverage-all used to do the integration with codecov
-	go test -race -covermode=atomic  -coverprofile=unit-coverage.out `go list ./... | grep -v e2e`
+	SKIP_GINKGO=1 go test -race -covermode=atomic  -coverprofile=unit-coverage.out `go list ./...`
```
06kellyjac, line 264: Is there a reason these e2e tests couldn't sit in the e2e dir?

naveensrinivasan: @06kellyjac  Nice suggestion, Thanks. Do you want to do a PR for the change?

azeemshaikh38: The tests added in this PR rely on some internal (non-exposed) fields. So, had to create the tests in the same package as it was testing.

06kellyjac: fair enough, thanks for the details

Issue #1621, opened 2022-02-10
Release Scorecard - Next release (v4.1.0)


naveensrinivasan:
In Scorecard bi-weekly meeting it was brought up about the next version of the scorecard which has some bug fixes. 

This is a tracking item to discuss this.

naveensrinivasan:
@azeemshaikh38 @laurentsimon @justaugustus FYI...

azeemshaikh38:
Are folks working on any new features/changes which we should wait for before creating a release?

Also, for some reason the weekly cron jobs haven't been running fine last 2 weeks. I haven't looked into them (might be some GitHub token issue and nothing to do with code itself). But it would be useful to have one successful run and tag that commitSHA as the new release.

azeemshaikh38:
Found the reason why cron job was having issues - #1623. Working on a fix for it, would be good to release after this fix.

azeemshaikh38:
> let's try blocking on [ossf/scorecard-action#97](https://github.com/ossf/scorecard-action/issues/97), if nobody has objections. wdut?

Do we need to block Scorecard release for this though? This will rely on scorecard-action release correct?

laurentsimon:
you're correct, no need to block scorecard release. I re-phrased my comment :-)

azeemshaikh38:
#1623 is now fixed. I think we are good to release the next minor version. To be safe we can release at [394789cf22887344f084efeffcfe235cc8aa390f](https://github.com/ossf/scorecard/tree/394789cf22887344f084efeffcfe235cc8aa390f) which has been tested using the cron job. If folks have any concerns let's discuss before making the release.

@naveensrinivasan could you help with creating the new release?

jeffmendoza:
Yes, getting #1644 in a release would be helpful for Allstar, but we can use hash if you don't want to wait. Thanks

justaugustus:
> Yes, getting #1644 in a release would be helpful for Allstar, but we can use hash if you don't want to wait. Thanks

Merged!

justaugustus:
I'll echo what I mentioned in [Slack](https://openssf.slack.com/archives/C0235AR8N2C/p1644969016218859?thread_ts=1644968819.756379&cid=C0235AR8N2C):

With the logging change in https://github.com/ossf/scorecard/pull/1516, we're going to break consumers, so this should be a major version bump.

Updated the tracking title to reflect v5.0.0!

naveensrinivasan:
[...]

PR #1644, opened 2022-02-15
:sparkles: githubrepo: Allow providing an already authenticated transport

jeffmendoza:
Way back in 2.0 the constructor took a github.Client directly, which is what Allstar instantiated to then run scorecard checks. With the current code, there is no way to override the built-in roundtripper.NewTransport which uses its own methods to authenticate.

To allow Allstar to use the current code, we need to create this client and provide a pre-authenticated transport (or http client, or github client).

cc @justaugustus

### DIFF
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -177,10 +177,8 @@ func (client *Client) Close() error {
 	return client.tarball.cleanup()
 }
 
-// CreateGithubRepoClient returns a Client which implements RepoClient interface.
-func CreateGithubRepoClient(ctx context.Context, logger *log.Logger) clients.RepoClient {
-	// Use our custom roundtripper
-	rt := roundtripper.NewTransport(ctx, logger)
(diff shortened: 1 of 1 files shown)


justaugustus:
@jeffmendoza -- Hehe, I just opened a branch for this! Mind if I push a few things on top of this PR?

azeemshaikh38 (approved):

azeemshaikh38:
Will we need to make a Scorecard release for AllStar to be able to consume it? If so, please do chime in #1621. We can wait for this PR to go in before cutting the release.

justaugustus:
> @jeffmendoza -- Hehe, I just opened a branch for this! Mind if I push a few things on top of this PR?



> Will we need to make a Scorecard release for AllStar to be able to consume it? If so, please do chime in #1621. We can wait for this PR to go in before cutting the release.

Merged. Will push any updates as a follow-up! Thanks Jeff!

PR #1656, opened 2022-02-18
📖 Adding missing documentation for Token-Permissions

chrismcgehee:

Docs update

The Token-Permissions check documentation does not include documentation on when the score is reduced due to certain write permissions being present on a job.


azeemshaikh38 (approved):
Looks good to me. Will also wait for @laurentsimon's approval.

docs/checks.md
```diff
@@ -566,9 +566,22 @@ and the required write permissions are declared at the
...
+### Write permissions causing a small reduction
+* `statuses` - May allow an attacker to change the result of pre-submit checks and get a PR merged.
+* `checks` - May allow an attacker to remove pre-submit checks and introduce a bug.
+* `security-events` - May allow an attacker to read vulnerability reports before a patch is available. However, points are not reduced if the job utilizes a recognized action for uploading SARIF results.
+* `deployments` - May allow an attacker to charge repo owner by triggering VM runs, and tiny chance an attacker can trigger a remote service with code they own if server accepts code/location variables unsanitized.
+
+### Write permissions causing a large reduction
+* `contents` - Allows an attacker to commit unreviewed code. However, points are not reduced if the job utilizes a recognized packaging action or command.
+* `packages` - Allows an attacker to publish packages. However, points are not reduced if the job utilizes a recognized packaging action or command.
+* `actions` - May allow an attacker to steal GitHub secrets by adding a malicious workflow or action.
```
laurentsimon, line 584: do you mind double checking the `actions` permissions (I ought to take a closer look but have not fond the time to do so)? I don't think it allows adding a workflow, for example. I think it allows running an existing workflow, which could by pass the LGTM that is used before running workflows on pull request.

laurentsimon (approved):

docs/checks.md
justaugustus: @chrismcgehee -- Ping on Laurent's q above

chrismcgehee: If https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-actions exhaustively lists what can be done with the `actions:write` permission, then it allows:
* Deleting artifacts from a previous action run
* Approving to run an action that needs approval
* Canceling an action that's currently running
* Deleting all logs from a workflow run
* Re-run a workflow that was previously run

I think approving an action may allow an attacker to steal GitHub secrets. Does that seem like enough to keep it as a high risk permission?

laurentsimon: I'd say we can keep it as high for now, since secrets typically give access to resources like ppg keys, cloud services, etc. If we get complaints of false positives, we could double-check that there is at least one workflow that uses a secret in the repo.
Wdut?

chrismcgehee: That sounds good. I left the scoring the same but updating the verbiage for the documentation.

laurentsimon (approved):
Thanks Chris!

Issue #1254, opened 2021-11-13
Feature: differentiate between `package` and `contents` permission in Token-Permission 


laurentsimon:
We currently accepts `package:write` and `contents:write` if the workflow is a packaging workflow. We don't differentiate between a workflow that writes to a branch for the release (e.g. by pushing a binary for the tagged branch - needs `contents`) vs a workflow that pushes to a registry (need `package`).

laurentsimon:
you're correct. I'm just not 100% sure whether some actions can do both or not.

PR #1663, opened 2022-02-19
✨ Token-Permissions, Allow `contents: write` permission only for jobs that are releasing

chrismcgehee:

Feature

If a GitHub workflow job has `contents: write` permission and it is a publishing job, it will not have its score reduced.
https://github.com/ossf/scorecard/issues/1254

If a GitHub workflow job has `contents: write` permission and it is a publishing job, it will have its score reduced. If a GitHub workflow job has `contents: write` permission and it is a releasing job, it will not have its score reduced. 

No

* **Other information**:
I looked through all of `isPackagingWorkflow` and determined the only ones that looked like they needed `contents: write` were `goreleaser` and `python-semantic-release`. I left `goreleaser` in `isPackagingWorkflow` so as not to alter the behavior of the Packaging check, even though that action does not need the `packages: write` permission. `python-semantic-release` requires both `packages: write` and `contents: write`.


checks/permissions_test.go
```diff
@@ -94,8 +94,8 @@ func TestGithubTokenPermissions(t *testing.T) {
 				Error:         nil,
 				Score:         checker.MaxResultScore,
 				NumberOfWarn:  0,
-				NumberOfInfo:  3,
-				NumberOfDebug: 3,
+				NumberOfInfo:  2,
+				NumberOfDebug: 4,
```
chrismcgehee, line 98: The reason there is one less info and more debug log (here and in `package workflow write`) is because `requiresContentsPermissions` now returns false.

checks/fileparser/github_workflow.go
```diff
@@ -330,8 +330,36 @@ type JobMatcherStep struct {
 	Run string
 }
 
-// Matches returns true if the job matches the job matcher.
-func (m *JobMatcher) Matches(job *actionlint.Job) bool {
+// AnyJobsMatch returns true if any of the jobs have a match in the given workflow.
```
naveensrinivasan, line 333: `*actionlint.Workflow` Does it have to be a pointer? Harder to maintain and also moves to heap which mean GC pressure.

checks/fileparser/github_workflow.go
```diff
@@ -330,8 +330,36 @@ type JobMatcherStep struct {
...
+		Path:   fp,
+		Type:   checker.FileTypeSource,
+		Offset: checker.OffsetDefault,
+		Text:   logMsgNoMatch,
+	})
+	return false
+}
+
+// matches returns true if the job matches the job matcher.
+func (m *JobMatcher) matches(job *actionlint.Job) bool {
```
naveensrinivasan, line 362: Similar comment , should this be a `job *actionlint.Job` pointer?

naveensrinivasan (commented):
Thanks

checks/fileparser/github_workflow.go
chrismcgehee: My impression is that it's better for performance to pass struct parameters as pointers because the entire struct would have to be copied otherwise if it were not a pointer.

All the places that call this function have a `*actionlint.Workflow` available, so it would be a fairly large refactor to make it not a pointer. If you feel strongly these should all not use pointers, then I think that change should be made in a separate PR.

azeemshaikh38 (approved):
Thanks Chris!

checks/fileparser/github_workflow.go
azeemshaikh38: I have seen a bunch of linters complain when we pass a large struct by value instead of pointer. I think passing by pointer here is fine.

azeemshaikh38 (approved):

checks/testdata/.github/workflows/github-workflow-permissions-contents-writes-release.yaml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2021 Security Scorecard Authors
```
justaugustus, line 1:
```suggestion
# Copyright 2022 Security Scorecard Authors
```

checks/fileparser/github_workflow.go
justaugustus: Yep, passing by pointer is fine here.

justaugustus (approved):
Thanks Chris!

PR #1675, opened 2022-02-24
✨ checks: add GitHub Webhook check

### LINKED ISSUE #1655: Web hook secret used
See if we can add a check to verify that web hooks defined in the repo have a secret configured for authentication https://docs.github.com/en/developers/webhooks-and-events/webhooks/creating-webhooks#secret

Unfortunately, I don't think the APIs let us query them without a PAT https://docs.github.com/en/rest/reference/webhooks#repository-webhooks :/

Note, however, that the PAT only needs to be read-only. 

cc @azeemsgoogle @jeffmendoza

cpanato:

This adds a new check to validate the GitHub repositories Webhooks if the webhook has a secret configured.
Tested with a token with `public_repo` permission only and worked.

Fixes #1655 

### TODO:
- [ ] create a repo with webhooks configured to set up the e2e tests (@laurentsimon this is something that you can help with?)

NONE

Add a new check to validate GitHub Webhook

No

* **Other information**:

### DIFF
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -79,8 +79,8 @@ type WebhooksData struct {
 // WebhookData contains the raw results
 // for webhook check.
 type WebhookData struct {
-	ID        *int64
-	HasSecret *bool
+	ID             int64
+	UsesAuthSecret bool
 }
 
 // BranchProtectionsData contains the raw results
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -25,8 +25,8 @@ type RawResults struct {
 	DependencyUpdateToolResults DependencyUpdateToolData
 	BranchProtectionResults     BranchProtectionsData
 	CodeReviewResults           CodeReviewData
-	MaintainedResults           MaintainedData
 	WebhookResults              WebhooksData
+	MaintainedResults           MaintainedData
 }
 
 // MaintainedData contains the raw results
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -70,6 +70,19 @@ type DependencyUpdateToolData struct {
 	Tools []Tool
 }
 
+// WebhooksData contains the raw results
+// for the Webhook check.
+type WebhooksData struct {
+	Webhook []WebhookData
+}
+
+// WebhookData contains the raw results
+// for webhook check.
+type WebhookData struct {
+	ID        *int64
+	HasSecret *bool
+}
+
 // BranchProtectionsData contains the raw results
 // for the Branch-Protection check.
 type BranchProtectionsData struct {
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -26,6 +26,7 @@ type RawResults struct {
 	BranchProtectionResults     BranchProtectionsData
 	CodeReviewResults           CodeReviewData
 	MaintainedResults           MaintainedData
+	WebhookResults              WebhooksData
 }
 
 // MaintainedData contains the raw results
@@ -79,6 +80,7 @@ type WebhooksData struct {
 // WebhookData contains the raw results
 // for webhook check.
 type WebhookData struct {
+	Path           string
 	ID             int64
 	UsesAuthSecret bool
 }
--- a/checks/raw/webhook.go
+++ b/checks/raw/webhook.go
@@ -15,6 +15,8 @@
 package raw
 
 import (
+	"fmt"
+
 	"github.com/ossf/scorecard/v4/checker"
 	sce "github.com/ossf/scorecard/v4/errors"
 )
@@ -33,10 +35,10 @@ func WebHook(c *checker.CheckRequest) (checker.WebhooksData, error) {
 
 	hooks := []checker.WebhookData{}
 	for _, hook := range hooksResp {
-
 		v := checker.WebhookData{
 			ID:             hook.ID,
 			UsesAuthSecret: hook.UsesAuthSecret,
+			Path:           fmt.Sprintf("https://%s/settings/hooks/%d", c.RepoClient.URI(), hook.ID),
 			// Note: add fields if needed.
 		}
 		hooks = append(hooks, v)
--- a/checks/raw/webhook.go
+++ b/checks/raw/webhook.go
@@ -1,0 +1,45 @@
+// Copyright 2022 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package raw
+
+import (
+	"github.com/ossf/scorecard/v4/checker"
+	sce "github.com/ossf/scorecard/v4/errors"
+)
+
+// WebHook retrieves the raw data for the WebHooks check.
+func WebHook(c *checker.CheckRequest) (checker.WebhooksData, error) {
+	hooksResp, err := c.RepoClient.ListWebhooks()
+	if err != nil {
+		return checker.WebhooksData{},
+			sce.WithMessage(sce.ErrScorecardInternal, "Client.Repositories.ListWebhooks")
+	}
+
+	if len(hooksResp) < 1 {
+		return checker.WebhooksData{}, nil
+	}
+
+	hooks := []checker.WebhookData{}
+	for _, hook := range hooksResp {
+		v := checker.WebhookData{
+			ID:        &hook.ID,
+			HasSecret: &hook.HasSecret,
+			// Note: add fields if needed.
+		}
+		hooks = append(hooks, v)
+	}
+
+	return checker.WebhooksData{Webhook: hooks}, nil
+}
--- a/checks/raw/webhook.go
+++ b/checks/raw/webhook.go
@@ -33,9 +33,10 @@ func WebHook(c *checker.CheckRequest) (checker.WebhooksData, error) {
 
 	hooks := []checker.WebhookData{}
 	for _, hook := range hooksResp {
+
 		v := checker.WebhookData{
-			ID:        &hook.ID,
-			HasSecret: &hook.HasSecret,
+			ID:             hook.ID,
+			UsesAuthSecret: hook.UsesAuthSecret,
 			// Note: add fields if needed.
 		}
 		hooks = append(hooks, v)
--- a/checks/webhook.go
+++ b/checks/webhook.go
@@ -1,0 +1,66 @@
+// Copyright 2022 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checks
+
+import (
+	"fmt"
+
+	"github.com/ossf/scorecard/v4/checker"
+	sce "github.com/ossf/scorecard/v4/errors"
+)
+
+const (
+	// CheckWebHooks is the registered name for WebHooks.
+	CheckWebHooks = "Webhooks"
+)
+
+//nolint:gochecknoinits
+func init() {
+	if err := registerCheck(CheckWebHooks, WebHooks, nil); err != nil {
+		// this should never happen
+		panic(err)
+	}
+}
+
+// WebHooks run Contributors check.
+func WebHooks(c *checker.CheckRequest) checker.CheckResult {
+	hooks, err := c.RepoClient.ListWebhooks()
+	if err != nil {
+		e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("Client.Repositories.ListWebhooks: %v", err))
+		return checker.CreateRuntimeErrorResult(CheckWebHooks, e)
+	}
+
+	if len(hooks) < 1 {
+		return checker.CreateMaxScoreResult(CheckWebHooks, "no webhooks defined")
+	}
+
+	hasSecretToCount := 0
+	for _, hook := range hooks {
+		if !hook.HasSecret {
+			hasSecretToCount++
+		}
+	}
+
+	if hasSecretToCount == 0 {
+		return checker.CreateMaxScoreResult(CheckWebHooks, "all webhooks have secrets defined")
+	}
+
+	if len(hooks) == hasSecretToCount {
+		return checker.CreateMinScoreResult(CheckWebHooks, "webhooks with no secrets configured detected")
+	}
+
+	return checker.CreateProportionalScoreResult(CheckWebHooks,
+		"webhooks with no secrets configured detected", hasSecretToCount, len(hooks))
+}
--- a/checks/webhook.go
+++ b/checks/webhook.go
@@ -16,8 +16,13 @@ package checks
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/olekukonko/tablewriter"
 
 	"github.com/ossf/scorecard/v4/checker"
+	"github.com/ossf/scorecard/v4/checks/raw"
 	sce "github.com/ossf/scorecard/v4/errors"
 )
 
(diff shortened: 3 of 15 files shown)


checks/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+		if !hook.HasSecret {
+			hasSecretToCount++
+		}
+	}
+
+	if hasSecretToCount == 0 {
+		return checker.CreateMaxScoreResult(CheckWebHooks, "all webhooks have secrets defined")
+	}
+
+	if len(hooks) == 1 && hasSecretToCount == 1 {
```
laurentsimon, line 60: should this be `len(hooks) > hasSecretToCount`?

clients/mockclients/repo_client.go
```diff
@@ -20,109 +20,108 @@
 package mockrepo
 
 import (
-	reflect "reflect"
-
 	gomock "github.com/golang/mock/gomock"
 	clients "github.com/ossf/scorecard/v4/clients"
+	reflect "reflect"
```
laurentsimon, line 25: the linter is going to complain if this line is at the end, I think

docs/checks.md
```diff
@@ -588,3 +588,15 @@ possible.
 **Remediation steps**
 - Fix the vulnerabilities. The details of each vulnerability can be found on <https://osv.dev>.
 
+## Webhooks 
+
+Risk: `High`  (possible service can be accessed by third-party)
+
+This check validate if the webhook defined in the repository have a token configured to send as validation to make sure that is a trusted request.
```
laurentsimon, line 595: `validate` -> `validates`
`to send as validation...` -> `to authenticate the origin of the request`

docs/checks.md
```diff
@@ -588,3 +588,15 @@ possible.
...
+ 
+
+**Remediation steps**
```
laurentsimon, line 598: @olivekl can you advise on the text?

laurentsimon (approved):

checks/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+const (
+	// CheckWebHooks is the registered name for WebHooks.
+	CheckWebHooks = "Webhooks"
+)
+
+//nolint:gochecknoinits
+func init() {
+	if err := registerCheck(CheckWebHooks, WebHooks, nil); err != nil {
```
azeemshaikh38, line 31: Could you flag-guard the `registerCheck` call using `SCORECARD_V6` env variable? We want to make sure we test it before releasing.

laurentsimon: should we do this in this flag or in https://github.com/ossf/scorecard/blob/504f13441636dcc6273d45d9dc1e7ea1ef080ae2/cmd/root.go#L317 (since we already have the function https://github.com/ossf/scorecard/blob/504f13441636dcc6273d45d9dc1e7ea1ef080ae2/cmd/root.go#L261)?

clients/mockclients/repo_client.go
cpanato: this was generated automaticaly

checks/webhook.go
cpanato: set to 
```
	if len(hooks) == hasSecretToCount {
		return checker.CreateMinScoreResult(CheckWebHooks, "webhooks with no secrets configured detected")
	}

```

that means all webhooks does not have secret configured

checks/webhook.go
cpanato: did in the root.go, but can change if that is better outside

cpanato: is this good?
when is not set 
```

File: checks/raw/webhook.go [hunk 6]
```diff
@@ -0,0 +1,44 @@
...
+			sce.WithMessage(sce.ErrScorecardInternal, "Client.Repositories.ListWebhooks")
+	}
+
+	if len(hooksResp) < 1 {
+		return checker.WebhooksData{}, nil
+	}
+
+	hooks := []checker.WebhookData{}
+	for _, hook := range hooksResp {
+		v := checker.WebhookData{
```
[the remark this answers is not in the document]
cpanato: they don't have a new (actually they have but during the tests, all are called `web`) but they have an ID

I can add the ID in the struct and we output that if we want to

clients/mockclients/repo_client.go
cpanato: but it is fixed :)

checks/raw/webhook.go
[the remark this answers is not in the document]
cpanato: do you would like to have a special output check if the id of the webhook and if that have a secret or not?

like

Webhook ID | Has secret? |
--- | --- |
12345 | yes |
67890 | no |

checks/webhook.go
azeemshaikh38: It's probably best to flag guard this in the `webhook.go` file since without it the Scorecard cron job (in `cron` package) will start using this. You could separately disable it in `cron` too, but that's just more work for you and might be simpler if we flag guard it here.

checker/raw_result.go
```diff
@@ -61,6 +61,19 @@ type DependencyUpdateToolData struct {
...
+// for the Webhook check.
+type WebhooksData struct {
+	Webhook []WebhookData
+}
+
+// WebhookData contains the raw results
+// for webhook check.
+type WebhookData struct {
+	ID        *int64
+	HasSecret *bool
```
azeemshaikh38, line 74: Nit: should we maybe name this to something like `UsesAuthSecret` or `UsesAuthToken`? It'll be bit more generic instead of being GitHub-specific and will make it the code more readable.

checks/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+		// this should never happen
+		panic(err)
+	}
+}
+
+// WebHooks run Contributors check.
```
azeemshaikh38, line 37: Typo in comment? `Contributors` -> `Webhooks`?

checks/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+			hasSecretToCount++
+		}
+	}
+
+	if hasSecretToCount == 0 {
+		return checker.CreateMaxScoreResult(CheckWebHooks, "all webhooks have secrets defined")
+	}
+
+	if len(hooks) == hasSecretToCount {
+		return checker.CreateMinScoreResult(CheckWebHooks, "webhooks with no secrets configured detected")
```
laurentsimon, line 61: make the message more specific: `%d hooks out of %d do not have...`

checks/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("Client.Repositories.ListWebhooks: %v", err))
+		return checker.CreateRuntimeErrorResult(CheckWebHooks, e)
+	}
+
+	if len(hooks) < 1 {
+		return checker.CreateMaxScoreResult(CheckWebHooks, "no webhooks defined")
+	}
+
+	hasSecretToCount := 0
```
azeemshaikh38, line 49: Nit: this variable counts "number of webhooks with *no* secrets". Consider renaming to reflect that - something like `hasNoSecretCount`. Right now the var name and the value it holds are different.

checks/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, hook := range hooks {
+		if !hook.HasSecret {
+			hasSecretToCount++
```
laurentsimon, line 52: we need to log thru `c.Dlogger.Warn3()` here. Let's give the name of the webhook and its URL: this will help users remediate the issue

checks/webhook.go
azeemshaikh38, line 61: Update message to: "no webhooks with secrets configured detected" or "all webhooks do not have a secret configured".

checks/webhook.go
laurentsimon: re: when is not set -> I think that's fine. It's unsupported until we remove the env variable guard, so it's fine to get the panic.

azeemshaikh38 (commented):
Looks good to me overall. Thanks for the awesome PR and for adding tests! Left some comments, please take a look.

laurentsimon (commented):
see comments inline

checks/webhook.go
cpanato: i will update, thanks so much for the reviews and feeback

laurentsimon:
`UsesAuthSecret` (as suggested by @azeemshaikh38) sounds good: it disambiguate the use case. `HasSecret` may be interpreted as `a secret is stored and should be removed`

docs/checks.md
olivekl: Suggested rewording:
"Risk: `High`  (service possibly accessible to third parties)

This check determines whether the webhook defined in the repository has a token configured to authenticate the origins of requests."

Can we add something about how the check is scored?

docs/checks.md
```diff
@@ -588,3 +588,15 @@ possible.
...
 
+## Webhooks 
+
+Risk: `High`  (possible service can be accessed by third-party)
+
+This check validates if the webhook defined in the repository have a token configured to authenticate the origin of the request to make sure that is a trusted request.
+ 
+
+**Remediation steps**
+- Check if your service supports the token authentication.
```
olivekl, line 599: `Check whether your service supports token authentication.` [ADD: how to check this]
`- If there is support for token authentication, set the secret in the webhook configuration. `[can we link to instructions for this?]
`- If the is no support for token authentication, consider implementing it by following [these directions](https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks).`

laurentsimon (commented):
amu update you'd like me to review?

checks/raw/webhook.go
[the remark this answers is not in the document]
cpanato: only the id, at least at this stage, maybe in the future GitHub make this more customized

docs/checks.md
cpanato: thanks for the review and feedback, addressed those.
the first one is very dependant on the customer/user hook server.

checks/webhook.go
cpanato: I've implemented this @laurentsimon, let me know your thoughts

```console
$ ./scorecard --repo github.com/cpanato/testing-ci-providers --checks Webhooks
Starting [Webhooks]
|--------------------------------------------------------------------------|------------------|
|                              GITHUB WEBHOOK                              | USES AUTH SECRET |
|--------------------------------------------------------------------------|------------------|
| https://github.com/cpanato/testing-ci-providers/settings/hooks/289347313 | true             |
| https://github.com/cpanato/testing-ci-providers/settings/hooks/345108191 | false            |
|--------------------------------------------------------------------------|------------------|

Finished [Webhooks]

RESULTS
-------
Aggregate score: 5.0 / 10

Check scores:
|--------|----------|--------------------------------|---------------------------------------------------------------------------------------------------------|
| SCORE  |   NAME   |             REASON             |                                        DOCUMENTATION/REMEDIATION                                        |
|--------|----------|--------------------------------|---------------------------------------------------------------------------------------------------------|
| 5 / 10 | Webhooks | webhooks with no secrets       | https://github.com/ossf/scorecard/blob/477737b08abd2c33ca3ff76f50d7db84641db2dd/docs/checks.md#webhooks |
|        |          | configured detected -- score   |                                                                                                         |
|        |          | normalized to 5                |                                                                                                         |
|--------|----------|--------------------------------|---------------------------------------------------------------------------------------------------------|
```

checks/raw/webhook.go
```diff
@@ -0,0 +1,45 @@
...
+
+package raw
+
+import (
+	"github.com/ossf/scorecard/v4/checker"
+	sce "github.com/ossf/scorecard/v4/errors"
+)
+
+// WebHook retrieves the raw data for the WebHooks check.
+func WebHook(c *checker.CheckRequest) (checker.WebhooksData, error) {
```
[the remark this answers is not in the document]
cpanato: thanks!

checks/webhook.go
```diff
@@ -0,0 +1,96 @@
...
+	sce "github.com/ossf/scorecard/v4/errors"
+)
+
+const (
+	// CheckWebHooks is the registered name for WebHooks.
+	CheckWebHooks = "Webhooks"
+)
+
+//nolint:gochecknoinits
+func init() {
```
laurentsimon, line 35: overall PR is almost read to merge.

There's one thing you need to update (it may be confusing). The file `checks/webhook.go` need not contain much logic. 

The way the implementation works is the following: 
1. `checks/webhook.go` should call `checks/raw/webhook.go` to collect the data. You already have this file implemented, so I don't expect further changes there
2. `checks/webhook.go`  verifies data was retrieved properly, then calls `checks/evaluation/webhook.go`. See binary-artifact code for example https://github.com/ossf/scorecard/blob/main/checks/binary_artifact.go.
3. `checks/evaluation/webhook.go` reads the results and uses its logic to decide what should be logged, using `Warn`, `Info`, etc. The "evaluation" is essentially where decision takes place regarding what to do with the results retrieved by `checks/evaluation/webhook.go`. The logic you have in this file should basically be moved to `checks/evaluation/webhook.go` instead. (see https://github.com/ossf/scorecard/blob/main/checks/evaluation/binary_artifact.go as example)
4. 
Does this clarify things?

checks/webhook.go
```diff
@@ -0,0 +1,96 @@
...
+			c.Dlogger.Warn(&checker.LogMessage{
+				Path: fmt.Sprintf("https://%s/settings/hooks/%d", c.RepoClient.URI(), hook.ID),
+				Type: checker.FileTypeURL,
+				Text: "Webhook with no secret configured",
+			})
+			hasNoSecretCount++
+		}
+	}
+
+	generateWebhookTable(c.RepoClient.URI(), rawData.Webhook)
```
laurentsimon, line 66: you don't need this. Once the results are logged using `c.Dlogger.Warn` in the file `checks/evaluation/webhook.go`, the results will automatically be available in the scorecard reults: both the table, json, and sarif format.

Let me know if you have further questions.

laurentsimon (commented):
LG. Just need to move some code to `checks/evaluation/webhook.go` and we'll be ready to merge. 
Thanks for the PR, this is great!

azeemshaikh38 (changes requested):
Since we haven't reached a consensus on #1655 yet, I'm against merging this in without the support of other maintainers. If this can wait, how about we bring this up in the Scorecard bi-weekly and can merge it after that?

laurentsimon:
added an item for next meeting. Also sent a friendly ping the the tracking issue

checks/webhook.go
cpanato: thanks!

checks/webhook.go
cpanato: thanks for the feedback
made the changes PTAL

docs/checks.md
laurentsimon: Let's add the the full score is given when all webhooks have a secret used for authentication. Otherwise, the score is calculated as a (the number of webhooks with secrets / total number of webhooks configured) * 10 .

docs/checks.md
```diff
@@ -601,3 +601,13 @@ possible.
 **Remediation steps**
 - Fix the vulnerabilities. The details of each vulnerability can be found on <https://osv.dev>.
 
+## Webhooks 
+
+Risk: `High` (service possibly accessible to third parties)
```
laurentsimon, line 606: I think we could even set it to critical, because the endpoint can be reached by arbitrary users on the Internet

laurentsimon (approved):

laurentsimon:
@azeemshaikh38 can you take a second look?

docs/checks.md
cpanato: done

cpanato:
> just a small CI error `checks/evaluation/webhooks.go:58`, the line is too long and you need to split it :-) Can you do that? Also, update your branch and add a gating env variable before `register()`'ing the check.

please let me know if that is ok for you, thanks again for your review

laurentsimon:
all looks good. @azeemshaikh38 will take a final look and we'll merge. Thanks again for this great PR; I know it's been a lot of work!

checks/webhook.go
azeemshaikh38: I don't see the flag-guard anymore, maybe this got removed due to recent refactoring? We need something like:

```
// This is psuedo-code
if env("SCORECARD_V6") == 1 {
  registerCheck(CheckWebhook);
}
```

azeemshaikh38 (approved):
LGTM, except this check isn't flag-guarded anymore. Please do add that before submitting.

checks/webhook.go
cpanato: @azeemshaikh38 i did here https://github.com/ossf/scorecard/pull/1675/files#diff-a52e58522f73e84652650c20022c6e11f2935c7add0723d8d7322df9bb21e59cR42

I tried on the registercheck but that failed to compile, because we complain about 

that is why I move to another point. but if that is not correct let me know

```
panic: check present in checks.yaml is not part of `checks.AllChecks`: Webhooks

goroutine 1 [running]:
main.main()
        /Users/cpanato/code/src/github.com/ossf/scorecard/docs/checks/internal/validate/main.go:61 +0x268
exit status 2
```

azeemshaikh38: Makes sense, thanks!

cpanato:
i will rebase

cpanato:
all green now :)

PR #1703, opened 2022-03-03
✨ cmd: Allow new scorecard to be instantiated with options

justaugustus:

One part feature, one part bug fix?

- cmd: Allow new scorecard commands to be instantiated with options
- options: Default flags to struct field values
- options: Use constants for flag names
- options: Simplify SARIF check

With https://github.com/ossf/scorecard/pull/1696, we enabled the ability to import the scorecard cobra command to support use cases like https://github.com/ossf/scorecard-action/pull/122.

Unfortunately, because of the way we're currently handling defaulting for the flags, it means any initialization of the scorecard options gets discarded in favor of the flag defaults.

Where this comes up is when we attempt to build an options set via environment variables.

- [x] (N/A) Tests for the changes have been added (for bug fixes/features)

First off, we allow the scorecard command to be initialized with options, which means a downstream consumer can now choose the way they build the options (ideally with `o := options.New()` + additional logic).

After that, we change how the flags default.
Instead of defaulting to empty values, we now use whatever value was initially set within `Options`.

BEFORE:

```go
	cmd.Flags().StringVar(
		&o.Repo,
		FlagRepo,
		"repo",
		"repository to check",
	)
```

AFTER:

```go
	cmd.Flags().StringVar(
		&o.Repo,
		FlagRepo,
		o.Repo,
		"repository to check",
	)
```

h/t to @n3wscott for rubber-ducking with me!

the `release-note`

```release-note
- cmd: Allow new scorecard commands to be instantiated with options
- options: Default flags to struct field values
- options: Use constants for flag names
- options: Simplify SARIF check
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -35,8 +35,6 @@ import (
 	"github.com/ossf/scorecard/v4/policy"
 )
 
-var opts = options.New()
-
 const (
 	scorecardLong = "A program that shows security scorecard for an open source software."
 	scorecardUse  = `./scorecard [--repo=<repo_url>] [--local=folder] [--checks=check1,...]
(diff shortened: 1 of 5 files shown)


naveensrinivasan (approved):
Thank you!

PR #1708, opened 2022-03-04
Draft: 🐛 Only download repo tarball when necessary

### LINKED ISSUE #1697: Remove download of unnecessary tarball
**Describe the bug**
Scorecard downloads the `google/oss-fuzz` repo even though it's unnecessary.

**Reproduction steps**
1. Run a single check against a small repo, e.g. `scorecard --repo=chrismcgehee/test-scorecard --checks=Pinned-Dependencies`
2. Notice that it takes a few seconds (or more depending on your internet speed) before it displays `Starting [Pinned-Dependencies]`.

**Expected behavior**
Scorecard can start the actual checks immediately after invoking the command.

**Additional context**
Before running any checks, scorecard [initializes](https://github.com/ossf/scorecard/blob/738b246fe97dfc555bb9a082434d85dee4011503/checker/client.go#L61) the `google/oss-fuzz` repo for later use in the Fuzzing check. One of the steps of the initialization is to [download and extract](https://github.com/ossf/scorecard/blob/738b246fe97dfc555bb9a082434d85dee4011503/clients/githubrepo/client.go#L74) the tarball for that repo. The Fuzzing check doesn't even utilize this tarball; it makes an [api call](https://github.com/ossf/scorecard/blob/738b246fe97dfc555bb9a082434d85dee4011503/checks/fuzzing.go#L61) to check if the current repo is one of the projects fuzzed by OSS-Fuzz.

On my laptop with a 40 Mbps connection, downloading and extracting the OSS-Fuzz tarball adds 20 seconds to the runtime of scorecard. When Scorecard is run as part of cron, I don't think this penalty is paid for every repo that it evaluates; the worker will [initialize OSS-Fuzz](https://github.com/ossf/scorecard/blob/738b246fe97dfc555bb9a082434d85dee4011503/cron/worker/main.go#L191) once and then evaluate many projects after that. I think it's still worth making this performance improvement for other users who run the Scorecard CLI tool.

I'll assign this to myself, and I'd like to double-check my plan of implementation with one or two of you. My plan is to make a new interface called something like `RepoSearcher` which will have a single method `Search(request SearchRequest) (SearchResponse, error)`. Then in `CheckRequest`, instead of the `OssFuzzRepo` field being a `clients.RepoClient`, it will be a `RepoSearcher`. When initializing the `RepoSearcher`, we will not need to download the tarball for the repo.

chrismcgehee:
Draft because my changes might be causing the OSS-Fuzz E2E to fail.

Bug fix

Fixes #1697

Yes, scorecard will run faster.

the `release-note`

```release-note
NONE
```

### DIFF
--- a/clients/githubrepo/client.go
+++ b/clients/githubrepo/client.go
@@ -71,9 +71,7 @@ func (client *Client) InitRepo(inputRepo clients.Repo, commitSHA string) error {
 	}
 
 	// Init tarballHandler.
-	if err := client.tarball.init(client.ctx, client.repo, commitSHA); err != nil {
-		return fmt.Errorf("error during tarballHandler.init: %w", err)
-	}
+	client.tarball.init(client.ctx, client.repo, commitSHA)
 
 	// Setup GraphQL.
 	client.graphClient.init(client.ctx, client.repourl)
--- a/clients/githubrepo/tarball.go
+++ b/clients/githubrepo/tarball.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/google/go-github/v38/github"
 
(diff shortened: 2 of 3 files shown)


naveensrinivasan (approved):
Thanks, @chrismcgehee! I will wait for @azeemshaikh38 👀.

azeemshaikh38 (approved):
Thanks @chrismcgehee!

PR #1720, opened 2022-03-09
:seedling: Unit tests for packagemanagers

### LINKED ISSUE #1568: BUG: Add unit tests for `cmd/package_managers.go`

naveensrinivasan:
- Related to https://github.com/ossf/scorecard/issues/1568

Unit tests

Fixes #1568 

the `release-note`

```release-note
NONE
```


clients/repo_client.go
```diff
@@ -42,3 +45,8 @@ type RepoClient interface {
 	Search(request SearchRequest) (SearchResponse, error)
 	Close() error
 }
+
+// PackageManagerClient interface is used by Scorecard checks to access a package manager.
+type PackageManagerClient interface {
+	Get(URI string, packagename string) (*http.Response, error)
+}
```
azeemshaikh38, line 52: Since this interface is only used in the `cmd` package, let's define this as an *unexposed* interface there?

naveensrinivasan: The reason I decided to have the interface in the `clients` package is to be consistent with other `clients`. If we don't expose it can't be consumed as a parameter in the `cmd` package.

naveensrinivasan: Agreed, moved it to the `cmd` package and changed the scope of the interface to be not public. PTAL.

Makefile
```diff
@@ -123,7 +123,9 @@ clients/mockclients/cii_client.go: clients/cii_client.go
 checks/mockclients/vulnerabilities.go: clients/vulnerabilities.go
 	# Generating MockCIIClient
 	$(MOCKGEN) -source=clients/vulnerabilities.go -destination=clients/mockclients/vulnerabilities.go -package=mockrepo -copyright_file=clients/mockclients/license.txt
-
+checks/mockclients/packagemanager.go: cmd/packagemanager_client.go
```
azeemshaikh38, line 126: How would this work? Since the interface is not exposed shouldn't the mock also live in the same package as the interface?

azeemshaikh38 (approved):
LGTM. Left a comment to move the mock defn to the same package where the interface lives.  Keeping the mock in a package from where the interface is not visible is confusing.

PS: disabled auto-merge.

Makefile
naveensrinivasan: Sounds good. I can make that change and then merge it in. Thanks

cmd/root.go
```diff
@@ -71,10 +73,23 @@ func New(o *options.Options) *cobra.Command {
 	return cmd
 }
 
+type packageManager struct{}
+
+// nolint: noctx
+func (c *packageManager) Get(url, packageName string) (*http.Response, error) {
```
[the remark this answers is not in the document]
naveensrinivasan: I agree, I will move it.

cmd/package_managers.go
```diff
@@ -29,24 +27,24 @@ type packageMangerResponse struct {
 	exists         bool
 }
 
-// TODO(#1568): Add unit tests for this.
-func fetchGitRepositoryFromPackageManagers(npm, pypi, rubygems string) (packageMangerResponse, error) {
+func fetchGitRepositoryFromPackageManagers(npm, pypi, rubygems string,
```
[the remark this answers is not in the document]
naveensrinivasan: Yes. Thanks

cmd/root.go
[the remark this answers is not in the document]
naveensrinivasan: Moved it. Thanks

PR #1749, opened 2022-03-18
:seedling: Auto merge dependabot for minor release

naveensrinivasan:

- Auto merge dependabot updates with minor release
- Dependabot will wait until all your status checks pass before merging.
This is a function of Dependabot itself, and not this Action.
- https://github.com/marketplace/actions/dependabot-auto-merge

NONE

the `release-note`

```release-note

Auto merge dependabot PR's

```

### DIFF
--- a/.github/workflows/dependabotautomerge.yml
+++ b/.github/workflows/dependabotautomerge.yml
@@ -1,0 +1,30 @@
+# Copyright 2021 Security Scorecard Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: auto-merge
+
+on:
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@629c2de402a417ea7690ca6ce3f33229e27606a5 # v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@5d98f10d910782d0bb431fc03ae26532543ae184 # v2
+        with:
+          target: minor
+          github-token: ${{ secrets.mytoken }}
--- a/.github/auto-merge.yml
+++ b/.github/auto-merge.yml
@@ -1,0 +1,17 @@
+# Copyright 2021 Security Scorecard Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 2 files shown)


.github/workflows/dependabotautomerge.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@629c2de402a417ea7690ca6ce3f33229e27606a5 # v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@5d98f10d910782d0bb431fc03ae26532543ae184 # v2
```
laurentsimon, line 27: can we have this specific action a special case where dependabot PRs are not automatically merged? Since it has high privileges, we should be give en the change to review those, otherwise this action could merge anything they want

.github/workflows/dependabotautomerge.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          target: minor
+          github-token: ${{ secrets.mytoken }}
```
laurentsimon, line 30: does this work with branch protection enabled?

.github/workflows/dependabotautomerge.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
```
laurentsimon, line 25: Can we add an additional step here that validate PRs are from dependabot? This would reduce the attack surface.

.github/workflows/dependabotautomerge.yml
laurentsimon, line 27: I'm curious what the 3P action actually does, and whether we could easily emulate it via an API call ourselves instead?

.github/workflows/dependabotautomerge.yml
```diff
@@ -0,0 +1,30 @@
...
+name: auto-merge
+
+on:
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
```
laurentsimon, line 24: any idea whether dependabot supports "batched" PRs? Instead of having a single PR per version update, we would have one per week or so, which seems reasonable for minor/patch updates.
The downside is that if an update breaks builds, we'll have to go thru each update to figure out which is the culprit.

laurentsimon: found https://github.com/dependabot/dependabot-core/issues/2233

.github/workflows/dependabotautomerge.yml
laurentsimon: actually I'm confused what this action does, because the description of the PR says `Dependabot will wait until all your status checks pass before merging. This is a function of Dependabot itself, and not this Action.`

.github/workflows/dependabotautomerge.yml
laurentsimon: the action seems to be doing that in https://github.com/ahmadnassri/action-dependabot-auto-merge/blob/master/action/index.js#L21
What I don't understand is how the `login` is set: it does not seem to correspond to an actual GitHub user (`github.com/dependabot[bot]` or even `github.com/dependabot-preview` do not exist). So there seems to be a way to set it. Is this a feature only accessible to GitHub or to arbitrary users?

Maybe I understand this all wrong, please feel free to correct me

PR #1758, opened 2022-03-21
📖 Add schema for the raw JSON

laurentsimon:
Add JSON schema for cron and scorecard's raw results
no breaking changes
```release-notes
Add JSON schema for cron and scorecard's raw results
```

PR #1760, opened 2022-03-21
🐛 Fix schemas from https://github.com/ossf/scorecard/pull/1758

laurentsimon:
Inadvertently forgot to include certain fields in the JSON schema for raw results.
Noo rbaking changes
```release-notes
Update JSON schema for raw results
```


azeemshaikh38 (approved):
Not needed for this PR, but is there a way to automate the maintenance of this schema file? Can this be auto-generated based on the Golang struct that is used to generate this?

PR #1761, opened 2022-03-21
✨ Make Vuln ID field lower case in raw results

laurentsimon:
Make ID field lower case
No breaking changes
```release-notes
Make ID field lower case in Vul for raw results
```


laurentsimon:
another protoc error.

laurentsimon:
cc @azeemshaikh38 this keeps failing with protoc error

PR #1790, opened 2022-03-29
✨ Raw results for license

laurentsimon:
Raw results for license
No breaking changes
```release-notes
Raw results for license
```


checks/evaluation/license.go
```diff
@@ -0,0 +1,50 @@
...
+	if r == nil {
+		e := sce.WithMessage(sce.ErrScorecardInternal, "empty raw data")
+		return checker.CreateRuntimeErrorResult(name, e)
+	}
+	// We expect a single license.
+	if len(r.Files) > 1 {
+		e := sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("invalid number of results: %d",
+			len(r.Files)))
+		return checker.CreateRuntimeErrorResult(name, e)
+	}
```
azeemshaikh38, line 37: Do we really care that strongly if some repos have >1 license files? Why not just log all of them?

azeemshaikh38: Basically, is it possible that some repos have different license requirements for different parts of the code (per-directory basis let's say). So instead of returning a runtime error maybe we should handle this more gracefully?

laurentsimon: the code never returns more than one license file even where there are several licenses.  Code in raw folder uses `fileparser.OnAllFilesDo(c.RepoClient, isLicenseFile, &path)` where path is a string, not an array. So the case where > 1 license in this file should never happen.

Let me know if I mis-understood something

azeemshaikh38: I see. So:

1) It should still not be creating a runtime error because we are baking in the assumption about how `raw` package is implemented into the `evaluation` package. The `evaluation` package works on `rawData` and should only handle that.
2) Maybe we should consider expanding the `rawData` logic to return >1 license files if there is possibility of this happening? Not needed in this PR, can consider for future change.

laurentsimon: > I see. So:
> 
> 1. It should still not be creating a runtime error because we are baking in the assumption about how `raw` package is implemented into the `evaluation` package. The `evaluation` package works on `rawData` and should only handle that.

What would you like to return? We do similar sanity checking for security policy, for example. I think what you're saying is that we should get rid of this?

laurentsimon:
Actually we don't check for a single file in security policy. I'll make the changes anyway because I think it's better. Thanks for suggesting!

laurentsimon:
error during compilation:
```
#11 57.12 go mod download: github.com/docker/cli@v20.10.12+incompatible: stream error: stream ID 5819; INTERNAL_ERROR
#11 ERROR: executor failed running [/bin/sh -c go mod download]: exit code: 1

```

Issue #897, opened 2021-08-24
Feature: add other fuzzing platforms to the Fuzzing check


laurentsimon:
see title

oliverchang:
Another one to add would be clusterfuzzlite which is launching soon: https://github.com/google/clusterfuzzlite.

naveensrinivasan:
On a different note should we add this check to scorecard?

oliverchang:
> Thanks Oliver. To detect clusterfuzzlite, we need to check for a workflow which uses `google/clusterfuzzlite/actions/run_fuzzers` action, is this correct? Do you expect users to enable it on PRs or push to main or something else?

Yep! Just check for the `google/clusterfuzzlite/actions/run_fuzzers` action. I think we can be flexible about where it's enabled (PR or push or otherwise) so we don't need to check that. 

> What about results of this check from this Fuzz?
> On a different note should we add this check to scorecard?

Could you clarify what you mean here? Are you suggesting a more detailed check beyond just checking if Fuzzing is enabled?

naveensrinivasan:
> > Thanks Oliver. To detect clusterfuzzlite, we need to check for a workflow which uses `google/clusterfuzzlite/actions/run_fuzzers` action, is this correct? Do you expect users to enable it on PRs or push to main or something else?
> 
> Yep! Just check for the `google/clusterfuzzlite/actions/run_fuzzers` action. I think we can be flexible about where it's enabled (PR or push or otherwise) so we don't need to check that.
> 
> > What about results of this check from this Fuzz?
> > On a different note should we add this check to scorecard?
> 
> @naveensrinivasan Could you clarify what you mean here? Are you suggesting a more detailed check beyond just checking if Fuzzing is enabled?

Yes , how do we report the results of these fuzz runs?

laurentsimon:
> > Thanks Oliver. To detect clusterfuzzlite, we need to check for a workflow which uses `google/clusterfuzzlite/actions/run_fuzzers` action, is this correct? Do you expect users to enable it on PRs or push to main or something else?
> 
> Yep! Just check for the `google/clusterfuzzlite/actions/run_fuzzers` action. I think we can be flexible about where it's enabled (PR or push or otherwise) so we don't need to check that.
> 
 great! Do you have cycles to add it before you officially release clusterfuzzlite? Or shall we add this to v4 milestone (~EOY)?

laurentsimon:
[...]

PR #1816, opened 2022-04-04
✨ Return inconclusive result if no fuzzing is detected

laurentsimon:
Return inconclusive result if no fuzzing is detected
No breaking changes
```release-notes
Return inconclusive result if no fuzzing is detected
```

### DIFF
--- a/checks/fuzzing.go
+++ b/checks/fuzzing.go
@@ -86,5 +86,5 @@ func Fuzzing(c *checker.CheckRequest) checker.CheckResult {
 			"project is fuzzed in OSS-Fuzz")
 	}
 
-	return checker.CreateMinScoreResult(CheckFuzzing, "project is not fuzzed")
+	return checker.CreateInconclusiveResult(CheckFuzzing, "could not determine if project is fuzzed")
 }
--- a/docs/checks/internal/checks.yaml
+++ b/docs/checks/internal/checks.yaml
@@ -374,9 +374,9 @@ checks:
       Risk: `Medium` (possible vulnerabilities in code)
 
       This check tries to determine if the project uses
-      [fuzzing](https://owasp.org/www-community/Fuzzing) by checking if the repository
-      name is included in the [OSS-Fuzz](https://github.com/google/oss-fuzz) project
-      list.
+      [fuzzing](https://owasp.org/www-community/Fuzzing). The current implementation
+      check if the repository name is included in the [OSS-Fuzz](https://github.com/google/oss-fuzz) project
+      list or if [ClusterFuzzLite](https://google.github.io/clusterfuzzlite/) is enabled.
 
       Fuzzing, or fuzz testing, is the practice of feeding unexpected or random data
       into a program to expose bugs. Regular fuzzing is important to detect


checks/fuzzing.go
```diff
@@ -86,5 +86,5 @@ func Fuzzing(c *checker.CheckRequest) checker.CheckResult {
 			"project is fuzzed in OSS-Fuzz")
 	}
 
-	return checker.CreateMinScoreResult(CheckFuzzing, "project is not fuzzed")
+	return checker.CreateInconclusiveResult(CheckFuzzing, "could not determine if project is fuzzed")
```
azeemshaikh38, line 89: Let's make the message explicit - `project not fuzzed through any platforms recognizable by Scorecard: oss-fuzz, CFLite. File a support request [here](https://github.com/ossf/scorecard/issues/897) if you would like support for other fuzzing platforms.`. Can even keep this as a failing case instead of inconclusive.

azeemshaikh38 (approved):

checks/fuzzing.go
laurentsimon: If we want to have a generic message `File a support request`, I think we should do this in a another PR and for all checks. The details has the details on which platforms are not found

laurentsimon: mhh, the details don't have this but it should. Ill update this PR to fix tat too

evverx:
FWIW I think the fuzzing check is kind of broken at this point because projects integrated into OSS-Fuzz along with CIFuzz/CFLite and actually maintained receive lower scores due to the "unpinned" dependencies and bogus "vulnerabilities" that end up in the OSV database triggering the "vulnerabilities" check. `scorecard` on the other hand has been failing to even build on OSS-Fuzz for more than a month and `scorecard` still thinks that it's fuzzed.

laurentsimon:
> FWIW I think the fuzzing check is kind of broken at this point because projects integrated into OSS-Fuzz along with CIFuzz/CFLite and actually maintained receive lower scores due to the "unpinned" dependencies and bogus "vulnerabilities" that end up in the OSV database triggering the "vulnerabilities" check. 

Can you explain what the "bogus" OSV entries are?

/cc @oliverchang 

`scorecard` on the other hand has been failing to even build on OSS-Fuzz for more than a month and `scorecard` still thinks that it's fuzzed.

Can you elaborate on `been failing to even build on OSS-Fuzz`?

evverx:
> Can you explain what the "bogus" OSV entries are?

I think the latest examples would be https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2022q1/016161.html and https://github.com/google/oss-fuzz/issues/7434

> Can you elaborate on been failing to even build on OSS-Fuzz?

According to https://oss-fuzz-build-logs.storage.googleapis.com/index.html#ossf-scorecard `scorecard` has been failing to compile on OSS-Fuzz with
```
Step #3 - "compile-libfuzzer-address-x86_64": github.com/ossf/scorecard/v4/policy
Step #3 - "compile-libfuzzer-address-x86_64": github.com/ossf/scorecard/v4/policy
Step #3 - "compile-libfuzzer-address-x86_64": Running go-fuzz -tags gofuzz -func FuzzParseFromYAML -o fuzz_parse_from_yaml.a github.com/ossf/scorecard/v4/policy
Step #3 - "compile-libfuzzer-address-x86_64": # github.com/ossf/scorecard/v4/policy
Step #3 - "compile-libfuzzer-address-x86_64": policy/yaml_fuzzer.go:19:9: undefined: ParseFromYAML
Step #3 - "compile-libfuzzer-address-x86_64": 2022/04/18 11:52:44 failed to build packages:exit status 2
Step #3 - "compile-libfuzzer-address-x86_64": ********************************************************************************
Step #3 - "compile-libfuzzer-address-x86_64": Failed to build.
Step #3 - "compile-libfuzzer-address-x86_64": To reproduce, run:
Step #3 - "compile-libfuzzer-address-x86_64": python infra/helper.py build_image ossf-scorecard
Step #3 - "compile-libfuzzer-address-x86_64": python infra/helper.py build_fuzzers --sanitizer address --engine libfuzzer --architecture x86_64 ossf-scorecard
Step #3 - "compile-libfuzzer-address-x86_64": ********************************************************************************
Finished Step #3 - "compile-libfuzzer-address-x86_64"
```
for almost two months.

PR #1819, opened 2022-04-05
:seedling: Pins the golang version in the workflow

naveensrinivasan:
Hopefully this fixes the make linter failures

https://github.com/ossf/scorecard/runs/5834278035?check_suite_focus=true

I noticed while trying to debug , which was using go 1.18 in the
workflow log.

Which made me decide to pin it to specific version of go 1.17.7
```
go env -w GOFLAGS=-mod=mod
  make check-linter
  shell: /usr/bin/bash -e {0}
  env:
    PROTOC_VERSION: 3.17.3
    GOROOT: /opt/hostedtoolcache/go/1.18.0/x64
```

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,6 +29,9 @@ on:
   paths-ignore:
   - "*.md"
 
+env:
+  GO_VERSION: 1.17.7
+
 jobs:
   scorecard:
     name: scorecard-docker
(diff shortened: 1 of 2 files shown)


naveensrinivasan:
I think this should fix the problem. This is using version `1.17.7` instead of `1.18`

```
go env -w GOFLAGS=-mod=mod
  make check-linter
  shell: /usr/bin/bash -e {0}
  env:
    PROTOC_VERSION: [3](https://github.com/ossf/scorecard/runs/5839847436?check_suite_focus=true#step:8:3).17.3
    GO_VERSION: 1.17.7
    GOROOT: /opt/hostedtoolcache/go/1.17.7/x6[4](https://github.com/ossf/scorecard/runs/5839847436?check_suite_focus=true#step:8:4)
 ```

azeemshaikh38 (approved):

PR #1822, opened 2022-04-06
✨ Support `Security-Policy` with `--local`

### LINKED ISSUE #1752: BUG: Security-Policy throws internal error if `--local .` passed
**Describe the bug**
`$ scorecard --local .` throws `input repo should be of type repoURL` error.

**Reproduction steps**
1. `$ git clone github.com/chainguard-dev/apko`
2. `$ scorecard --local .`
3. See: `internal error: input repo should be of type repoURL: <nil>`

**Expected behavior**
We should not print out `Security-Policy` section in case we don't pass `--repo` flag?

**Additional context**
Version: `4.1.0`

azeemshaikh38:

Add localclient support for `Security-Policy` check.

Fixes #1752

the `release-note`

```release-note
Support `Security-Policy` with `--local`
```

### DIFF
--- a/checks/raw/security_policy.go
+++ b/checks/raw/security_policy.go
@@ -29,8 +29,6 @@ import (
 
 // SecurityPolicy checks for presence of security policy.
 func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error) {
-	// TODO: not supported for local clients.
-
 	files := make([]checker.File, 0)
 	err := fileparser.OnAllFilesDo(c.RepoClient, isSecurityPolicyFile, &files)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


laurentsimon (approved):
LGTM, does this check support local repos? It's downloading the org repo if it exists, so it would return an error when security file is missing with the `--local` option, no?

azeemshaikh38:
> LGTM, does this check support local repos? It's downloading the org repo if it exists, so it would return an error when security file is missing with the `--local` option, no?

That was the behavior before this PR. With this PR, it only looks at the org repo if the `Repo.Org() != nil` and returns a min score otherwise.

laurentsimon:
I think the problem was introduced in https://github.com/ossf/scorecard/pull/1822. This PR declared this check as File-based, even though the checks.yml says it's not https://github.com/ossf/scorecard/blob/main/docs/checks/internal/checks.yaml#L539

azeemshaikh38:
> I think the problem was introduced in #1822. This PR declared this check as File-based, even though the checks.yml says it's not https://github.com/ossf/scorecard/blob/main/docs/checks/internal/checks.yaml#L539
> 
> I think we need to make this check _not_ file/commit type.

Discussed offline. `--local` will not run `Security-Policy` check. We'll only use this check for GitHub-like VCS platforms.

PR #1849, opened 2022-04-19
✨ Raw results for dangerous workflow

laurentsimon:
Raw results for dangerous workflow
no breaking changes

Note: need to finalize the unit tests

```release-notes
Raw results for dangerous workflow
```


pkg/json_raw_results.go
```diff
@@ -137,8 +138,28 @@ type jsonLicense struct {
 	// TODO: add fields, like type of license, etc.
 }
 
+type jsonWorkflows struct {
+	UntrustedCheckouts   []jsonUntrustedCheckout `json:"untrustedCheckouts"`
+	ScriptInjections     []jsonScriptInjection   `json:"scriptInjections"`
+	SecretInPullRequests []jsonEncryptedSecret   `json:"secretInPullRequests"`
+}
+
+type jsonUntrustedCheckout struct {
```
laurentsimon, line 147: shall I simplify here and make each of these `jsonUntrustedCheckout`, et a `jsonFile`?

laurentsimon:
friendly ping

checks/raw/dangerous_workflow.go
```diff
@@ -0,0 +1,506 @@
...
+	return data, err
+}
+
+// Check file content.
+var validateGitHubActionWorkflowPatterns fileparser.DoWhileTrueOnFileContent = func(path string,
+	content []byte,
+	args ...interface{},
+) (bool, error) {
+	if !fileparser.IsWorkflowFile(path) {
+		fmt.Println("here", path)
```
azeemshaikh38, line 86: Remove?

checker/raw_result.go
```diff
@@ -256,3 +257,27 @@ type ReleaseAsset struct {
...
+
+// ScriptInjection represents a script injection.
+type ScriptInjection struct {
+	File File
+}
+
+// EncryptedSecret represents an encrypted secret.
+type EncryptedSecret struct {
+	File File
+}
```
azeemshaikh38, line 283: Could we add more details to these structs like - the secret we think is at risk of leaking, job name, script injection value etc.?

checks/evaluation/dangerous_workflow.go
```diff
@@ -0,0 +1,90 @@
...
+		})
+
+		score -= checker.MaxResultScore
+	}
+
+	if score < checker.MinResultScore {
+		score = checker.MinResultScore
+	}
+
+	return createResult(name, score)
```
azeemshaikh38, line 78: Score calculation can be simplified as:

```
if len(r.ScriptInjections) > 0 ||
   len(r.UntrustedCheckout) > 0 ||
   ... {
  return checker.MinResultScore
}
return checker.MaxResultScore
```

azeemshaikh38 (approved):

checks/raw/dangerous_workflow.go
laurentsimon: good catch, yes!

checker/raw_result.go
laurentsimon: ah yes that slipped thru the crack. The value is already present in the `File.Snippet` (can it here instead), but job names, etc should be added. I'll do that, thanks

laurentsimon: Done:
```
"untrustedCheckouts": [
        {
          "file": {
            "snippet": "${{ github.event.pull_request.head.sha }}",
            "path": ".github/workflows/integration.yml",
            "offset": 45
          },
          "job": {
            "name": null,
            "id": "integration-trusted"
          }
        }
```

checks/evaluation/dangerous_workflow.go
laurentsimon: Done.

laurentsimon:
temporarily disabling `build-proto` pre-submit to merge, because it's failing on protbuf

PR #1865, opened 2022-04-26
✨ Update raw format for Dangerous workflows

laurentsimon:
Using a `type` for the dangerous workflow results. This will avoid updating the structure/BQ table every time we add a new type of check.

Note: need https://github.com/ossf/scorecard/pull/1864 to be merged

```release-notes
Update raw format for Dangerous workflows
```

### DIFF
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -296,14 +296,19 @@ const (
 // DangerousWorkflowData contains raw results
 // for dangerous workflow check.
 type DangerousWorkflowData struct {
-	Workflows []Workflow
+	Workflows []DangerousWorkflow
 }
 
-// Workflow represents a result for a workflow.
+// DangerousWorkflow represents a dangerous workflow.
+type DangerousWorkflow struct {
+	Workflow Workflow
+	Type     DangerousWorkflowType
+}
+
+// Workflow represents a workflow.
 type Workflow struct {
 	Job  *WorkflowJob
 	File File
-	Type DangerousWorkflowType
 }
 
 // WorkflowJob reprresents a workflow job.
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -283,24 +283,27 @@ type CIIBestPracticesData struct {
 	Badge CIIBadge
 }
 
+// DangerousWorkflowType represents a type of dangerous workflow.
+type DangerousWorkflowType int
+
+const (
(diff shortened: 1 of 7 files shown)


azeemshaikh38:
could you rebase with the latest changes?

laurentsimon:
friendly ping for review

checker/raw_result.go
```diff
@@ -283,24 +283,27 @@ type CIIBestPracticesData struct {
...
+	Workflows []Workflow
 }
 
-// ScriptInjection represents a script injection.
-type ScriptInjection struct {
+// Workflow represents a result for a workflow.
+type Workflow struct {
 	Job  *WorkflowJob
 	File File
+	Type DangerousWorkflowType
```
azeemshaikh38, line 306: Maybe rename `Type` -> `DetectedDangerousPattern` (or something similar) to make it clear what this field represents?

Also, should this be part of `WorkflowJob` since each job can have it's own unique dangerous pattern?

azeemshaikh38 (approved):

checker/raw_result.go
laurentsimon: > Maybe rename `Type` -> `DetectedDangerousPattern` (or something similar) to make it clear what this field represents?

I'll update the structure from `Workflow` to `DangerousWorkflow` instead.

> Also, should this be part of `WorkflowJob` since each job can have it's own unique dangerous pattern?

`WorkflowJob` should be generic so that we can re-use it for Token-Permisison and Pinned-Dependencies check, o I think `Type` should not be part of it.

azeemshaikh38: How do we plan to surface information about multiple dangerous patterns in a workflow?

laurentsimon: using the Type:
```
type DangerousWorkflowData struct {
	Workflows []DangerousWorkflow
}

// DangerousWorkflow represents a dangerous workflow.
type DangerousWorkflow struct {
	Workflow Workflow
	Type     DangerousWorkflowType
}

// Workflow is generic and re-usable.
type Workflow struct {
	Job  *WorkflowJob
	File File
}

```

Wdut?

azeemshaikh38: Makes sense. LGTM.

azeemshaikh38 (approved):

PR #1875, opened 2022-04-27
🐛 Propagate error in log

laurentsimon:
Propagate error in log
```release-notes
Propagate error in log
```


naveensrinivasan (approved):

laurentsimon:
interesting error: `Error: API rate limit exceeded for installation ID 12341091.`

Issue #1868, opened 2022-04-26
BUG Binary Artifacts False positive scripts/Makefile.lib


jeffmendoza:
**Describe the bug**
The file scripts/Makefile.lib from Linux Kernel forks is triggering as a binary artifact, when it is a Makefile.

**Reproduction steps**
Example repo: 
https://github.com/google/capsicum-linux/blob/capsicum/scripts/Makefile.lib

Run scorecards, scripts/Makefile.lib is reported as a binary artifact.

**Expected behavior**
Not to be reported.

laurentsimon:
Is this a standard naming convention Makefile.lib, or that's what they picked?

PR #1879, opened 2022-05-02
🐛 Fix .lib false positives in binary artifacts

laurentsimon:
Fix .lib false positives in binary artifacts
no breaking changes
```release-notes
Fix .lib false positives in binary artifacts
```
closes https://github.com/ossf/scorecard/issues/1868

### DIFF
--- a/checks/raw/binary_artifact.go
+++ b/checks/raw/binary_artifact.go
@@ -107,6 +107,7 @@ var checkBinaryFileContent fileparser.DoWhileTrueOnFileContent = func(path strin
 		})
 	}
 
+	fmt.Println(*pfiles)
 	return true, nil
 }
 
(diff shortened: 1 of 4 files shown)


naveensrinivasan (approved):
Thanks!

laurentsimon:
> Let's do the sanity test _iff_ `exists1 == false`. I would trust the magic number inference more than the character search logic we have here and we can fallback to it if the magic number inference fails.
> 
> ```
> if exists1 {
>   // Binary file
> }
> 
> if !isText(content) && exists2 {
>   // Binary file
> }
> ```

done.

checks/raw/binary_artifact.go
```diff
@@ -103,3 +113,16 @@ var checkBinaryFileContent fileparser.DoWhileTrueOnFileContent = func(path strin
 
 	return true, nil
 }
+
+// TODO: refine this function.
```
azeemshaikh38, line 117: Nit: make the TODO more actionable. If the refinement refers to using Makefile/YAML parsers, then mention that explicitly here or create an issue and link to it here.

azeemshaikh38 (approved):
Thanks!

PR #1881, opened 2022-05-04
🐛 Fix `scorecard version` not showing build info

azeemshaikh38:
Bug fix the issue where `scorecard version` was not showing any build info. We were relying on `k8s/release-utils` for this but build info is only available through the binaries that invoke them.

NONE

NONE
the `release-note`

```release-note

```


laurentsimon (approved):

laurentsimon:
@naveensrinivasan @azeemsgoogle I'm curious why we need this package. What feature does it provide that we cannot achieve with the previous code?

Issue #1850, opened 2022-04-19
Feature: better remediation steps


laurentsimon:
It's pretty hard for a user to remediate some of the results:
- permissions: which to use?
- pinned dependencies: what's the hash for each version used?
- dependabot: which options to enable in the config file + settings?

We should try to do better. Some thoughts:
1. Provide a link to the StepSecurity in the results presented in the scorecard action
2. Provide a snippet of code that a user can copy/paste into a new PR
3. Provide a link to create a new PR with the diff pre-populated
4. Provide diff in SARIF if possible, and let GitHub surface this information

Starting with option (1) may be the easiest to start with.

/cc @varunsh-coder

varunsh-coder:
> It's pretty hard for a user to remediate some of the results:
> 
> * permissions: which to use?
> * pinned dependencies: what's the hash for each version used?
> * dependabot: which options to enable in the config file + settings?
> 
> We should try to do better. Some thoughts:
> 
> 1. Provide a link to the StepSecurity in the results presented in the scorecard action
> 2. Provide a snippet of code that a user can copy/paste into a new PR
> 3. Provide a link to create a new PR with the diff pre-populated
> 4. Provide diff in SARIF if possible, and let GitHub surface this information
> 
> Starting with option (1) may be the easiest to start with.
> 
> /cc @varunsh-coder

Adding more details to the above options:
1. There is already a link to app.stepsecurity.io to fix token permissions and pin Action commits, but users need to click the link, find their workflow, copy it, paste it in etc. All these steps can be eliminated by having a link to app.stepsecurity.io/secureworkflows with repo and workflow as query string inputs. The UI can then fetch the workflow for open source projects and present the fixed workflow result, which the user can copy and paste back.
2. This option improves on the previous one by having the fixed workflow snippet be in the SARIF file, so one does not have to go to another link. Copying this diff might not be user-friendly, so this option can also be combined with #1, where user gets both the diff and a link to view the diff on app.stepsecurity.io/secureworkflows and copy from there. 
3. w.r.t creating PR we need to investigate if we can do this without need for an App that has write access to the repo. If anyone knows how other solutions do this, please share that info. 
4. w.r.t diff in SARIF, I will check with GitHub if there is an option to surface remediation information if it is included in SARIF

@zwass since you originally requested this, please share insights on what would work best for you/ ideas to improve these options.

laurentsimon:
[...]

PR #1885, opened 2022-05-04
✨ Add custom remediation for workflow permissions/pinned dependencies

laurentsimon:
Add custom remediation for workflow permissions/pinned dependencies
```release-notes
Add custom remediation for workflow permissions/pinned dependencies
```
see https://github.com/ossf/scorecard/issues/1850

Screenshot:
image

### DIFF
--- a/checks/remediations.go
+++ b/checks/remediations.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Security Scorecard Authors
+// Copyright 2022 Security Scorecard Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/checks/remediations.go
+++ b/checks/remediations.go
@@ -25,9 +25,10 @@ import (
 )
 
 var (
-	remediationBranch string
-	remediationRepo   string
-	remediationOnce   *sync.Once
+	remediationBranch   string
+	remediationRepo     string
+	remediationOnce     *sync.Once
+	remediationSetupErr error
 )
 
 //nolint:gochecknoinits
@@ -36,12 +37,11 @@ func init() {
 }
 
 func remdiationSetup(c *checker.CheckRequest) error {
-	var setupErr error
 	remediationOnce.Do(func() {
 		// Get the branch for remediation.
 		b, err := c.RepoClient.GetDefaultBranch()
 		if err != nil && !errors.Is(err, clients.ErrUnsupportedFeature) {
-			setupErr = err
+			remediationSetupErr = err
 			return
 		}
 		if b.Name != nil {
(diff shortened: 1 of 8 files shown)


laurentsimon:
Fyi, the protobuf-related checks have started failing again over the past few days

checks/remediations.go
```diff
@@ -0,0 +1,86 @@
...
+
+//nolint:gochecknoinits
+func init() {
+	remediationOnce = new(sync.Once)
+}
+
+func remdiationSetup(c *checker.CheckRequest) error {
+	remediationOnce.Do(func() {
+		// Get the branch for remediation.
+		b, err := c.RepoClient.GetDefaultBranch()
```
varunsh-coder, line 42: Should this be the default branch or the branch on which Scorecard Action was run? As an example, if one creates a PR with a new workflow, which is missing permissions, this will still have main (/default branch). But when app.stepsecurity.io tries to fix it, it will not find the workflow in that branch.

checks/remediations.go
```diff
@@ -0,0 +1,86 @@
...
+	return createWorkflowRemediation(filepath, "pin")
+}
+
+func createWorkflowRemediation(path, t string) *checker.Remediation {
+	p := strings.TrimPrefix(path, ".github/workflows/")
+	if remediationBranch == "" || remediationRepo == "" {
+		return nil
+	}
+
+	text := fmt.Sprintf("update your workflow using https://app.stepsecurity.io/secureworkflow/%s/%s/%s?enable=%s",
```
varunsh-coder, line 76: Should these literals be set as constants at top of file? Else will need to be modified at multiple places.

varunsh-coder:
@laurentsimon thanks a lot for the PR! Great to see remediation being added. I added couple of comments. Do you have a screenshot of how this looks right now? I am curious how remediation info will show up in the GitHub Code Scanning UI.

checks/remediations.go
laurentsimon: You're correct. However we only support the default branch for non pull_request events https://github.com/ossf/scorecard-action/blob/main/entrypoint.sh#L101 and we don't officially support (https://github.com/ossf/scorecard-action below the table of content).

So I think we only need to support the default branch for now. Let me know if I missed some cases, it's possible

checks/remediations.go
laurentsimon: yes I can do that. Thanks!

laurentsimon:
> @laurentsimon thanks a lot for the PR! Great to see remediation being added. I added couple of comments. Do you have a screenshot of how this looks right now? I am curious how remediation info will show up in the GitHub Code Scanning UI.

I added the screenshot i the issue PR description. PTAL

checks/remediations.go
varunsh-coder: ok, thanks for the info!

checks/remediations.go
```diff
@@ -0,0 +1,89 @@
...
+
+func createWorkflowPermissionRemediation(filepath string) *checker.Remediation {
+	return createWorkflowRemediation(filepath, "permissions")
+}
+
+func createWorkflowPinningRemediation(filepath string) *checker.Remediation {
+	return createWorkflowRemediation(filepath, "pin")
+}
+
+func createWorkflowRemediation(path, t string) *checker.Remediation {
```
azeemshaikh38, line 76: Let's pass the repo name and branch as arguments to these fns directly. The checks already take care of ensuring only supported repo types call them so it'll avoid all the extra complexity of checking it here again.

azeemshaikh38 (approved):

PR #1898, opened 2022-05-09
🐛 Remove Options that belong to the Action

laurentsimon:
See https://github.com/ossf/scorecard-action/issues/261
Remove Options that belong to the Action

```release-notes
Remove Options that belong to the Action
```


naveensrinivasan (approved):
Thanks !

PR #1912, opened 2022-05-13
✨ Add raw results for Token-Permissions

laurentsimon:
Add raw results for Token-Permissions

```release-notes
Add raw results for Token-Permissions
```

PR #1913, opened 2022-05-13
✨ Raw results for Packaging check

laurentsimon:
Raw results for Packaging check

```release-notes
Raw results for Packaging check
```


checker/raw_result.go
```diff
@@ -34,6 +35,52 @@ type RawResults struct {
 	LicenseResults              LicenseData
 }
 
+// TODO: Add Msg to all results.
+
+// OutcomeType represents the type of the raw result.
+type OutcomeType int
```
azeemshaikh38, line 41: IMO, severity should get tagged on during evaluation phase. Severity is usually associated with a policy failure and raw results should remain agnostic to any policy. Wdut?

laurentsimon: changed made. PTAL

checker/check_result.go
```diff
@@ -63,6 +63,32 @@ const (
 	FileTypeURL
 )
 
+// OutcomeType represents the outcome of a result.
+type OutcomeType int
```
azeemshaikh38, line 67: Where is this being exposed?

checker/check_result.go
```diff
@@ -63,6 +63,32 @@ const (
...
+
+const (
+	// OutcomeTypeUnknown is for unknown type. Used for debugging and catching unset types.
+	OutcomeTypeUnknown OutcomeType = iota
```
azeemshaikh38, line 71: Nit: type is already `OutcomeType`, so rename the enums to `Unknown`, `Debug` etc. to avoid redundancy.

Also, note that when this gets printed out it'll only print out an integer. Consider using a string type to make this clearer.

azeemshaikh38, line 71: Nit: type is already `OutcomeType`, so rename the enums to `Unknown`, `Debug` etc. to avoid redundancy.

Also, note that when this gets printed out it'll only print out an integer. Consider using a string type to make this clearer.

checks/permissions.go
```diff
@@ -613,3 +613,128 @@ func isReleasingWorkflow(workflow *actionlint.Workflow, fp string, dl checker.De
 
 	return fileparser.AnyJobsMatch(workflow, jobMatchers, fp, dl, "not a releasing workflow")
 }
+
+// TODO: remove when migrated to raw results.
+// Should be using the definition in raw/packaging.go
+func isPackagingWorkflow(workflow *actionlint.Workflow, fp string, dl checker.DetailLogger) bool {
```
azeemshaikh38, line 619: Nit: `isPackagingWorkflow` can be made part of `fileparser` package and be shared across both the `permissions` and `packaging` checks?

azeemshaikh38, line 619: Nit: `isPackagingWorkflow` can be made part of `fileparser` package and be shared across both the `permissions` and `packaging` checks?

checker/check_result.go
laurentsimon: not used yet. I can remove it for this PR if you want. Shall I do that?

checker/check_result.go
laurentsimon: I tend to use `OutcomeType` as a namespace, so it does not collide with other names. I can change that, no problem.
We don't print this value yet, the plan is to have a `outcomeToString()` function for it.

checks/permissions.go
laurentsimon: in the future yes. For now it's not needed because it's going to be removed in https://github.com/ossf/scorecard/pull/1912. The code here is a temporary solution until the permission check is migrated.

laurentsimon: This function will be removed as part of the permission raw PR https://github.com/ossf/scorecard/pull/1912, so I'll deal with sharing in the other PR. Thanks for thee suggestion

checker/check_result.go
laurentsimon: removed it from this PR.

checker/check_result.go
laurentsimon: removed.

laurentsimon:
PTAL. Thanks

azeemshaikh38 (approved):

checker/check_result.go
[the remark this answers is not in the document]
laurentsimon: this has been removed. You must have looked at a commit in-between the change.

PR #1917, opened 2022-05-16
✨ Raw results for Fuzzing check

laurentsimon:
Raw results for Fuzzing check
```release-notes
Raw results for Fuzzing check
```


checker/raw_result.go
```diff
@@ -31,9 +31,48 @@ type RawResults struct {
...
+
+// FuzzingData represents different fuzzing done.
+type FuzzingData struct {
+	Fuzzers []Fuzzer
+}
+
+// Fuzzer represent the use of a fuzzer.
+type Fuzzer struct {
+	Job      *WorkflowJob
+	File     *File
```
azeemshaikh38, line 58: How does this capture the case where `oss-fuzz` is being used to fuzz? Should we add a `Repository` (or similar) field to signify where we found the evidence for fuzzing?

checks/raw/fuzzing.go
```diff
@@ -0,0 +1,79 @@
...
+	if usingCFLite {
+		fuzzers = append(fuzzers, checker.Fuzzer{Name: checker.FuzzerNameCIFuzz})
+	}
+
+	usingOSSFuzz, e := checkOSSFuzz(c)
+	if e != nil {
+		return checker.FuzzingData{}, fmt.Errorf("%w", e)
+	}
+	if usingOSSFuzz {
+		fuzzers = append(fuzzers, checker.Fuzzer{Name: checker.FuzzerNameOSSFuzz})
```
azeemshaikh38, line 42: This does not seem to populate any location data?

pkg/json_raw_results.go
```diff
@@ -193,10 +221,33 @@ type jsonRawResults struct {
...
+		case checker.FuzzerNameCIFuzz:
+			fuzzer.Name = fuzzerNameCIFuzz
+		case checker.FuzzerNameOSSFuzz:
+			fuzzer.Name = fuzzerNameOSSFuzz
+		case checker.FuzzerNameGoBuiltin:
+			fuzzer.Name = fuzzerNameGoBuiltin
+		default:
+			return fmt.Errorf("%w: %d", errorInvalidFuzzer, f.Name)
+		}
+		r.Results.Fuzzers = append(r.Results.Fuzzers, fuzzer)
```
azeemshaikh38, line 246: It's better to have the type as `string` instead of converting here.

checks/raw/fuzzing.go
laurentsimon: I've not added things that we not already present in the existing codebase. This PR was intended to be a mere migration. Shall I add support in this PR?

pkg/json_raw_results.go
laurentsimon: problem with strings is that we don't get the linter to warn us when we forget a case in a switch statement

checker/raw_result.go
```diff
@@ -31,9 +31,48 @@ type RawResults struct {
...
+	FuzzingResults              FuzzingData
 	LicenseResults              LicenseData
 }
 
+// FuzzerName represents a fuzzing service.
+type FuzzerName int
+
+const (
+	// FuzzerNameCIFuzz is CIFuzz.
+	FuzzerNameCIFuzz FuzzerName = iota
```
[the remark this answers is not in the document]
azeemshaikh38: > I tend to re-use the type name to avoid name collision and to make it clearer what type it refers to in general. Shall I change that?

Upto you. Feel free to leave it as is. 

> I prefer enums because the linter will warn us if we forget to. There will be a stringify function when we need to print it, like in the json_raw_results.go file. Wdut?

Linter will still catch string enums if the switch case is not exhaustive. If you want to continue using integer however, I suggest adding a `String()` function. It'll make the enum <-> string conversion more robust (i.e will not depend on the caller).

```
func (f FuzzerName) String() string {
  switch f {
    // Linter ensures all integer values have a string name. 
  } 
}
```

checker/raw_result.go
azeemshaikh38: How about leaving out these fields then? We can use another PR to add and populate them. It's better than having empty fields with no info.

checker/raw_result.go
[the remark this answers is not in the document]
laurentsimon: if the string are caught by linters, then I'm fine with strings.

checker/raw_result.go
laurentsimon: I'll remove them and add a TODO. Thanks

pkg/json_raw_results.go
laurentsimon: I was wrong. string enums work well!

checker/raw_result.go
laurentsimon: I've removed them.

laurentsimon:
I've addressed the comments. PTAL

checker/raw_result.go
```diff
@@ -32,9 +32,45 @@ type RawResults struct {
...
+	FilesCoverage []FileCoverage
+	// TODO: aggregate coverage, etc
+}
+
+// FileCoverage represent the coverage of a file.
+type FileCoverage struct {
+	Lines []uint
+	File  File
+}
+
```
azeemshaikh38, line 73: `CodeCoverage` and `FileCoverage` not used anywhere so maybe remove it in this PR?

azeemshaikh38 (approved):

checker/raw_result.go
laurentsimon: SG

PR #1920, opened 2022-05-17
:warning: Remove the oldjson format from cron

naveensrinivasan:
- removed the old json format from cron
fix https://github.com/ossf/scorecard/issues/1393 


laurentsimon (approved):
Let's be sure to state that we've removed it in the next release

PR #1926, opened 2022-05-18
✨ Use string enum for dangerous workflow type

laurentsimon:
Use string enum for dangerous workflow type
```release-notes
Use string enum for dangerous workflow type
```


laurentsimon:
protobuf checks are failing again :/

azeemshaikh38 (approved):
Thanks!

PR #1932, opened 2022-05-19
✨ Raw results for Pinned-Dependencies

laurentsimon:
Raw results for Pinned-Dependencies
```release-notes
Raw results for Pinned-Dependencies
```


checker/raw_result.go
```diff
@@ -60,6 +61,40 @@ type Fuzzer struct {
...
+	// DependencyUseTypeDownloadThenRun is a download followed by a run.
+	DependencyUseTypeDownloadThenRun DependencyUseType = "download-then-run"
+	// DependencyUseTypeGoCommand is a go command.
+	DependencyUseTypeGoCommand DependencyUseType = "go command"
+	// DependencyUseTypeChocoCommand is a choco command.
+	DependencyUseTypeChocoCommand DependencyUseType = "choco command"
+	// DependencyUseTypeNpmCommand is an npm command.
+	DependencyUseTypeNpmCommand DependencyUseType = "npm command"
+	// DependencyUseTypePipCommand is a pipp command.
+	DependencyUseTypePipCommand DependencyUseType = "pip command"
```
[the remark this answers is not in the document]
laurentsimon: the type here is using camel case like for other types we use, so it seems consistent. Sorry, I think I'm mis-understanding something.

checker/raw_result.go
```diff
@@ -60,6 +61,40 @@ type Fuzzer struct {
...
+	Dependencies []Dependency
+}
+
+// Dependency represents a dependency.
+type Dependency struct {
+	// TODO: unique dependency name.
+	// TODO: Job         *WorkflowJob
+	File *File
+	Msg  *string // Only for debug messages.
+	Type DependencyUseType
```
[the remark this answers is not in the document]
laurentsimon: `DependencyUseType` and `Name` are the same in your example, but yes we can add `PinnedAt`. I hesitated because its kinda redundant and the consumer can just split the string. But I see it could be useful, so I'll add it.

checker/raw_result.go
```diff
@@ -60,6 +61,40 @@ type Fuzzer struct {
...
+// PinningDependenciesData represents pinned dependency data.
+type PinningDependenciesData struct {
+	Dependencies []Dependency
+}
+
+// Dependency represents a dependency.
+type Dependency struct {
+	// TODO: unique dependency name.
+	// TODO: Job         *WorkflowJob
+	File *File
```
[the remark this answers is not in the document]
laurentsimon: sg, will do.

azeemshaikh38:
Let me know when the PR is ready for review.

laurentsimon:
PTAL. I'm populating the `PinnedAt` and `Name` only for containers and GHActions to start with.

checks/evaluation/pinned_dependencies.go
```diff
@@ -0,0 +1,310 @@
...
+			dl.Debug(&checker.LogMessage{
+				Path:      rr.Location.Path,
+				Type:      rr.Location.Type,
+				Offset:    rr.Location.Offset,
+				EndOffset: rr.Location.EndOffset,
+				Text:      *rr.Msg,
+				Snippet:   rr.Location.Snippet,
+			})
+		} else {
+			// Warn for inpinned dependency.
```
azeemshaikh38, line 79: Typo: `unpinned`

checks/evaluation/pinned_dependencies.go
```diff
@@ -0,0 +1,310 @@
...
+	// Update other result types.
+	var p pinnedResult
+	addPinnedResult(&p, false)
+	pr[rr.Type] = p
+}
+
+func generateText(rr *checker.Dependency) (string, error) {
+	if err := validateType(rr.Type); err != nil {
+		return "", err
+	}
```
azeemshaikh38, line 166: Do we need a `validateType` fn? Isn't that already done by Golang (i.e only enums of type `DependencyUseType` are allowed)?

checks/evaluation/pinned_dependencies.go
```diff
@@ -0,0 +1,310 @@
...
+	if rr.Type == checker.DependencyUseTypeGHAction {
+		// Check if we are dealing with a GitHub action or a third-party one.
+		gitHubOwned := fileparser.IsGitHubOwnedAction(rr.Location.Snippet)
```
azeemshaikh38, line 169: Shouldn't we be passing `rr.Name` here instead of `Snippet`?

azeemshaikh38 (approved):

checks/evaluation/pinned_dependencies.go
laurentsimon: Snippet contains `name@pin`, which is the original code. I'll update with `Name` or add a comment. Thanks

checks/evaluation/pinned_dependencies.go
laurentsimon: thanks.

PR #1939, opened 2022-05-22
✨  Update message for org-level security policy files

### LINKED ISSUE #1908: BUG: org-defined security policy link is incorrect
The org-defined security policy file is reported as `SECURITY.md` instead of `github.com/org/.github/SECURITY.md`

aidenwang9867:

Introduce a non-breaking feature.

As described in #1908, the org-defined security policy file is reported as `SECURITY.md` instead of `github.com/org/.github/SECURITY.md`.

Scorecards can report more granular information/messages for an org-defined (global) security policy as `github.com/ORG_NAME/.github/SECURITY.md` in the detailed logger.

No new unit tests have been added for this specific bug fix since (1) the changes don't affect the original correct judgment logic of the security policy file; (2) the root cause for this bug is trivial: when an org-defined security policy is detected, it doesn't set the ```file.Type``` to ```4``` (an URL) and the ```file.Path``` is not filled with the client URI (instead, it is just a file path in the repo with ```file.Type=1```)

Manual testcases like ```go run . --repo=aidenwang9867/scorecard --format json --checks Security-Policy --show-details --verbosity debug | jq``` (security policy in repo), ```go run . --repo=aidenwang9867/vld --format json --checks Security-Policy --show-details --verbosity debug | jq``` (security policy in org) can be used to verify the correctness of my changes.

- [x] Additional unit testcases have been added (see changes in ```checks/security_policy_test.go```)

1. > case insensitive testing for a security policy file name
2. > testing for the failure of a non-security policy filename (error triggered, might be a potential bug, see in issue #1954)

#1908 

N/A

No.

```release-note
Update message for org-level security policy files
```


checks/evaluation/security_policy.go
```diff
@@ -28,6 +28,7 @@ func SecurityPolicy(name string, dl checker.DetailLogger, r *checker.SecurityPol
 
 	// Apply the policy evaluation.
 	if r.Files == nil || len(r.Files) == 0 {
+		// if file is null or file has zero lengths, directly return not detected
```
laurentsimon, line 31: nit: start your comment with capital letter, and end with a full dot

checks/evaluation/security_policy.go
```diff
@@ -38,12 +39,15 @@ func SecurityPolicy(name string, dl checker.DetailLogger, r *checker.SecurityPol
 			Offset: f.Offset,
 		}
 		if msg.Type == checker.FileTypeURL {
-			msg.Text = "security policy detected in org repo"
+			msg.Text = "security policy detected in org global repo"
+			dl.Info(&msg)
 		} else {
-			msg.Text = "security policy detected"
+			// security policy detected in repo, return immediately since it has higher priority
```
laurentsimon, line 45: same comments for the comment :-)

checks/evaluation/security_policy.go
```diff
@@ -38,12 +39,15 @@ func SecurityPolicy(name string, dl checker.DetailLogger, r *checker.SecurityPol
 			Offset: f.Offset,
 		}
 		if msg.Type == checker.FileTypeURL {
-			msg.Text = "security policy detected in org repo"
+			msg.Text = "security policy detected in org global repo"
```
laurentsimon, line 42: "org repo" seems just fine instead of "org global repo"

checks/raw/security_policy.go
```diff
@@ -61,7 +61,13 @@ func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error)
 		return checker.SecurityPolicyData{}, err
 	}
 
-	// Return raw results.
+	// Modify the path to include org name and return raw results.
+	for i := range files {
+		// Add the client URI before the path
+		files[i].Path = dotGitHubClient.URI() + "/" + files[i].Path
+		// Modify the file type to 4 (meaning it's an URL)
+		files[i].Type = 4
```
laurentsimon, line 69: don't hard-code this value, use [checker.FileTypeURL](https://github.com/ossf/scorecard/blob/main/checker/check_result.go#L63)

checks/raw/security_policy.go
```diff
@@ -61,7 +61,13 @@ func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error)
 		return checker.SecurityPolicyData{}, err
 	}
 
-	// Return raw results.
+	// Modify the path to include org name and return raw results.
+	for i := range files {
+		// Add the client URI before the path
+		files[i].Path = dotGitHubClient.URI() + "/" + files[i].Path
```
laurentsimon, line 67: Use https://pkg.go.dev/path#Join to join paths.

checks/security_policy_test.go
```diff
@@ -113,6 +113,16 @@ func TestSecurityPolicy(t *testing.T) {
 				NumberOfInfo: 1,
 			},
 		},
+		{
+			name: "pass_case_testing_case_insensitive",
```
laurentsimon, line 117: you can use spaces in the name, no ned for `_`

checks/raw/security_policy.go
```diff
@@ -61,7 +61,13 @@ func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error)
 		return checker.SecurityPolicyData{}, err
 	}
 
-	// Return raw results.
+	// Modify the path to include org name and return raw results.
+	for i := range files {
```
laurentsimon, line 65: instead of patching the results, it may be cleaner to update isSecurityFile function to generate the right data. Wdut?

aidenwang9867: Modifying the func ```isSecurityPolicyFile``` including its input params and its call chain was my first idea, but my concerns are:

1. if we would like to include the orgname checking along with the filename checking into this func, at least one extra param is needed. Since the type of this func is `` fileparser.DoWhileTrueOnFilename``` which doesn't include the client contexts (especially the needed org name). AFAIK, in golang, what I can do is to create a new type like ```fileparser.DoWhileTrueOnFilenameWithClientContexts``` since I find golang doesn't support neither optional parameters nor method overloading. And if I do this, it seems the following modifications along with this would be much more.

2. Maybe we should keep the atomicity of this func? That is, only to check the filename itself. Otherwise,  it would be a little confusing for checking both ```filename``` and ```orgname + hostname + filename in this func.

3. If the information is only used for the logger, in my view the policy name results patching would be enough (and clean).

wdut? @laurentsimon

checks/evaluation/security_policy.go
aidenwang9867: modified, thanks

checks/evaluation/security_policy.go
aidenwang9867: modified, thanks

checks/evaluation/security_policy.go
aidenwang9867: Yes, I agree with this and have already set it back to the original one ;P

checks/raw/security_policy.go
aidenwang9867: Modified, thanks

checks/raw/security_policy.go
aidenwang9867: Modified, thanks

checks/security_policy_test.go
aidenwang9867: Got it, thanks

checks/raw/security_policy.go
aidenwang9867: I'll look into it and see which solution is better.

checks/evaluation/security_policy.go
```diff
@@ -39,11 +40,14 @@ func SecurityPolicy(name string, dl checker.DetailLogger, r *checker.SecurityPol
 		}
 		if msg.Type == checker.FileTypeURL {
 			msg.Text = "security policy detected in org repo"
+			dl.Info(&msg)
 		} else {
-			msg.Text = "security policy detected"
+			// Security policy detected in current repo, return immediately since it has higher priority.
+			msg.Text = "security policy file detected in current repo"
+			dl.Info(&msg)
+			return checker.CreateMaxScoreResult(name, "security policy file detected in current repo")
```
laurentsimon, line 48: you can keep the code the same in each if-statements. No need to return here. Since we allow multiple files for the repo, let's do the same for org repos.

checks/evaluation/security_policy.go
```diff
@@ -39,11 +40,14 @@ func SecurityPolicy(name string, dl checker.DetailLogger, r *checker.SecurityPol
...
 		}
-		dl.Info(&msg)
 	}
-
-	return checker.CreateMaxScoreResult(name, "security policy file detected")
```
laurentsimon, line 48: can keep the message the same in all cases. Users who want to know the details can inspect the `details` field.

laurentsimon (approved):
LGTM. I added just 2 comments. Let's address them and then merge.

checks/evaluation/security_policy.go
aidenwang9867: Updated, thanks

checks/evaluation/security_policy.go
aidenwang9867: Addressed, thanks

checks/raw/security_policy.go
```diff
@@ -71,14 +84,25 @@ var isSecurityPolicyFile fileparser.DoWhileTrueOnFilename = func(name string, ar
 	if len(args) != 1 {
 		return false, fmt.Errorf("isSecurityPolicyFile requires exactly one argument: %w", errInvalidArgLength)
 	}
-	pfiles, ok := args[0].(*[]checker.File)
+	contexts, ok := args[0].(struct {
```
laurentsimon, line 87: you may declare this as `type SomeName struct{}`, this way it's defined in one place.
You can simplify the struct as:
```
type SomeName struct{
  files []checker.File
  uri   string
}
```

then pass it as:
```
var data SomeName
...
fileparser.OnAllFilesDo(dotGitHubClient, isSecurityPolicyFile, &data)
...

OnAllFilesDo(){
 ...
  pdata, ok := args[0].(SomeName)
  pdata.files = append(pdata.files, ...)
}
```

aidenwang9867: Addressed, type SecurityPolicyFilesWithURI is defined in the current file, please see: https://github.com/ossf/scorecard/pull/1939/files#diff-a6da886b4a695f5dee2ea974759e5fc7bb15ed91e1b2f180365e2d5d7e431376R31-R35

checks/raw/security_policy.go
```diff
@@ -28,23 +28,25 @@ import (
...
 func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error) {
-	files := make([]checker.File, 0)
-	err := fileparser.OnAllFilesDo(c.RepoClient, isSecurityPolicyFile, struct {
-		FilePointer *[]checker.File
-		ClientURI   string
-	}{
-		&files,
-		"",
-	})
+	// files := make([]checker.File, 0)
```
laurentsimon, line 38: remove this comment.

checks/raw/security_policy.go
```diff
@@ -84,23 +80,19 @@ var isSecurityPolicyFile fileparser.DoWhileTrueOnFilename = func(name string, ar
 	if len(args) != 1 {
 		return false, fmt.Errorf("isSecurityPolicyFile requires exactly one argument: %w", errInvalidArgLength)
 	}
-	contexts, ok := args[0].(struct {
-		FilePointer *[]checker.File
-		ClientURI   string
-	})
+	pdata, ok := args[0].(*SecurityPolicyFilesWithURI)
+	// fmt.Println(pdata)
```
laurentsimon, line 84: remove comment

checks/raw/security_policy.go
```diff
@@ -84,23 +80,19 @@ var isSecurityPolicyFile fileparser.DoWhileTrueOnFilename = func(name string, ar
...
 	}
-	pfiles := contexts.FilePointer
-	clientURI := contexts.ClientURI
 	if isSecurityPolicyFilename(name) {
 		tempPath := name
 		tempType := checker.FileTypeSource
-		if clientURI != "" {
-			tempPath = path.Join(clientURI, "/", tempPath)
+		if pdata.uri != "" {
+			tempPath = path.Join(pdata.uri, "/", tempPath)
```
laurentsimon, line 92: you don't need the "/", path.Join automatically does this for you

checks/raw/security_policy.go
```diff
@@ -27,17 +28,24 @@ import (
...
+	uri   string
+}
+
 // SecurityPolicy checks for presence of security policy.
 func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error) {
-	files := make([]checker.File, 0)
-	err := fileparser.OnAllFilesDo(c.RepoClient, isSecurityPolicyFile, &files)
+	data := SecurityPolicyFilesWithURI{
+		files: make([]checker.File, 0),
+		uri:   "",
```
laurentsimon, line 40: Note: by default the string is the empty string, so no need for that

laurentsimon (approved):

checks/raw/security_policy.go
aidenwang9867: Got it, thanks. For this one, should I keep it as it is now, or delete the empty string `uri` and commit again?

laurentsimon: it's fine like that it. One might argue it's more explicit. Let's merge in!

aidenwang9867: No problem, thanks!

laurentsimon:
Some pre-submits are failing, PTAL https://github.com/ossf/scorecard/runs/6598997286?check_suite_focus=true

If you see errors like 
```
make: *** [Makefile:108: cron/data/request.pb.go] Error 127
Warning: Attempt 1 failed. Reason: Child_process exited with error code 2
```

this is a known problem. Let me know if all other pre-submits are failing and I'll merge in

Issue #1942, opened 2022-05-23
BUG: nil-ptr dereference 


azeemshaikh38:
Reproduce using:

```
SCORECARD_V6=1 ./scorecard --repo=ossf-tests/scorecard-check-token-permissions-e2e --commit=35a3425d1e682c32946b7d36adcfd772cf772e63 --checks=Token-Permissions
```

@laurentsimon seems to be happening in the `remediation.go` code that was added recently. Could you take a look.

On a side note, it still boggles me how the e2e tests don't catch this but fail when we run locally.

laurentsimon:
interesting. I'm surprised the e2e tests did not catch it.

laurentsimon:
fix in https://github.com/ossf/scorecard/pull/1947/files

PR #1947, opened 2022-05-23
🐛 Fix remediation code

laurentsimon:
Fix remediation code. The if statement was incorrect: we were not returning in case there was an error. `b` was nil in this case and caused a nil pointer.

```release-notes
Fix remediation code
```

closes https://github.com/ossf/scorecard/issues/1942


azeemshaikh38 (approved):
Not sure how this fixes the nil-ptr dereference, but code LG to submit. For the sake of posterity, could you explain in the PR description why this was causing code to panic?

laurentsimon:
> Not sure how this fixes the nil-ptr dereference, but code LG to submit. For the sake of posterity, could you explain in the PR description why this was causing code to panic?

Done in PR description.

Issue #1943, opened 2022-05-23
Share common data structures between `checker` and `client` packages


azeemshaikh38:
Tracking bug to see if common data structures like `Issue`, `Commit` etc. can be shared across `checker` and `client` packages instead of re-defining them.

PR #1956, opened 2022-05-24
🌱 Replace `checker.User` with `clients.User`

azeemshaikh38:
Part of #1943 

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -65,7 +65,7 @@ type CodeReviewData struct {
 
 // ContributorsData represents contributor information.
 type ContributorsData struct {
-	Users []User
+	Users []clients.Contributor
 }
 
 // VulnerabilitiesData contains the raw results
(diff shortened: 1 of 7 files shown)


checker/raw_result.go
```diff
@@ -65,7 +65,7 @@ type CodeReviewData struct {
 
 // ContributorsData represents contributor information.
 type ContributorsData struct {
-	Users []User
+	Users []clients.Contributor
```
laurentsimon, line 68: anything wrong with a contributor being a user? They are the same, some fields remain `nil`

azeemshaikh38: I didn't want to attempt combining `clients.Contributor` and `clients.User` struct in this PR. If you think removing `clients.Contributor` in favor of `clients.User` makes more sense, I can attempt that in a different PR.

laurentsimon: gocha. Let's try in a follow-up PR then. Unless you have reservation. Wdut?

azeemshaikh38: Nope, combining the 2 structs makes sense to me.

Issue #1874, opened 2022-04-27
Feature: add severity to each Warn result


laurentsimon:
We have severity at the level of a check today.
However, within a check, severity may vary. For example, contents: write is more critical than status: write, for example.

This could then be exposed in SARIF in the action https://github.com/ossf/scorecard-action/issues/210

laurentsimon:
@azeemsgoogle @naveensrinivasan @justaugustus wdut?

laurentsimon:
friendly ping: would love to hear comments or feedback. I think this would help users prioritize their remediation. It's also something thats been asked at least a couple times in the issues

azeemshaikh38:
SGTM.

justaugustus:
SGTM!

laurentsimon:
I think to be even broader, we need a flag that is either positive or negative. For example, some results are positive `Info()`.
These can be reported also in the raw results. So I think we need something like an enum to the effect of:

```
PositiveLow, PositiveMedium, PositiveHigh, PositiveOutstanding,
NegativeLow, NegativeMedium, NegativeHigh, NegativeCritical
```

laurentsimon:
An first use case is to lower the severity of `contents: write` declared at the top-level if there's only one job and the jobs is a release job (needs the permission). Today score is 10 :/

laurentsimon:
Below is how I think we could augment the existing results to be more structured:

```go
type Outcome string // PositiveLow, etc - see https://github.com/ossf/scorecard/issues/1874#issuecomment-1126265840
type Path struct{
	Type // File, URL, etc
	Value // will let you create links to original repo file, e.g. deps.dev when they show the results
	LineStart *uint
	LineEnd *uint
}

type RemediationEffort string // Minutes, Hours, Days, Weeks, etc. or Low/Medium/High/etc?

type Remediation struct{
	Text string	// For human
	Patch string	// For machines, e.g to create a pull request
	Effort RemediationEffort
}

type DetailID struct{
	Short string // short description
	URL string
	ID string // Will help with policy/configuration, e.g. to disable a type of detail results
}

type DevAnnotation struct {
   Text string //  string provided by developers about the check results, e.g., why they don't pass it
   Path Path // location of the annotation
}

type CheckResult struct {
  	Name    string
	Outcome *Outcome
	Remediation *Remediation // Next iteration
	DevAnnotation *DevAnnotation 
        Details [] struct{
		Path Path
                Outcome *Outcome
		Remediation *Remediation
[...]

PR #1977, opened 2022-06-01
✨ Silence scorecard warning

laurentsimon:
Scorecard complains because the top level permissions are not set to read.
Ill tackle this false positive as part of https://github.com/ossf/scorecard/issues/1874


naveensrinivasan (approved):
Thanks!

PR #1981, opened 2022-06-03
:sparkles: SAST Check for TFSec

naveensrinivasan:
`tfsec` is the standard for validating terraform code security.

It was requested in the Scorecard meeting on 6/2/2022.

Also, GitHub recommends these tools for code scanning https://github.blog/2021-07-28-new-code-scanning-integrations-open-source-security-tools/

### DIFF
--- a/checks/sast.go
+++ b/checks/sast.go
@@ -25,7 +25,7 @@ import (
 // CheckSAST is the registered name for SAST.
 const CheckSAST = "SAST"
 
-var sastTools = map[string]bool{"github-code-scanning": true, "lgtm-com": true, "sonarcloud": true}
+var sastTools = map[string]bool{"github-code-scanning": true, "lgtm-com": true, "sonarcloud": true, "tfsec": true}
 
 var allowedConclusions = map[string]bool{"success": true, "neutral": true}
 
@@ -49,10 +49,16 @@ func SAST(c *checker.CheckRequest) checker.CheckResult {
 		return checker.CreateRuntimeErrorResult(CheckSAST, codeQlErr)
 	}
 
+	tfsecScore, tfsecErr := tfsecCheckDefinitions(c)
+	if tfsecErr != nil {
+		return checker.CreateRuntimeErrorResult(CheckSAST, tfsecErr)
+	}
+
 	// Both results are inconclusive.
 	// Can never happen.
 	if sastScore == checker.InconclusiveResultScore &&
-		codeQlScore == checker.InconclusiveResultScore {
+		codeQlScore == checker.InconclusiveResultScore &&
+		tfsecScore == checker.InconclusiveResultScore {
 		// That can never happen since sastToolInCheckRuns can never
 		// retun checker.InconclusiveResultScore.
 		return checker.CreateInconclusiveResult(CheckSAST, "internal error")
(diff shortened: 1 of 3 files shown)


naveensrinivasan:
I am aware of there is an existing PR for SAST https://github.com/ossf/scorecard/pull/1487 , but it has been WIP for almost 6 months. The community is looking for tools like these and it is important that we provide them with these tools.

checks/sast.go
```diff
@@ -25,7 +25,7 @@ import (
 // CheckSAST is the registered name for SAST.
 const CheckSAST = "SAST"
 
-var sastTools = map[string]bool{"github-code-scanning": true, "lgtm-com": true, "sonarcloud": true}
+var sastTools = map[string]bool{"github-code-scanning": true, "lgtm-com": true, "sonarcloud": true, "tfsec": true}
```
laurentsimon, line 28: this is fine on the condition that the repo only contains these sort of files, ie it's not both a development repo and a deployment repo.

@nadgowdas can you confirm the above is correct? Could you also provide us with some example repos with terraform files? Is there a way to differentiate between the two "kinds" of repos?

checks/sast.go
```diff
@@ -25,7 +25,7 @@ import (
 // CheckSAST is the registered name for SAST.
 const CheckSAST = "SAST"
 
-var sastTools = map[string]bool{"github-code-scanning": true, "lgtm-com": true, "sonarcloud": true}
```
laurentsimon, line 28: this assumes that the tool is installed an a GH App. Is this how it works? I thought it was a CLI users need to run. if so, we may need to wait for https://github.com/ossf/scorecard/pull/1487 to land.

Wdut?

checks/sast.go
naveensrinivasan: Sorr I don't understand. Why can't it have both?

laurentsimon: it may have both, but it won't be reflected in the check results in the current implementation. The check uses an OR to pass, see https://github.com/ossf/scorecard/blob/d4dec7fdf1dd86238df448e0a4a116d5c36e056d/checks/sast.go#L162

checks/sast.go
naveensrinivasan: It is a GH Action. So will it be an issue? I don't think so.

laurentsimon: sorry I missed something when reading the PR. You can remove `tfsec` from the list. This list is only for GH apps.
The code `tfsecCheckDefinitions()` takes care of declaration of GH actions.

checks/sast.go
```diff
@@ -49,10 +49,16 @@ func SAST(c *checker.CheckRequest) checker.CheckResult {
 		return checker.CreateRuntimeErrorResult(CheckSAST, codeQlErr)
 	}
 
+	tfsecScore, tfsecErr := tfsecCheckDefinitions(c)
```
laurentsimon, line 52: if tfsecCheckDefinitions() succeeds, we're returning early. If repos with such files never have other type of code and don't need to have codeQL enabled, the code works. If it's possible they have both terraform code and other programming languages code, I think this will create false positive, ie high score even though dependabot is not enabled.

Wdut?

Is there a way to tweak the logic to accommodate both scenarios?

checks/sast.go
nadgowdas: Yes, I agree with @laurentsimon 
I think we should have a pre-scanner that parses code from given repo and identifies applicable checks. E.g. if repo has '.tf` code, then we select `tfsec` to be applicable. If code contains both app code (e.g. .go and .tf) then we should select checks applicable for both types. 
Also, if repo contains `.yaml`/helm charts then we could have provision to run tools like [kubelinter](https://github.com/stackrox/kube-linter). 
I general, it can be made pluggable for any "kind" of artifact in the code repository. Thoughts ?

laurentsimon: makes sense to me. I suggested a similar approach in https://github.com/ossf/scorecard/pull/1981#discussion_r890245194: would be great if we can use the language API (https://docs.github.com/en/rest/repos/repos#list-repository-languages). If not we'll have to list files ourselves for `.tf`, etc.

checks/sast.go
nadgowdas: I think that's a great idea to use that  API 👍

Issue #2008, opened 2022-06-28
Feature: Scorecard Dependency-diff API, CLI and Action


aidenwang9867:
**Is your feature request related to a problem? Please describe.**
The scorecard Action only reports the Scorecard results for the repository under analysis. The project will surface scorecard results for the dependencies for new pull requests.

**Describe the solution you'd like**
1. Add an API design that can fetch **direct** dependency changes (diffs) from a pull request using the [GitHub Dependency Review API](https://docs.github.com/en/rest/dependency-graph/dependency-review#get-a-diff-of-the-dependencies-between-commits), surface Scorecard checks and scores for those dependencies and return the results.

2. Add a new Scorecard sub-command CLI, `scorecard dependency-diff`, which uses the above API and outputs the result dependencies as JSON.

3. Add a new Scorecard Action running on a pull request, use the API to visualize the dependency-diff and their Scorecard checking results in the PR comments as markdown.

**Additional context**
Assignee: @aidenwang9867

naveensrinivasan:
> **Is your feature request related to a problem? Please describe.** The scorecard Action only reports the Scorecard results for the repository under analysis. The project will surface scorecard results for the dependencies for new pull requests.
> 
> **Describe the solution you'd like**
> 
> 1. Add a design that can fetch **direct** dependency changes (diffs) from a pull request using the [GitHub Dependency Review API](https://docs.github.com/en/rest/dependency-graph/dependency-review#get-a-diff-of-the-dependencies-between-commits), get dependencies of these direct dependencies (**indirect** ones) and vulnerabilities of both direct and indirect dependencies.
> 2. Information regarding `Dependencies` and `vulnerabilities` can be retrieved from the [BigQuery](https://cloud.google.com/bigquery) dataset [Open Source Insights](https://deps.dev/).
> 3. Visualize the result to the PR comments like what [CodeCov](https://docs.codecov.com/docs/pull-request-comments) does, with information showing if a dependency is `removed` or `added`. Also, give vulnerability alerts (`VulnID`, `VulnSource`, `CVSS3Score`, `Severity`, etc.) either together in the comments, or in the repo `Security` tag.
> 4. Blocks the PR if vulnerabilities found.
> 
> **Describe alternatives you've considered**
> 
> **Additional context** This design is Scorecard-focused and is expected to be used in a pull request as a part of the Scorecard Action pre-submit check automation.
> 
[...]

PR #2030, opened 2022-07-08
✨ Feature DependencyDiff CLI (Version 0 Part 1)

aidenwang9867:

This PR introduces the v0, p1 of the design [Dependency-diff CLI](https://docs.google.com/document/d/13UGTr7r_kYDdmpTxePZSJxEhAfBa694KFcSR23-eJQs/edit?usp=sharing),

including definitions of structures `Dependency`, `Vulnerability` and `ScorecardResult`;
> Currently in `ScorecardResult`, only the Scorecard aggregate score is used.

The current Scorecard Action only reports the Scorecard results for the repository under analysis. This design will surface Scorecard results for the arriving dependencies in new pull requests.

Since this PR only includes static definitions of structures, no new behaviors are introduced. However, with Version Zero of the feature, users can use the CLI to surface the scorecard results (aggregate scores) for the dependency diffs (changes) between a user-specified head commit and a base commit in different output format.

Part1 does not include functions and APIs. Tests will be added in incoming parts.

issue [#2008](https://github.com/ossf/scorecard/issues/2008).

No.

```release-note
NONE
```


cmd/depdiff/dependencies.go
```diff
@@ -0,0 +1,78 @@
...
+	IsDirect bool
+
+	// ChangeType indicates whether the dependency is added or removed.
+	ChangeType ChangeType `json:"change_type"`
+
+	// ManifestFileName is the name of the manifest file of the dependency, such as go.mod for Go.
+	ManifestFileName string `json:"manifest"`
+
+	// Ecosystem is the name of the package management system, such as NPM, GO, PYPI.
+	Ecosystem string `json:"ecosystem" bigquery:"System"`
```
laurentsimon, line 50: Do we need the `bigquery` metadata to the definition?

cmd/depdiff/dependencies.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Name is the name of the dependency.
+	Name string `json:"name" bigquery:"Name"`
```
laurentsimon, line 53: @oliverchang please give us some feedback on the best naming convention we should follow. Ideally it would be generic.

cmd/depdiff/dependencies.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Version is the package version of the dependency.
+	Version string `json:"version" bigquery:"Version"`
+
+	// Package URL is a short link for a package.
+	PackageURL string `json:"package_url"`
+
+	// License is the license of the dependency.
+	License string `json:"license"`
```
laurentsimon, line 62: This is present in scorecard results, so we should not need it, correct?

cmd/depdiff/dependencies.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// SrcRepoURL is the source repository URL of the dependency.
+	SrcRepoURL string `json:"source_repository_url"`
+
+	// ScResults is the Scorecard scanning result of the dependency package repository.
+	ScResults ScorecardResult `json:"scorecard_results"`
+
+	// Vulnerabilities is a list of Vulnerability.
+	Vulnerabilities []Vulnerability `json:"vulnerabilities"`
```
laurentsimon, line 71: This is present in Scorecard, so should not be needed, correct?

cmd/depdiff/dependencies.go
laurentsimon, line 53: Maybe make all these fields pointers, because in some cases they won't be populated?

cmd/depdiff/dependencies.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Dependencies is the list of dependencies of the current dependency,
+	// e.g. indirect (transitive) dependencies.
+	// TODO: this is not a version-zero feature, and will be used to analyze transitive
+	// dependencies in future versions.
+	Dependencies []Dependency `json:"dependencies"`
```
laurentsimon, line 77: Let's remove this field since it's not used.

cmd/depdiff/scorecard_results.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package depdiff
+
+// ScorecardResult is the Scorecard scanning result of a repository.
+type ScorecardResult struct {
```
laurentsimon, line 18: This is already defined in scorecard `pkg` folder, so we should not need to re-define it, right?

cmd/depdiff/vulnerabilities.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2022 Security Scorecard Authors
```
laurentsimon, line 1: I think we can remove this since it's part of Scorecard results

cmd/depdiff/dependencies.go
aidenwang9867: For v0, since we don't use any bigquery data sources to fetch data fields, I think the `bigquery` tags are not necessary. Should I remove them in v0, or leave them there for future versions to use?

cmd/depdiff/dependencies.go
aidenwang9867: yes i agree

cmd/depdiff/dependencies.go
aidenwang9867: Yes, I agree.

cmd/depdiff/dependencies.go
aidenwang9867: Are you referring to both `PackageURL` and `License`, or just `License`?

cmd/depdiff/dependencies.go
laurentsimon: yes, remove everything that's not necessary for this PR. Thanks

cmd/depdiff/errors.go
```diff
@@ -0,0 +1,27 @@
...
+
+package depdiff
+
+import (
+	"errors"
+)
+
+var (
+	// ErrInvalidDepDiffFormat indicates the specified dependency diff output format is not valid.
+	ErrInvalidDepDiffFormat = errors.New("invalid depdiff format")
```
laurentsimon, line 23: those don't seem to be used, so let's remove this file and add it in the commit that introduces code that uses them.

cmd/depdiff/dependencies.go
aidenwang9867: np, thx

cmd/depdiff/errors.go
aidenwang9867: np

cmd/depdiff/dependencies.go
aidenwang9867: Update: since the `License` field is duplicated and has been removed from the `Dependency` struct, only `package_url` and `source_repository_url` are changed to pointers for now.

cmd/depdiff/dependencies.go
laurentsimon: License.

cmd/depdiff/dependencies.go
laurentsimon: The structure should be generic and not depend on the GH API. Right now we populate it with the GH API but this may change in the future.

Spikebady:
Tha.7

cmd/depdiff/dependencies.go
Spikebady: Thank you so much

Spikebady (commented):
I have yet to see my videos

pkg/check-depdiff/dependencies.go
```diff
@@ -0,0 +1,58 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+// Dependency is a dependency.
+type Dependency struct {
+	// ChangeType indicates whether the dependency is added, updated, or removed.
+	ChangeType ChangeType `json:"change_type"`
```
laurentsimon, line 39: In the scorecard codebase, we use camel case for this, so let's use that here too. Example: `changeType`

pkg/check-depdiff/dependencies.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// ManifestFileName is the name of the manifest file of the dependency, such as go.mod for Go.
+	ManifestFileName string `json:"manifest"`
```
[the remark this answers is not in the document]
aidenwang9867: confirmed it is a path - changing this to `ManifestPath`, thanks!

pkg/check-depdiff/dependencies.go
laurentsimon: I mean the json tag, yes. We agreed that this structure should be generic and not tied to an API, right? My understanding is that this structure is the one used tby scorecard to return the results, not the structure used to query the GH API.

Let me know if I mis-understood something

aidenwang9867: @laurentsimon  PTAL at the latest changes, thx :D

pkg/check-depdiff/dependencies.go
```diff
@@ -0,0 +1,91 @@
...
+	case Added, Updated, Removed:
+		return true
+	default:
+		return false
+	}
+}
+
+// rawDependency is the Dependency structure that is used to receive
+// the raw results from the GitHub Dependency Review API.
+type rawDependency struct {
```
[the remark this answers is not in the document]
aidenwang9867: np

pkg/check-depdiff/dependencies.go
```diff
@@ -0,0 +1,66 @@
...
+	PackageURL *string `json:"packageUrl"`
+
+	// SrcRepoURL is the source repository URL of the dependency.
+	SrcRepoURL *string `json:"srcRepoUrl"`
+
+	// ChangeType indicates whether the dependency is added, updated, or removed.
+	ChangeType *ChangeType `json:"changeType"`
+
+	// ManifestFileName is the name of the manifest file of the dependency, such as go.mod for Go.
+	ManifestPath *string `json:"manifest"`
```
laurentsimon, line 53: not: update `json:"manifest"` to `json:"manifestPath"`

pkg/check-depdiff/dependencies.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+// DependencyCheckResult is the dependency structure used in the returned results.
+type DependencyCheckResult struct {
+	// Package URL is a short link for a package.
+	PackageURL *string `json:"packageUrl"`
+
+	// SrcRepoURL is the source repository URL of the dependency.
+	SrcRepoURL *string `json:"srcRepoUrl"`
```
laurentsimon, line 47: nit: let's not abbreviate and use `SourceRepository`

pkg/check-depdiff/dependencies.go
```diff
@@ -0,0 +1,66 @@
...
+	Ecosystem *string `json:"ecosystem"`
+
+	// Name is the name of the dependency.
+	Name string `json:"name"`
+
+	// Version is the package version of the dependency.
+	Version *string `json:"version"`
+
+	// ScReresults is the scorecard result for the dependency repo.
+	ScReresults *pkg.ScorecardResult `json:"scorecardResults"`
```
laurentsimon, line 65: typo in `ScReresults `? Let's jut call it `ScorecardResults`

pkg/check-depdiff/dependencies.go
aidenwang9867: done, thx

pkg/check-depdiff/dependencies.go
aidenwang9867: done

pkg/check-depdiff/dependencies.go
```diff
@@ -44,13 +44,13 @@ type DependencyCheckResult struct {
 	PackageURL *string `json:"packageUrl"`
 
 	// SrcRepoURL is the source repository URL of the dependency.
-	SrcRepoURL *string `json:"srcRepoUrl"`
+	SrcRepoURL *string `json:"sourceRepositoryURL"`
```
laurentsimon, line 47: the variable name too. Note: `sourceRepository` is probably enough

aidenwang9867: oops sorry :/  done!

laurentsimon (approved):
Thanks!

aidenwang9867:
> Thanks!

thank u so much!

spikeb4 (approved):

PR #2033, opened 2022-07-11
:seedling: Fix for protoc validate script

naveensrinivasan:
- Fix for validate script failures


laurentsimon:
The pre-submit is still failing.
If this is a rate limit problem, how about we keep the binary on a GCS bucket and download it (can verify the hash)?

naveensrinivasan:
> The pre-submit is still failing. If this is a rate limit problem, how about we keep the binary on a GCS bucket and download it (can verify the hash)?
I agree that will probably be a better choice.

naveensrinivasan:
This isn't applicable any more thanks to @raghavkaul fix!

PR #2037, opened 2022-07-11
✨ Use crane to add hash suggestion to unpinned Docker images

### LINKED ISSUE #967: Feature - Provide docker image sha for failed pinned dependencies
Right now scorecard provides a warning message for  `Warn: unpinned dependency detected in Dockerfile: ' alpine:3.14` when the docker file does not have a pinned SHA.

Instead of just providing the warning it would be helpful to provide the actual SHA that would have to be used to address this issue. 
This can be achieved by using https://github.com/google/go-containerregistry/tree/main/cmd/crane and running something like `crane digest alpine:3.14` which provides the SHA
`sha256:e1c082e3d3c45cccac829840a25941e679c25d438cc8412c2fa221cf1a824e6a`

spencerschrock:

This PR attempts to address an enhancement in #967 

A warning is generated when a Docker dependency is not pinned.

> Warn: containerImage not pinned by hash: Dockerfile:1

A hash is now provided, as determined by `crane`

> Warn: containerImage not pinned by hash. Fix by updating golang:1.17 to golang:1.17@sha256:e4d220556e80a686fad8874b40ccfa2f3172dfe5c6ca7f36f8ee79d594a9c959: Dockerfile:1

Fixes #967 

NONE

the `release-note`

```release-note
Hash suggestions are now made for unpinned images in Dockerfiles
```

### DIFF
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -18,8 +18,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/google/go-containerregistry/pkg/crane"
-
 	"github.com/ossf/scorecard/v4/checker"
 	"github.com/ossf/scorecard/v4/checks/fileparser"
 	sce "github.com/ossf/scorecard/v4/errors"
@@ -140,10 +138,7 @@ func generateRemediation(rr *checker.Dependency) *checker.Remediation {
 	case checker.DependencyUseTypeGHAction:
 		return remediation.CreateWorkflowPinningRemediation(rr.Location.Path)
 	case checker.DependencyUseTypeDockerfileContainerImage:
-		if rr.Name == nil {
-			return nil
-		}
-		return remediation.CreateDockerfilePinningRemediation(*rr.Name)
+		return remediation.CreateDockerfilePinningRemediation(rr.Name)
 	default:
 		return nil
 	}
@@ -167,21 +162,13 @@ func updatePinningResults(rr *checker.Dependency,
 }
 
 func generateText(rr *checker.Dependency) string {
-	switch rr.Type {
-	case checker.DependencyUseTypeGHAction:
+	if rr.Type == checker.DependencyUseTypeGHAction {
 		// Check if we are dealing with a GitHub action or a third-party one.
 		gitHubOwned := fileparser.IsGitHubOwnedAction(rr.Location.Snippet)
 		owner := generateOwnerToDisplay(gitHubOwned)
 		return fmt.Sprintf("%s %s not pinned by hash", owner, rr.Type)
-	case checker.DependencyUseTypeDockerfileContainerImage:
-		if rr.Name == nil {
-			break
-		}
-		if hash, err := crane.Digest(*rr.Name); err == nil { // if NO error
-			return fmt.Sprintf("%s not pinned by hash. Fix by updating %[2]s to %[2]s@%s", rr.Type, *rr.Name, hash)
-		}
-	default:
 	}
+
 	return fmt.Sprintf("%s not pinned by hash", rr.Type)
 }
 
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -136,10 +136,17 @@ func PinningDependencies(name string, dl checker.DetailLogger,
 }
 
 func generateRemediation(rr *checker.Dependency) *checker.Remediation {
-	if rr.Type == checker.DependencyUseTypeGHAction {
+	switch rr.Type {
+	case checker.DependencyUseTypeGHAction:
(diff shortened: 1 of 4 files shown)


spencerschrock:
I was aware of the `crane` options, but wasn't sure how to determine things on the `scorecard` side of things. From what I can tell, the same unpinned Dockerfile can be used on different platforms and work accordingly. We also can't make assumptions based on the platform where `scorecard` is running as it may differ from where the Dockerfile is deployed. So I was thinking we need some sort of option or metadata added. Is there anything already existing?

> On that note, we should have standard comment blocks for these remediations.
> We would need some standard prefix for our recommendations so that customers can grep for it.

+1 
I think #1874 would cover this

checks/evaluation/pinned_dependencies.go
```diff
@@ -158,13 +167,21 @@ func updatePinningResults(rr *checker.Dependency,
...
+	case checker.DependencyUseTypeGHAction:
 		// Check if we are dealing with a GitHub action or a third-party one.
 		gitHubOwned := fileparser.IsGitHubOwnedAction(rr.Location.Snippet)
 		owner := generateOwnerToDisplay(gitHubOwned)
 		return fmt.Sprintf("%s %s not pinned by hash", owner, rr.Type)
+	case checker.DependencyUseTypeDockerfileContainerImage:
+		if rr.Name == nil {
+			break
+		}
+		if hash, err := crane.Digest(*rr.Name); err == nil { // if NO error
```
laurentsimon, line 180: The remediation text only needs to live in the remediation.go file. You have 2 options to make the remediation results part of the final JSON results:
1. Start updating the JSON structure we use in https://github.com/ossf/scorecard/blob/main/pkg/json.go, which is step towards achieving https://github.com/ossf/scorecard/issues/1874
2. Use the remediation results when constructing the final JSON string in  https://github.com/ossf/scorecard/blob/main/pkg/json.go, which is done in https://github.com/ossf/scorecard/blob/main/pkg/common.go#L30

As an example, you can see we do exactly that for SARIF output in https://github.com/ossf/scorecard/blob/main/pkg/sarif.go#L305-L308

For this PR, option 2 is the one I'd choose since option 1 is part of a bigger issue.

spencerschrock: Thanks for the feedback.

> The remediation text only needs to live in the remediation.go file

Sounds good. It felt weird having the logic in two places. I'll focus on `remediation.go` and revert the other commits.

> For this PR, option 2 is the one I'd choose since option 1 is part of a bigger issue.
I'll take a look at your provided links and take a stab at it

laurentsimon:
> > On that note, we should have standard comment blocks for these remediations.
> > We would need some standard prefix for our recommendations so that customers can grep for it.
> 
> +1 I think #1874 would cover this

+1, the JSON should do that for free. /cc @raghavkaul

spencerschrock:
And to address @naveensrinivasan 's earlier point:

> Is this PR going to provide only SHA for Linux/amd64?
> Crane provides an option ...

`crane`'s default should be good for multi-arch support. more detail in #967

pkg/common.go
```diff
@@ -32,16 +32,24 @@ func DetailToString(d *checker.CheckDetail, logLevel log.Level) string {
...
-	switch {
-	case d.Msg.Path != "" && d.Msg.Offset != 0 && d.Msg.EndOffset != 0 && d.Msg.Offset < d.Msg.EndOffset:
-		return fmt.Sprintf("%s: %s: %s:%d-%d", typeToString(d.Type), d.Msg.Text, d.Msg.Path, d.Msg.Offset, d.Msg.EndOffset)
-	case d.Msg.Path != "" && d.Msg.Offset != 0:
-		return fmt.Sprintf("%s: %s: %s:%d", typeToString(d.Type), d.Msg.Text, d.Msg.Path, d.Msg.Offset)
-	case d.Msg.Path != "" && d.Msg.Offset == 0:
-		return fmt.Sprintf("%s: %s: %s", typeToString(d.Type), d.Msg.Text, d.Msg.Path)
-	default:
-		return fmt.Sprintf("%s: %s", typeToString(d.Type), d.Msg.Text)
+	var sb strings.Builder
```
laurentsimon, line 35: Looks good. Shall we add a few unit tests?

laurentsimon (approved):
Thanks. Maybe add a few unit tests for the formatting part.

pkg/common.go
spencerschrock: Probably a good idea. The existing behavior should be easy to write a test for.
I played around earlier at how I might mock the `crane` call, so I should have something tomorrow.

laurentsimon: SG. Also for the output string format (nothing crazy complicated, just a few combinations of Text, Path, etc being set or not)

spencerschrock: Added some tests, which the original code passes except for the new remediation one.

PR #2039, opened 2022-07-12
✨ Binary artifact exception for gradle-wrapper.jar when using validation action

### LINKED ISSUE #1815: Binary Artifacts should allow well-known artifacts, if best practices are followed

> The gradle-wrapper.jar is a binary blob of executable code that is checked into nearly [2.8 Million GitHub Repositories](https://github.com/search?l=&q=filename%3Agradle-wrapper.jar&type=Code).

This is the intended use of this jar

https://github.com/marketplace/actions/gradle-wrapper-validation
https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification

Binary artifacts knows about gradle-wrapper.jar and allows it if the repo has the Wrapper Validation Action installed (or another community-approved mitigation best practice).

An alternative could be fully user configurable, like providing the name and checksum (https://gradle.org/release-checksums/).  The public scorecard results wouldn't have this, but user-run scorecard checks could exempt certain binaries.

ethanent:

Feature

Scorecard considers `gradle-wrapper.jar` files to be harmful binary artifacts even when they are verified by the [gradle/wrapper-validation-action](https://github.com/gradle/wrapper-validation-action) action.

`gradle-wrapper.jar` files will no longer be flagged by the binary artifacts check iff all conditions are met:
- There is a workflow using the gradle/wrapper-validation-action action in a step in a job
- That workflow passed for the latest commit on default branch

Fixes #1815 

the `release-note`

```release-note
gradle-wrapper.jar files are no longer flagged by the Binary Artifacts check if the Gradle Wrapper Validation Action is enabled as part of a passing workflow.
```


checks/raw/binary_artifact_test.go
```diff
@@ -21,54 +21,136 @@ import (
...
 func TestBinaryArtifacts(t *testing.T) {
 	t.Parallel()
 	tests := []struct {
-		name   string
-		err    error
-		files  []string
-		expect int
+		name                   string
+		err                    error
+		files                  [][]string
```
loosebazooka, line 38: could this be 
```
   files       []string
   actionFiles []string
```
instead?

checks/raw/binary_artifact.go
```diff
@@ -40,6 +44,27 @@ func BinaryArtifacts(c clients.RepoClient) (checker.BinaryArtifactData, error) {
 		return checker.BinaryArtifactData{}, fmt.Errorf("%w", err)
 	}
 
+	// Indices of any gradle-wrapper.jar files
+	var gradleWrappers []int
```
loosebazooka, line 48: I know I'm a java man, but would it make sense to call these `gradleWrapperIndexes` (or `gwIndexes` I dunno)?

checks/raw/binary_artifact.go
```diff
@@ -40,6 +44,27 @@ func BinaryArtifacts(c clients.RepoClient) (checker.BinaryArtifactData, error) {
...
+			}
+		}
+	}
+	if len(gradleWrappers) > 0 {
+		// Gradle wrapper JARs present, so check that they are validated
+		if ok, err := gradleWrapperValidationOK(c); ok && err == nil {
+			// It has been confirmed that latest commit has validated JARs!
+			// Remove Gradle wrapper JARs from files.
+			offset := 0
+			for _, ji := range gradleWrappers {
```
laurentsimon, line 62: This part of the code is a little hard to follow. Any simplification we can do?
```golang
binaries := []checker.File{}
err := fileparser.OnMatchingFileContentDo(c, fileparser.PathMatcher{
		Pattern:       "*",
		CaseSensitive: false,
	}, checkBinaryFileContent, &binaries)
...

files :=  []checker.File{}
for f := range binaries{
  if isGraddleJAR() {
   continue
  }

  // for other ecosystems
  if isMavenJAR() {
   continue
  }

  files = append(files, f)
}
```

Wdut?

laurentsimon (commented):
Thanks for the PR. As @loosebazooka said, there is demand for this!

checks/raw/binary_artifact.go
```diff
@@ -40,6 +44,27 @@ func BinaryArtifacts(c clients.RepoClient) (checker.BinaryArtifactData, error) {
...
+	// Indices of any gradle-wrapper.jar files
+	var gradleWrappers []int
+	if len(files) > 0 {
+		for i, f := range files {
+			if filepath.Base(f.Path) == "gradle-wrapper.jar" {
+				gradleWrappers = append(gradleWrappers, i)
+			}
+		}
+	}
+	if len(gradleWrappers) > 0 {
```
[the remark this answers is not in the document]
ethanent: #2048 created

checks/raw/binary_artifact.go
```diff
@@ -40,6 +57,33 @@ func BinaryArtifacts(c clients.RepoClient) (checker.BinaryArtifactData, error) {
...
+			// It has been confirmed that latest commit has validated JARs!
+			// Remove Gradle wrapper JARs from files.
+			filterFiles := []checker.File{}
+			for _, f := range files {
+				if filepath.Base(f.Path) != "gradle-wrapper.jar" {
+					filterFiles = append(filterFiles, f)
+				}
+			}
+			files = filterFiles
+		} else if err != nil {
```
laurentsimon, line 82: let's put that early:
```
ok, err := gradleWrapperValidated(c);
if err != nil {return}
...
```

checks/binary_artifact.go
```diff
@@ -50,5 +49,5 @@ func BinaryArtifacts(c *checker.CheckRequest) checker.CheckResult {
 	}
 
 	// Return the score evaluation.
-	return evaluation.BinaryArtifacts(CheckBinaryArtifacts, c.Dlogger, &rawData)
+	return evaluation.BinaryArtifacts(CheckBinaryArtifacts, c.Dlogger, &rawData, c.RepoClient)
```
laurentsimon, line 52: I know the change is well intentioned - really appreciate the effort. But I don't think we want make that change unfortunately. I should have told you earlier.

I know you're trying to move the evaluation of the results into the evaluation (which is what we want long-term), but the evaluation should *not* require any further access to the API: The raw results should contain all the information necessary to evaluate the results; it's the main reason we created them. For example, imagine someone retrieves them from our BQTable, they should be able to parse and evaluation using their "policy".

There's no simple solution today. You can think about where this would live in the raw structure https://github.com/ossf/scorecard/blob/main/checker/raw_result.go. We have the same limitation today for permissions: we have "exceptions" in the sense that we allow write permission if a certain action is used (e.g., create a release needs `content:write` permission). We would like to move them to the evaluation, but we have not decided which structure we should keep that data in.

If you have some ideas where this information could live in the raw results: new structure or in an existing structure, that'd be really nice. But for this PR I think it's fine to create the exception in the `raw/` instead.

Does this make sense?

/cc @raghavkaul something to be aware that we need to fix.

laurentsimon: Maybe it could live under a generic `Workflows` structure with a list actions found for each job. Ideally that would also contain permission information, so that both the permission job, dangerous workflow and permission checks could use it. But it's a pretty big change so we need to think about it a bit more.

ethanent: That makes sense, thanks for the explanation!

laurentsimon:
Let us know when you need a review.

checks/raw/binary_artifact_test.go
ethanent: I see what you mean. Eventually, once #2048 is solved and the new test cases are moved to checks/evaluation from checks/raw, it'll no longer be helpful to use [][]string here.

In the meantime, with the current test cases, it might make sense to use [][]string because each []string represents a ListFiles() call

checks/raw/binary_artifact.go
```diff
@@ -39,11 +56,52 @@ func BinaryArtifacts(c clients.RepoClient) (checker.BinaryArtifactData, error) {
...
 	// No error, return the files.
 	return checker.BinaryArtifactData{Files: files}, nil
 }
 
+// excludeValidatedGradleWrappers returns the subset of files not confirmed
+// to be Action-validated gradle-wrapper.jar files.
+func excludeValidatedGradleWrappers(c clients.RepoClient, files []checker.File) ([]checker.File, error) {
+	// Check if gradle-wrapper.jar present
+	hasGradleWrappers := false
+	for _, f := range files {
```
laurentsimon, line 74: nit: create a function `filePresent(string)` (or a better name). Then when we add support for another ecosystem, we can re-use `filePresent("other-wrapper.bla")`. This way you can have:
```
if !filePresent("gradle-wrapper.jar"){
  return files, nil
}
```

laurentsimon (approved):
Thanks, LGTM!

PR #2044, opened 2022-07-12
🌱 Adjust 'exhaustive' linter to consider 'default' as exhaustive

spencerschrock:

Configuration change for `golangci-lint`. With the current linter settings, `default` is useless and satisfying the linter involves adding empty cases that clutter the code more than `default`. I realize this is subjective, but the current setting is overly cautious.

Switch statements on enums currently require cases for all enum members.

`default` will satisfy the linter now. 

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -81,6 +81,8 @@ linters-settings:
   errcheck:
     check-type-assertions: true
     check-blank: true
+  exhaustive:
+    default-signifies-exhaustive: true
   govet:
     enable:
       - fieldalignment
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -82,6 +82,7 @@ linters-settings:
     check-type-assertions: true
     check-blank: true
   exhaustive:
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -81,6 +81,8 @@ linters-settings:
   errcheck:
     check-type-assertions: true
     check-blank: true
+  exhaustive:
+    default-signifies-exhaustive: true
```
naveensrinivasan, line 85: Can you please add a link to the docs?

naveensrinivasan (approved):
Thank you!

.golangci.yml
spencerschrock:
```suggestion
    # https://golangci-lint.run/usage/linters/#exhaustive
    default-signifies-exhaustive: true
```

wasn't sure if you meant here or in the config. assumed the latter

PR #2046, opened 2022-07-13
✨ Feature DependencyDiff (Version 0 Part 2)

aidenwang9867:
This is a follow-up PR of PR #2030, introducing the main Dependency-diff API to the Scorecard repo.

Detailed changes include:

1. Adding the function API `GetDependencyDiffResults` used to fetch Scorecard results for dependency changes between two code commits and return the dependency check results (with scorecard results included);
2. Adding the function API `FetchDependencyDiffData` used to get dependency-diffs between two code commits;
3. Add the raw `Dependency` struct used for receiving raw data from the dependency-diff source (currently we are using the GitHub Dependency Review API);
4. Adding `AsJSON` support for `DependencyCheckResult`, making it JSON-dump-able;
5. Adding unit tests for corresponding functions;
6. Minor directory and naming changes.

In version 0 part 1 (PR #2030), only the exported result structure `DependencyCheckResult` was introduced.

Introduce subsequent functions and structures of this [feature](https://github.com/ossf/scorecard/issues/2008).

This is a part of the issue [#2008](https://github.com/ossf/scorecard/issues/2008).

No.

the `release-note`

```release-note
NONE
```


dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error fetching dependency changes: %w", err)
+	}
+
+	ghRepo, err := githubrepo.MakeGithubRepo(path.Join(ownerName, repoName))
+	if err != nil {
+		return nil, fmt.Errorf("error creating the github repo: %w", err)
+	}
+	ghRepoClient := githubrepo.CreateGithubRepoClient(ctx, nil)
+	ossFuzzRepoClient, err := githubrepo.CreateOssFuzzRepoClient(ctx, nil)
```
laurentsimon, line 47: do we need this? Something breaks if we set it to `nil`?
Same question for `DefaultVulnerabilitiesClient` below

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating the oss fuzz repo client: %w", err)
+	}
+	vulnsClient := clients.DefaultVulnerabilitiesClient()
+
+	results := []pkg.DependencyCheckResult{}
+	for _, d := range deps {
+		if d.SourceRepository != nil {
+			if *d.SourceRepository == "" {
```
laurentsimon, line 56: How about appending a result with empty scorecard results, to make it explicit we could not find the result?
This way the caller can decide what do do with it: ignore it, or display a message to the user that it could not be retrieved (e.g., in the GHA).

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,119 @@
...
+				continue
+			}
+			// Running scorecard on this dependency and fetch the result.
+			// TODO: use the Scorecare REST API to retrieve the Scorecard result statelessly.
```
laurentsimon, line 60: Feel free to create an issue and update the TODO with a link to the issue `//TODO(link): do something`

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,119 @@
...
+				checks.AllChecks,
+				ghRepoClient,
+				ossFuzzRepoClient,
+				nil,
+				vulnsClient,
+			)
+			if err != nil {
+				return nil, fmt.Errorf("error fetching the scorecard result: %w", err)
+			}
+			depCheckResult := pkg.DependencyCheckResult{
```
laurentsimon, line 74: see comment above: we may want to append a result even if the scorecard result is `nil`. 
I'm not sure what we should do if scorecard run returns an error: I'm inclined to not fail immediately and instead return the scorecard return error. This way even if one run fails, it still provides the results for other dependencies.

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,119 @@
...
+			results = append(results, depCheckResult)
+		}
+		// Skip those without source repo urls.
+		// TODO: use the BigQuery dataset to supplement null source repo URLs
+		// so that we can fetch the Scorecard results for them.
+	}
+	return results, nil
+}
+
+// FetchDependencyDiffData fetches the depednency-diffs between the two code commits
```
laurentsimon, line 93: nit: typo depednency-diffs

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,119 @@
...
+// using the GitHub Dependency Review API, and returns a slice of Dependency.
+func FetchDependencyDiffData(ctx context.Context, owner, repo, base, head string) ([]Dependency, error) {
```
laurentsimon, line 95: this seems to be a private API for this file, so let's not export it.

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,119 @@
...
+	reqURL, err := url.Parse("https://api.github.com")
+	if err != nil {
+		return nil, fmt.Errorf("error parsing the url: %w", err)
+	}
+	reqURL.Path = path.Join(
```
laurentsimon, line 100: will this fail if someone run scorecard on a Windows machine? Since there will be `\\` instead of `/`?
Maybe that's a problem with other APIs we have !?
There must be a proper http path API that exist for this purpose maybe, https://pkg.go.dev/net/url#QueryEscape?

dependencydiff/raw_dependencies.go
```diff
@@ -0,0 +1,41 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dependencydiff
+
+import "github.com/ossf/scorecard/v4/pkg"
+
+// Dependency is a raw dependency fetched from the GitHub Dependency Review API.
+type Dependency struct {
```
laurentsimon, line 20: How about implementing `FetchDependencyDiffData()` in this file?
This way we can hide `Dependency` and make it private. This function can populate the `DependencyCheckResult`

laurentsimon (commented):
Looking good. Added a few comments.
Thank you!

dependencydiff/dependencydiff.go
aidenwang9867: yes, i think it makes sense that we add empty scorecard results to those without a srcRepoURL.

dependencydiff/dependencydiff.go
aidenwang9867: I agree that we can return the scorecard error and continue to the next dependency if the run fails. 

When it fails, what will the function `RunScorecards`-returned `ScorecardResult` be like? Can I append the failed `scorecardResult` directly to the results, or should I generate another error message using another module?

dependencydiff/dependencydiff.go
aidenwang9867: um, i tested it on my personal windows PC and it seems to be fine.

still looking for documents and APIs that can be more properly used for this url construction purpose.

laurentsimon: Let's not spend too much time on this, it's fine for now. Thanks for checking.

dependencydiff/dependencydiff.go
laurentsimon: Scorecard results should have an `error` field we can use to populate it. Can you check?

dependencydiff/dependencydiff.go
laurentsimon: Gotcha. Add a comment.
@azeemsgoogle , was this intended? Shall we file an issue for it?

dependencydiff/raw_dependencies.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// Ecosystem is the name of the package management system, such as NPM, GO, PYPI.
+	Ecosystem *string `json:"ecosystem"`
+
+	// Version is the package version of the dependency.
+	Version *string `json:"version"`
+
+	// Name is the name of the dependency.
+	Name string `json:"name"`
+}
```
azeemshaikh38, line 54: You shouldn't need a new struct for parsing the response JSON. `pkg.DependencyCheckResult` can be used here directly.

azeemshaikh38 (commented):
Added some comments. PTAL.

dependencydiff/dependencydiff.go
aidenwang9867: re: do we really want to run all the checks here?

I think for this API, we might want to give all the scorecard check results for users themselves to decide which one to use in their projects. A better way is undoubtedly the [REST query](https://api.securityscorecards.dev/projects/github.com/ossf-tests/scorecard-action), but since we cannot do that yet, running all scorecard checks for every dependency might be an "expediency"?

azeemshaikh38: Sure, I'm not against running all the checks. Just something to consider to help scale this better at the cost of removing checks that may not be too impactful.

laurentsimon: Is this not something the caller will provide? In CLI, via `--checks`, in the Action, via whatever we decide to do there...?

aidenwang9867: Should I add a `chekcsToRun checker.CheckNameToFnMap` param to the API `GetDependencyDiffResults`? 

such that if users specify `--checks` when calling this API, we can run Scorecards on those checks; otherwise we run all the checks?

dependencydiff/raw_dependencies.go
```diff
@@ -0,0 +1,130 @@
...
+			ChangeType:       d.ChangeType,
+			ManifestPath:     d.ManifestPath,
+			Ecosystem:        d.Ecosystem,
+			Version:          d.Version,
+			Name:             d.Name,
+		}
+		// For now we skip those without source repo urls.
+		// TODO: use the BigQuery dataset to supplement null source repo URLs
+		// so that we can fetch the Scorecard results for them.
+		if d.SourceRepository != nil {
```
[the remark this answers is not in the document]
aidenwang9867: ah yes, forgot to add the "skip removed pkgs" code here, thanks

dependencydiff/raw_dependencies.go
```diff
@@ -0,0 +1,130 @@
...
+			if *d.SourceRepository != "" {
+				// If the srcRepo is valid, run scorecard on this dependency and fetch the result.
+				// TODO: use the Scorecare REST API to retrieve the Scorecard result statelessly.
+				scorecardResult, err := pkg.RunScorecards(
+					ctx,
+					ghRepo,
+					clients.HeadSHA,
```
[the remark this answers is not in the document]
aidenwang9867: np, thanks

dependencydiff/raw_dependencies.go
aidenwang9867: @laurentsimon and I discussed this before. The `dependency` struct here with JSON tags in snakecase (underslines) is used for receiving raw data from the GH API, while the `pkg.DependencyCheckReult` is the exported one with JSON tags in camelcase following the Scorecard repo's naming convention.

dependencydiff/dependencydiff.go
aidenwang9867: sorry for not explaining my idea very clearly.

i think that's possible. i can use the error returned by `RunScorecards` to decide whether:
(1) it succeeds with `err==nil` (the situation when my API returns the result normally), or 
(2) it fails with `err!=nil` (my API returns an empty result for this dependency rather than directly returns and skip all the following dependencies).

dependencydiff/raw_dependencies.go
```diff
@@ -0,0 +1,130 @@
...
+					checks.AllChecks,
+					ghRepoClient,
+					ossFuzzRepoClient,
+					nil,
+					vulnsClient,
+				)
```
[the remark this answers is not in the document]
aidenwang9867: thank u so much! i'll start with the critical checks and may expand it later.

dependencydiff/raw_dependencies.go
```diff
@@ -0,0 +1,180 @@
...
+				checksToRun,
+				ghRepoClient,
+				ossFuzzClient,
+				ciiClient,
+				vulnsClient,
+			)
+			// "err==nil" suggests the run succeeds, so that we record the scorecard check results for this dependency.
+			// Otherwise, it indicates the run fails and we leave the current dependency scorecard result empty
+			// rather than letting the entire API return nil since we still expect results for other dependencies.
+			if err == nil {
```
laurentsimon, line 172: where do you set the error if scorecard returns an error?

spikeb4 (approved):

dependencydiff/raw_dependencies.go
aidenwang9867: can i record the scorecard failing errors into the logger?

dependencydiff/raw_dependencies.go
```diff
@@ -0,0 +1,180 @@
...
+			vulnsClient = clients.DefaultVulnerabilitiesClient()
+		case checks.CheckCIIBestPractices:
+			ciiClient = clients.DefaultCIIBestPracticesClient()
+		}
+	}
+
+	results := []pkg.DependencyCheckResult{}
+	for _, d := range deps {
+		// Skip removed dependencies and don't run scorecard checks on them.
+		if *d.ChangeType == pkg.Removed {
```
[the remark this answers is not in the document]
aidenwang9867: sync from 1:1: adding a new input param to the API, letting the caller decide the scorecard checks to run on which types of dependencies

aidenwang9867:
PTAL at the latest changes @laurentsimon , thanks!

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,162 @@
...
+		case checks.CheckVulnerabilities:
+			dCtx.vulnsClient = clients.DefaultVulnerabilitiesClient()
+		case checks.CheckCIIBestPractices:
+			dCtx.ciiClient = clients.DefaultCIIBestPracticesClient()
+		}
+	}
+	return nil
+}
+
+func initScorecardChecks(checkNames []string) checker.CheckNameToFnMap {
```
laurentsimon, line 106: Can we call the existing function that already exist in scorecard?

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,162 @@
...
+	// Initialize the repo and client(s) corresponding to the checks to run.
+	err = initRepoAndClientByChecks(&dCtx)
+	if err != nil {
+		return nil, fmt.Errorf("error in initRepoAndClientByChecks: %w", err)
+	}
+	getScorecardCheckResults(&dCtx)
+	return dCtx.results, nil
+}
+
+func initRepoAndClientByChecks(dCtx *dependencydiffContext) error {
```
laurentsimon, line 79: I suspect this function is also in scorecard. Can we call it?

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,162 @@
...
+			ChangeType:       d.ChangeType,
+			ManifestPath:     d.ManifestPath,
+			Ecosystem:        d.Ecosystem,
+			Version:          d.Version,
+			Name:             d.Name,
+		}
+		// For now we skip those without source repo urls.
+		// TODO: use the BigQuery dataset to supplement null source repo URLs to fetch the Scorecard results for them.
+		if d.SourceRepository != nil && *d.SourceRepository != "" {
+			if d.ChangeType != nil && (dCtx.changeTypesToCheck[*d.ChangeType] || dCtx.changeTypesToCheck == nil) {
```
laurentsimon, line 135: is it every possible that `dCtx.changeTypesToCheck == nil` happens?

dependencydiff/dependencydiff_test.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/ossf/scorecard/v4/checks"
+	"github.com/ossf/scorecard/v4/clients"
+	"github.com/ossf/scorecard/v4/log"
+	"github.com/ossf/scorecard/v4/pkg"
+)
+
+// Test_fetchRawDependencyDiffData is a test function for fetchRawDependencyDiffData.
+func Test_fetchRawDependencyDiffData(t *testing.T) {
+	t.Parallel()
+	//nolint
```
laurentsimon, line 31: those feel more like e2e tests. Maybe move them to the e2e tests instead.

laurentsimon (approved):
LGTM, thanks. Added a few comments

dependencydiff/dependencydiff.go
aidenwang9867: I was thinking of making this the same as the default logic of the CLI (running all checks if the caller specifies nothing). So if the caller calls my `GetDependencydiffResults` API and give the input map`changeTypesToCheck` a nil, then we'll run checks on all types (added, removed, updated) of dependencies as well

laurentsimon: sorry, I confused `changeTypesToCheck` and `checksToRun` which i set in the function above. Resolving this command

dependencydiff/dependencydiff_test.go
aidenwang9867: Yeah I was also suspicious if these should be a part of the e2e test. Since those functions are called in the main API `GetDependencydiffResult` and the API has already been tested in e2e, maybe I can simply get them removed?

dependencydiff/dependencydiff.go
aidenwang9867: found `policy/GetEnabled` that does the same thing, removed mine, thanks!

dependencydiff/dependencydiff.go
aidenwang9867: Yeah I found `checker/GetClients()` that initalizes the clients, but when it does that, it returns all the clients. I think mine here is better - creating clients depending on which checks to perform :D

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,158 @@
...
+					clients.HeadSHA,
+					checksToRun,
+					dCtx.ghRepoClient,
+					dCtx.ossFuzzClient,
+					dCtx.ciiClient,
+					dCtx.vulnsClient,
+				)
+				// If the run fails, we leave the current dependency scorecard result empty and record the error
+				// rather than letting the entire API return nil since we still expect results for other dependencies.
+				if err != nil {
```
laurentsimon, line 148: we could do:
```
if err != nil {
  depCheckResult.ScorecardResultsWithError.Error = ...
  continue
}

depCheckResult.ScorecardResultsWithError.ScorecardResults = &scorecardResult
```

the caller is responsible for checking error.

dependencydiff/dependencydiff.go
```diff
@@ -0,0 +1,158 @@
...
+					depCheckResult.ScorecardResultsWithError.Error = fmt.Errorf("error running the scorecard checks: %w", err)
```
laurentsimon, line 149: can you create the error using the way we do it in the rest of the code? `sce.WithMessage(sce.ErrScorecardInternal, err.Error())`

dependencydiff/dependencydiff.go
aidenwang9867: if we do a `continue` when the scorecard checking `err != nil`, the `depCheckResult` will be lost, and in the returned result, we won't be able to see this dependency's info

i think we still need to append the depcheck result even if there is an error? the caller can use the `ScorecardResultsWithError` field to check if the scorecard run of this dependency succeeds - it either has a valid result and an empty error or has an empty result and an error.

laurentsimon: > if we do a `continue` when the scorecard checking `err != nil`, the `depCheckResult` will be lost, and in the returned result, we won't be able to see this dependency's info
>
> i think we still need to append the depcheck result even if there is an error? the caller can use the `ScorecardResultsWithError` field to check if the scorecard run of this dependency succeeds - it either has a valid result and an empty error or has an empty result and an error.

ah yes. So 
```
depCheckResult.ScorecardResultsWithError.ScorecardResults = &scorecardResult
// caller is responsible for checking error
if err != nil {
  depCheckResult.ScorecardResultsWithError.Error = ...
}

```

It's a detail anyway. Tell me when you've updated the error creation, and I'll merge.

PR #2057, opened 2022-07-14
✨ Add CODEOWNERS branch protection check

raghavkaul:
Enhance Branch-Protection check to require CODEOWNER review.

This adds 'bonus points' to the Branch-Protection check, but doesn't detract from the score if it's not enabled.

Today, the Branch-Protection check requires that code review must be completed before merge. In GitHub, CODEOWNERS can be assigned to subpaths in a. Without CODEOWNER review, anyone with write access to the repo can approve any PR. In large repos with many modules and many maintainers, CODEOWNERS allows maintainers to delegate code reviews and limit the scope of approval. 

Added a unit test; updated existing unit tests.

```release-note
Scorecard will evaluate whether CODEOWNER review is required for protected branches.
```

### DIFF
--- a/checks/evaluation/branch_protection.go
+++ b/checks/evaluation/branch_protection.go
@@ -30,6 +30,8 @@ const (
 	nonAdminContextLevel        = 2 // Level 3.
 	nonAdminThoroughReviewLevel = 1 // Level 4.
 	adminThoroughReviewLevel    = 1 // Level 5.
+	codeownerReviewLevel        = 1 // Level 5.
+
 )
 
 type scoresInfo struct {
@@ -40,6 +42,7 @@ type scoresInfo struct {
 	context             int
 	thoroughReview      int
 	adminThoroughReview int
+	codeownerReview     int
 }
 
 // Maximum score depending on whether admin token is used.
@@ -76,6 +79,7 @@ func BranchProtection(name string, dl checker.DetailLogger,
 		score.scores.thoroughReview, score.maxes.thoroughReview = nonAdminThoroughReviewProtection(&b, dl)
 		// Do we want this?
 		score.scores.adminThoroughReview, score.maxes.adminThoroughReview = adminThoroughReviewProtection(&b, dl)
+		score.scores.codeownerReview, score.maxes.codeownerReview = codeownersBranchProtection(&b, dl)
 
 		scores = append(scores, score)
 	}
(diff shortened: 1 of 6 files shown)


checks/evaluation/branch_protection.go
```diff
@@ -30,6 +30,8 @@ const (
 	nonAdminContextLevel        = 2 // Level 3.
 	nonAdminThoroughReviewLevel = 1 // Level 4.
 	adminThoroughReviewLevel    = 1 // Level 5.
+	codeownerReviewLevel        = 1 // Level 5.
```
laurentsimon, line 33: this should be a new level if you want a new level. The sum of all the levels should be 10. 

How bout packing in an existing level? 
One issue with adding a level 6 is that it implies a dependence on the previous levels,  meaning it will only be satisfied if all the previous levels are satisfied. 
In practice ,it's hard for us to argue that these levels make sense... we tried to adjust them in a way that sort of does, but it's hard :)

That's why I'm trying to find a place within a previous level to do it. I think we could put it as part of level 4. Level 5 is virtually non-attainable unless you enable force LGTM after each commit in a single PR (which is un-usable)

checks/evaluation/branch_protection.go
```diff
@@ -104,60 +108,76 @@ func BranchProtection(name string, dl checker.DetailLogger,
...
 	score := 0
-	for _, s := range scores {
+	for i := range scores {
+                s := scores[i]
 		score += s.scores.review
 	}
 	return score
 }
 
 func computeAdminReviewScore(scores []levelScore) int {
```
laurentsimon, line 136: these function could be re-written with go generic now that they are supported ;)

laurentsimon (commented):
LGTm, looks great.
I think we just need to try to put the changes as part of an existing "level", see my comment.

Wdut?

checks/evaluation/branch_protection.go
laurentsimon: SGTM.

docs/checks/internal/checks.yaml
```diff
@@ -199,6 +199,7 @@ checks:
       
       Tier 5 Requirements (10/10 points):
         - For administrators: Dismiss stale reviews
```
olivekl, line 201: Minor nit: can you format this so the list hierarchy is clear? Are these both under `For administrator`? If separate, put `dismiss stale review` under line 201 and indented. If both apply to admins, indent the `require CODEOWNER review` as well.

olivekl (commented):
Docs look good to me, with a minor nit for formatting.

olivekl (approved):

naveensrinivasan:
@raghavkaul  The linter check is failing. Could you please fix that? We should be able to merge after that.

naveensrinivasan:
@raghavkaul  A friendly ping!

docs/checks/internal/checks.yaml
raghavkaul: Yes! Fixed.

PR #2077, opened 2022-07-19
✨ Feature [experimental]: The Scorecard Dependencydiff CLI (Version 0 Part 1)

### LINKED ISSUE #2008: Feature: Scorecard Dependency-diff API, CLI and Action

aidenwang9867:
This PR introduces a v0 experimental feature - **the Dependencydiff CLI** powered by the Scorecard API `GetDependencydiffResults` (PR #2046).

Update: This PR (v0-p1) only introduces the option changes, cmd changes are made in PR #

Minor fixes: some potential misleading var names in the GetDependencydiffResults API, such as changing `baseSHA/headSHA` to `base/head`, since we can use branch names as the inputs.

In PR #2046, the v0 of the `GetDependencydiffResults` function API was introduced to the Scorecard repo, but Scorecard hasn't gotten use of it.

Now with the `SCORECARD_EXPERIMENTAL` env var set in system or CLI, users can use **the `dependency-diff` sub-command** to specify two code commits either by commitSHA or branch name, to surface Scorecard check result of those dependencies to better understand the security posture of their repo. This works as a separate sub-command so it won't affect the original root usage mode of Scorecard (running checks on a repo).

A possible CLI usage might be:
```shell
SCORECARD_EXPERIMENTAL=1 \
./scorecard dependency-diff \
--repo=aidenwang9867/go_manifest_test \
--base main \
--head 112997623f24d5f677ff9155f3868964d4b81df3 \
--checks Branch-Protection,Fuzzing,License
```
where:
-  `SCORECARD_EXPERIMENTAL=1` enables the `dependency-diff` sub-command feature
-  `dependencydiff` is the sub-command and shifts the Scorecard from the original check mode to dependencydiff
- `--repo` specifies the repo to be checked (same usage as the original Scorecard CLI)
- `--base` specifies `main` as the BASE, and `--head` specifies `main` (the latest commit to the `112997623f24d5f677ff9155f3868964d4b81df3` branch) as the HEAD.
- `--checks` specifies the checks to run on **the dependencies** (not the repo itself) (same usage format as the original Scorecard CLI)

Issue #2008 

Note the v0 of the CLI runs pretty slow since we are running scorecard checks on every dependency with a valid srcRepo URI. I am considering adding `Parallel()` support for the scorecard running on dependencies, or an option (YES/NO) to ask if the user would like to continue when there are more than, for example, 20+ dependencies to check and tell the user it might be slow.
@naveensrinivasan @azeemshaikh38 @laurentsimon  wdut?

Yes.
the `release-note`

```release-note
Now with the SCORECARD_EXPERIMENTAL env var set in system or CLI, users can use the dependency-diff sub-command to specify two code commits either by commitSHAs or branch names, to surface Scorecard check result of those dependencies to better understand the security posture of their repo.
```


cmd/root.go
```diff
@@ -38,7 +39,7 @@ import (
 const (
 	scorecardLong = "A program that shows security scorecard for an open source software."
 	scorecardUse  = `./scorecard (--repo=<repo> | --local=<folder> | --{npm,pypi,rubygems}=<package_name>)
-	 [--checks=check1,...] [--show-details]`
+	 [--checks=check1,...] [--show-details] [--dependencydiff=<base>...<head>]`
```
laurentsimon, line 42: @azeemsgoogle @naveensrinivasan please chime in about the new option
LGTM, but maybe you had other ideas.

cmd/root.go
```diff
@@ -89,74 +90,97 @@ func rootCmd(o *options.Options) {
...
+		if !strings.EqualFold(o.Commit, clients.HeadSHA) {
+			requiredRequestTypes = append(requiredRequestTypes, checker.CommitBased)
+		}
+		enabledChecks, err := policy.GetEnabled(pol, o.ChecksToRun, requiredRequestTypes)
+		if err != nil {
+			log.Panic(err)
 		}
-	}
 
-	repoResult, err := pkg.RunScorecards(
```
laurentsimon, line 126: Would it be possible for the `GetDependencyDiffResult` API to have similar inputs as:
```
pkg.RunScorecards(
			ctx,
			repoURI,
			o.Commit,
			enabledChecks,
			repoClient,
			ossFuzzRepoClient,
			ciiClient,
			vulnsClient,
```
I mean all the client initialization part.

Looks like we're doing a lot of duplicated work / logic in the `GetDependencyDiffResult`, but it could be shared. Also, consumers like AllStar may want the same interface, since it seems to be working. I think it'd help to be consistent.

Wdut? Is there a blocker?

options/options.go
```diff
@@ -38,15 +38,17 @@ type Options struct {
...
+	Dependencydiff string
 
 	// Feature flags.
-	EnableSarif       bool `env:"ENABLE_SARIF"`
-	EnableScorecardV5 bool `env:"SCORECARD_V5"`
-	EnableScorecardV6 bool `env:"SCORECARD_V6"`
+	EnableSarif          bool `env:"ENABLE_SARIF"`
+	EnableScorecardV5    bool `env:"SCORECARD_V5"`
+	EnableScorecardV6    bool `env:"SCORECARD_V6"`
+	EnableDependencydiff bool `env:"ENABLE_DEPENDENCYDIFF"`
```
laurentsimon, line 51: shall we start using a common `SCORECARD_EXPERIMENTAL` flag? I think this is something that was proposed before to simplify. (Don't rename the other ones, we can start using it for new features only)

@naveensrinivasan @azeemsgoogle Wdut?

pkg/json.go
```diff
@@ -82,6 +83,18 @@ type JSONScorecardResultV2 struct {
 	Metadata       []string            `json:"metadata"`
 }
 
+// JSONDependencydiffResult exports dependency-diff results as JSON for new detail format.
+type JSONDependencydiffResult struct {
+	ChangeType           *ChangeType           `json:"changeType"`
```
laurentsimon, line 88: Let's add a comment + issue to update these fields to be consistent with OSV naming convention.

laurentsimon (commented):
LGTM overall!

Added a few comment

cmd/root.go
aidenwang9867: IMO, the API `GetDependencyDiffResult` is designed for the caller to initialize a list of checks to run, and this should be a string array and is easy for the caller to specify. If we modify the API to have all those clients as input, it'll increase the complexity of using this API - users have to declare those corresponding clients to the checks to run, and we need to validate if the input clients - checks to run are matched.

I do agree that the current design of the client initialization part (see [here](https://github.com/ossf/scorecard/blob/main/dependencydiff/dependencydiff.go#L81-L83)) of `GetDependencyDiffResult` is a little duplicated. What I am doing in the func `initRepoAndClientByChecks` is, firstly I initialize all the clients using the original API `checker.GetClients` used in the scorecard check runner. Then, I return the required clients corresponding to the checks to run. 

Since the [current initialization](https://github.com/ossf/scorecard/blob/096cbd0622cc28ba321f920fd3ab833f5cdf81ab/cmd/root.go#L92-L135) for the scorecard check runner is to simply initialize and use all the clients if a valid `repoURI` is given, even if checks like `Fuzzing` or  `CII-Best-Practices` are not specified to run. Should I optimize the original API using mine?

laurentsimon: > IMO, the API `GetDependencyDiffResult` is designed for the caller to initialize a list of checks to run, this is a string array and should be easy for the caller to specify. If we modify the API to have all those clients as input, it'll increase the complexity of using this API - users have to declare those clients corresponding to the checks to run, and we need to validate if the input `clients` --- `checks to run` are matched (since they are from users).
> 

@jeffmendoza any insights on the API? I'm curious if we should follow the same pattern, since I suspect it must have been discussed before.

cmd/root.go
```diff
@@ -23,22 +23,23 @@ import (
...
+
+	docs "github.com/ossf/scorecard/v4/docs/checks"
+	sclog "github.com/ossf/scorecard/v4/log"
 )
 
 const (
 	scorecardLong = "A program that shows security scorecard for an open source software."
 	scorecardUse  = `./scorecard (--repo=<repo> | --local=<folder> | --{npm,pypi,rubygems}=<package_name>)
-	 [--checks=check1,...] [--show-details]`
+	 [--checks=check1,...] [--show-details] [--dependencydiff=<base>...<head>]`
```
azeemshaikh38, line 42: Any reason we decided to do this as a flag instead of sub-command?

options/options.go
azeemshaikh38: Yes, please. Let's not add more env vars to gate experimental features.

options/options.go
```diff
@@ -85,6 +87,8 @@ const (
 	FormatDefault = "default"
 	// FormatRaw specifies that results should be output in raw format.
 	FormatRaw = "raw"
+	// FormatMarkdown specifies that results should be output in Markdown format.
+	FormatMarkdown = "markdown"
```
azeemshaikh38, line 91: Any discussion where we decided to add `markdown` format as a supported format? Also, maybe keep the support for adding `markdown` separate from this PR?

options/options.go
```diff
@@ -177,6 +197,17 @@ func (o *Options) Validate() error {
...
+	if o.Dependencydiff == "" && o.Format == FormatMarkdown {
+		errs = append(
+			errs,
+			fmt.Errorf(
+				"%w: markdown can be only used for the --dependencydiff option",
+				errFormatNotSupported,
+			),
+		)
+	}
+
```
azeemshaikh38, line 210: More such validation will be needed if we support this feature as a flag (like`--depdiff` and `--local` can't be used together). Instead a sub-command is cleaner.

pkg/dependencydiff_result.go
```diff
@@ -80,10 +85,100 @@ type DependencyCheckResult struct {
...
+				"invalid format flag: %v. Expected [default, json, markdown]",
+				opts.Format,
+			),
+		)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to output dependencydiff results: %w", err)
+	}
+	return nil
+}
```
azeemshaikh38, line 116: Repeating my question from the design doc - is it worth adding support for and maintaining so many output formats for `depdiff`? Couldn't we instead assume JSON as the default output and skip the `--format` flag altogether?

cmd/root.go
laurentsimon: that was not discussed, and I wanted some feedback on that. Thanks

options/options.go
laurentsimon: I'm also in favor of this being treated separately, maybe in the Action itself.

pkg/dependencydiff_result.go
laurentsimon: I also like not having a ton of options to support :)

aidenwang9867: For the default format, my intention is to give the users a table like the default scorecard check results, this way it might be easier for users to read the results (and JSON might be more friendly to programming calls).

The markdown is prepared for the Action to use. I agree it could be a bit redundant here.

cmd/root.go
aidenwang9867: @azeemshaikh38 @laurentsimon @naveensrinivasan 
sync from chat:
justifications of using a sub-command:

(1) it's easier to maintain redundant flag names than validation code for all different combinations of flags

(2) although we don't have sub-commands for Scorecard now, cobra has support for it and we can easily add a scorecard dependencydiff to separate the original Scorecard running mode (running checks on the current repo) and the dependencydiff running mode. (https://cobra.dev/)

(3) it doesn't affect the reuse of other flags, such as re-using --repo to get the repo. those still work for sub-commands

options/options.go
aidenwang9867: sync from chat: will remove the markdown support for now, and may add it to the Action in the future

pkg/dependencydiff_result.go
aidenwang9867: about the supported output format:

removing the markdown support, which is an unnecessary one for now

we don't have doubts on JSON, and it could be the default output format for `dependencydiff`.

For the default format (the plaintext table one), should we still leave it for users to use?

my justification: 
Since users can specify the base and head using the branch names such as `main...dev` to see dep changes in the dev branch (the API will handle the branch name and take the head commit of that branch to use), a user might want to use this CLI in the terminal and see the default result which is more human-readable than JSON.

cons of introducing this extra format:
the overhead maintenance: Once it's added someone will depend on it and we'll have to maintain it in the long run

Is my justification a strong reason? @laurentsimon @naveensrinivasan  wdut?

laurentsimon: I don't fully follow the reasoning. My 2ct: JSON is easy to read, especially when piped into `jq` you have colors. The table always has problem due to terminal size. Nobody has been able to fix it till now.

aidenwang9867: Thanks for your opinions! I'm inclined to only preserve JSON as the output format in v0, and we might discuss about this later whether we would add more format support to it.

pkg/json.go
```diff
@@ -168,6 +181,74 @@ func (r *ScorecardResult) AsJSON2(showDetails bool,
 	if err := encoder.Encode(out); err != nil {
 		return sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("encoder.Encode: %v", err))
 	}
+	return nil
+}
 
+// DependencydiffResultsAsJSON exports dependencydiff results as JSON.
+func DependencydiffResultsAsJSON(depdiffResults []DependencyCheckResult,
```
laurentsimon, line 188: can we make this a "class" function, like we do above for `ScorecardResult`?

dependencydiff/dependencydiff_test.go
```diff
@@ -40,8 +40,8 @@ func Test_fetchRawDependencyDiffData(t *testing.T) {
 				ctx:       context.Background(),
 				ownerName: "no_such_owner",
```
laurentsimon, line 41: in a follow-up PR, it would be useful to have better tests for getScorecardCheckResults(). There's a bunch of logic in there that would be nice to have tests for. It may require some code-refactoring.

cmd/root.go
jeffmendoza: Allstar needs to pass in a client/transport to handle authentication on its own. This is for when running scorecard checks as a library. I'm not sure we would plan to use this feature from Allstar directly.

aidenwang9867:
a test comment

pkg/json.go
aidenwang9867: > can we make this a "class" function, like we do above for `ScorecardResult`?

my initial idea is to make this func an oop-like one, but for the original scorecard run, we only return a single `ScorecardRusult`; while for this one, we return an array of `DependncydiffResult`s. 

then i just defined it this way since the `AsJSON` method does not seem to adapt to a single `DependncydiffResult`, and when we want the dependency checking results, we want an array of them (the CLI, the Action, etc.). Thus, IMO, this func should not be a class func and simply takes an array of  `DependncydiffResult` as the input.

aidenwang9867: > > can we make this a "class" function, like we do above for `ScorecardResult`?
> 
> my initial idea is to make this func an oop-like one, but for the original scorecard run, we only return a single `ScorecardRusult`; while for this one, we return an array of `DependncydiffResult`s.
> 
> then i just defined it this way since the `AsJSON` method does not seem to adapt to a single `DependncydiffResult`, and when we want the dependency checking results, we want an array of them (the CLI, the Action, etc.). Thus, IMO, this func should not be a class func and simply takes an array of `DependncydiffResult` as the input.

oh finally this comment is not pending anymore :/

options/flags.go
```diff
@@ -84,11 +84,14 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
...
-	cmd.Flags().StringVar(
+	cmd.PersistentFlags().StringVar(
 		&o.Commit,
 		FlagCommit,
 		o.Commit,
-		"commit to analyze",
+		"For the default scorecard run, this specifies a code commit to analyze. "+
+			`For dependencydiff, this includes the two commits BASE and HEAD, use "..." to separate them. `+
+			`Both commitSHAs (commit_A_SHA...commit_B_SHA) or branch names ("main...dev") or a mix of them `+
+			`(main...commit_A_SHA) are supported. Please don't use the default value for dependencydiff usage.`,
```
aidenwang9867, line 94: I was considering not using the `--commit` flag to take the `base...head` dependnecy-diff input, since it is used for taking exactly one commitSHA used for the original Scorecard check; otherwise, it could be ambiguous if we use it differently in the root command and the `dependency-diff` sub-command.

Instead, maybe we could create two sub-command flags `--base` and `--head` for `dependency-diff` to take the base and head inputs. wdut? @laurentsimon @azeemshaikh38

aidenwang9867: changed the flag usage: using the global flag `--commit` to receive `base...head` --> using two sub-command flags `--base` for `base` and `--head` for `head`

options/flags.go
```diff
@@ -59,6 +59,12 @@ const (
 
 	// FlagFormat is the flag name for specifying output format.
 	FlagFormat = "format"
+
+	// FlagBase is the flag name for specifying a dependency-diff base.
+	FlagBase = "base"
```
laurentsimon, line 64: I'm curious if base and head make sense to a user. Is this a common naming convention if bitbucket and gitlab as well?

Is there better term for these otherwise (`start-commit`, `end-commit`, etc)

@olivekl may have some advice

aidenwang9867: > I'm curious if base and head make sense to a user. Is this a common naming convention in bitbucket and gitlab as well?
> 
> Are there better terms for these otherwise (`start-commit`, `end-commit`, etc)
> 
> @olivekl may have some advice

Thanks. I've considered using some flag like `--ref` to receive the base and head. It is clear for receiving a branch name reference, but is `commitSHA` also a `ref`? I'm not sure if `--ref` will also lead to ambiguity.

Or we just create two flags `--ref` and `--sha` so users can use either of them.

options/options.go
```diff
@@ -39,14 +39,17 @@ type Options struct {
 	PolicyFile string
 	// TODO(action): Add logic for writing results to file
 	ResultsFile string
+	Base        string
+	Head        string
```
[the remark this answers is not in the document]
aidenwang9867: Thanks, this makes sense to me. I'll use a new `DependencyDiffOptions`.

options/flags.go
aidenwang9867: sync from the bi-weekly meeting:

`--base` and `--head` don't have the ambiguity of whether using a `reference` or using a `commitSHA`, and might be good enough for users to correctly use them if there is a clear document. 

Here's my doc in the `--help` option:
`--base: the base branch name reference or the base commitSHA to check. valid input examples: 
			"main" (branch name), "SHA_VALUE" (commitSHA)`,
			
`--head: the head branch name reference or the head commitSHA to check. valid input examples: 
			"dev" (branch name), "SHA_VALUE" (commitSHA)`,
			
Using "SHA_VALUE" to represent a hash string since it's dangerous to expose a real SHA in the doc.

laurentsimon: is `base` and `head` a GitHub concept or is it universally used across other platforms like GitLab, BitBuckets, etc?

aidenwang9867: I searched `base` and `head` for `GitLab` and `BitBuckets` online,

https://gitlab.com/gitlab-org/gitlab/-/issues/209071
https://stackoverflow.com/questions/53889502/base-vs-default-branch-in-bitbucket-version-of-git
https://stackoverflow.com/questions/2304087/what-is-head-in-git

found some issues and stackoverflow answers where the developers use BASE and HEAD to refer to code commits and branches, but i didn't find an official document using these two phrases.

olivekl: Are the terms being used as they normally would be in Git? If so, I think it's safe to assume they're well known enough and not platform dependent.

aidenwang9867: > Are the terms being used as they normally would be in Git? If so, I think it's safe to assume they're well known enough and not platform dependent.

Yes, `base` and `head` are normally used in Git, and the [GitHub Dependency Review doc](https://docs.github.com/en/rest/dependency-graph/dependency-review#get-a-diff-of-the-dependencies-between-commits) also makes these two explicit.

aidenwang9867: resolving this conversation

options/depdiff_flags.go
```diff
@@ -0,0 +1,57 @@
...
+		FlagHead,
+		depOptions.Head,
+		`The head code branch name or the head commitSHA to check. Valid input examples: 
+		dev (using a branch name), SHA_VALUE_2 (using a commitSHA)`,
+	)
+	cmd.Flags().StringSliceVar(
+		&depOptions.ChangeTypes,
+		FlagChangeTypes,
+		depOptions.ChangeTypes,
+		`Dependency change types for surfacing the scorecard results. This is not a required
```
laurentsimon, line 54: say it's a list. Also, state the default value

options/depdiff_options.go
```diff
@@ -0,0 +1,92 @@
...
+// NewDepdiff creates a new instance of `DependencydiffOptions`.
+func NewDepdiff() *DependencydiffOptions {
+	depdiffOpts := &DependencydiffOptions{}
+	// No need to do the env.Parse() for a dependency-diff option since there
+	// are no struct env tags for now.
+	return depdiffOpts
+}
+
+// ValidateDepdiff validates scorecard dependency-diff configuration options.
+func (depOptions *DependencydiffOptions) ValidateDepdiff() error {
```
laurentsimon, line 53: can we just call it `validate()`? The class name already implies that it validate deps options?

aidenwang9867: Makes sense, thanks. I'll change this after the meeting.

options/depdiff_flags.go
aidenwang9867: done, thanks

naveensrinivasan:
@laurentsimon @azeemshaikh38  What is pending in this PR?

naveensrinivasan:
@aidenwang9867 Are you still planning to work on this?

naveensrinivasan:
Closing this PR as it has been stale for a while. Please reopen.

PR #2090, opened 2022-07-22
✨ Feature: Dependency-diff API optimize: var re-naming, removing unused JSON tags

aidenwang9867:

1. `GetDependencydiffResults` API (PR #2046) input param optimization (renaming ambiguous ones).

2. Remove unnecessary struct JSON tags for `DependencyCheckResult`: will define a `JSONDependencyCheckResult` struct with JSON tags in PR #2077.

Var names such as `baseSHA` and `headSHA` might be confusing and ambiguous since users can use either SHAs or branch names as the inputs; changing them to `base` and `head`. There are some other minor changes on var names.

Struct JSON tags for `DependencyCheckResult` are unnecessary since we will define a `JSONDependencyCheckResult` struct with JSON tags in PR #2077 to output JSON. (removing these tags)

the `release-note`

```release-note
NONE
```


laurentsimon (approved):
LGTM, thanks

pkg/dependencydiff_result.go
```diff
@@ -44,40 +44,41 @@ func (ct *ChangeType) IsValid() bool {
...
 	// Error is an error returned when running the scorecard checks. A nil Error indicates the run succeeded.
-	Error error `json:"scorecardRunTimeError"`
+	Error error
 }
 
 // DependencyCheckResult is the dependency structure used in the returned results.
 type DependencyCheckResult struct {
 	// ChangeType indicates whether the dependency is added, updated, or removed.
-	ChangeType *ChangeType `json:"changeType"`
+	ChangeType *ChangeType
```
laurentsimon, line 60: I'm curious why the JSON fields are no longer need

aidenwang9867: We'll have a `JSONDependencyCheckReuslt` in PR #2077 to output JSON, like `ScorecardResult` and `JSONScorecardResultV2`, so IMO, we don't need json tags for `DependencyCheckResult`, is that right?

laurentsimon: ah, you mean this struct is copied into another one. Got it.

PR #2104, opened 2022-07-28
✨ Enhancement: letting the ecosystem naming convention mapping log the missing entry info rather than returning an error, typo fixes

aidenwang9867:
1. typo fixes
2. Letting the ecosystem naming convention mapping module no longer return an error. Instead, log the info when no mapping entries are found for an ecosystem name, so that missing one mapping won't cause the others to fail.

If no mapping entries are found for an ecosystem name, it returns an error and terminates the subsequent mapping execution.

The module logs the "mapping not found for dependency xxx" info instead of directly returning an error.

No.

NONE
```release-note
NONE
```

### DIFF
--- a/dependencydiff/mapping.go
+++ b/dependencydiff/mapping.go
@@ -15,7 +15,7 @@
 package dependencydiff
 
 import (
-	"fmt"
+	sclog "github.com/ossf/scorecard/v4/log"
 )
 
 // Ecosystem is a package ecosystem supported by OSV, GitHub, etc.
@@ -52,7 +52,7 @@ const (
 	ecosystemMaven ecosystem = "Maven"
 
 	// The NuGet package ecosystem.
-	ecosystemNuGet ecosystem = "Nuget"
+	ecosystemNuGet ecosystem = "NuGet"
 
 	// The Linux kernel.
 	ecosystemLinux ecosystem = "Linux" // nolint:unused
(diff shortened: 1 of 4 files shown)


dependencydiff/mapping.go
```diff
@@ -81,9 +81,23 @@ var (
 	}
 )
 
-func toEcosystem(e string) (ecosystem, error) {
-	if ecosystemOSV, found := gitHubToOSV[e]; found {
-		return ecosystemOSV, nil
+func mapDependencyEcosystemNaming(logger *sclog.Logger, deps []dependency) {
+	for i := range deps {
+		if deps[i].Ecosystem == nil {
+			logger.Info("dependency %s has a nil ecosystem", deps[i].Name)
```
laurentsimon, line 87: what happens when we ignore the ecosystem. I mean, what will the final result data be missing?

aidenwang9867: If we return an error when an GH ecosystem cannot be mapped to an OSV one: we'll have a nil `DependencyCheckResult` and an error `cannot find the mapping`. That said, we won't have any valid results if one ecosystem is missing its mapping. My concern is, when GH adds a new ecosystem naming and we haven't had the map updated in Scorecard yet, this API will fail and return nil when that new ecosystem shows in the raw results.

If we ignore the ecosystem mapping missing: we'll still have the final `DependencyCheckResult`, but the ecosystems might have different naming convention - some are mapped to OSV, those missing a mapping are still GH. We'll see the mapping missing information in the log though.

aidenwang9867: having the info in log: do `logger.Info("naming convention mapping not found for dependency xxx")`

laurentsimon: Info() is for positive things. Any logging for this should be Debug(). We have to decide if we want the check to return an error or not. I'm fine with the entire check failing: we want to know this problem is happening; and I doubt the names will change often.

aidenwang9867: > Info() is for positive things. Any logging for this should be Debug(). 
Re: Info() and Debug(): I see, thanks.

> We have to decide if we want the check to return an error or not. I'm fine with the entire check failing: we want to know this problem is happening; and I doubt the names will change often.

I agree with both: (1) we want to know if this problem is happening and (2) the names might change often, and I haven't gotten a better idea to let the caller know this problem is happening - returning an error solves this and is easy. 

If we keep returning an error for it as it is now, we do need timely and maybe frequent updates on that hashmap if a new GH name comes out; otherwise, the `dependency-diff` will keep failing because of lack of naming convention mapping entries.

aidenwang9867: I don't have better or stronger justifications to make it not return an error - so maybe let's leave it as it is?

I'll create another PR to log it into the logger and fix the typo.

Another issue is, GH has an ecosystem `actions` but OSV doesn't. For this, should I create a map entry `"actions": "actions"`?

laurentsimon: I'm in favor of returning an error. These names are not going to be changing all the time. They need backward compatibility so they can only *add* names. They support so many already that I think it's a small risk.

aidenwang9867: > I'm in favor of returning an error. These names are not going to be changing all the time. They need backward compatibility so they can only _add_ names. They support so many already that I think it's a small risk.

I see, thanks. Resolving this conversation & closing the PR.

aidenwang9867:
after discussion: keep returning an error as it is now
will send another one to fix the typo

PR #2109, opened 2022-07-29
✨ Enhancement: adding new entries for GH actions & Pub as ecosystems, typo fixes

aidenwang9867:
1. Adding entry for GH actions as an ecosystem
2. Ecosystem naming convention mapping type fixes

The GitHub ecosystem `actions` (for the GitHub Action) is not an ecosystem defined by OSV. Missing this special entry in the convention map will lead to the entire `GetDependencyDiffCheckResults` API to fail.

Adding the "action": "action" map entry for it not to fail.

No.

the `release-note`

```release-note
NONE
```


laurentsimon (approved):
LGTM.

dependencydiff/mapping.go
```diff
@@ -78,12 +81,35 @@ var (
 		"composer": ecosystemPackagist,
 		"rubygems": ecosystemRubyGems,
 		"nuget":    ecosystemNuGet,
+		"actions":  ecosystemActions, /* created this entry for the "actions" not to fail in the mapping process. */
 	}
 )
 
+func mapDependencyEcosystemNaming(deps []dependency) error {
+	for i := range deps {
+		if deps[i].Ecosystem == nil {
```
laurentsimon, line 90: why does this not return an error too?

aidenwang9867: hmm, since we allow the package ecosystem field to be null-able?

laurentsimon: ah, GH does not always populate it?
Let's add comments / notes to the code to explain.

aidenwang9867: > ah, GH does not always populate it? Let's add comments / notes to the code to explain.

for the GitHub API as the data source, the `ecosystem` is a required field to return - but in our previous discussion we'd like to make this more generale and not limited by an specific API, so only we only made the `name` field a `string` (required).

i'll definitely add a comment here to explain

PR #2112, opened 2022-07-31
🐛 Fix a bug in cron API data exporting

### LINKED ISSUE #1056: Feature - Web API for Scorecard bigquery data
**Is your feature request related to a problem?
The scorecard bigquery is accessible only via bigquery client/API. This is an issue in consuming the scorecard data.

Tools like `osv` are great because it has a REST endpoint https://osv.dev/docs/#section/Getting-Started and it can be consumed by other tools without having to jump through hoops. 

 For example, I was able to build https://github.com/naveensrinivasan/stunning-tribble to validate any new `go` dependencies against `osv` with the help of the rest API and not needing a google cloud account.  This brings an advantage of now new dependencies are checked against for OSV's (let's set aside the existing dependencies can have new OSV/CVE which we have to validate as a cron job and also this is an option https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/about-alerts-for-vulnerable-dependencies#detection-of-vulnerable-dependencies)

 HTTP is the standard for querying data.  Now with `150k` repositories, we will have collected significant data and it would be best only if it is accessible as an interface that the community wants in the format.

Provide a REST endpoint over Scorecard data that can be retrieved from the BigQuery and return as `JSON`

azeemshaikh38:

The current cron job writes its data to the releasetest bucket and also uses the wrong filename. This PR fixes that and fixes the naming convention used.

#1056 

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cron/internal/config/config.go
+++ b/cron/internal/config/config.go
@@ -79,7 +79,7 @@ type config struct {
 	// Raw results.
 	RawResultDataBucketURL string `yaml:"raw-result-data-bucket-url"`
 	RawBigQueryTable       string `yaml:"raw-bigquery-table"`
-	ExportResultsBucketURL string `yaml:"export-results-bucket-url"`
+	APIResultsBucketURL    string `yaml:"api-results-bucket-url"`
 }
 
 func getParsedConfigFromFile(byteValue []byte) (config, error) {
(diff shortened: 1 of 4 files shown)


naveensrinivasan (approved):
Thank you!

Issue #2085, opened 2022-07-21
Feature: Detect Sonar as SAST via pom.xml


edwardsph:
Can you add support for detecting the use of SonarQube as well as SonarCloud. This would tend to be used on private repositories when people run Sonar in-house. For Java projects you would see this running as part of the Maven build or similar.

naveensrinivasan:
Thanks, Would you like to a PR for this?

edwardsph:
I had a look at the code and wasn't sure what is was doing - I don't know Go

laurentsimon:
@edwardsph is the setup using a config file committed to a repository?
Do you use https://github.com/SonarSource/sonarqube-scan-action?

The only way scorecard can detect it is if _something_ is visible in the repository: files, directories, settings, etc

If it's only on a developer's machine, scorecard won't be able to detect it.

Can you tell us more how it works?

edwardsph:
We don't use the action. The Sonar analysis is run within the Maven build during the CI/CD process so there is a GitHub workflow which runs `./mvnw sonar:sonar`. The other evidence will be properties in the POM file such as `sonar.host.url`. Can that be used to detect Sonar?

edwardsph:
Here is a link to a POM on an open source repo using SonarCloud: https://github.com/solid-contrib/conformance-test-harness/blob/main/pom.xml - it has properties:
```
        <sonar.organization>solid-contrib</sonar.organization>
        <sonar.host.url>https://sonarcloud.io</sonar.host.url>
```
The `pom.xml` is expected to be in the project root.
For a SonarQube example here are examples of the properties in use (I can't link as it is a private repo):
```
    <sonar.coverage.jacoco.xmlReportPaths>target/jacoco-report/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
    <sonar.host.url>https://sonarqube.private.domain</sonar.host.url>
    <sonar.projectKey>${projectKey}</sonar.projectKey>
    <sonar.moduleKey>${project.artifactId}</sonar.moduleKey>
```

laurentsimon:
I've send https://github.com/ossf/scorecard/pull/2114

spencerschrock:
Closing as this PR was merged. Re-open if it wasn't addressed.

PR #2114, opened 2022-08-01
feat: Add pom.xml support for sonarype SAST

laurentsimon:
See https://github.com/ossf/scorecard/issues/2085

/cc @edwardsph

### DIFF
--- a/checks/sast.go
+++ b/checks/sast.go
@@ -245,6 +245,7 @@ func sonarEnabled(c *checker.CheckRequest) (int, error) {
 			Offset:    result.file.Offset,
 			EndOffset: result.file.EndOffset,
 			Text:      "Sonar configuration detected",
+			Snippet:   result.url,
 		})
 	}
 
@@ -260,13 +261,13 @@ var validateSonarConfig fileparser.DoWhileTrueOnFileContent = func(pathfn string
 	content []byte,
 	args ...interface{},
 ) (bool, error) {
-	if path.Base(pathfn) != "pom.xml" {
+	if !strings.EqualFold(path.Base(pathfn), "pom.xml") {
 		return true, nil
 	}
 
 	if len(args) != 1 {
 		return false, fmt.Errorf(
-			"validateSonarConfig requires exactly 1 arguments: %w", errInvalid)
+			"validateSonarConfig requires exactly 1 argument: %w", errInvalid)
 	}
 
 	// Verify the type of the data.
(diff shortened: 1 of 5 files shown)


checks/sast.go
```diff
@@ -205,3 +222,106 @@ func codeQLInCheckDefinitions(c *checker.CheckRequest) (int, error) {
...
+			Type:      checker.FileTypeSource,
+			Offset:    offset,
+			EndOffset: endOffset,
+		},
+	})
+
+	return true, nil
+}
+
+func findLine(content, data []byte) (uint, error) {
```
azeemshaikh38, line 309: What if there are multiple matching lines?

azeemshaikh38 (approved):
LGTM with comments.

checks/sast.go
```diff
@@ -205,3 +222,106 @@ func codeQLInCheckDefinitions(c *checker.CheckRequest) (int, error) {
...
+	content []byte,
+	args ...interface{},
+) (bool, error) {
+	if path.Base(pathfn) != "pom.xml" {
+		return true, nil
+	}
+
+	if len(args) != 1 {
+		return false, fmt.Errorf(
+			"validateSonarConfig requires exactly 1 arguments: %w", errInvalid)
```
aidenwang9867, line 269: nit: argument

checks/sast.go
```diff
@@ -205,3 +222,106 @@ func codeQLInCheckDefinitions(c *checker.CheckRequest) (int, error) {
...
+	}
+
+	// Verify the type of the data.
+	pdata, ok := args[0].(*[]sonarConfig)
+	if !ok {
+		return false, fmt.Errorf(
+			"validateSonarConfig expects arg[0] of type *[]sonarConfig]: %w", errInvalid)
+	}
+
+	regex := regexp.MustCompile(`<sonar\.host\.url>\s*(https:\/\/.*)\s*<\/sonar\.host\.url>`)
```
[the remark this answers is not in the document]
aidenwang9867: Found the default value of the host.url starts with `http://`:
https://docs.sonarqube.org/latest/analysis/analysis-parameters/

so I agree the regex could be more general ;)

checks/sast.go
laurentsimon: good catch

checks/sast.go
laurentsimon: For now it does not matter so I kept it simple. We get the first it's probably enough. I don't know if it's allowed to have multiple ones.

checks/sast.go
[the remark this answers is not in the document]
laurentsimon: I forgot to ignore https. That was in the back of my mind.. and I forgot :-)
Thanks

checks/sast.go
```diff
@@ -205,3 +222,106 @@ func codeQLInCheckDefinitions(c *checker.CheckRequest) (int, error) {
...
+
+	return checker.MinResultScore, nil
+}
+
+// Check file content.
+var validateSonarConfig fileparser.DoWhileTrueOnFileContent = func(pathfn string,
+	content []byte,
+	args ...interface{},
+) (bool, error) {
+	if path.Base(pathfn) != "pom.xml" {
```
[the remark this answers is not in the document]
laurentsimon: can do that. I thought it was overkill, but it
s cheap. Will do.

edwardsph:
Thanks for the quick change - looking forward to testing in the next release

PR #2131, opened 2022-08-08
✨  support for SLSA provenance in Signed-Release

laurentsimon:
Search for `.intoto.jsonl` in the release check.
```release-notes
Search for `.intoto.jsonl` in the release check.
```

### DIFF
--- a/checks/signed_releases_test.go
+++ b/checks/signed_releases_test.go
@@ -103,7 +103,7 @@ func TestSignedRelease(t *testing.T) {
 					TargetCommitish: "master",
 					Assets: []clients.ReleaseAsset{
 						{
-							Name: "foo.asc",
+							Name: "foo.intoto.jsonl",
 							URL:  "http://foo.com/v1.0.0/foo.intoto.jsonl",
 						},
 					},
--- a/checks/signed_releases_test.go
+++ b/checks/signed_releases_test.go
@@ -95,6 +95,25 @@ func TestSignedRelease(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


checks/signed_releases_test.go
```diff
@@ -94,6 +94,25 @@ func TestSignedRelease(t *testing.T) {
...
+		{
+			name: "Releases with assests with intoto SLSA provenance",
+			releases: []clients.Release{
+				{
+					TagName:         "v1.0.0",
+					URL:             "http://foo.com/v1.0.0",
+					TargetCommitish: "master",
+					Assets: []clients.ReleaseAsset{
+						{
+							Name: "foo.asc",
```
sethmlarson, line 106: Shouldn't the name be `foo.intoto.jsonl` as well?

laurentsimon: well, ideally yes. But it's possible to have one attestation attesting to multiple artifacts. I know it's not super accurate. The proper way is to verify it, we can try that in a follow-up PR to beef it up.

sethmlarson: I think this test case isn't testing the right thing, from [this line](https://github.com/ossf/scorecard/pull/2131/files#diff-ec7162da628e8587757286bab0848563d1be93f39b0e33facc307792d27ffe94L53) it looks like the `asset.Name` is what is being checked for the suffix, not the URL. Given this I believe this test case is passing only because the `Name` is `*.asc`, not due to `.intoto.jsonl`?

laurentsimon: +1, you're correct. Updated. Thanks for catching.

PR #2134, opened 2022-08-09
🌱 Limit access to registered checks

spencerschrock:
Limit access to registered checks except through `checks.GetAll()` and `checks.GetAllForEnvironment()`. 
The intention is to provide two versions, one which returns everything, and one which excludes experimental features (unless enabled via environment flags). This was causing problems in #2133, as WebHooks was being run by default.

`checks.AllChecks` and `checks.GetAll()` were both public and being used.
`checks.GetAll()` did not filter on environment variable constraints, despite its comment saying it did

`checks.allChecks` made private
`checks.GetAll()` had its comment updated to reflect the behavior
`checks.GetAllForEnvironment()` introduced to filter the checks to those supported by the environment variables.

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/checks/all_checks.go
+++ b/checks/all_checks.go
@@ -16,18 +16,43 @@
 package checks
 
 import (
+	"os"
+
 	"github.com/ossf/scorecard/v4/checker"
 )
 
-// AllChecks is the list of all security checks that will be run.
-var AllChecks = checker.CheckNameToFnMap{}
+// allChecks is the list of all registered security checks
+var allChecks = checker.CheckNameToFnMap{}
 
-// GetAll returns the full list of checks, given any environment variable
-// constraints.
-// TODO(checks): Is this actually necessary given `AllChecks` exists?
-func GetAll() checker.CheckNameToFnMap {
-	possibleChecks := AllChecks
+func getAll(overrideExperimental bool) checker.CheckNameToFnMap {
+	// need to make a copy or caller could mutate original map
+	possibleChecks := checker.CheckNameToFnMap{}
+	for k, v := range allChecks {
+		possibleChecks[k] = v
+	}
+
+	if overrideExperimental {
+		return possibleChecks
+	}
+
+	// TODO: remove this check when v6 is released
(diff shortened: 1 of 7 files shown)


laurentsimon (approved):

checks/all_checks.go
```diff
@@ -16,18 +16,43 @@
...
+	for k, v := range allChecks {
+		possibleChecks[k] = v
+	}
+
+	if overrideExperimental {
+		return possibleChecks
+	}
+
+	// TODO: remove this check when v6 is released
+	if _, v6 := os.LookupEnv("SCORECARD_V6"); !v6 {
```
azeemshaikh38, line 39: This PR might be a good chance for us to move to `SCORECARD_EXPERIMENTAL` instead of V6. If its not a lot of work, would you mind changing that? I don't think it should break any users.

checks/all_checks.go
```diff
@@ -16,18 +16,43 @@
...
 
-// GetAll returns the full list of checks, given any environment variable
+	return possibleChecks
+}
+
+// GetAllForEnvironment returns the full list of checks, given any environment variable
 // constraints.
-// TODO(checks): Is this actually necessary given `AllChecks` exists?
+func GetAllForEnvironment() checker.CheckNameToFnMap {
+	return getAll(false /*overrideExperimental*/)
```
azeemshaikh38, line 49: Nit: change fn names to `GetAllWithExperimental` and `GetAll`. Wdyt?

azeemshaikh38 (approved):
Thanks! LGTM with comments

checks/all_checks.go
spencerschrock: Sure

checks/all_checks.go
spencerschrock: I played around with a few naming ideas, including those two. 
Personally, it slightly felt off to have a function called `GetAll` which doesn't actually return everything.
But that's just how it felt, if this is a normal pattern I'm fine with adopting those names

PR #2151, opened 2022-08-16
🐛 Detect recently created Github repositories

### LINKED ISSUE #2150: BUG: Check for created_since in Maintained check
Given the recent [malicious forks attack](https://www.bleepingcomputer.com/news/security/35-000-code-repos-not-hacked-but-clones-flood-github-to-serve-malware/), one idea that came up was to use `created_since` signal and use it in the `Maintained` check. What we could do is, if a repo was created in the last X days (say 90), dock some score and add a `warn` detail to the logger. Else the behavior of the check remains the same.

raghavkaul:
Recently, Github repos that are forks of popular projects have been observed to contain malware and deceptively edited Git histories. This changeset adjusts the the score of a Github repo downward if it was created in the last 90 days with the aim of detecting when such repos are created without enough time to determine whether they are being actively maintained or not.

Today, Github repo age is not used to compute scorecard `Maintained` scores.

* Tests added for score change
* No tests for errors/local repos yet

Fixes #2150

```release-note
Repositories created in the last 90 days now score lower on the `Maintained` check.
```


checks/evaluation/maintained.go
```diff
@@ -49,6 +50,14 @@ func Maintained(name string, dl checker.DetailLogger, r *checker.MaintainedData)
 		}
 	}
 
+	// Emit a warning if this repo was created recently and is a fork
```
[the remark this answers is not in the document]
azeemshaikh38: +1

checks/evaluation/maintained.go
```diff
@@ -49,6 +50,14 @@ func Maintained(name string, dl checker.DetailLogger, r *checker.MaintainedData)
 		}
 	}
 
+	// Emit a warning if this repo was created recently
+	recencyThreshold := time.Now().AddDate(0 /*years*/, 0 /*months*/, -1*lookBackDays /*days*/)
+	weight := 0
+	if r.CreatedAt.After(recencyThreshold) {
+		dl.Warn(&checker.LogMessage{Text: fmt.Sprintf("repo was created in the last 90 days (Created at: %s)", r.CreatedAt.Format(time.RFC3339))})
```
inferno-chromium, line 57: repo was recently created in last 90 days, please evaluate content carefully

checks/evaluation/maintained.go
```diff
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	lookBackDays    = 90
-	activityPerWeek = 1
-	daysInOneWeek   = 7
+	lookBackDays     = 90
+	activityPerWeek  = 1
+	daysInOneWeek    = 7
+	newProjectWeight = -3
```
inferno-chromium, line 30: We should zero out the score if less than 90 days, not just -3. We can't trust if earlier than 90 days.

naveensrinivasan: >We should zero out the score if less than 90 days, not just -3. We can't trust if earlier than 90 days.

Why zero out the score? This means any new repository would get `0` score.

inferno-chromium: > > We should zero out the score if less than 90 days, not just -3. We can't trust if earlier than 90 days.
> 
> Why zero out the score? This means any new repository would get `0` score.

Yes that should be expected. We can't trust any new repository from day 1, its content needs to be evaluated and it needs to take time to build trust. Also, this will be just one check failure, rest everything should pass, so upto consumer to decide.

inferno-chromium:
You also need to update README.md with updated check description. Like maintained check looks for well maintained repos, exceeding 90 days minimum or something like that.

checks/evaluation/maintained.go
raghavkaul: Updated wording

checks/evaluation/maintained.go
raghavkaul: Updated wording to make this a bit more clear - being younger than 90 days makes it harder for us to assess the history of maintenance of a project

inferno-chromium (approved):

checks/maintained_test.go
```diff
@@ -298,6 +299,29 @@ func Test_Maintained(t *testing.T) {
 				Score: 0,
 			},
 		},
+		{
+			name:       "recently created fork",
```
azeemshaikh38, line 303: Nit: remove fork

azeemshaikh38 (approved):
LGTM with nit. Thanks for the quick turnaround here!

PR #2168, opened 2022-08-17
🐛 Fix remediation text when Scorecard is run multiple times within a program

### LINKED ISSUE #2152: Using Scorecard GitHub Action reduces the Token-Permissions score
**Describe the bug**

urllib3 recently added the [Scorecard GitHub Action](https://github.com/urllib3/urllib3/blob/main/.github/workflows/scorecards.yml) to its repository. After adding this we observed our "Token-Permissions" score being 9/10 instead of the previous value of 10/10. This was due to:

> Warn: jobLevel 'security-events' permission set to 'write': .github/workflows/scorecards.yml:16: update your workflow using https://app.stepsecurity.io/secureworkflow/syndicate-lang/syndicate-js/scorecards.yml/main?enable=permissions

This is because the action wants the permission `security-events: write` which is a permission that is docked points if given to any GitHub Action.

**Reproduction steps**
1. Add the Scorecard GitHub Action w/ `security-events: write`
2. Run Scorecard on the repository
3. Observe an imperfect score due to Scorecard GitHub Action config

**Expected behavior**

Using the Scorecard GitHub Action shouldn't reduce the project's scorecard score.

spencerschrock:
bug fix for incorrect repo info in remediation text. the problem affected the weekly scorecard cron job

The remediation text used to reflect the first repo / branch  processed by a scorecard cron worker.

The remediation text will now match the repo/branch being analyzed. 

#2152 

This is a temporary fix, and modifies the function signatures for `evaluation.PinningDependencies` and `evaluation.TokenPermissions`. I don't think anyone besides ourselves are using these exported functions.

the `release-note`

```release-note
NONE
```

### DIFF
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -42,7 +42,7 @@ type worklowPinningResult struct {
 }
 
 // PinningDependencies applies the score policy for the Pinned-Dependencies check.
-func PinningDependencies(name string, dl checker.DetailLogger,
+func PinningDependencies(name string, c *checker.CheckRequest,
 	r *checker.PinningDependenciesData,
 ) checker.CheckResult {
 	if r == nil {
@@ -52,6 +52,9 @@ func PinningDependencies(name string, dl checker.DetailLogger,
 
 	var wp worklowPinningResult
 	pr := make(map[checker.DependencyUseType]pinnedResult)
+	dl := c.Dlogger
+	//nolint:errcheck
+	remediaitonMetadata, _ := remediation.Setup(c)
 
 	for i := range r.Dependencies {
 		rr := r.Dependencies[i]
@@ -83,7 +86,7 @@ func PinningDependencies(name string, dl checker.DetailLogger,
 				EndOffset:   rr.Location.EndOffset,
 				Text:        generateText(&rr),
 				Snippet:     rr.Location.Snippet,
-				Remediation: generateRemediation(&rr),
+				Remediation: generateRemediation(remediaitonMetadata, &rr),
 			})
 
 			// Update the pinning status.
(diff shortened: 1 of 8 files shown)


azeemshaikh38:
Could we simply extend [`PinnedDepsRawData`](https://github.com/ossf/scorecard/blob/main/checker/raw_result.go#L92) and [`TokenPermissionsRawData`](https://github.com/ossf/scorecard/blob/main/checker/raw_result.go#L252) to include `RemediationMetadata` struct?

This isn't ideal, and really the raw data struct should have a common struct to hold this info. But atleast this avoids passing in the CheckRequest to `evaluation`.

remediation/remediations.go
```diff
@@ -42,53 +33,51 @@ var (
...
-		uri := c.RepoClient.URI()
-		parts := strings.Split(uri, "/")
-		if len(parts) != 3 {
-			setupErr = fmt.Errorf("%w: empty: %s", errInvalidArg, uri)
-			return
-		}
-		repo = fmt.Sprintf("%s/%s", parts[1], parts[2])
-	})
-	return setupErr
+func Setup(c *checker.CheckRequest) (RemediationMetadata, error) {
```
azeemshaikh38, line 43: Nit: change name to `New` ?

azeemshaikh38 (approved):

azeemshaikh38:
> > Could we simply extend [`PinnedDepsRawData`](https://github.com/ossf/scorecard/blob/main/checker/raw_result.go#L92) and [`TokenPermissionsRawData`](https://github.com/ossf/scorecard/blob/main/checker/raw_result.go#L252) to include `RemediationMetadata` struct?
> > This isn't ideal, and really the raw data struct should have a common struct to hold this info. But atleast this avoids passing in the CheckRequest to `evaluation`.
> 
> I was just following your quick fix comments in #2152. Also, I was hesitant to touch RawData for fear of messing up the json/cron job flow.
> 
> > But I know it's not an easy change and might take time. So, let's remove the init() fn and all the global vars for now (that should solve the data race issue too). We need to fix this before EOW so that the cron job picks up the fix.
> > Once the fix is in, let's reconsider how to solve it the right way using RawData. Wdyt?

SGTM, approved. Could we have a follow-up PR to address this?

spencerschrock:
> SGTM, approved. Could we have a follow-up PR to address this?

👍

checks/evaluation/pinned_dependencies.go
```diff
@@ -133,10 +136,10 @@ func PinningDependencies(name string, dl checker.DetailLogger,
 		"dependency not pinned by hash detected", score, checker.MaxResultScore)
 }
 
-func generateRemediation(rr *checker.Dependency) *checker.Remediation {
+func generateRemediation(remediaitonMd remediation.RemediationMetadata, rr *checker.Dependency) *checker.Remediation {
```
laurentsimon, line 139: nit: remediaitonMd

PR #2172, opened 2022-08-18
✨ Use sha256 for release hashes

laurentsimon:
This updates the hash from sha512 to sha256 for release hashes. That's what's expected by the SLSA generator.
sha256 is enough for security
```release-notes
Use sha256 for goreleaser config
```


naveensrinivasan (approved):
Thanks

Issue #2157, opened 2022-08-17
Improving Score Reporting: Empty repository


azeemshaikh38:
Running Scorecard on an empty repository (no commits) returns results which do not make sense:

```
./scorecard --repo=azeemshaikh38/empty-repo-test --format=json | jq .
2022/08/17 01:00:47 unable to get tarball tarball not found: https://api.github.com/repos/azeemshaikh38/empty-repo-test/tarball/. Skipping...                                                        
{                                                                                                                                                                                                    
  "date": "2022-08-17",                                                                                                                                                                              
  "repo": {                                                                                                                                                                                          
    "name": "github.com/azeemshaikh38/empty-repo-test",                                                                                                                                              
    "commit": "no commits found"                                                                                                                                                                     
  },                                                                                                                                                                                                 
  "scorecard": {                                                                                                                                                                                     
    "version": "v4.5.0-25-g887facf",                                                                                                                                                                 
    "commit": "887facf3cad2bd401ff5f6a51725225f8ec28c17"                                                                                                                                             
  },                                                                                                                                                                                                 
  "score": 5,                                                                                                                                                                                        
  "checks": [                                                                                                                                                                                        
     ...
   ]
}
```

The aggregate score of 5 for an empty repository does not make any sense and needs to be fixed.

@ossf/scorecard-maintainers @spencerschrock @raghavkaul fyi.

azeemshaikh38:
One idea might be to have the `Score` field to be a pointer. That way a `nil` value can indicate NA instead of `-1` which we today use to indicate inconclusive.

leec94:
Hi, in testing #2158 I was testing scorecard on an empty repository, but got an empty result. It surprised me that even the repo name wasn't present in the results, so I wasn't sure if my github auth token was expired or wrong, so it led me astray refreshing my token before trying to run scorecard on another repo. 

Suggestion, can the results still contain a valid date and repo fields in the results of an empty repo scan? -1 as the score is fine too. Fine with opening this up again or creating a new issue. 

```
carolinelee@Carolines-MBP scorecard % go run main.go --repo=github.com/azeemshaikh38/empty-repo-test --format=json --checks=Token-Permissions --show-details | jq .
{
  "date": "0001-01-01T00:00:00Z",
  "repo": {
    "name": "",
    "commit": ""
  },
  "scorecard": {
    "version": "",
    "commit": ""
  },
  "score": -1,
  "checks": null,
  "metadata": null
}
```

[...]

PR #2182, opened 2022-08-20
:seedling: Fix for empty repository

naveensrinivasan:
- Fixes the score to `0` for an empty repository.
- Fixes https://github.com/ossf/scorecard/issues/2157

### DIFF
--- a/pkg/json.go
+++ b/pkg/json.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/ossf/scorecard/v4/checker"
+
 	docs "github.com/ossf/scorecard/v4/docs/checks"
 	sce "github.com/ossf/scorecard/v4/errors"
 	"github.com/ossf/scorecard/v4/log"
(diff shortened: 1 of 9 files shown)


naveensrinivasan:
@azeemshaikh38  👀 . I haven't handled the SARIF yet but wanted feedback prior to that.

naveensrinivasan:
@azeemshaikh38  a friendly ping

azeemshaikh38 (commented):
The simpler fix might be to return an empty `ScorecardResult{}` struct here when there are no commits found - https://github.com/ossf/scorecard/blob/6fc08e7c9ce05bbff1329e777b2eca38cb41564f/pkg/scorecard.go#L97 ?

naveensrinivasan:
> The simpler fix might be to return an empty `ScorecardResult{}` struct here when there are no commits found -
> 
> https://github.com/ossf/scorecard/blob/6fc08e7c9ce05bbff1329e777b2eca38cb41564f/pkg/scorecard.go#L97
> 
> ?

Cool, Thanks took care of it here https://github.com/ossf/scorecard/pull/2207

Issue #1776, opened 2022-03-24
Feature: New check for SLSA provenance generation


laurentsimon:
This check may replace the `Signed-Release` checks and verify that the releases generate SLSA provenance using the official builders from https://github.com/slsa-framework/slsa. We can:
1. Verify the presence of `.intoto.jsonl` file
2. Verify a workflow exists calling an official action
3. Verify the signature by calling the API exposed by https://github.com/slsa-framework/slsa

This should allow us to also provide some SLSA compliance flags for scorecard, e.g. something like `--slsa=3`

azeemshaikh38:
Sounds like a good idea to me. Maybe expand the `Signed-Release` check to include this? The updated check could verify various methods which ensure release integrity - GitHub signatures, cosign signatures, the SLSA workflow etc?

naveensrinivasan:
This is great! 👍

azeemshaikh38:
Sure, not tied to the `Signed-Release` name at all. To confirm are you saying that if users sign their release (let's say with cosign), we should complain that it's not enough and instead drive them towards building provenance? 

I was thinking the new check will encompass all schemes - provenance but also signatures.

laurentsimon:
> Sure, not tied to the `Signed-Release` name at all. To confirm are you saying that if users sign their release (let's say with cosign), we should complain that it's not enough and instead drive them towards building provenance?

yes, so long as there is tooling to achieve it.

> I was thinking the new check will encompass all schemes - provenance but also signatures.

provenance also contains a signature that covers the binary and additional information (builder ID, commands, etc), so it does encompasses the signature.

sethmlarson:
Chiming in to say I'd like to start using SLSA with urllib3 but losing out on the `Signed-Releases` check we're currently getting from Sigstore would be unfortunate by changing to SLSA.

Could we can get Scorecard to recognize `*.intoto.jsonl` as the same as `*.sig` on a GitHub Release? I don't know if Scorecard is verifying the signatures we're currently publishing as the info only mentions the existence of a `*.sig` file so if we're going off presence of a file that'd be an easier first step than implementing full end-to-end verification?

laurentsimon:
We're not verifying signatures at all, because it's hard t find out the correct pubic key in practice. SO we can definitely look for an .intoto.jsonl.

sethmlarson:
I think that makes sense, especially as a means to convince people to start using SLSA.

laurentsimon:
[...]

PR #2186, opened 2022-08-22
:bug: Signed-Releases - Check releases for signatures even if they only have source code

### LINKED ISSUE #2169: Improve Score Reporting: Signed-Releases looks at old release data
In some cases, Signed-Releases check looks at older releases instead of looking at the latest 5 releases. Some failing repos:

- https://github.com/pallets/jinja
- http://github.com/scikit-build/cmake-python-distributions
- http://github.com/hardbyte/python-can
- http://github.com/scikit-build/scikit-build
- http://github.com/koalalorenzo/python-digitalocean

raghavkaul:
Fix issue where we skipped recent releases when looking for signature/provenance.

Previously we checked for the presence of any assets. However, source code zips/tarballs aren't considered assets; they are created by default at the GitHub tag.

Source code zips/tarballs 

Fixes #2169.

```release-note
NONE
```

### DIFF
--- a/checks/evaluation/signed_releases.go
+++ b/checks/evaluation/signed_releases.go
@@ -30,8 +30,8 @@ var (
 
 const releaseLookBack = 5
 
-//SignedReleases applies the score policy for the Signed-Releases check.
-//nolint
+// SignedReleases applies the score policy for the Signed-Releases check.
+// nolint
 func SignedReleases(name string, dl checker.DetailLogger, r *checker.SignedReleasesData) checker.CheckResult {
 	if r == nil {
 		e := sce.WithMessage(sce.ErrScorecardInternal, "empty raw data")
@@ -42,7 +42,7 @@ func SignedReleases(name string, dl checker.DetailLogger, r *checker.SignedRelea
 	total := 0
 	score := 0
 	for _, release := range r.Releases {
-		if release.TagName == "" {
+		if release.ZipballURL == "" && release.TarballURL == "" && len(release.Assets) == 0 {
 			continue
 		}
 
--- a/checks/evaluation/signed_releases.go
+++ b/checks/evaluation/signed_releases.go
@@ -42,7 +42,7 @@ func SignedReleases(name string, dl checker.DetailLogger, r *checker.SignedRelea
 	total := 0
 	score := 0
 	for _, release := range r.Releases {
-		if len(release.Assets) == 0 {
+		if release.TagName == "" {
 			continue
 		}
 
@@ -127,6 +127,6 @@ func SignedReleases(name string, dl checker.DetailLogger, r *checker.SignedRelea
 	}
 
 	score = int(math.Floor(float64(score) / float64(totalReleases)))
-	reason := fmt.Sprintf("%d out of %d artifacts are signed or have provenance", total, totalReleases)
+	reason := fmt.Sprintf("%d out of %d most recent artifacts are signed or have provenance", total, totalReleases)
 	return checker.CreateResultWithScore(name, reason, score)
 }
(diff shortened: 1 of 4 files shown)


checks/evaluation/signed_releases.go
```diff
@@ -42,7 +42,7 @@ func SignedReleases(name string, dl checker.DetailLogger, r *checker.SignedRelea
 	total := 0
 	score := 0
 	for _, release := range r.Releases {
-		if len(release.Assets) == 0 {
+		if release.ZipballURL == "" && release.TarballURL == "" && len(release.Assets) == 0 {
```
azeemshaikh38, line 45: Let's add a `log.Warn` here to say for release X, no assets or source code was found.

checks/evaluation/signed_releases.go
```diff
@@ -42,7 +42,7 @@ func SignedReleases(name string, dl checker.DetailLogger, r *checker.SignedRelea
...
 			continue
```
azeemshaikh38, line 46: Is there anyway to sign or provide provenance for source code? @laurentsimon fyi

azeemshaikh38 (approved):

checks/evaluation/signed_releases.go
raghavkaul: I believe you can PGP-sign a tarball or zipfile (we already check for the `.asc` file extension, including `.tar.gz.asc`). However we could expand the set of `signatureExtensions` and `provenanceExtensions` if there are any tarball or zip-specific formats that we don't already have.

laurentsimon: I suppose it's possible to sign too. The 2 default assets are generated by GitHub and cannot be updated by maintainers. So signing them does not improve security by much, unless we distrust GH.

laurentsimon: Any update? I'm not entirely sold on the benefit of this change. If there are no assets in the GH releases, it'd be fine to return a `-1`. Forcing signature that adds little to no security guarantees feels like asking users to do something that's not necessary and a burden.
Happy to chat more.

raghavkaul: The benefit of signing releases is that someone can be sure an artifact comes from the maintainer of a repo. It feels like a better approach would be to have a "not enough data" or "inconclusive" result (as per #1874), because there are no real artifacts in recent releases. I'm not even sure what benefit it provides to sign an artifact that's already being released through GitHub. If it were a package, Package integrity would be better captured by the Packaging check instead of this check.

In short, the way forward would be to wrap this kind of finding into #1874.

laurentsimon: > The benefit of signing releases is that someone can be sure an artifact comes from the maintainer of a repo. 

But you need to be a maintainer to release. Are you saying that the benefit is that someone takes over a GH workflow to cut a release? Even then, the code is still from the repository, ie from a maintainer.

> I'm not even sure what benefit it provides to sign an artifact that's already being released through GitHub. 

In transit, these artifacts can be tampered with. I agree it's not perfect, and finding the public keys to verify is something nobody does in practice. I actually proposed getting rid of signature in https://github.com/ossf/scorecard/issues/1776 as well.

> If it were a package, Package integrity would be better captured by the Packaging check instead of this check.

Agreed, but no ecosystem really supports this. In golang, you don't even need signatures because code is fetched directly from repo.

azeemshaikh38: This PR still fixes a bug in our code - instead of looking at the latest 5 releases *with* a release artifact, look for the latest 5 releases irrespective of whether it has a release artifact or not. The only question that remains is, how we should treat those. IMO, we have 3 options:

1) Skip these releases and return an `-1`
2) Treat releases with source code only as signed and return a score of 10.
3) Expect a signature even if it is source code only.

(1) should be the easiest/most acceptable fix. We can follow up separately to decide b/w (2) and (3). What do folks thinks?

laurentsimon: +1 on fixing the bug in the code. 1 seems fine for now. 

I think we should discuss 2 and 3 more broadly in the context of what guarantees / goals this check should provide, and how it fits in the improvement of supply-chain. Today this check is not actionable due to lack of ecosystem support, and it's not "ecosystem-aware".

Wdut?

Issue #1916, opened 2022-05-16
Bug Fix Request: Re-enable 3 checks in cron job


Parth59:
**Describe the bug**
Currently, the CRON pushes scores of 16 checks to bigQuery and does not include scores for the following 3 checks CI_TEST, SAST, and CONTRIBUTOR. 

**Expected behavior**
We expect all the scores to be available in the processed data dump. If the scorecards application was unable to compute it then possibly assigning a score value like -1 or any other pre-defined value can help us ignore those scores.

**Additional context**
I can think of the following 2 reasons -

A) There was a design decision taken to exclude these metrics. That's might be one of the reasons why previous issues that asked to re-enable this  https://github.com/ossf/scorecard/issues/1302 were put to backlog. 

B) This has been missed out to include somehow. If that is the case, can we enable those now?

laurentsimon:
Some of these checks were taken out because they are API intensive so rate limiting can be a bottleneck.
@azeemsgoogle any plan to add these checks?

azeemshaikh38:
Disabling these checks is a conscious decision for being able to scale the weekly job (see https://github.com/ossf/scorecard/blob/main/cron/config/config.yaml#L27). We need to optimize these checks for API usage before we enable them in the cron job. As of now, I don't plan to work on adding these back. Happy if someone wants to take this on. 

We should probably explicitly call out in our documentation though that not all checks are run during the weekly cron job.

laurentsimon:
We could add an extra column in the https://github.com/ossf/scorecard#scorecard-checks. @Parth59 interested in sending a quick PR for it?

spencerschrock:
I can take a look at this

spencerschrock:
My plan is to expand the graphQL API call to fetch all the check info in the initial API call. Something like this:

```
query {
  repository(owner:"ossf", name:"scorecard") {
    object(expression: "heads/main") {
      ... on Commit {
        checkSuites(first: 10) {
          edges {
            node {
              app{
                slug
              }
              conclusion
              status
            }
          }
        }
      }
    }
  }
}
```

which will produced the same info used by `Client.Checks.ListCheckRunsForRef`

```
{
  "node": {
    "app": {
      "slug": "dependabot"
    },
    "conclusion": null,
    "status": "QUEUED"
  }
}
```

azeemshaikh38:
[...]

PR #2191, opened 2022-08-24
✨ Optimize SAST check

spencerschrock:
Optimization

`CheckRuns` info is fetched via 30 REST API calls.

Caches info about `CheckRuns` which is pulled in during the initial graphQL step without an increase in that API cost (currently cost=1). This makes the info available for the `SAST` check without needing 30 REST API calls. 

If any uncached info is requested, REST API calls are performed, same as before.

Partial fix for #1916

Might run into uncached scenarios for repos that don't squash on merge.

the `release-note`

```release-note
NONE
```


spencerschrock:
Need to double check E2E behavior on a repo that doesn't squash commits. When run on `github.com/ossf/scorecard`, everything is cached.

clients/githubrepo/client.go
```diff
@@ -192,7 +192,14 @@ func (client *Client) ListSuccessfulWorkflowRuns(filename string) ([]clients.Wor
 
 // ListCheckRunsForRef implements RepoClient.ListCheckRunsForRef.
 func (client *Client) ListCheckRunsForRef(ref string) ([]clients.CheckRun, error) {
-	return client.checkruns.listCheckRunsForRef(ref)
+	if crs, ok := client.graphClient.checkRuns[ref]; ok {
```
azeemshaikh38, line 195: Nit: IMO, cross-file var references reduce code readability - you need context from the other file to understand what the var is. Consider replacing with `graphClient.listCheckRunsForRef()` or similar self-explanatory fn name instead.

azeemshaikh38 (approved):
Awesome work @spencerschrock with the quick turn around here! Can't wait to see this running in prod.

clients/githubrepo/graphql.go
```diff
@@ -89,6 +90,21 @@ type graphqlData struct {
...
+													App struct {
+														Slug githubv4.String
+													}
+													Conclusion githubv4.CheckConclusionState
+													Status     githubv4.CheckStatusState
+												}
+											} `graphql:"checkSuites(first: $checksToAnalyze)"`
+										}
+									}
+								} `graphql:"commits(last:1)"`
```
azeemshaikh38, line 107: Nit: can consider returning the merge commit here and comparing against `headRefOid` and log differences. Can be used in debugging later.

azeemshaikh38 (approved):

clients/githubrepo/graphql.go
azeemshaikh38: Nvm this - we've the other log (comparing this against old `listCheckRuns`) that should be enough

clients/pull_request.go
```diff
@@ -19,14 +19,16 @@ import (
...
-	Author   User
-	Labels   []Label
-	Reviews  []Review
+	Number    int
+	MergedAt  time.Time
+	HeadSHA   string
+	Author    User
+	Labels    []Label
+	Reviews   []Review
+	CheckRuns []CheckRun
```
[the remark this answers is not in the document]
spencerschrock: SG. That was left over from an initial implementation

clients/githubrepo/client.go
```diff
@@ -192,7 +192,14 @@ func (client *Client) ListSuccessfulWorkflowRuns(filename string) ([]clients.Wor
...
+		return crs, nil
+	}
+	crs, err := client.checkruns.listCheckRunsForRef(ref)
+	if err == nil {
+		client.graphClient.checkRuns[ref] = crs
+	}
+	return crs, err
```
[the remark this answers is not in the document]
spencerschrock: Any suggestion on how to log it?
I had tried something like this, using the scorecard log package, and it seems like it would log things for everyone, not just the cron job
```
msg := fmt.Sprintf("listCheckRunsForRef cache miss: %s/%s:%s", handler.repourl.owner, handler.repourl.repo, ref)
logger.Info(msg)
```

clients/githubrepo/client.go
spencerschrock: Is the update any better?

clients/githubrepo/client.go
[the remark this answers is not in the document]
azeemshaikh38: I'm not too familiar with the logger package in Scorecard. Seems like it should [support](https://github.com/ossf/scorecard/blob/main/log/log.go) `logger.Debug` option? Could we use that instead?

clients/githubrepo/client.go
azeemshaikh38: LGTM.

azeemshaikh38:
> Looks like some consistent failures in the e2e token tests. I'll take a look

May need to include `checks: read` permission.

clients/githubrepo/client.go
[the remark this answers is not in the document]
spencerschrock: Staying with `logger.Info` as there is no `logger.Debug` available

azeemshaikh38 (approved):

PR #2195, opened 2022-08-24
:sparkles: Improved Security Policy Check

### LINKED ISSUE #2137: Feature: Improved Security-Policy checks
The mere existence of a `SECURITY.md` file (or other file(s) searched such as `.adoc`) with 1 or more bytes in the repo rewards the project's score to the fullest possible score (`10`).

Gradate the Security-Policy check which would incrementally improve the Security-Policy score if the `SECURITY.md` (or found file) contained links to pages or email addresses. That is detecting `https:`, or `mailto:`, or some other pattern (perhaps using regex) on the contents of that file. There would be no need to actually determine if the `https:`, or `mailto:` actually resolved (DNS) or didn't bounce (SMTP). Perhaps:

no Security-Policy: score 0
one found link content: score += 3
more than one found link content: score += 6
more bytes than the sum of the length of all the link content found: score += 4

the score of `10` here could then assume that around all that link content are intelligible words. for instance:

```
Sed ut perspiciatis unde omnis iste https://example.com/security or email security@example.com natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni
```

would score 10 as it has 2 link contents (score = 6) and a length of content (~379) which is greater than the sum of the length of the two link content (48 = strlen of the https (28) + strlen of the email (20)) would be (score = 6 + 4). If the content had no link content, then the final score would be score = 4 or had it only 1 link content, score = 7.

In practice, ossf's own [scorecard file](https://github.com/ossf/scorecard/blob/main/SECURITY.md) would score = 7 (it has one email address and many bytes beyond the length of that email). If ossf's scorecard is intended to be the exemplar, then perhaps another scale could be devised to detect portions of timelines, days, and other disclosure practices - which would require language analysis.

If a score for a project was non-zero, pull the appropriate security-policy file and inspect/parse myself. scorecard already details in the report that `"reason": "security policy file detected"` but does not display the file/path/name/URI of what was actually detected--therefore, the burden is on the consumer to 'go hunting' (first) then analyze later.

**Additional context**
The same could be said for the checks `License` and `Signed-Releases` as long as the representative file in GitHub is 1 byte or more, the project is award full points. Although checking the validity of `Signed-Releases` is troublesome, a single byte length file seems too small for any signature. For `License`, more NLP might be called for, but with further investigation known strings could be searched and recommended for those that want to be scorecard friendly. Such as [https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository](url) or [https://opensource.org/licenses](url)

shissam:

Improvement on the Security Policy check which goes beyond the simply existence of a recognized Security Policy filename, and examines the contents of the found Security Policy filename seeking hits on key elements of a community security policy (linked content, emails, time frames and specifics about disclosures or vulnerability reporting)

Security Policy will score either 0 (zero) if no file is found in the repo or the org's repo, or will award a maximum score of 10 (ten) simply on the existence of the file and a length greater than 0 (zero) - regardless if there is anything in it.

If a Security Policy file does not exist or is less that 1 byte in length, the behavior of scorecard is unchanged. Otherwise:

```
#1: found one linked (email/http) content: score += 3
    rationale: someone to collaborate with or link to
    information (strong for community)

#2: more than one unique (email/http) linked content found: score += 3
    rationale: if more than one link, even stronger for the community

#3: more bytes than the sum of the length of all the linked content found: score += 3
    rationale: there appears to be information and context around those links
    no credit if there is just a link to a site or an email address (those given above)

#4: found whole number(s) and or match(es) to "Disclos" and or "Vuln": score += 1
    rationale: works towards the intent of the security policy file
    regarding whom to contact about vuls and disclosures and timing
    e.g., we'll disclose, report a vulnerabily, 30 days, etc.
    looking for at least 2 hits
```

Fixes #2137

The fact that Security-Policy scores will now be gradiated (and not just 0 or 10) is user facing, in this PR, there is an update to `checks.md` proposed.

the `release-note`

```release-note
This check examines the contents of the security policy file awarding points for those policies that express vulnerability process(es), disclosure timelines, and have links (e.g., URL(s) and email(s) to support the users.
```


olivekl:
Thanks for this! The documentation in checks.md is autogenerated. Please don't manually update checks.md. Instead, make the changes in checks.yaml and run `make generate-docs`.

shissam:
> Thanks for this! The documentation in checks.md is autogenerated. Please don't manually update checks.md. Instead, make the changes in checks.yaml and run `make generate-docs`.

hi, thanks, I see that now, sorry I missed that. I have also seen that `make unit-test` (locally) fails just like the integration test in the workflow in this PR.  In the `CONTRIBUTING.md` for scorecard, there is a comment on `make e2e` (which for me (locally) reports `make: Nothing to be done for 'e2e'`. There was nothing in that guide about `make unit-test`.

So, I have more work to do

- [ ] fix the checks.yaml and do `make generate-docs`
- [ ] update the unit tests for Security-Policy to support GetFileContents and revise the test to reflect new potential scores in the mock site (i.e., pass `make unit-test` locally).

given this "depth" of additional work I need to complete, how should this PR proceed? Should I delete the PR and re-fork and re-PR after I get those two things fixed? I am not sure what is the best way to proceed. If this PR is kept, how to I submit those changes (like `checks.yaml`) into this PR? (I am a n00b to this overall process.

laurentsimon:
@shissam feel free to include the update of checks.yaml / checks.md + unit tests in this PR. That's what we usally do. Thanks again for your contribution

checks/evaluation/security_policy.go
```diff
@@ -15,10 +15,59 @@
...
+func scoreSecurityCriteria(contentLen, linkedContentLen, urls, emails, discvuls int) (int, string) {
+	score := 0
+	reason := ""
+
+	// #1: found one linked (email/http) content: score += 3
+	//     rationale: someone to collaborate with or link to
+	//     information (strong for community)
+	if urls >= 1 || emails >= 1 {
+		score += 3
+		reason += "linked content, "
```
laurentsimon, line 33: Instead of building the reason, you can use `log.Info()` for things that are positive, and `log.Warn()` for things that are negative. Something along the lines of:
```golang
if urls >= 1 || emails >= 1 {
  log.Info("Found link in description: %s", link)
} else {
  log.Warn("no email or URL found in description)
}
```

Also note that the logging function allows you to pass file / line numbers as well, which you should try to use. These are reflected in the SARIF format (GH scanning dashboard). They are also accessible via `-show-details` and in the weekly scan.

Same comment for the other checks below in the code.

checks/evaluation/security_policy.go
```diff
@@ -15,10 +15,59 @@
...
 	sce "github.com/ossf/scorecard/v4/errors"
 )
 
+func scoreSecurityCriteria(contentLen, linkedContentLen, urls, emails, discvuls int) (int, string) {
+	score := 0
+	reason := ""
+
+	// #1: found one linked (email/http) content: score += 3
+	//     rationale: someone to collaborate with or link to
+	//     information (strong for community)
```
laurentsimon, line 30: Maybe add a link to an example .md file that satisfies this, to help readers of the code.

checks/evaluation/security_policy.go
```diff
@@ -15,10 +15,59 @@
...
+	if urls >= 1 || emails >= 1 {
+		score += 3
+		reason += "linked content, "
+	}
+
+	// #2: more than one unique (email/http) linked content found: score += 3
```
laurentsimon, line 36: Is it really necessary to differentiate between 1 vs multiple links? Why would one email address not be enough?

checks/evaluation/security_policy.go
```diff
@@ -43,7 +95,17 @@ func SecurityPolicy(name string, dl checker.DetailLogger, r *checker.SecurityPol
 		} else {
 			msg.Text = "security policy detected in current repo"
 		}
+
+		var contentLen, linkedContentLen, urls, emails, discvuls int
+		fmt.Sscanf(f.Snippet, "%d,%d,%d,%d,%d", &contentLen, &linkedContentLen, &urls, &emails, &discvuls)
```
laurentsimon, line 100: It's fine to use multiple Info/Warn, I think.

checks/raw/security_policy.go
```diff
@@ -112,3 +137,87 @@ func isSecurityPolicyFilename(name string) bool {
...
+	if len(content) == 0 {
+		// perhaps there are more policy files somewhere else,
+		// keep looking (true)
+		return true, nil
+	}
+
+	// TODO: there is an assertion here that if content has length
+	//       then (*pfiles)[0] is not nil (can that be checked)
+	//       (so is this test necessary?)
+	if (*pfiles) != nil {
```
laurentsimon, line 163: return an error if it's nil, because this means a bug in our code.

checks/raw/security_policy.go
```diff
@@ -85,8 +108,10 @@ var isSecurityPolicyFile fileparser.DoWhileTrueOnFilename = func(name string, ar
 	}
 	if isSecurityPolicyFilename(name) {
 		tempPath := name
+		// TODO: really should be FileTypeText (.md, .adoc, etc)
```
laurentsimon, line 111: Good catch. Let's use FileTypeText instead.

checks/raw/security_policy.go
```diff
@@ -112,3 +137,87 @@ func isSecurityPolicyFilename(name string) bool {
...
+			Snippet: contentMetrics,
+		}
+	}
+
+	// stop here found something, no need to look further (false)
+	return false, nil
+}
+
+func countUniquePolicyHits(policyContent string) (int, int, int, int) {
+	var urls, strlenOfUrls, emails, strlenOfEmails, discvuls int
```
laurentsimon, line 181: Let's record the lines of the hit, so that we can add them later to details via `Info()`

checks/raw/security_policy.go
```diff
@@ -44,7 +45,15 @@ func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error)
 		return checker.SecurityPolicyData{}, err
 	}
 	// If we found files in the repo, return immediately.
+	// TODO: // assert that at this point only 1 file is returned by isSecurityPolicyFile
 	if len(data.files) > 0 {
+		err := fileparser.OnMatchingFileContentDo(c.RepoClient, fileparser.PathMatcher{
```
laurentsimon, line 50: I think you can update the existing code to perform the matching as well, instead of doing that at the end.
What I mean is that you can update previous calls to `OnAllFilesDo` to `OnMatchingFileContentDo`

checks/raw/security_policy.go
```diff
@@ -112,3 +137,87 @@ func isSecurityPolicyFilename(name string) bool {
...
+	if (*pfiles) != nil {
+		// preserve file type
+		tempType := (*pfiles)[0].Type
+		linkedContentLen, urls, emails, discvuls := countUniquePolicyHits(string(content))
+		contentMetrics := fmt.Sprintf("%d,%d,%d,%d,%d", len(content), linkedContentLen, urls, emails, discvuls)
+		(*pfiles)[0] = checker.File{
+			Path:    path,
+			Type:    tempType,
+			Offset:  checker.OffsetDefault,
+			Snippet: contentMetrics,
```
laurentsimon, line 172: Snippet is typically used for exact snippet of code. I suggest we instead update this structure https://github.com/ossf/scorecard/blob/main/checker/raw_result.go#L141-L144 to contain the information gathered in this file. Then in the evaluation part, we use that.

How about:
```golang

type SecurityPolicyInformationType string

const (
	// SecurityPolicyInformationTypeEmail blabla
	SecurityPolicyInformationType SecurityPolicyInformationTypeEmail = "emailAddress"
	SecurityPolicyInformationType SecurityPolicyInformationTypeLink = "httpLink"
        ...
)
type SecurityPolicyFileInformation struct {
   File File
   Information SecurityPolicyInformationType
}
type SecurityPolicyData struct {
	// Files contains a list of files.
	InformationFiles []SecurityPolicyFileInformation
}
```

This is not perfect, and I think we may want to update the `File` structure to
```golang
type FileLocation struct{
  Snippet   string   // Snippet of code
  Offset    uint     // Offset in the file of Path (line for source/text files).
  EndOffset uint     
}
type File struct {
	Path      string
	Locations []FileLocation
	Type      FileType // Type of file.
	// TODO: add hash.
}
```

but this is out of the scope for this PR. /cc @raghavkaul 

We can also update https://github.com/ossf/scorecard/blob/32d6ba27757fcaf6145b03195f84c74b311e4121/pkg/json_raw_results.go#L42-L47 once we've agreed what the format should be.

laurentsimon (commented):
Thanks for the PR. I've suggested some changes:
1. Rename the structure in https://github.com/ossf/scorecard/blob/32d6ba27757fcaf6145b03195f84c74b311e4121/checker/raw_result.go#L141-L144
2. Gather all the information needed
3. Parse that data, add calls to Warn/Info and compute the score as you do right now.

Let me know if this does not make sense. I understand there's a lot to unpack.

checks/evaluation/security_policy.go
shissam: this point along with the related `SecurityPolicyFileInformation` noted in one of the other suggested changes sounds perfect (wish I'd thought of it). I'll make that change.

checks/evaluation/security_policy.go
shissam: So, in this latest commit there are two unit-test cases which I used as exemplars `10_realworld` (earns 10 points) and `07_realworld` (earns 7 points) pulled from `kubernetes` and `ossf/scorecard` respectively.

When you say "add a link" - do you mean to the `checks.yaml` which would get incorporated into `checks.md` (by way of `make generate-docs`? Or something else?

checks/evaluation/security_policy.go
shissam: So - this is where I am happy to "do anything"... in my mind (hence the _rational_ in the code comments and this PR) is that more linked content is more (perhaps) helpful or insightful for the contributor wishing to communicate vuls and be informed of the disclosure process.

As such I valued more linked content (say over number of words (symbols (think groups of [:print:] separated by [:space:])--which in hindsight would make `SecurityPolicyFileInformation` really messy)

After that, then (again) without counting words, I considered the overall content length of the text file as being more than the sum of the length of the links (emails/urls) as an hint/indicator of useful content. The final plus up, is that within that texts I can find hints of numbers and words which look like disclosure and/or vulnerability.

So - in the end, the intent was to have a fairly nice gradation for the scoring over 0 or 10 (now I am just reminded I need a unit-test for only text with no hits and text which will score only 1).

So... I am open to suggestions.

checks/evaluation/security_policy.go
shissam: On this one--I am not sure what you are hinting at. I think I need more clarification on this comment/suggestion please.

checks/raw/security_policy.go
shissam: Will do!

checks/raw/security_policy.go
shissam: Will do... I was not sure what other problems might be causes by changing the file type (I don't know all the dependencies)

checks/raw/security_policy.go
shissam: Check... got it, will do!

checks/raw/security_policy.go
shissam: This is one that I am not sure of.

What I found with `OnMatchingFileContentDo` (OMFCD), I would set the (regex) pattern to be the name of **the** file I want, e.g., `security.md` (like that returned from `isSecurityPolicyFilename`) having to "pattern" (per se), and OMFCD would return that **and** any other file within the repo that had that name.... I recall seeing `/<dir>/<dir>/<dir>/<dir>/security.md` which were not wanted.

This is because OMFCD uses `isMatchingPath` walking the entire repo, and finds **all** files which (in this case) have `security.md`--because `isMatchingPath` uses two tests (see lines 36-47 of `checks/fileparser/listing.go`) one on the fullpath and the other on the filename (or basename) for the search. `OnAllFilesDo` (OAFD) does not do that (use `isMatchingPath`)

So, for the security policy, say I was searching for `security.adoc` and not using (OAFD) **and** `security.adoc` was not in `repos root`, `.github/`, or `docs/` but did appear deep in the repo, say `/contrib/test/other/data/whatif/security.adoc`, OMFCD would actually provide the contents of that file which would then be parsed/evaluated.

What seems to work is keying off `OnAllFilesDo` finding what I am looking for, using the specific pattern to OMFCD and stopping at the first file match.

Perhaps I am using OMFCD incorrectly, and need hint as to a better way to use OMFCD.

checks/raw/security_policy.go
shissam: I like this very much.. I new I has co-opted Snippet as a channel to get info from `raw` to `evaluation` code and did so (perhaps) violating intent of that struct field without having to change more to be cleaner -- this suggestion is much cleaner. So, Will do!

checks/raw/security_policy.go
shissam: wait... actually, (I wonder) if OMFCD could be parameterized (perhaps a `bool` like `CaseSensitive`) to NOT do the backup `filename` and stick to the first search for full path, then I think OMFCD could / should work. That would mean that `isSecurityPolicyFilename` behavior would change (unless OMFCD could also be parameterized with a predicate like `isSecurityPolicyFilename`) -- just a thought

checks/evaluation/security_policy.go
laurentsimon: I meant an http link to an actual repository that satisfies it. Maybe that's not needed.

checks/evaluation/security_policy.go
laurentsimon: > As such I valued more linked content (say over number of words (symbols (think groups of [:print:] separated by [:space:])--which in hindsight would make `SecurityPolicyFileInformation` really messy)

One link / email address is probably enough, I'd say. This way we err on the side of caution to avoid false negatives. Wdut?

Btw, what do links typically point to in the policy file? A webpage to submit a vuln instead of an email address?

> After that, then (again) without counting words, I considered the overall content length of the text file as being more than the sum of the length of the links (emails/urls) as an hint/indicator of useful content. The final plus up, is that within that texts I can find hints of numbers and words which look like disclosure and/or vulnerability.

That sounds reasonable to me. It's similar to pre-submits that look for good description for PRs.

checks/evaluation/security_policy.go
laurentsimon: sorry. I mean that when reporting what we find, we can have `Info("found an email: %s")`, another one `Info("we found a URL: %s)`, etc. We don't need to add everything to a single snippet.

checks/evaluation/security_policy.go
shissam: > One link / email address is probably enough, I'd say. This way we err on the side of caution to > avoid false negatives. Wdut?
>
> Btw, what do links typically point to in the policy file? A webpage to submit a vuln instead of an email address?

I will think more about that - let me work these comments/suggestions, with the intent of credit for 1 url or 1 email to get a larger plus up on score.

As for what I've see, for example, Kubernetes has NO emails but numerous links, to a) point user where to read about vuls; b) how to report; and c) support-kinda-things. ossf/scorecard only has one email (NO urls) but the disclosure/vulnerability highlights are right there!

shissam: At this point the security policy scoring is now:

```
1: one or more unique (email/http) linked content found: score += 6

2: content in the file that is more than the sum of all the links found in the first test score += 3

3: found whole number(s) and or match(es) to "Disclos" and or "Vuln" (hints to disclosure or vulnerability reporting process) score += 1
```

I think that'll suffice for a good baseline

checks/evaluation/security_policy.go
shissam: So I did not try to log.Info() that actual content found, just the check that was satisfied (scored 10):

```
    "checks": [
        {
            "details": [
                "Info: Found linked content in security policy: SECURITY.md",
                "Info: Found text in security policy: SECURITY.md",
                "Info: Found disclosure, vulnerability, and/or timelines in security policy: SECURITY.md",
                "Info: security policy detected in current repo: SECURITY.md:1"
            ],
```

and here is an example where a site scored 0 (zero)

```
    "checks": [
        {
            "details": [
                "Warn: no email or URL found in security policy: SECURITY.md",
                "Warn: No text (beyond any linked content) found in security policy: SECURITY.md",
                "Warn: One or no descriptive hints of disclosure, vulnerability, and/or timelines in security policy: SECURITY.md",
                "Info: security policy detected in current repo: SECURITY.md:1"
            ],

```

checks/raw/security_policy.go
shissam: in revising `SecurityPolicyData` to include `SecurityPolicyInformationType`/`SecurityPolicyInformation` I went the route to note for this policy check there is only on `File` (as opposed to `[]File`) - as such I am now checking for empty `File` (i.e., `(File{})`) - but I do have to check for `nil` on `(*pinfo)` for the new policy information.

So .. the intent is done.

checks/raw/security_policy.go
shissam: Done.

checks/raw/security_policy.go
shissam: Do, I did not do line hits, and I suppose it could be done.

Right now, I am using `regexp.FindAllString()` on the buffered content and not going "line by line". If this is a non-starter, I can go back and do finer grained parsing/telemetry to id the line (and col/offset per line) of the matches.

Your call...

checks/raw/security_policy.go
shissam: I have not tried to optimize this (yet) - it is an optimization which can be performed.

One potential thought I had was rather than use OMFCD, keep using OAFD with `repoClient.GetFileContent` (which is used within OMFCD) and just get the contents directly.

checks/raw/security_policy.go
shissam: The security policy check, unit-test, e2e tests have all be revised to reflect the change to `SecurityPolicyData`

checks/evaluation/security_policy.go
shissam: log.Info/Warn - done

checker/raw_result.go
```diff
@@ -136,11 +136,28 @@ type VulnerabilitiesData struct {
 	Vulnerabilities []clients.Vulnerability
 }
 
+type SecurityPolicyInformationType string
+
+const (
+	// forms of security policy hints being evaluated.
+	SecurityPolicyInformationTypeEmail SecurityPolicyInformationType = "emailAddress"
+	SecurityPolicyInformationTypeLink  SecurityPolicyInformationType = "httpLink"
+	SecurityPolicyInformationTypeText  SecurityPolicyInformationType = "vuln & disclosure text"
```
laurentsimon, line 145: keep this camelcase, without spaces. This will ultimately be used in a field "name".

checker/raw_result.go
```diff
@@ -136,11 +136,28 @@ type VulnerabilitiesData struct {
...
+)
+
+type SecurityPolicyInformation struct {
+	InformationType  SecurityPolicyInformationType
+	InformationValue []string
```
laurentsimon, line 150: how will you keep track of lines each of the `InformationValue` were found?

checks/evaluation/security_policy.go
```diff
@@ -19,6 +19,81 @@ import (
 	sce "github.com/ossf/scorecard/v4/errors"
 )
 
+func scoreSecurityCriteria(f checker.File, info []checker.SecurityPolicyInformation, dl checker.DetailLogger) int {
+	var urls, emails, discvuls, linkedContentLen, score int
+	// for Security-Policy, EndOffset denotes the
```
laurentsimon, line 24: `EndOffset` is typically used to identify the end of the relevant file where a Snippet value was found. In your case, Snippet would be an email. address, for example.

checks/raw/security_policy.go
```diff
@@ -29,23 +30,38 @@ import (
 )
 
 type securityPolicyFilesWithURI struct {
-	uri   string
-	files []checker.File
+	info []checker.SecurityPolicyInformation
+	uri  string
+	file checker.File
```
laurentsimon, line 35: You may have to make fields fields public, so that we can access them from `pkg/` folder.

pkg/json_raw_results.go
```diff
@@ -609,11 +609,9 @@ func (r *jsonScorecardRawResult) addBinaryArtifactRawResults(ba *checker.BinaryA
 //nolint:unparam
 func (r *jsonScorecardRawResult) addSecurityPolicyRawResults(sp *checker.SecurityPolicyData) error {
 	r.Results.SecurityPolicies = []jsonFile{}
-	for _, v := range sp.Files {
-		r.Results.SecurityPolicies = append(r.Results.SecurityPolicies, jsonFile{
-			Path: v.Path,
-		})
-	}
+	r.Results.SecurityPolicies = append(r.Results.SecurityPolicies, jsonFile{
```
laurentsimon, line 612: Ideally information about SecurityPolicyInformationTypeEmail, etc should be reported here. This way someone who fetches results from the REST API (once the raw results are exposed) will be able to parse and create their own policies based on what kind of information was found.
One suggestion it to create a `jsonSecurityFile`:
```golang
type jsonSecurityFile struct{
  jsonFile
  Type string --> this will take a string like SecurityPolicyInformationTypeEmail, etc
}
```

Later we can "fix" the jsonFile structure to contain a list of Snippet/lines. @raghavkaul fyi

laurentsimon (commented):
Thanks. I've added comments. Let me know if they make sense or not.
Sorry for the back and forth. You're basically working on the heart of raw results, and they are not well polished. We can hop on a call if you want.

Thanks

shissam:
I’ll return to this PR when I return from leave

checker/raw_result.go
shissam: changed `"vuln & disclosure text"` to `"vulnDisclosureText"`

checker/raw_result.go
shissam: changed `InformationValue []string` to a new struct to track "telemetry" of matches/hits in the policy file aligning with the `SecurityPolicyInformationType` constants as in:

```
type SecurityPolicyValueType struct {
       Match      string   // Snippet of match
       LineNumber uint     // Line number in policy file of match
       Offset     uint     // Offset in the line of the match
}
```

as such, the new `SecurityPolicyInformation` type tracks this telemetry for each match/hit as in:

```
 type SecurityPolicyInformation struct {
       InformationType  SecurityPolicyInformationType
       InformationValue SecurityPolicyValueType
 }
```

checks/evaluation/security_policy.go
shissam: I agree - my "overload" using `EndOffset` is not well polished. for my proposed security check improvement, I want to ensure that there is some text beyond simply emails and urls (linked content). To do this, I'd like to know just how much content is in the found policy file at the time of scoring.

I have two options as I see it (perhaps there is a better option)

1. add a new attribute/field/member to the struct `File` - something like (`FileLength`) but that would break the purpose used for code snippets--so likely not a good option.
2. (my choice) add an explicit attribute to the `SecurityPolicyData` struct which conveys that files content size

To that end, the new `SecurityPolicyData` struct appears as:
```
type SecurityPolicyData struct {
        // security policy information found in repo or org
        Information []SecurityPolicyInformation
        // total size in bytes of the security policy file contents
        SecurityContentLength uint
        // file that contains the security policy information
        // only looking for one file
        File File
}
```

checks/raw/security_policy.go
shissam: if you are referring to `securityPolicyFilesWithURI`, that struct (internal to `checks/raw/security_policy.go`) were NOT public prior to my changes for this PR

```
type securityPolicyFilesWithURI struct {
        uri   string
        files []checker.File
}

if you want all those fields public from this point on, please let me know and I will make them public---I was only following the pattern I saw when I started this PR.
```

shissam:
today (20 OCT) I thought I had successfully resolved the conflict with `docs/checks.md`, committed that resolution to this PR... but it still shows a conflict which I cannot seem to understand.  On my local git when comparing my `docs/checks.md` with a fresh pull from the official main scorecard repo (branch `main`) I see this:

image

The `diff` seen above is the change I want to make to reflect the change made by this PR.

I am at a loss on as to what I continue to see the "This branch has conflicts that must be resolved" in this PR for this file.

checks/raw/security_policy.go
shissam: closing this comment for now - we can revisit if necessary

pkg/json_raw_results.go
shissam: closing this comment for now - we can revisit if necessary

laurentsimon: tell me what the confusion is. I can try to explain or we can chat offline

checker/raw_result.go
```diff
@@ -136,11 +136,36 @@ type VulnerabilitiesData struct {
...
+	InformationType  SecurityPolicyInformationType
+	InformationValue SecurityPolicyValueType
+}
+
 // SecurityPolicyData contains the raw results
 // for the Security-Policy check.
 type SecurityPolicyData struct {
-	// Files contains a list of files.
-	Files []File
+	// security policy information found in repo or org
```
laurentsimon, line 162: why not just:
```golang
type PolicyFile {
  File File
  InformationType
  ContentLength
}
type SecurityPolicyData struct {
   PolicyFiles []PolicyFile
}
```

shissam: I will do that, just to be specific:

```
type PolicyFile {
   File File
   Information []SecurityPolicyInformation
   ContentLength uint
}
type SecurityPolicyData struct {
   PolicyFiles []PolicyFile
}
```

pkg/json_raw_results.go
shissam: I would like to chat... I simply don't have my head into json raw results - my sense looking here is that I could create a pattern for security policy hits that follows TokenPermissions or DangerousWorkFlow or DependencyPinning which appears to emit Snippet/offset/file information -- but in this case to would be SecurityInformationType/Match/linenumber/File information in a `jsonSecurityFile`

If I got that right - great, I can do that, if I am still missing the point - we should chat

checker/raw_result.go
shissam: one additional comment - looking back on this new spec... for Security-Policy check, it is only looking for ONE policy file (by name), then the search finds an existing file (by name) either in the repo or the org's repo, the search stops--hence there would not be any more than one file.

therefore `PolicyFiles []PolicyFile` should/would/could be simply `PolicyFile PolicyFile` - it was that observation when I started this PR that the `SecurityPolicyData struct` is the way it is with the last commit reviewed here.

pkg/json_raw_results.go
shissam: after studying raw results - I now understand and have changed the raw results to emit all the findings from the security policy check.

checker/raw_result.go
laurentsimon: my reasoning was to re-use the info in File (Match, LineNumber, Offset) instead of creating a new struct for it. I fully agree with you it's not optimal either due to your observation. It's probably fine to keep as-is. 

One thing we should change is allow for reporting multiple policy files. Today the code reports only one policy file, but if we decide to change it in the future we don't want to break this structure. So we just need to make a list of *some-struct* with the struct being the fields you currently have in `SecurityPolicyData`.

checks/raw/security_policy.go
```diff
@@ -112,3 +151,114 @@ func isSecurityPolicyFilename(name string) bool {
...
+		tempType := pfiles.Type
+		policyHits := collectPolicyHits(content)
+
+		*pfiles = checker.File{
+			Path:   path,
+			Type:   tempType,
+			Offset: checker.OffsetDefault,
+			// convey the length/amount of content using
+			// the EndOffset as the len to EOF used in eval
+			EndOffset: uint(len(content)),
```
laurentsimon, line 191: I think it's fine to add a `uint * size` in the File structure. EndOffset is for code snippet. We use in in SARIF, for example.

laurentsimon (approved):
LGTM overall. Left a few comments you can address before merging. Thanks for the PR again!

olivekl (commented):
Doc changes LGTM, thanks for these improvements!

checks/raw/security_policy.go
```diff
@@ -70,7 +81,24 @@ func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error)
 	}
 
 	// Return raw results.
-	return checker.SecurityPolicyData{Files: data.files}, nil
+	if len(data.files) > 0 {
+		for idx := range data.files {
+			filePattern := data.files[idx].File.Path
+			// undo path.Join in isSecurityPolicyFile
+			if data.files[idx].File.Type == checker.FileTypeURL {
+				filePattern = strings.Replace(data.files[idx].File.Path, data.uri+"/", "", 1)
```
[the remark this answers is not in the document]
shissam: I have removed (and tested) needing the actual URI in the path.

checks/raw/security_policy.go
shissam: Added `FileSize` to `File struct` to hold this information thereby allowing removal of `SecurityContentLength` which I think nicely cleaned up that part of the code.

checks/raw/security_policy.go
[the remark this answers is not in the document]
laurentsimon: wdym?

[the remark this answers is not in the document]
shissam: In the original code if the security policy file was not in the actual repo but in the repo's org `.github` folder - the `File.Type` would be set to `checker.FileTypeURL` and the URI would be prepended/joined to the found file name. In the new security policy which looks at the content, when trying to `fileparser.OnMatchingFileContentDo` the method would fail because doing that joining (in the original code) was incorrect--so as a temporary solution, I removed that which was joined so I could get the contents properly. HTH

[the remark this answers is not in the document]
shissam: that is *incorrect* as a parameter to `fileparser.OnMatchingFileContentDo`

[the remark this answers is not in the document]
shissam: I need to put that back the way it was... yes, I need to remove the `uri` portion of the path when getting the contents of the file using `OnMatchingFileContentDo` - but the `uri` portion needs to remain on the path for the `raw` output--when I did that work, I was not aware of the `raw` json output capability, so that part has to be restored. I will do that tonight. 

current official `raw` json reports `path` for an organization security policy as:

```
        "securityPolicies": [
            {
                "path": "github.com/gorilla/.github/SECURITY.md"
            }
```

where the new code on the same site reports `path` without the `uri` as (partial json here):

```
                ...}
                ],
                "path": "SECURITY.md"
```

my bad :-(

spencerschrock:
Sorry about yet another merge conflict. That ones on me. Will take a look at this later today.

shissam:
> Sorry about yet another merge conflict. That ones on me. Will take a look at this later today.

@spencerschrock the one failure I stumbled upon was the introduction of spaces in the stanza near `new e2e attestor` in `.github/workflows/integration.yml` - I repaired that in my PR's workflow just to keep this PR moving, now I have a huge failure which I cannot seem to account for regarding `Timeout of 1800000ms hit` in [https://github.com/ossf/scorecard/actions/runs/3373001227](url) - so, are you talking about the `new e2e attester` problem or the Timeout?

spencerschrock:
> > Sorry about yet another merge conflict. That ones on me. Will take a look at this later today.
> 
> @spencerschrock the one failure I stumbled upon was the introduction of spaces in the stanza near `new e2e attestor` in `.github/workflows/integration.yml` - I repaired that in my PR's workflow just to keep this PR moving

I believe the failure was actually from a PR I merged that had a syntax issue in it which I missed. See #2417 for details.

>  now I have a huge failure which I cannot seem to account for regarding `Timeout of 1800000ms hit` in [https://github.com/ossf/scorecard/actions/runs/3373001227](url)

that timeout is from rate limiting. when we have too many PRs, we exhaust our API quota and the code will sleep until it can continue. You can see it [here](https://github.com/ossf/scorecard/actions/runs/3373001227/jobs/5597103155#step:7:3730)
```
time="2022-11-01T21:32:08Z" level=info msg="Rate limit exceeded. Waiting 40m2.862527371s to retry..."
```

shissam:
> > > Sorry about yet another merge conflict. That ones on me. Will take a look at this later today.
> > 
> > 
> > @spencerschrock the one failure I stumbled upon was the introduction of spaces in the stanza near `new e2e attestor` in `.github/workflows/integration.yml` - I repaired that in my PR's workflow just to keep this PR moving
> 
> I believe the failure was actually from a PR I merged that had a syntax issue in it which I missed. See #2417 for details.
> 
> > now I have a huge failure which I cannot seem to account for regarding `Timeout of 1800000ms hit` in [https://github.com/ossf/scorecard/actions/runs/3373001227](url)
> 
> that timeout is from rate limiting. when we have too many PRs, we exhaust our API quota and the code will sleep until it can continue. You can see it [here](https://github.com/ossf/scorecard/actions/runs/3373001227/jobs/5597103155#step:7:3730)
> 
> ```
> time="2022-11-01T21:32:08Z" level=info msg="Rate limit exceeded. Waiting 40m2.862527371s to retry..."
> ```

understood - I think - with `2 Failed` after reporting `FAIL! - Interrupted by User` I see other errors around lines 3771 where it reports `Expect(err).Should(BeNil())` for `fuzzing_test.go:38` are these also because of the interrupted API? Assuming so, then, how does the job/integration test get restarted or triggered to rerun?

spencerschrock:
> understood - I think - with 2 Failed after reporting FAIL! - Interrupted by User I see other errors around lines 3771 where it reports Expect(err).Should(BeNil()) for fuzzing_test.go:38 are these also because of the interrupted API?

Yep. The interrupted by user is from the 30m timeout stoping the other tests. 

> Assuming so, then, how does the job/integration test get restarted or triggered to rerun?

Basically any change to your branch, a new commit, merge, rebase etc. Or a maintainer can manually re-run jobs.

shissam:
> > understood - I think - with 2 Failed after reporting FAIL! - Interrupted by User I see other errors around lines 3771 where it reports Expect(err).Should(BeNil()) for fuzzing_test.go:38 are these also because of the interrupted API?
> 
> Yep. The interrupted by user is from the 30m timeout stoping the other tests.
> 
> > Assuming so, then, how does the job/integration test get restarted or triggered to rerun?
> 
> Basically any change to your branch, a new commit, merge, rebase etc. Or a maintainer can manually re-run jobs.

changes for the workflow were successful - thanks!

spencerschrock (approved):

checker/raw_result.go
```diff
@@ -236,6 +262,7 @@ type File struct {
 	Snippet   string   // Snippet of code
 	Offset    uint     // Offset in the file of Path (line for source/text files).
 	EndOffset uint     // End of offset in the file, e.g. if the command spans multiple lines.
+	FileSize  uint     // Total size of file.
```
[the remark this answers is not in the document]
shissam: `FileSize` vs. `Size` - this is a bit of a discombobulation for me. Yes, `FileSize` *is* in the `File` struct, but reading the entire struct it is really about the `Snippet` and its (beginning) `Offset` (in the `File`) and the end of the offset for that `Snippet` in the file (`EndOfOffset`). So, reporting the size as `FileSize` it is clear (or clearer for me that) this is the size of the `File` (hence `File.FileSize` — yeah, at the expense of being repetative) and avoids potential confusion of `Size` (like sibling elements of `Offset` and `EndOffset`) being the size of the `Snippet` (which is computable by `EndOffset - Offset` anyhow).  But I could change — it’s y’alls final call.

checks/evaluation/security_policy.go
```diff
@@ -19,6 +19,105 @@ import (
...
+}
+
+func countSecInfo(secInfo []checker.SecurityPolicyInformation,
+	infoType checker.SecurityPolicyInformationType,
+	unique bool,
+) int {
+	keys := make(map[string]bool)
+	count := 0
+	for _, entry := range secInfo {
+		if _, value := keys[entry.InformationValue.Match]; !value && entry.InformationType == infoType {
```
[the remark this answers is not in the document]
shissam: concur - much better (I like `!present`)

checks/evaluation/security_policy.go
```diff
@@ -19,6 +19,105 @@ import (
...
+}
+
+func findSecInfo(secInfo []checker.SecurityPolicyInformation,
+	infoType checker.SecurityPolicyInformationType,
+	unique bool,
+) []checker.SecurityPolicyInformation {
+	keys := make(map[string]bool)
+	var secList []checker.SecurityPolicyInformation
+	for _, entry := range secInfo {
+		if _, value := keys[entry.InformationValue.Match]; !value && entry.InformationType == infoType {
```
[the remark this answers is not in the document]
shissam: concur - much better (I like `!present`)

checks/evaluation/security_policy.go
```diff
@@ -19,6 +19,105 @@ import (
...
+		linkedContentLen += len(i.InformationValue.Match)
+	}
+
+	msg := checker.LogMessage{
+		Path: f.Path,
+		Type: f.Type,
+		Text: "",
+	}
+
+	// #1: more than one unique (email/http) linked content found: score += 6
```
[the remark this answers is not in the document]
shissam: concur (i did miss the comment on the change)--done!

checks/evaluation/security_policy.go
```diff
@@ -19,6 +19,105 @@ import (
...
+		msg.Text = "No text (beyond any linked content) found in security policy"
+		dl.Warn(&msg)
+	}
+
+	// #3: found whole number(s) and or match(es) to "Disclos" and or "Vuln": score += 1
+	//     rationale: works towards the intent of the security policy file
+	//     regarding whom to contact about vuls and disclosures and timing
+	//     e.g., we'll disclose, report a vulnerabily, 30 days, etc.
+	//     looking for at least 2 hits
+	if discvuls > 1 {
```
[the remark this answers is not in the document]
shissam: on this final `+1` the regex is looking for digits and portions of the word ‘vuln’ and ‘disclos’ (seen in `checks/raw/security_policy.go`). In this check, there is no test for uniqueness. I did not want to trigger a false positive here where some text have a number in it (like “this is my 1 st security policy”) and nothing else (but I would grant a `+1` with “this is my 1 st and 2 nd security policy”). The “gold” test is looking for hints like “… 90 days for disclosure “.

I felt a single hit was just too easy. Now, the risk is a `SECURITY.md` which has “To report a vulnerability send email to fred@flinstones.com” - this would give you 9 points (the length of free text (+3), the email (+6)), to get the full 10 points you’d improve that to “To report a vulnerability send email to fred@flinstones.com and expect resolution within 90 days” - that’d get you the full 10 points. For me, I was inspired by the Security-Policy check risk statement “possible insecure reporting of vulnerabilities” — so looking for “enough” hints (more than just 1) was my rationale without getting into NLP or following links to other websites to scrape what might be in those files.

Not sure this is resolved here (yet)

docs/checks/internal/checks.yaml
```diff
@@ -559,6 +559,10 @@ checks:
       A security policy (typically a `SECURITY.md` file) can give users information
       about what constitutes a vulnerability and how to report one securely so that
       information about a bug is not publicly visible.
+
+      This check examines the contents of the security policy file awarding points
+      for those policies that express vulnerability process(es), disclosure timelines,
+      and have links (e.g., URL(s) and email(s)) to support the users.
```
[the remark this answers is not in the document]
shissam: concur - updated using inspiration from branch protection.

docs/checks.md
```diff
@@ -527,6 +527,10 @@ well-known directories.
 A security policy (typically a `SECURITY.md` file) can give users information
 about what constitutes a vulnerability and how to report one securely so that
 information about a bug is not publicly visible.
+
+This check examines the contents of the security policy file awarding points
```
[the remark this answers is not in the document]
shissam: done.

checks/raw/security_policy.go
```diff
@@ -30,22 +32,30 @@ import (
...
 	data := securityPolicyFilesWithURI{
-		uri:   "",
-		files: make([]checker.File, 0),
+		uri: "", files: make([]checker.SecurityPolicyFile, 0),
 	}
 	err := fileparser.OnAllFilesDo(c.RepoClient, isSecurityPolicyFile, &data)
 	if err != nil {
 		return checker.SecurityPolicyData{}, err
 	}
 	// If we found files in the repo, return immediately.
```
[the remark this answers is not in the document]
spencerschrock: No, you're right. I was finding the immediately to be conflicting with the `checkSecurityPolicyFileContent` work. But compared to the parent org checks, it still returns early.

spencerschrock:
Thanks @shissam !

PR #2202, opened 2022-08-26
✨  Allow contents: write for Token-Permissions when doing mvn release 

katzj:

This allows workflow steps that call 'mvn release:prepare' to use `contents: write` permissions as they are needed for doing the maven release process.

Somewhere between a bug and a feature

Current behavior is that a workflow that is doing mvn release will get dinged on the Token-Permissions check due to `contents: write` permission for that workflow

A workflow using `mvn release:prepare` will be allowed the `contents: write` permission without lowering the score

NONE

```release-note
The Token-Permissions check no longer penalizes workflows with the `contents: write` permission if they are using it to run `mvn release:prepare`.
```

### DIFF
--- a/checks/raw/permissions.go
+++ b/checks/raw/permissions.go
@@ -491,6 +491,16 @@ func isReleasingWorkflow(workflow *actionlint.Workflow, fp string, pdata *permis
 			},
 			LogText: "candidate SLSA publishing workflow using slsa-github-generator",
 		},
+		{
+			// Running mvn release:prepare requires committing changes.
+			// https://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html
+			Steps: []*fileparser.JobMatcherStep{
+				{
+					Run: ".*mvn.*release:prepare.*",
+				},
+			},
+			LogText: "candidate mvn release workflow",
+		},
 	}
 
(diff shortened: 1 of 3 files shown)


checks/raw/permissions.go
```diff
@@ -491,6 +491,16 @@ func isReleasingWorkflow(workflow *actionlint.Workflow, fp string, pdata *permis
 			},
 			LogText: "candidate SLSA publishing workflow using slsa-github-generator",
 		},
+		{
+			// Running mvn release:prepare requires committing changes.
+			// https://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html
+			Steps: []*fileparser.JobMatcherStep{
+				{
+					Run: ".*mvn.*release:prepare.*",
```
laurentsimon, line 499: should this be `mvnw` instead of `mvn`?

laurentsimon (approved):
Thank you, this is great!

checks/raw/permissions.go
katzj: Not everyone uses mvnw; some people just depend on an installed mvn instead. The regex will allow both to work

laurentsimon: Got it, thanks.

PR #2207, opened 2022-08-28
:seedling: Fix for empty repository

naveensrinivasan:
- Fixes the score to `0` for an empty repository.
- Fixes https://github.com/ossf/scorecard/issues/2157

### DIFF
--- a/pkg/scorecard.go
+++ b/pkg/scorecard.go
@@ -74,7 +74,7 @@ func getRepoCommitHash(r clients.RepoClient) (string, error) {
 	if len(commits) > 0 {
 		return commits[0].SHA, nil
 	}
-	return "no commits found", nil
+	return "", nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/scorecard.go
```diff
@@ -98,6 +98,10 @@ func RunScorecards(ctx context.Context,
 	if err != nil {
 		return ScorecardResult{}, err
 	}
+	// zero commits which is not an error and return an empty ScorecardResult
+	if commitSHA == "" {
+		return ScorecardResult{}, err
+	}
```
azeemshaikh38, line 104: Nit: simplify to:

```
if err != nil && commitSHA == "" {
  return ScorecardResult{}, err
}
```

azeemshaikh38 (approved):

pkg/scorecard.go
naveensrinivasan: Thanks addressed it.

Issue #2160, opened 2022-08-17
Improve Score Reporting: Check for maintainer activity upto last year


azeemshaikh38:
Maintained check only looks for activity within the last 90 days which might be too short of a time frame for stable projects and we unfairly penalize them. Let's increase the activity period we look for to 1 year.

LappleApple:
Looks like we would need to do more testing to surface impact. @spencerschrock's Golden testing tool should help here.

spencerschrock:
API quota may or may not be relevant here. We fetch 30 commits, and currently only count commits newer than 90d. The same would be true if its 1 year. (similar for issues i believe)

What may need to change is the scoring threshold, (1 commit per week is mentioned somewhere, which over a year long period is 52 commits, which is more than we look back).

PR #2216, opened 2022-08-29
:seedling: Fix maintainer activity upto last year

naveensrinivasan:
- Fixes https://github.com/ossf/scorecard/issues/2160

### DIFF
--- a/checks/evaluation/maintained.go
+++ b/checks/evaluation/maintained.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	lookBackDays    = 90
+	lookBackDays    = 365
 	activityPerWeek = 1
 	daysInOneWeek   = 7
 )
--- a/docs/checks/internal/checks.yaml
+++ b/docs/checks/internal/checks.yaml
@@ -25,7 +25,7 @@ checks:
 
       This check determines whether the project is actively maintained. If the project
       is archived, it receives the lowest score. If there is at least one commit per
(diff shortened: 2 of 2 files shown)


laurentsimon (approved):
Thank. Hope this does not increase API usage too much.

azeemshaikh38 (changes requested):
Could we hold off on these changes? This would completely change the output displayed on deps.dev so would be good to do a bit of analysis before merging this in. One simple analysis might be to use the RawData dumped out for 1M+ repos and get an histogram of last updated activity across the ecosystem. It should quickly tell us how changing to 1year will reflect on deps.dev. What do folks think?

naveensrinivasan:
> Could we hold off on these changes? This would completely change the output displayed on deps.dev so would be good to do a bit of analysis before merging this in. One simple analysis might be to use the RawData dumped out for 1M+ repos and get an histogram of last updated activity across the ecosystem. It should quickly tell us how changing to 1year will reflect on deps.dev. What do folks think?

Sounds good. 

On a different note, Can we expose the raw data in the scorecards API? We have it in the bucket.

azeemshaikh38:
> On a different note, Can we expose the raw data in the scorecards API? We have it in the bucket.

I don't think its in a state to be made public yet - some checks are yet to be migrated to raw data and we don't have a clearly defined schema for the results. Probably best to wait till we work these out.

PR #2233, opened 2022-09-07
:book: Rename CII to OpenSSF

david-a-wheeler:
The "CII Best Practices badge" project was renamed to "OpenSSF Best Practices badge" a while back. Update this page to reflect the update. Also, update the URL so it uses the user's preferred language ("/en" forces English).

Docs update.


olivekl (approved):
Thanks for catching both of these!

PR #2235, opened 2022-09-08
🌱 cron: allow controller to read CSVs from cloud storage (1/n)

spencerschrock:

The cron job controller uses `cron/internal/data/projects.csv` as input for projects to analyze in the cron infrastructure.

The controller will now read multiple CSV files from a configurable bucket if ran with no additional arguments. 
If passed one additional argument (which is how the controller is currently configured), the behavior is identical to the current behavior.
If passed multiple additional arguments, multiple local CSV files will be used to populate the pubsub topic.

NONE

the `release-note`

```release-note
NONE
```


spencerschrock:
@azeemshaikh38 @calebbrown 

This is the current state which handles importing multiple CSV files, either locally or from a single bucket (but not both).

Possible enhancements (could do both)
1. Structure the YAML in a way to define multiple buckets and pull CSVs from all of them
2. Pull from local files **AND** bucket files in the same run


What is the expected scale of project URLs? Currently the bucket implementation stores all files in the bucket in memory, which isn't a problem for the ~68MB scorecard `projects.csv` file, but could be a problem in the future.

cron/internal/data/iterator.go
```diff
@@ -64,3 +64,30 @@ func (reader *csvIterator) Next() (RepoFormat, error) {
...
+		if i.iterators[i.current].HasNext() {
+			return true
+		}
+		i.current++
+	}
+	return false
+}
+
+func (i *nestedIterator) Next() (RepoFormat, error) {
+	r, err := i.iterators[i.current].Next()
```
azeemshaikh38, line 88: Replace with `return i.iterators[i.current].Next()` ?

azeemshaikh38 (approved):
Looks great!

cron/internal/data/iterator.go
spencerschrock: I originally had that, but the `wrapcheck` linter complained.
> error returned from interface method should be wrapped: sig: func

calebbrown:
Just a heads up, this adds support for CSV files that use the `data.RepoFormat` struct, and require a header row that matches.

Currently the Criticality Score enumerate_github tool outputs a TXT file with no header row - so I suspect this won't work for the current output as-is.

oliverchang:
Would it make sense to support not having header rows, especially given that this is a machine readable format?

azeemshaikh38:
> Would it make sense to support not having header rows, especially given that this is a machine readable format?

@spencerschrock fyi - https://github.com/jszwec/csvutil#but-my-csv-file-has-no-header-

spencerschrock:
Will take a look and address in another PR

PR #2250, opened 2022-09-09
🐛 Wrap check errors with distinct error for scorecard-action to ignore.

spencerschrock:

bug fix

Runtime errors in individual checks bubble up to any cobra command created via `cmd.New()` after #2133
This unintentionally broke scorecard-action https://github.com/ossf/scorecard-action/issues/856

Runtime errors are wrapped in a new error, `ErrorCheckRuntime` so that `scorecard-action` can ignore them. Which will be implemented in a separate PR after the next scorecard release.

This is a pretty difficult thing to test, forcing runtime errors to occur in a check.

Partially fixes https://github.com/ossf/scorecard-action/issues/856

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ossf/scorecard/v4/checker"
 	"github.com/ossf/scorecard/v4/clients"
 	docs "github.com/ossf/scorecard/v4/docs/checks"
+	sce "github.com/ossf/scorecard/v4/errors"
 	sclog "github.com/ossf/scorecard/v4/log"
 	"github.com/ossf/scorecard/v4/options"
 	"github.com/ossf/scorecard/v4/pkg"
@@ -162,7 +163,7 @@ func rootCmd(o *options.Options) error {
 	// intentionally placed at end to preserve outputting results, even if a check has a runtime error
 	for _, result := range repoResult.Checks {
 		if result.Error != nil {
-			return fmt.Errorf("one or more checks had a runtime error: %w", result.Error)
+			return sce.WithMessage(sce.ErrorCheckRuntime, fmt.Sprintf("%s: %v", result.Name, result.Error))
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -162,7 +163,7 @@ func rootCmd(o *options.Options) error {
 	// intentionally placed at end to preserve outputting results, even if a check has a runtime error
 	for _, result := range repoResult.Checks {
 		if result.Error != nil {
-			return fmt.Errorf("one or more checks had a runtime error: %w", result.Error)
+			return sce.WithMessage(sce.ErrorCheckRuntime, fmt.Sprintf("%s: %v", result.Name, result.Error))
```
laurentsimon, line 166: do we need to make this a list of errors, since multiple checks may have failed?

laurentsimon (approved):

cmd/root.go
spencerschrock: It depends on what we want. For the purposes of `scorecard-action` the error is just getting matched with `errors.Is` so the details don't matter.
It also wasn't necessary in #2133 since the info for all errors were visible if you ran with `--show-details`. The main point was returning a singular error so the status code would be 1 when any checks had failures.

laurentsimon: Thanks for the clarification. I suppose the problem is if there are multiple checks failing, each with a different error. Is it a concern?

spencerschrock: From a behavior point of view: no. As long as there is one error, the CLI exit code is non-zero and `scorecard-action` can ignore the errors. The code is at the very end of `rootCmd` so it's purely information for status code.

From an informational point of view, collating multiple errors would benefit CLI users that don't have `--show-details`, but that's about it. The error info still makes it to `scorecard-action`. For example, if you look at the [current badge for scorecard](https://api.securityscorecards.dev/projects/github.com/ossf/scorecard), you'll see the info already there:  `{"name":"Branch-Protection","score":-1,"reason":"internal error: error during branchesHandler.setup: .......`

laurentsimon: One limitation is that scorecard action does not have access to a structured version of the results, just to a file. Correct? So I'm wondering if it's a problem if the Action filters based on the single error, even though there may be multiple distinct errors.
I agree it's fine for a human who can read the results.

laurentsimon: Ignore my comment: action will only filter for `ErrorCheckRuntime`, so not a problem.

PR #2251, opened 2022-09-09
:seedling: Format the JSON output

naveensrinivasan:
- This helps with this issue https://github.com/ossf/scorecard-webapp/issues/206

### DIFF
--- a/cron/internal/format/json.go
+++ b/cron/internal/format/json.go
@@ -86,6 +86,7 @@ type jsonScorecardResultV2 struct {
 // AsJSON exports results as JSON for new detail format.
 func AsJSON(r *pkg.ScorecardResult, showDetails bool, logLevel log.Level, writer io.Writer) error {
 	encoder := json.NewEncoder(writer)
+	encoder.SetIndent("", "  ")
 
 	out := jsonScorecardResult{
 		Repo:     r.Repo.Name,
--- a/cron/internal/format/json_raw_results.go
+++ b/cron/internal/format/json_raw_results.go
@@ -265,6 +265,7 @@ func fillJSONRawResults(r *jsonScorecardRawResult, raw *checker.RawResults) erro
 // AsRawJSON exports results as JSON for raw results.
 func AsRawJSON(r *pkg.ScorecardResult, writer io.Writer) error {
 	encoder := json.NewEncoder(writer)
+	encoder.SetIndent("", "  ")
 	out := jsonScorecardRawResult{
 		Repo: jsonRepoV2{
 			Name:   r.Repo.Name,
--- a/pkg/json.go
+++ b/pkg/json.go
@@ -86,6 +86,7 @@ type JSONScorecardResultV2 struct {
 // AsJSON exports results as JSON for new detail format.
 func (r *ScorecardResult) AsJSON(showDetails bool, logLevel log.Level, writer io.Writer) error {
 	encoder := json.NewEncoder(writer)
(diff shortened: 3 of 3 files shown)


pkg/json.go
```diff
@@ -124,6 +124,7 @@ func (r *ScorecardResult) AsJSON2(showDetails bool,
 	}
 
 	encoder := json.NewEncoder(writer)
+	encoder.SetIndent("", "  ")
```
laurentsimon, line 127: There are tools like BigQuery transfers that expect the JSON output to be on a single line. I'm not sure it's our job to prettify the result (breaking change). I fully agree we need more human-friendly results for the badges; but is this the right place for the fix?

naveensrinivasan: This is a start. @azeemshaikh38 will this cause breaking charges for BigQuery?

laurentsimon: I dont think it will for our integration, because we JSONify the results in a different file (under cron/). Some users may be using this output, though.

azeemshaikh38: +1. Very likely that some users get broken by this.

naveensrinivasan: OK, If that is the case. I will close it 👍

naveensrinivasan: So @azeemshaikh38 If I change the BQ will it be an issue?

naveensrinivasan: @azeemshaikh38  We discussed this in our bi-weekly. We agreed formatting the JSON output shouldn't be an issue.

cron/internal/format/json.go
```diff
@@ -86,6 +86,7 @@ type jsonScorecardResultV2 struct {
 // AsJSON exports results as JSON for new detail format.
 func AsJSON(r *pkg.ScorecardResult, showDetails bool, logLevel log.Level, writer io.Writer) error {
 	encoder := json.NewEncoder(writer)
+	encoder.SetIndent("", "  ")
```
spencerschrock, line 89: > If I change the BQ will it be an issue?

As far as I know, yes. [BQ expects newline delimited JSON](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json) and refers to the [linked spec](https://jsonlines.org/).

"Each Line is a Valid JSON Value". I can try to confirm with a test upload, but I expect this to break the BQ transfer.

spencerschrock: The files in the API bucket don't get transferred as far as I can tell. Would a pretty-print argument (bool?) be acceptable?

cron/internal/format/json_raw_results.go
```diff
@@ -265,6 +265,7 @@ func fillJSONRawResults(r *jsonScorecardRawResult, raw *checker.RawResults) erro
 // AsRawJSON exports results as JSON for raw results.
 func AsRawJSON(r *pkg.ScorecardResult, writer io.Writer) error {
 	encoder := json.NewEncoder(writer)
+	encoder.SetIndent("", "  ")
```
spencerschrock, line 268: same as above comment

spencerschrock:
This PR fell off, but for future attempts I'm not sure this was the right place to implement it. The [webapp](https://github.com/ossf/scorecard-webapp) unmarshalls the file, so I don't think the output in the bucket is what matters.

https://github.com/ossf/scorecard-webapp/blob/e105bd123d601b29b676f18988e3c0a64f8b893b/app/server/get_results.go#L40-L52

PR #2254, opened 2022-09-12
🌱 cron: generalize config and create optional values for scorecard and criticality (2/n)

spencerschrock:

Every config value was a top level yaml field

Only fields which are general enough to be re-used both by `scorecard` and `criticality_score` remain as top level fields. Anything specific to either program have been moved into nested fields called `scorecard` and `criticality`.  These will be parsed as key, value pairs (both strings). 

The values can still be overwritten by environment variables. Consider the example below, which can be overridded by setting the environment variable `SCORECARD_EXAMPLE_FIELD`:
```
scorecard:
  example-field: foo
```

NONE

the `release-note`

```release-note
NONE
```


spencerschrock:
@calebbrown @oliverchang fyi

The map can be fetched with `config.GetCriticalityValues()`. And the PR/code documents how to override the default values with environment variables. When you add your config values, some unit tests will need to change, but that should more or less be it.

cron/internal/config/config.go
```diff
@@ -66,24 +61,20 @@ var (
...
+	BigQueryDataset        string            `yaml:"bigquery-dataset"`
+	BigQueryTable          string            `yaml:"bigquery-table"`
+	CompletionThreshold    float32           `yaml:"completion-threshold"`
+	WebhookURL             string            `yaml:"webhook-url"`
+	MetricExporter         string            `yaml:"metric-exporter"`
+	ShardSize              int               `yaml:"shard-size"`
+	InputBucketURL         string            `yaml:"input-bucket-url"`
+	InputBucketPrefix      string            `yaml:"input-bucket-prefix"`
+	Criticality            map[string]string `yaml:"criticality"`
+	Scorecard              map[string]string `yaml:"scorecard"`
```
azeemshaikh38, line 77: This can just be `AdditionalParams map[string]string` ?

azeemshaikh38 (approved):

azeemshaikh38:
Thanks! LGTM with comment

cron/internal/config/config.go
spencerschrock: It could be yes. I think the benefit of having them separate would be the env var prefixes. Something like

```
criticality:
  foo: bar
scorecard:
  foo: bar
```

would be set with `CRITICALITY_FOO` and `SCORECARD_FOO`, respectively. But I could always rename individual fields:

```
optional-params:
  scorecard-foo: bar
  criticality-foo: bar
```

I think something like that could work, but I would need to make changes such that the env name is based entirely off the key, so you end up with  `SCORECARD_FOO` instead of `OPTIONAL_PARAMS_SCORECARD_FOO`. WDUT?

spencerschrock: A third option is:

```
optional-params:
  scorecard:
    foo: bar
  criticality:
      foo: bar
```
Which would basically just cut out the repetition of the project prefix. I would need to explore how the parsing would happen though.

cron/internal/config/config.go
```diff
@@ -66,24 +61,20 @@ var (
...
-	InputBucketPrefix      string `yaml:"input-bucket-prefix"`
+	ProjectID              string            `yaml:"project-id"`
+	ResultDataBucketURL    string            `yaml:"result-data-bucket-url"`
+	RequestTopicURL        string            `yaml:"request-topic-url"`
+	RequestSubscriptionURL string            `yaml:"request-subscription-url"`
+	BigQueryDataset        string            `yaml:"bigquery-dataset"`
+	BigQueryTable          string            `yaml:"bigquery-table"`
+	CompletionThreshold    float32           `yaml:"completion-threshold"`
+	WebhookURL             string            `yaml:"webhook-url"`
+	MetricExporter         string            `yaml:"metric-exporter"`
```
azeemshaikh38, line 72: `MetricExporter` can be optional param too?

cron/internal/config/config.go
azeemshaikh38: I just realized the env vars for the required params also start with `SCORECARD_`. We should rename these to be a bit more generic. And having `OPTIONAL_PARAMS_SCORECARD_FOO` should be fine IMO.

It might take some refactoring, so let's do this after the criticality pipeline work is unblocked.

PR #2260, opened 2022-09-13
🌱 Feature: Group commits into changesets

raghavkaul:

Introduce a `Changeset` construct, which contains several Commits that were reviewed as a unit, along with any associated review activity.

Today, scorecard inconsistently uses either Commits or Pull Requests. Commits are more universal than PRs, but evaluating each Commit in a vacuuum may lead to scoring incorrectly when there are multiple, unsquashed commits in a repo's history which all really were part of a singular "code review". This is also the case with CI test runs, where we look at tests run on individual commits, when tests are really run on all commits that are part of an unsquashed PR.

Scorecard will batch all related commits that are part of a single GitHub Merge request or Phabricator differential revision into a single Changeset. If a commit is not linked to any external SCM or we can't find the associated code review, the commit is part of a standalone changeset.

* Not all SCMs are supported equally
* For Phabricator, we now consider a Differential Revision "reviewed" even if there is no "Reviewed By" line
* Scoring should not be affected by this PR

```release-note
Raw Results: Commits are now grouped into changesets and associated with review activity, if any
```

### DIFF
--- a/checks/code_review.go
+++ b/checks/code_review.go
@@ -37,7 +37,7 @@ func init() {
 
 // CodeReview will check if the maintainers perform code review.
 func CodeReview(c *checker.CheckRequest) checker.CheckResult {
-	rawData, err := raw.CodeReview(c.RepoClient)
+	rawData, err := raw.CodeReview(c.RepoClient, c.Dlogger)
 	if err != nil {
 		e := sce.WithMessage(sce.ErrScorecardInternal, err.Error())
 		return checker.CreateRuntimeErrorResult(CheckCodeReview, e)
--- a/checks/code_review.go
+++ b/checks/code_review.go
@@ -37,7 +37,7 @@ func init() {
 
 // CodeReview will check if the maintainers perform code review.
 func CodeReview(c *checker.CheckRequest) checker.CheckResult {
-	rawData, err := raw.CodeReview(c.RepoClient, c.Dlogger)
+	rawData, err := raw.CodeReview(c.RepoClient)
 	if err != nil {
 		e := sce.WithMessage(sce.ErrScorecardInternal, err.Error())
 		return checker.CreateRuntimeErrorResult(CheckCodeReview, e)
--- a/checks/code_review_test.go
+++ b/checks/code_review_test.go
@@ -229,7 +229,7 @@ func TestCodereview(t *testing.T) {
 				},
 			},
 			expected: checker.CheckResult{
-				Score: 0,
+				Score: checker.MaxResultScore,
 			},
 		},
 		{
--- a/checks/evaluation/code_review.go
+++ b/checks/evaluation/code_review.go
@@ -23,78 +23,38 @@ import (
 	sce "github.com/ossf/scorecard/v4/errors"
 )
 
+type ReviewScore = int
+
+// TODO More partial credit? E.g. approval from non-contributor, discussion liveness,
+// number of resolved comments, number of approvers (more eyes on a project)
 const (
-	reviewPlatformGitHub      = "GitHub"
-	reviewPlatformProw        = "Prow"
-	reviewPlatformGerrit      = "Gerrit"
-	reviewPlatformPhabricator = "Phabricator"
-	reviewPlatformPiper       = "Piper"
+	NoReview              ReviewScore = 0 // No approving review by contributors before merge
+	Reviewed                          = 1 // Changes were reviewed by contributor w/ write access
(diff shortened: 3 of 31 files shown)


checks/code_review.go
```diff
@@ -37,7 +37,7 @@ func init() {
 
 // CodeReview will check if the maintainers perform code review.
 func CodeReview(c *checker.CheckRequest) checker.CheckResult {
-	rawData, err := raw.CodeReview(c.RepoClient)
+	rawData, err := raw.CodeReview(c.RepoClient, c.Dlogger)
```
laurentsimon, line 40: Can we avoid passing the logger here? The raw result collects data only, it should be the evaluator's job to create the final detail information. Also note that raw results don't use the Details in https://github.com/ossf/scorecard/blob/main/pkg/json_raw_results.go

azeemshaikh38:
I took a quick glance, will take a closer tomorrow and provide my detailed feedback. 

Just wanted to comment to say this is an awesome change and will really help improve Scorecard results - thanks for working through this!

checks/raw/code_review.go
```diff
@@ -15,21 +15,210 @@
...
+			j += 1
+		}
+		i = j
+	}
+
+	// Add data to Changeset raw result (if available)
+	// Do this per changeset, instead of per-commit, to save effort
+	for i := range changesets {
+		augmentChangeset(&changesets[i])
+	}
```
azeemshaikh38, line 206:
```suggestion
	for i := range commits {		
		plat, rev, err := getCommitRevisionByPlatform(&commits[i])
		if err != nil {
		  // Handle error.
		}
		
		if exists, changeset := foundChangesets[reviewPlatform][revisionID]; !exists {
		    foundChangesets[reviewPlatform][revisionID] = ChangeSet{
		      // Add revision and commit info
		    })   
		    continue
		}
		
		// Part of a previously found changeset.
		changesetCommits := foundChangesets[reviewPlatform][revisionID].Commits
		changesetCommits = append(changesetCommits, commit[i])
        }

```

checks/raw/code_review.go
```diff
@@ -15,21 +15,210 @@
...
+		Text: fmt.Sprintf("commit %s reviewed through piper revision %s", c.SHA, match[1]),
+	})
+
+	return match[1]
+}
+
+func getCommitRevisionByPlatform(
+	c *clients.Commit,
+	dl checker.DetailLogger,
+) (string, string, error) {
```
azeemshaikh38, line 139: It might be helpful to define a struct which holds both `reviewPlatform` and `revisionID`. It'll improve the return signature of this function and also can be used to define helper fns like: `newlyDefinedStruct.isEqual()`

clients/pull_request.go
```diff
@@ -37,6 +37,13 @@ type Label struct {
 
 // Review represents a PR review.
 type Review struct {
```
azeemshaikh38, line 39: IIUC, none of the changes in `clients` are being used in this PR. Can we remove these unsused changes from this PR?

checks/evaluation/code_review.go
```diff
@@ -16,210 +16,59 @@ package evaluation
...
+// number of resolved comments, number of approvers (more eyes on a project).
 const (
-	reviewPlatformGitHub      = "GitHub"
-	reviewPlatformProw        = "Prow"
-	reviewPlatformGerrit      = "Gerrit"
-	reviewPlatformPhabricator = "Phabricator"
-	reviewPlatformPiper       = "Piper"
+	NoReview              ReviewScore = 0 // No approving review before merge
+	Reviewed              ReviewScore = 1 // Changes were reviewed
+	ReviewedOutsideGithub ReviewScore = 1 // Full marks until we can check review platforms outside of GitHub
```
azeemshaikh38, line 31: No need to expose these consts?

azeemshaikh38 (approved):
LGTM with comments. Thanks for fixing this!

laurentsimon (approved):
LGTM

checks/code_review_test.go
```diff
@@ -229,7 +229,7 @@ func TestCodereview(t *testing.T) {
 				},
 			},
 			expected: checker.CheckResult{
-				Score: 0,
+				Score: checker.MaxResultScore,
```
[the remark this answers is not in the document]
raghavkaul: We automatically give full points if the commit was associated with a Phabricator revision even if there's no Reviewed-By. For now, everyone using Phabricator + GitHub will get a score bump - since this is only on the default branch, we already expect that those commits have a level of scrutiny even if they haven't been reviewed on Phabricator.

PR #2261, opened 2022-09-13
🌱 cron: make CSV header optional (3/n)

spencerschrock:

CSV files read by the controller needed to have a header line:
```
repo,metadata
```

The header line is now optional, and is skipped by the `data.csvIterator`.

Follow up PR for #2235 

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cron/internal/data/iterator.go
+++ b/cron/internal/data/iterator.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"reflect"
 
 	"github.com/jszwec/csvutil"
 
@@ -36,7 +37,11 @@ type Iterator interface {
 func MakeIteratorFrom(reader io.Reader) (Iterator, error) {
 	csvReader := csv.NewReader(reader)
 	csvReader.Comment = '#'
-	dec, err := csvutil.NewDecoder(csvReader)
+	header, err := csvutil.Header(RepoFormat{}, "csv")
+	if err != nil {
+		return nil, fmt.Errorf("error in csvutil.Header: %w", err)
+	}
+	dec, err := csvutil.NewDecoder(csvReader, header...)
 	if err != nil {
 		return nil, fmt.Errorf("error in csvutil.NewDecoder: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


spencerschrock:
@oliverchang @calebbrown fyi

cron/internal/data/iterator.go
```diff
@@ -47,10 +52,29 @@ type csvIterator struct {
...
+	}
 }
 
 func (reader *csvIterator) HasNext() bool {
 	reader.err = reader.decoder.Decode(&reader.next)
+	// only compares the first read against the csv header for performance
+	if reader.count == 0 {
+		reader.ignoreHeader()
+	}
+	reader.count++
```
azeemshaikh38, line 77:
```suggestion
	reader.err = reader.ignoreIfHeader(); 
	reader.err = reader.decoder.Decode(&reader.next)
```

azeemshaikh38, line 77: `reader.count` can simply be a `bool` which is updated inside `ignoreIfHeader` fn?

azeemshaikh38 (approved):
LGTM with comments.

cron/internal/data/iterator.go
spencerschrock: I'll make it similar to this, but I'll need to swap the order of the two calls back to what I had since `reader.ignoreIfHeader()` relies on `decoder.Decode` being called.

azeemshaikh38: My suggestion was to avoid the `decoder.Decode` call inside `ignoreIfHeader`. Could `ignoreIfHeader` simply `return reflect.DeepEqual()` ?

azeemshaikh38: Or maybe, this should be:

```
if header, reader.err := reader.isHeader(); header {
  _ = reader.decoder.Decode(nil) // ignore header
}
reader.err = reader.decoder.Decode(&reader.next)
```

Wdyt?

azeemshaikh38: Sorry, I see what you mean. Ignore me.

spencerschrock:
```go
// returns true on the first call if the most recently decoded record is a header.
// always returns false on subsequent calls for performance.
func (reader *csvIterator) isHeader() bool {
	if reader.afterHeader {
		return false
	}
	header, err := csvutil.Header(RepoFormat{}, "csv")
	if err != nil {
		reader.err = err
		return false
	}
	lastRead := reader.decoder.Record()
	reader.afterHeader = true
	return reflect.DeepEqual(header, lastRead)
}

func (reader *csvIterator) HasNext() bool {
	reader.err = reader.decoder.Decode(&reader.next)
	if reader.isHeader() {
		reader.err = reader.decoder.Decode(&reader.next)
	}
	return !errors.Is(reader.err, io.EOF)
}
```
something like that?

azeemshaikh38: Yep, that LGTM

PR #2279, opened 2022-09-20
🌱 cron: pass config as an argument to binaries (4/n)

spencerschrock:

The `cron/internal/config/config.yaml` file is embedded inside cron binaries, and packaged with the various docker files.
This was not flexible for re-using the cron infrastructure across projects.

1. All cron binaries which interact with the `config` package now support a `config` flag.
```
  --config string
        Location of config file (default "/etc/scorecard/config.yaml")
```

2. The config file is now volume mounted for k8s cronjobs at `/etc/scorecard/config.yaml`

NONE

I made a few naming choices, (e.g. storing configs at `/etc/scorecard`), feel free to nit and I can patch.

the `release-note`

```release-note
NONE
```


cron/internal/bq/main.go
```diff
@@ -31,6 +32,8 @@ import (
 	"github.com/ossf/scorecard/v4/cron/internal/data"
 )
 
+var configFilename = flag.String(config.ConfigFlag, config.ConfigDefault, config.ConfigUsage)
+
```
azeemshaikh38, line 36: Can define this flag in `config.go` so that multiple definitions are not required?

azeemshaikh38: Basically,

```
// config.go

var configFileFlag = ...

func ReadConfig() err {
  // Read input file from `--config`
}
```

```
// main.go

flag.Parse()
if err := config.ReadConfig(); err != nil {
  // handle error
}
```

cron/internal/config/config.go
```diff
@@ -35,6 +33,14 @@ const (
 	ShardNumFilename string = ".shard_num"
 	// TransferStatusFilename file identifies if shard transfer to BigQuery is completed.
 	TransferStatusFilename string = ".transfer_complete"
+
+	// ConfigFlag is the name of the CLI flag to specify a config file.
+	ConfigFlag string = "config"
+	// ConfigDefault is the default value for the config file if not provided as a CLI arg.
+	ConfigDefault string = "/etc/scorecard/config.yaml"
+	// ConfigUsage is a description of the config CLI flag.
+	ConfigUsage string = "Location of config file"
```
azeemshaikh38, line 42: Need not expose these consts if we are defining the flag in this file.

Also, let's keep the default value of `ConfigDefault` to empty "" and make the flag required to ensure is being passed in.

azeemshaikh38 (approved):

oliverchang:
@calebbrown FYI

cron/internal/bq/main.go
spencerschrock: I think it makes the `flag.Parse()` call opaque, but it would work

azeemshaikh38 (approved):

azeemshaikh38 (approved):

cron/k8s/controller.yaml
```diff
@@ -52,13 +52,21 @@ spec:
           containers:
           - name: controller
             image: gcr.io/openssf/scorecard-batch-controller:stable
-            args: ["cron/internal/data/projects.csv"]
+            args: ["--config=/etc/scorecard/config.yaml", "cron/internal/data/projects.csv"]
```
azeemshaikh38, line 55: Since the controller uses `os.Args[1]`, the argument positions should be swapped?

azeemshaikh38 (approved):

cron/k8s/controller.yaml
spencerschrock: The controller uses `reader = localFiles(flag.Args())` where 
> flag.Args returns the non-flag command-line arguments.

PR #2297, opened 2022-09-22
:seedling: Reduce the number of PR's opened by dependabot

naveensrinivasan:
- Reduce the number of PR's opened by dependabot


laurentsimon (approved):
s on increasing the schedule? Is daily really needed? How about weekly or bi-weekly? Not sure this makes a difference if only a few PRs are opened at once. Curious :)

PR #2298, opened 2022-09-23
🌱 cron: Fix k8s yaml errors and document how to prevent them. (5/n)

spencerschrock:

Two yaml files from #2279 had a parsing error that stopped them from being applied with `kubectl`

All files parse

NONE

This PR has two kinds of changes:
1. Necessary changes to make the controller yamls parse, involving location of the `volumes` tag
2. Whitespace changes  to fix the linter warnings (which `kubectl` will ignore anyway)

the `release-note`

```release-note
NONE
```


spencerschrock:
We could also make this a build workflow step to prevent this sort of thing in the future, but it would add a new step to that job since we'd need to `apt install yamllint`. As long as it's limited by something like #2293 I think the effect would be small.

azeemshaikh38 (approved):

PR #2308, opened 2022-09-28
🐛 Prevent partial cron transfers caused by controller failures

### LINKED ISSUE #2307: Missing BigQuery data
**Describe the bug**
The cron job has partial results for the 9/19 partition.

**Reproduction steps**
Not user reproducible, but occurs when the controller crashes without publishing shard metadata.

**Expected behavior**
Every weekly cron produces complete data.

**Additional context**
1. the controller started, and pushed some subset of messages to the pubsub topic (at least 109k shards of the expected 124k were pushed). Batch 1
2. the controller crashed within the first 12 seconds. and the shard metadata file was never created
3. the controller restarted and pushed another 124k shards to the pubsub topic. Batch 2
4. the workers tried to work on all 200k+ requests and only finished when the messages died after 1 week in the pubsub

This leads to two bugs:
1. the bq transfer tool transferred partial results on Thursday since the first batch didn't have a metadata file
2. the bq transfer tool never transferred results for batch 2 because workers only processed 10% of the batch 2 shards which is less than the transfer threshold.

spencerschrock:

Bug fix

See #2307 for details.

1. the BQ transfer tool will not transfer results for jobs without a `.shard_metadata` file.
2. cron workers won't process requests for jobs without a `.shard_metadata` file.

Note: This makes assumptions about how the pubsub subscription is configured, namely that messages redeliveries are limited. The default value (5 re-tries) works, and is currently being used by scorecardcron.

#2307 

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cron/internal/bq/main_test.go
+++ b/cron/internal/bq/main_test.go
@@ -1,0 +1,62 @@
+// Copyright 2022 Security Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+)
+
+func TestIsCompleted(t *testing.T) {
+	t.Parallel()
+	testcases := []struct {
+		name               string
+		inputExpected      int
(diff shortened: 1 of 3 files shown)


cron/internal/bq/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+)
+
+func TestIsCompleted(t *testing.T) {
```
azeemshaikh38, line 21: Would be very helpful if we added tests for `getBucketSummary` fn. Can do in a separate PR.

azeemshaikh38 (approved):
Thanks! Great finds :)

cron/internal/bq/main_test.go
azeemshaikh38: Resolving so we can merge.

PR #2311, opened 2022-09-29
🐛 Ignore Bot commits when checking for Code Review

### LINKED ISSUE #2302: Feature: Code Review Check handle commits made by version bump bots

Discussion started at https://github.com/electron/electron/pull/35741#pullrequestreview-1118960274

Just oppening the discussion about the commits made automatically by some bots that, depending on its access rights and on its complexity, can not represent a big security risk.

We have some situations where it is not reasonable for the project to review every single commit made by the bot, since it is a very frequent and automatic job. But, by the other hand, allowing an ignore list for bots also leaves the door open to other projects excluding bots that might be more susceptible to compromise.

The Code Review Check could handle, somehow, commits made by version bump bots, that just updates the version of the project itself, not of its dependencies.

I've consider to allow the user to exclude commits made by the bot, but this has the turnaround mentioned above of leaving the door open to other projects excluding bots that might be more susceptible to compromise.

**Additional context**

raghavkaul:
Ignore bot commits when checking for code review. This is an update to the evaluation logic as well as clients. It checks for Bots without relying solely on committer names.

Today, bot commits that are pushed directly to the default branch are considered unreviewed.

Fixes #2302

Raw result format is also updated.

### DIFF
--- a/checks/evaluation/code_review.go
+++ b/checks/evaluation/code_review.go
@@ -21,7 +21,7 @@ import (
 	sce "github.com/ossf/scorecard/v4/errors"
 )
 
-type reviewScore = int
+type reviewScore int
 
 // TODO(raghavkaul) More partial credit? E.g. approval from non-contributor, discussion liveness,
 // number of resolved comments, number of approvers (more eyes on a project).
@@ -29,6 +29,7 @@ const (
 	noReview              reviewScore = 0 // No approving review before merge
 	changesReviewed       reviewScore = 1 // Changes were reviewed
 	reviewedOutsideGithub reviewScore = 1 // Full marks until we can check review platforms outside of GitHub
+	reviewNotNeeded       reviewScore = 1 // Review not needed (e.g. bot commits)
 )
 
 // CodeReview applies the score policy for the Code-Review check.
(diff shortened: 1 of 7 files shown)


checks/evaluation/code_review.go
```diff
@@ -57,6 +58,13 @@ func CodeReview(name string, dl checker.DetailLogger, r *checker.CodeReviewData)
 }
 
 func reviewScoreForChangeset(changeset *checker.Changeset) (score reviewScore) {
+	if changeset.Commits[0].Committer.IsBot {
```
laurentsimon, line 61: is it right to always ignore these bots? Say, AllStar / Dependabot sends a PR to fix something.. should it be manually reviewed? Ignoring unilaterally creates a precedent that any random bots someone sets up won't need reviews.
Is there a way to distinguish?

Also, I think it's feasible for someone to set the bot name themselves when creating a PR / commit. 

Shall we use the User.id instead https://github.com/raghavkaul/scorecard/blob/5b490d3e583bbb1b774a75c459caaf0496292a90/clients/user.go#L23 to identify "real" bots? I think this one cannot be forged... but should verify my claim.

If this works, shall we keep a list of "acceptable" bots? And add it to our documentation of the check?

laurentsimon (commented):
LGTM. Left one main question

PR #2319, opened 2022-09-30
📖 Clarifications about the pinned dependencies check

katzj:
The pinned dependencies check is confusing to authors of libraries, who shouldn't be pinning the dependencies of the library. But it is still valuable for flagging things which are used as part of the build and release process for a library.

Signed-off-by: Jeremy Katz <jeremy@tidelift.com>

Docs update

Currently library maintainers are confused thinking that this check wants them to pin their dependencies but that is not the real intent. This tries to make the documentation a little clearer.

NONE

```release-note
NONE
```

### DIFF
--- a/docs/checks/internal/checks.yaml
+++ b/docs/checks/internal/checks.yaml
@@ -447,17 +447,18 @@ checks:
     risk: Medium
     tags: supply-chain, security, dependencies
     repos: GitHub, local
-    short: Determines if the project has declared and pinned its dependencies.
+    short: Determines if the project has declared and pinned the dependencies of its build process.
     description: |
       Risk: `Medium` (possible compromised dependencies)
 
-      This check tries to determine if the project pins its dependencies.
+      This check tries to determine if the project pins dependencies used during its build and release process.
       A "pinned dependency" is a dependency that is explicitly set to a specific hash instead of
       allowing a mutable version or range of versions. It
       is currently limited to repositories hosted on GitHub, and does not support
       other source hosting repositories (i.e., Forges).
 
-      The check works by looking for unpinned dependencies in Dockerfiles, shell scripts and GitHub workflows.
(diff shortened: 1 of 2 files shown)


docs/checks/internal/checks.yaml
```diff
@@ -447,17 +447,18 @@ checks:
...
     tags: supply-chain, security, dependencies
     repos: GitHub, local
-    short: Determines if the project has declared and pinned its dependencies.
+    short: Determines if the project has declared and pinned the dependencies of its build process.
     description: |
       Risk: `Medium` (possible compromised dependencies)
 
-      This check tries to determine if the project pins its dependencies.
+      This check tries to determine if the project pins dependencies used during its build and release process.
       A "pinned dependency" is a dependency that is explicitly set to a specific hash instead of
```
laurentsimon, line 455: Can you squeeze in a link to https://github.com/ossf/package-manager-best-practices/blob/main/published/npm.md for npm?

laurentsimon (approved):

azeemshaikh38 (commented):
Generate the `checks.md` file using `make generate-docs`

docs/checks/internal/checks.yaml
katzj: I think that actually would increase the confusion as it dives deep into package-lock.json, yarn.lock, etc and that's where library maintainers are currently feeling that this check is not relevant for them.

katzj:
> Generate the `checks.md` file using `make generate-docs`

Done (note I had to explicitly remove the file and then run the command... it didn't notice the change; makefile rule looks right to me though 🤷‍♂️ )

azeemshaikh38 (approved):

docs/checks/internal/checks.yaml
azeemshaikh38: Resolving for now. Can create separate PR if needed.

PR #2333, opened 2022-10-06
🌱 Fixup: list GitHub check runs of MergeRequest.HeadSHA instead of Commit.SHA

raghavkaul:
This reduces the number of uncached GitHub GraphQL API calls. It also ensures that scorecard ignores CI tests that only run on commits after they've already been merged + reviewed.


e2e/ci_tests_test.go
```diff
@@ -77,5 +77,30 @@ var _ = Describe("E2E TEST:"+checks.CheckCITests, func() {
 			Expect(scut.ValidateTestReturn(nil, "CI tests run", &expected, &result, &dl)).Should(BeTrue())
 			Expect(repoClient.Close()).Should(BeNil())
 		})
+		It("Should return absence of CI tests in a repo with unsquashed merges", func() {
```
azeemshaikh38, line 80: Want to clarify my understanding here - so if a repo has unsquashed merges, even if the repo runs CI tests before merging PRs, Scorecard will return absence of CI tests?

checks/raw/ci_tests.go
```diff
@@ -45,19 +44,21 @@ func CITests(c clients.RepoClient) (checker.CITestData, error) {
...
 
-		crs, err := c.ListCheckRunsForRef(commit.SHA)
-		if err != nil {
-			return checker.CITestData{}, sce.WithMessage(
-				sce.ErrScorecardInternal,
-				fmt.Sprintf("Client.Repositories.ListCheckRunsForRef: %v", err),
-			)
-		}
+		// HeadSHA is the last commit before the merge. if squashing enabled,
+		// multiple commit SHAs will map to a single HeadSHA
```
azeemshaikh38, line 48: So in case squashing is not enabled, HeadSHA is the final commit the PR author made. Correct?

e2e/ci_tests_test.go
raghavkaul: In this case, there were CI tests run only on the merge commit of an unsquashed batch of commits (the merge commit is created automatically by Github after the PR is accepted and merged). If CI tests are run before the merge (i.e. they are run on any commit before the merge commit), scorecard would say that CI tests are present

azeemshaikh38 (approved):

PR #2354, opened 2022-10-14
🌱  Remediation cleanup

### LINKED ISSUE #2152: Using Scorecard GitHub Action reduces the Token-Permissions score

spencerschrock:
Followup PR to clean up some work done in #2168  

The function signatures for `TokenPermissions` and `PinningDependencies` was different from the rest of the checks because it included the check request. This wasn't ideal as it should ideally operate just on raw data.

More remediation specific types/functions have been moved to the `remediation` package.

All the remediation metadata is now part of the raw data structures for the checks that need it.
The function signatures are now restored.

Fixes #2152 

the `release-note`

```release-note
NONE
```


azeemshaikh38 (commented):
Some high-level comments:

-> `RawResult` should be data only. So let's not add any non-serializable logic to that structure.
-> `Remediation` struct only has 2 fields. Is it worth having a new package just for that?  Could this struct simply be in `raw` package?
-> Having a common place for all remediation generation logic doesn't seem very scalable/maintainable. Could we instead refactor this such that each check owns its remediation generation logic?

laurentsimon:
I'm not sure remediation logic can live in `raw` since raw it non opinionated?
As part of structured results, I think it makes sense to have a dedicated remediation file. The plan is to have a config files for each "rule" to be able to scale, so the remediation code should be simple.

Wdut?

Do we think we should wait till we start the implementation of the structured results to update this code?

azeemshaikh38:
> I'm not sure remediation logic can live in `raw` since raw it non opinionated?

I meant the `Remediation` metadata struct (which just has branch and repo name). That should be ok to live in `raw` ?

> As part of structured results, I think it makes sense to have a dedicated remediation file. The plan is to have a config files for each "rule" to be able to scale, so the remediation code should be simple.

IMO each check should own its logic (including raw, remediation, rules etc.). Its easier to scale the number of checks that way. 

> Do we think we should wait till we start the implementation of the structured results to update this code?

Yes and no. IIUC, this PR is trying to address the fact that some files in `raw` package depend on `CheckRequest` instead of `RepoClient` only (it currently doesn't accomplish that and we should definitely fix that in this PR). But we can avoid adding more packages to the project until we have reached a consensus on the structured results design.

laurentsimon:
> > I'm not sure remediation logic can live in `raw` since raw it non opinionated?
> 
> I meant the `Remediation` metadata struct (which just has branch and repo name). That should be ok to live in `raw` ?

IMO, remediation has nothing to do with `raw`. raw does non-opinionated and has no concept of alerts / bugs, so should not be aware of remediations. 

> 
> > As part of structured results, I think it makes sense to have a dedicated remediation file. The plan is to have a config files for each "rule" to be able to scale, so the remediation code should be simple.
> 
> IMO each check should own its logic (including raw, remediation, rules etc.). Its easier to scale the number of checks that way.

With structured results, there will be no remediation-specific code to write in the checks. The checks will simply instantiate a remediation / rule object with the name of the rules. The remediation package will have the common logic which will be independent of the rules. The rules will be defined in a config file: we can decide to put them under the directory of each check (if we think this helps scale / maintenance) or in a rule directory shared across all checks. 

> > Do we think we should wait till we start the implementation of the structured results to update this code?
> 
> Yes and no. IIUC, this PR is trying to address the fact that some files in `raw` package depend on `CheckRequest` instead of `RepoClient` only (it currently doesn't accomplish that and we should definitely fix that in this PR). But we can avoid adding more packages to the project until we have reached a consensus on the structured results design.

spencerschrock:
> Some high-level comments:
> 
> -> `RawResult` should be data only. So let's not add any non-serializable logic to that structure. 

I assume you're talking about the `SetupRemediationMetadata` method defined on the token and pinning structs?


> Yes and no. IIUC, this PR is trying to address the fact that some files in raw package depend on CheckRequest instead of RepoClient only (it currently doesn't accomplish that and we should definitely fix that in this PR). But we can avoid adding more packages to the project until we have reached a consensus on the structured results design.

Just to be clear `remediations` is already a package in the codebase. I'm. not opposed to each check owning its remediation logic, but I'm also not familiar with the code Laurent is talking about.

> IMO, remediation has nothing to do with raw. raw is non-opinionated and has no concept of alerts / bugs, so should not be aware of remediations. I don't mind keeping it in raw temporarily, though. But long-term I don't think it should live in raw.

It's a bit of a misnomer currently, as it's just information about the repository that's needed to generate remediation text. If we're trying to make it so evaluation code only depends on raw data, but evaluation is where the remediation text gets generated, then that metadata about the project needs to live in the raw data somewhere.

spencerschrock:
@azeemshaikh38 I chatted with @laurentsimon. What do you think of adding the `remediation.Metadata` as an argument to the `evaluation` function signatures. `md` would be optional (could be nil), and just wouldn't generate remediations if not present.

e.g.
```go
func TokenPermissions(name string, dl checker.DetailLogger, 
    r *checker.TokenPermissionsData, md remediation.Metadata) checker.CheckResult
```

This could go a few places:
1. Inside pkg.RunScorecards() if we just want to generate the data once to be used by every check. 
2. Inside the various `checks/checkname.go` files

Not a fan of (1) as it would require additional arguments added to multiple functions to bubble down. And the logic would be centralized, and pretty far from where its used.
I think (2) works well, as each check would own the code responsible for maintaining its own remediation metadata. It would look something like this for `checks/permissions.go`: 
```go
// TokenPermissions will run the Token-Permissions check.
func TokenPermissions(c *checker.CheckRequest) checker.CheckResult {
	rawData, err := raw.TokenPermissions(c)
	if err != nil {
		e := sce.WithMessage(sce.ErrScorecardInternal, err.Error())
		return checker.CreateRuntimeErrorResult(CheckTokenPermissions, e)
	}

	// Return raw results.
	if c.RawResults != nil {
		c.RawResults.TokenPermissionsResults = rawData
	}

        md := extractRemediationMD(c)

	// Return the score evaluation.
	return evaluation.TokenPermissions(CheckTokenPermissions, c.Dlogger, &rawData, md)
}
```

PR #2355, opened 2022-10-14
⚠️ Warn and not fail if job perms are content:write

### LINKED ISSUE #2338: Explicitly allowed action permissions?
Hello.

I'm trying to max out my score for the scorecard, especially in the security section. One of the issues I'm running into is that I use the google release-please action to cut releases and then push those tags back into the repo.  Currently, there is no mechanism to explicitly mark that action as "safe".

I would like to check in a yaml file into my repository which says "I really do trust the following actions" and not have to specify a sha / version.

Alternatively, we can update your tool to also say "release-please is safe".

- [My thread in slack](https://openssf.slack.com/archives/C0235AR8N2C/p1665074071700439)
- [Someone else's, looking to allow a different tool](https://openssf.slack.com/archives/C0235AR8N2C/p1665158708840929)

eddie-knight:
Following discussion in #2338, it has been determined that the permissions check should no longer deduct when a job is set to write. This code is a minimalist change, and will require additional polish/cleanup of obsolete logic later.

Signed-off-by: Eddie Knight <iv.eddieknight@gmail.com>

Changes how scoring is done for token permissions

The Token-Permissions check is opinionated regarding job-level permissions.

The Token-Permissions check now will only issue a warning regarding job-level permissions.

Fixes #2338

I wasn't able to get the Token-Permissions check to report anything lower than 10/10 when running locally, even for repos that are less than 10 when checked via the API. As such, my validation checks were limited. I suggest a regular contributor or maintainer validates these changes before merging.

```release-note
Job Level Permissions will no longer be evaluated on the Scorecard, but warnings will still be issued.
```

### DIFF
--- a/checks/evaluation/permissions.go
+++ b/checks/evaluation/permissions.go
@@ -241,8 +241,9 @@ func calculateScore(result map[string]permissions) int {
 
 		// contents.
 		// Allows attacker to commit unreviewed code.
+		// Scoring does not apply to job-level permissions, as this is a common place to use third-party actions.
 		// High risk: -10
-		if permissionIsPresent(perms, "contents") {
+		if permissionIsPresentInTopLevel(perms, "contents") {
 			score -= checker.MaxResultScore
 		}
 
--- a/checks/permissions_test.go
+++ b/checks/permissions_test.go
@@ -251,7 +251,7 @@ func TestGithubTokenPermissions(t *testing.T) {
 			filenames: []string{"./testdata/.github/workflows/github-workflow-permissions-contents-writes-no-release.yaml"},
 			expected: scut.TestReturn{
 				Error:         nil,
-				Score:         checker.MinResultScore,
+				Score:         checker.MaxResultScore,
 				NumberOfWarn:  1,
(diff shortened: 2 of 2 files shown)


spencerschrock:
If we want to warn on write permissions, but not affect scoring if defined at the run level, this might be better addressed in the scoring code. Curious what others think.

https://github.com/ossf/scorecard/blob/36d6a340ed9fcc5d5a894c722cca9526a9a4e40a/checks/evaluation/permissions.go#L191

The various calls to `permissionIsPresent` could be replaced with `permissionIsPresentInTopLevel` for example.

eddie-knight:
Thanks @spencerschrock, I think it's a good idea to shift the change into the calculate function instead.

naveensrinivasan (approved):
Thanks! I'll let @spencerschrock 👀

spencerschrock (approved):
LGTM overall. May be worth adding `packages` as that's another permission which results in a `-10` deduction if not on the allowlist. 

There's also an allowlist for `security-events`, but I imagine that's much less commonly used and only a `-1` deduction.

PR #2362, opened 2022-10-17
🌱 Add soft mem limit to controller k8s spec

spencerschrock:

* Docker images now use go1.19 like the rest of scorecard. (which was also needed for `GOMEMLIMIT`)
* The controller now has a soft memory limit (with a 5-10% overhead as suggested by the [GC guide](https://tip.golang.org/doc/gc-guide)). The controller failed in a recent run due to `OOMKilled`

NONE

Will apply the config change to the release test first, and then the production controller after a few days

the `release-note`

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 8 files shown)


naveensrinivasan (approved):
Thanks

PR #2413, opened 2022-11-01
🌱 Code Review: treat merging a PR as code review

### LINKED ISSUE #2391: BUG: Some merged PRs that should qualify as "reviewed" are not being tagged as such
**Describe the bug**
spring-core is getting reported as not doing code review, but it appears that recently-merged changes do qualify as reviewed based on the description of that check.  Here’s the API results:
https://api.securityscorecards.dev/projects/github.com/spring-projects/spring-framework
And here is an example of a PR that I think should qualify (merged by someone other than the committer (implicit review)):
https://github.com/spring-projects/spring-framework/pull/28753
Here is the description of the "code review" check:
https://github.com/ossf/scorecard/blob/main/docs/checks.md#code-review
And the relevant excerpt from the description:
"If this fails, the check determines whether the most recent (~30) commits have a Github-approved review or if the merger is different from the committer (implicit review)"

**Reproduction steps**
1. View the API link above
2. View the PR linked above

**Expected behavior**
I had expected the linked PR to count as code review (maybe with a low score -- not as good as branch protection with required review but better than nothing).

**Additional context**
None

raghavkaul:

Count any linked GitHub PR as a code review. Currently, we require an explicit approval on a GitHub PR, but maintainers often save time by merging a PR without approving it.

Fixes #2391 

We don't check reviewer or PR author identities yet, because the PR opener may not be the commit author, and commit authorship can be spoofed. 

```release-note
NONE
```


checks/evaluation/code_review.go
```diff
@@ -61,12 +61,8 @@ func reviewScoreForChangeset(changeset *checker.Changeset) (score reviewScore) {
...
 	}
 
-	for i := range changeset.Commits {
-		for _, review := range changeset.Commits[i].AssociatedMergeRequest.Reviews {
-			if review.State == "APPROVED" {
-				return changesReviewed
-			}
-		}
+	if changeset.ReviewPlatform == checker.ReviewPlatformGitHub {
+		return changesReviewed
```
spencerschrock, line 65: This doesn't check for "if the merger is different from the committer". Tested it on a PR I created and merged here: https://github.com/spencerschrock/codereview-test.

`1 out of last 2 changesets reviewed before merge -- score normalized to 5`
The first commit was pushed straight to main and isn't marked as reviewed. The second commit was added via a PR authored and merged by me.

spencerschrock: There was also some additional discussion about verified commits to try to avoid faking other users in the [slack thread](https://app.slack.com/client/T019QHUBYQ3/C0235AR8N2C/thread/C0235AR8N2C-1665602612.497779)

> One caveat - note that we should be ensuring we honor verified  commits only when checking for mergedBy . It is super easy for someone to fake another user

raghavkaul: Yeah, that's true - verified commits, when available, should be the source of truth for the Code Review check (and requiring Verified commits could also be good as a standalone check). I do think the change is what maintainers would find useful in the short term, and we could have a separate issue for 'hardening' the Code Review check to rely on verified authors. Wdyt?

spencerschrock: > we could have a separate issue for 'hardening' the Code Review check to rely on verified authors

I think it's fine to harden in a followup.

But the code as written is too permissive. It counts any PR as reviewed even when the PR author is the same as the person who merges.

raghavkaul: Makes sense, added the check back.

spencerschrock (approved):
Thanks!

Issue #2412, opened 2022-10-31
Code hotspots


spencerschrock:
A significant portion (5.5%) of the cron job is spent in `raw.isText`
https://github.com/ossf/scorecard/blob/1fa791043273aae2dfce4ee54cf4035be6802486/checks/raw/binary_artifact.go#L170-L180

There's some experimental code that also does this, and has better performance. I plan to import/copy the implementation over.

```go
// IsText reports whether a significant prefix of s looks like correct UTF-8;
// that is, if it is likely that s is human-readable text.
func IsText(s []byte) bool {
	const max = 1024 // at least utf8.UTFMax
	if len(s) > max {
		s = s[0:max]
	}
	for i, c := range string(s) {
		if i+utf8.UTFMax > len(s) {
			// last char may be incomplete - ignore
			break
		}
		if c == 0xFFFD || c < ' ' && c != '\n' && c != '\t' && c != '\f' {
			// decoding error or control character - not a text file
			return false
		}
	}
	return true
}
```

Should be an easy performance win, but would change the behavior slightly as it **only checks the first 1024 bytes**. 

Since you implemented the heuristic, thoughts @laurentsimon ? I know the code was introduced to reduce false positives, so should still be valid for that.

Benchmarking shows a 2x speedup on small files (1KiB). 6x speedup on medium files (3KiB). And on gigantic files (like our 50MB `projects.csv` file), the difference is 5-6 orders of magnitude.
```
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkIsText-8                     79         295087760 ns/op
BenchmarkExternalIsText-8       13462120              1763 ns/op
```

laurentsimon:
Do you know why it is faster? What's the source of slowdown in the current version?
Looking at the first 1k byte would be OK of we have data (from cron) that the results are the same?

Btw, how did you perform the benchmark?

laurentsimon:
Thanks for the additional info. So inlining code is what speed things up. I'm surprised there's a 2x speedup on 1k files due to that. In fact that seems weird to me because there's a single stack frame created to call the function...
It would be great to have cone numbers confirming we're not making things worse. Maybe we can add some logging code temporarily to call both function in the cron (gated with an env variable) and mine the logs for different results?

spencerschrock:
> In fact that seems weird to me because there's a single stack frame created to call the function...

In a loop, called for each rune in the file contents. 

[...]

PR #2414, opened 2022-11-01
🌱 Enable comparison for alternative isText implementation

spencerschrock:
Optimization / comparison

The Binary Artifacts check uses `raw.isText`, which is inefficient as written.

The Binary Artifacts still uses `raw.isText`, but if `SCORECARD_COMPARE_ISTEXT` is set as an environment variable the result is compared to the implementation in `golang.org/x/tools/godoc/util`.
This value is set for the release-test worker to gather data.

In the event of a mismatch, a log message will be displayed.

For example, running scorecard on a test repo which has a file with binary data appended after 1024 bytes: https://github.com/spencerschrock/binary-test
```
SCORECARD_COMPARE_ISTEXT=1 go run main.go --repo spencerschrock/binary-test --checks=Binary-Artifacts
```

```
2022/11/01 12:59:23 isText implementations differ for file: script-pkg-managers
```

Part of #2412 

the `release-note`

```release-note
NONE
```


laurentsimon (approved):
Thanks for doing all this validation work!

spencerschrock:
There are quite a few differences in the worker, so far it seems like the main difference is due to how we treat carriage returns (which are semi frequent).

`IsText` from `golang.org/x/tools/godoc/util` allows the`\n`, `\t` and `\f` control characters.
Where as we allow `\n`, `\t` and `\r`.

Modifying `IsText` from `golang.org/x/tools/godoc/util` to match our allowed control characters seems to resolve the mismatches I've tested locally. I'll send a follow-up PR later so we can test more in the release test

PR #2433, opened 2022-11-07
🌱 enable more performant isText

### LINKED ISSUE #2412: Code hotspots

spencerschrock:

performance optimization

Code relies on calling `unicode.IsPrint`

Fixes #2412

Over the last 3 days, there have been ~40k files in the cron worker release test where the two `isText` implementations have differed.
 
98% of these differences were the old `raw.isText` returning `false` and new implementation returning `true`. The difference is due to space characters that aren't the ASCII space character (which is the only space where `unicode.IsPrint()` returns true). The following spaces were prevalent in the test repos:
*  [Byte Order Mark](https://en.wikipedia.org/wiki/Byte_order_mark)
* [NBSP](https://en.wikipedia.org/wiki/Non-breaking_space)

The other 2% were the new implementation returning `false` and the old implementation returning `true`. This was due to UTF-8 decoding errors on files which were not UTF-8 encoded (e.g. simplified Chinese GB2312). As a result, the condition which failed on `utf8.RuneError` (`0xFFFD`) was removed.

the `release-note`

```release-note
NONE
```

### DIFF
--- a/checks/raw/binary_artifact.go
+++ b/checks/raw/binary_artifact.go
@@ -16,12 +16,9 @@ package raw
 
 import (
 	"fmt"
-	"log"
-	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
-	"unicode"
 	"unicode/utf8"
 
 	semver "github.com/Masterminds/semver/v3"
@@ -158,12 +155,7 @@ var checkBinaryFileContent fileparser.DoWhileTrueOnFileContent = func(path strin
 	}
 
 	exists2 := binaryFileTypes[strings.ReplaceAll(filepath.Ext(path), ".", "")]
-	// TODO remove the comparison after testing in release-test worker
-	isTextFile := isText(content)
-	if _, enabled := os.LookupEnv("SCORECARD_COMPARE_ISTEXT"); enabled && isTextFile != isText2(content) {
-		log.Printf("isText implementations differ (raw.isText: %t) for file: %s", isTextFile, path)
-	}
-	if !isTextFile && exists2 {
+	if !isText(content) && exists2 {
 		*pfiles = append(*pfiles, checker.File{
 			Path:   path,
 			Type:   checker.FileTypeBinary,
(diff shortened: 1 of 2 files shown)


spencerschrock:
@jeffmendoza FYI

The implementation is guaranteed to be more permissive than the previous implementation, so there won't be an increase in false positives.

naveensrinivasan (approved):
Cool! Thanks

PR #2438, opened 2022-11-08
:warning: Remove attestor as its own module

### LINKED ISSUE #2418: Keeping go.mod up-to-date for both `scorecard` and `scorecard-attestor` modules
This seems like it could be a frequent pain point going forward.
 
https://github.com/ossf/scorecard/actions/runs/3373560170/jobs/5598256385#step:7:95
```
Error: ../../checks/raw/binary_artifact.go:30:2: missing go.sum entry for module providing package golang.org/x/tools/godoc/util (imported by github.com/ossf/scorecard/v4/checks/raw); to add:
```
Or the number of `go mod tidy` commits in the original PR https://github.com/ossf/scorecard/pull/2309/commits

@raghavkaul Is there a benefit to having the attestor as its own module? Is there anything blocking us from doing what we do with some of the cron binaries [add](https://github.com/ossf/scorecard/tree/main/cron/internal/data/add) or [worker](https://github.com/ossf/scorecard/tree/main/cron/internal/worker)?

spencerschrock:
Nested module cleanup

The scorecard attestor is in a nested module, which breaks many tools (`golangci-lint`, vscode, etc) and has issues keeping the modules in sync.

The scorecard attestor is now just a sub package.

Fixes #2418 

* I listed this as a breaking change as we're removing a module / changing import path on a few `attestor` packages. But I don't see the packages in deps.dev or pkg.go.dev, so I don't think it's being depended upon.
* I fixed the easy linter warnings, but left some as a TODO to avoid breaking logic. @raghavkaul FYI
* Unit and e2e tests for the attestor are part of `make unit-test` and `make e2e-pat`/`make e2e-gh`

the `release-note`

```release-note
NONE
```


spencerschrock:
@raghavkaul already had a PR for this, with the changes in #2430 included

PR #2439, opened 2022-11-08
:book: :sparkles: metrics.openssf.org using scorecard for grafana metrics dashboard 

singhsaurabh:

[prominent-scorecards-users](https://github.com/ossf/scorecard/#prominent-scorecards-users) not having above user information 

the `release-note`

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ metrics. Prominent projects that use Scorecards include:
 -   [Flutter](https://github.com/flutter/flutter)
 -   [sos.dev](https://sos.dev)
 -   [deps.dev](https://deps.dev)
+-   [metrics.openssf](https://metrics.openssf.org/grafana/d/default/metric-dashboard?orgId=1)
 
 ### Public Data
 


naveensrinivasan (commented):
This doesn't update the scorecard data in their dashboard.

image

[https://metrics.openssf.org/grafana/d/default/metric-dashboard?orgId=1&var-PackageURL=pkg%3Agithub/curl/curl](https://metrics.openssf.org/grafana/d/default/metric-dashboard?orgId=1&var-PackageURL=pkg%3Agithub/curl/curl)

singhsaurabh:
@naveensrinivasan Thank You for the feedback, 

I tried making it work locally. BQ was failing so I used the scorecard binary to collect the evidence to update the grafana database.  Still required more work to do. If you suggest I will work on these changes to make it work. Thanks

The Scorecard metrics dashboard screenshot,
 
Metrics-Dashboard

naveensrinivasan:
> @naveensrinivasan Thank You for the feedback,
> 
> I tried making it work locally. BQ was failing so I used the scorecard binary to collect the evidence to update the grafana database. Still required more work to do. If you suggest I will work on these changes to make it work. Thanks
> 
> The Scorecard metrics dashboard screenshot,
> 
> Metrics-Dashboard

It cannot work using the binary it will run out of tokens. The one of the other ways is to use the API and I would recommend that  it be fixed in the other repository. Thanks

azeemshaikh38 (commented):
Note that metrics.openssf isn't maintained as of today and there is a new SIG that was formed (I believe its named Stream 2 - Dashboard) to figure out the future of OSSF Metrics dashboard. I would wait until discussions there are resolved before adding this. Closing this PR for that reason.

PR #2442, opened 2022-11-09
:sparkles: Add Additional Details to License Check

### LINKED ISSUE #1369: Add gradation to license analysis

Currently scorecard looks for a license, but it simply gives a 10 if there's a license found at all and a 0 if there's no license found at all. That doesn't give any gradation.

Add more gradations for licenses. I think scorecard was intended for OSS projects, so I propose giving "10" for projects with a license known to be OSS per OSI or Free Software as defined by the FSF. Otherwise, give it a 5. For projects on GitHub you can just reuse the GitHub analysis, as GitHub has an API that provides license info. You can see an example of this use in the CII Best Practices' [github_basic_detective](https://github.com/coreinfrastructure/best-practices-badge/blob/main/app/lib/github_basic_detective.rb) and [floss_license_detective](https://github.com/coreinfrastructure/best-practices-badge/blob/main/app/lib/floss_license_detective.rb).

Further discussion here: https://github.com/ossf/scorecard/issues/1038

Different people have different views of copyleft, I don't think scorecard should force a particular view of whether copyleft or not-copyleft is better (less risk). You could argue either way, which to me suggests you shouldn't make the argument.  Even those who prefer BSD-style licenses will often use the Linux kernel (GPL), because *how* you use software is very important. As a result, I think "is it approved by OSI or FSF" should be enough, don't force further gradations.

**Additional context**

Again, further discussion here: https://github.com/ossf/scorecard/issues/1038

shissam:

This is a proposed feature enhancement to the License Check which servers two purposes
- Provide specific details as to which license was detected
- Provide the basis to add gradation to license analysis 

License check looks in the repo for specific license files and some matching content. Score is either min or max depending if a file is found - regardless of the license content.

- introduce `clients/licenses.go` which embodies this github api (similar to `clients/languages.go`) - it would be plural as it may come down to the point where gitlab returns multiple licenses for a `repo` (I have a real-world example).
- This github client would populate a structure like the following:
```
type License struct {
        Key      string // RepositoryLicense.GetLicense().GetKey()
        Name     string // RepositoryLicense.GetLicense().GetName()
        Path     string // RepositoryLicense.GetName()
        Size     int    // RepositoryLicense.GetSize()
        SPDXId   string // RepositoryLicense.GetLicense().GetSPDXID()
        Type     string // RepositoryLicense.GetType()
}
```
where (example from ossf/scorecard):
```
{
 Key:apache-2.0
 Name:Apache License 2.0
 Path:LICENSE
 Size:11356
 SPDXId:Apache-2.0 
Type:file
}
```
- modify `checks/raw/license.go` to follow one of two possible paths:
-- `pathOne` try the `repoClient` (as in `c.RepoClient.ListLicenses()`), if a license is reported by github **USE** that and return
-- `pathTwo` if the `repoClient` returns `nil`, continue with the `legacy` logic and return what is always returned
- there are a number of advantages to this approach:
-- _speedup_: for github, my test of `c.RepoClient.ListLicenses()` returns in less than 5 seconds, when compared to other repos with many files (like `torvalds/linux` this check takes over a minute
-- _extensibility_: this test with github is promising, and gitlab does appear to have a api for getting this information (https://docs.gitlab.com/ee/api/templates/licenses.html)
-- _confidence_: using the github api for 'sensing' which specific license is in use I assume is higher (more review on their algorithm which might use NLP)--I tried a license on my site which changed the work `Apache` to `Apacje` in the complete `Apache License 2.0` file and github still sensed it was APL-2.0 (very cool).
-- _standardization_: I am encouraged by the use of the `SPDXId` in the github API, if it is true that gitlab using the same standard identifier for licenses - this would be great for the community
-- _usability_: using `scorecard` to retrieve the specific license detected would a) make this information more readily available to the end-user; and b) would support develop pipelines that need to sense the license (by name) as a risk management activity without having to move to other means to acquire such license information
- with this approach, propose changing `checker.LicenseData{}`:
```
type License struct {
        Key      string // from clients.License.Key
        Name     string // from clients.License.Name
        Size     int    // clients.License.Size (Size maybe not here but in type File struct
        SPDXId   string // clients.License.SPDXId
        Attribution     AttributionType // license sourced from LicenseRepoAPI or ScorecardEngine
}

// one file contains one license.
type LicenseFile struct {
        License License
        File    File
}

// LicenseData contains the raw results
// for the License check.
// Some repos may have more than one license.
type LicenseData struct {
        LicenseFiles []LicenseFile
}
```

Fixes #1369

With a more complete proposal at [https://github.com/ossf/scorecard/issues/1369#issuecomment-1304831531](url)

I would like to have a discussion with the owners/maintainer about idea for gradating the evaluation, with additional information from the repo API, that may lead to greater confidence in scores. This PR would take a step in that direction.

`format=raw` would have the entire LicenseData found by scorecard in the file or the repo API
`format=[json|default]` would have the rational for any such decided on gradation.

the `release-note`

```release-note

```


checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
 	ArchivedStatus       ArchivedStatus
 }
 
+type LicenseAttributionType string
+
+const (
+	// forms of security policy hints being evaluated.
+	LicenseAttributionTypeOther     LicenseAttributionType = "otherAttribution"
+	LicenseAttributionTypeRepo      LicenseAttributionType = "repoAttribution"
```
laurentsimon, line 133: do you have a sense of whether the GH API always outperform our current code? Ie, if we look at the weekly cron results, are there cases where out current code finds a license that the GH APIs do not?

I'm trying to understand whether we should just switch to the GH API entirely. I suppose it's good to have a backup if the platform does not have an API (GitLab, BitBucket, etc), so probably better to keep it.

checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
 	ArchivedStatus       ArchivedStatus
 }
 
+type LicenseAttributionType string
+
+const (
+	// forms of security policy hints being evaluated.
+	LicenseAttributionTypeOther     LicenseAttributionType = "otherAttribution"
```
laurentsimon, line 132: Do you think we need this level of granularity to be exposed? It feels like an implementation detail to me, and consumers should not have to know about it. I'm worried that exposing this would cause breaking changes as soon as the implementation changes. I would suggest we remove it. Wdut?

checks/raw/license.go
```diff
@@ -69,16 +69,49 @@ func License(c *checker.CheckRequest) (checker.LicenseData, error) {
...
+	}
+
+	// no licenses reported by repo API, continue looking for files
 	err := fileparser.OnAllFilesDo(c.RepoClient, isLicenseFile, &path)
 	if err != nil {
 		return results, fmt.Errorf("fileparser.OnAllFilesDo: %w", err)
 	}
 
+	// scorecard search stops at first candidate (isLicenseFile) license file found
 	if path != "" {
```
laurentsimon, line 101: do we need change the score calculation? I've not seen a change, but I may have missed it.
Please also add unit tests to verify the change of behavior

clients/gitlabrepo/licenses.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (handler *licensesHandler) init(repourl *repoURL) {
+	handler.repourl = repourl
+	handler.errSetup = nil
+	handler.once = new(sync.Once)
+}
+
+func (handler *licensesHandler) setup() error {
+	handler.once.Do(func() {
+		// TODO: find actual GitLab API, data type, and fields
```
laurentsimon, line 46: feel free to create an issue on the repo and tackle this in a follow-up PR.

checks/raw/license.go
```diff
@@ -69,16 +69,49 @@ func License(c *checker.CheckRequest) (checker.LicenseData, error) {
 	var results checker.LicenseData
 	var path string
```
laurentsimon, line 71: you also need to update https://github.com/ossf/scorecard/blob/6f84cb92bb88aef0db98813f127f3981f74775c4/checks/license.go#L29-L31 since the check no longer support file based (it needs the API).

laurentsimon (commented):
Thanks for the PR! I left a few comments. Don't forget to add unit tests and update the docs.

checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
...
+	LicenseAttributionTypeRepo      LicenseAttributionType = "repoAttribution"
+	LicenseAttributionTypeScorecard LicenseAttributionType = "scorecardAttribution"
+)
+
+// license details.
+type License struct {
+	Key         string                 // repo specified key
```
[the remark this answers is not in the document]
shissam: Redundant--perhaps (it is until it isn't)

So the GH REST API is silent on just what this is (specifically), and the go-github pkg also seems silent. I can only infer based on the responses I've seen from the GH REST API (via curl).. as in the example returned for ossf/scorecard:

```
{
 Key:apache-2.0
 Name:Apache License 2.0
 Path:LICENSE
 Size:11356
 SPDXId:Apache-2.0 
 Type:file
}
```

In the handful of examples I pulled, `Key == strings.ToLower(SPDXId)` always seems to be the answer.

Now - if I exercise the GH API to get information about a specific license [(https://docs.github.com/en/rest/licenses)](url) I can use `Key` as in:

`curl -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${GITHUB_AUTH_TOKEN}" https://api.github.com/licenses/apache-2.0`

But, alas, `SPDXId` also works but other forms like `apache`, `apache2`, `apache20` don't work.

The go-github pkg source code basically states `Get extended metadata for one license `. Further [https://docs.github.com/en/rest/licenses](url) is just as vague.

The other thing I inferred from the docs is that the GH REST API appears to use [https://github.com/benbalter/licensee](url) on the backend (in some way)--not sure if that `Key` is needed for that purpose -- I just cannot fathom how.
 
Perhaps there is a hubber we can ask for insight.

checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
...
+	Name        string                 // OSI standardized license name
+	SpdxID      string                 // SPDX standardized identifier
+	Attribution LicenseAttributionType // source of licensing information
+	Size        int                    // size of the license file found (default: 0)
```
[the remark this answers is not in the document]
shissam: I struggled as to whether or not to put this here. I always like to know if there is something there without having to really "read" it - with size (greater than the length of the license) allows one to infer there is more there than simply the string "Apache License 2.0"

I'll wack it for now.

checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
...
+	LicenseAttributionTypeOther     LicenseAttributionType = "otherAttribution"
+	LicenseAttributionTypeRepo      LicenseAttributionType = "repoAttribution"
+	LicenseAttributionTypeScorecard LicenseAttributionType = "scorecardAttribution"
+)
+
+// license details.
+type License struct {
+	Key         string                 // repo specified key
+	Name        string                 // OSI standardized license name
+	SpdxID      string                 // SPDX standardized identifier
```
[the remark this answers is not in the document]
shissam: let me look... I tried to find the SpdxID specification (alluded to at https://docs.github.com/en/rest/licenses which simply points to https://spdx.org/) so more digging is needed. I can also look at the [Ruby Gem Licensee](https://github.com/benbalter/licensee) site which is reported to be on the backend -- their code may provide more insight.

Say I find it... when you say "_ add a link to the list of SPDX ID we may expect to see_" is the for the code (embedded comment and documentation) or for another purpose?

checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
...
+
+const (
+	// forms of security policy hints being evaluated.
+	LicenseAttributionTypeOther     LicenseAttributionType = "otherAttribution"
+	LicenseAttributionTypeRepo      LicenseAttributionType = "repoAttribution"
+	LicenseAttributionTypeScorecard LicenseAttributionType = "scorecardAttribution"
+)
+
+// license details.
+type License struct {
```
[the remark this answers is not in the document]
shissam: will do... (it is on my TODOs)

checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
...
+	Key         string                 // repo specified key
+	Name        string                 // OSI standardized license name
+	SpdxID      string                 // SPDX standardized identifier
+	Attribution LicenseAttributionType // source of licensing information
+	Size        int                    // size of the license file found (default: 0)
+}
+
+// one file contains one license.
+type LicenseFile struct {
+	License License
```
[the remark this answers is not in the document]
shissam: Check!

clients/githubrepo/licenses.go
```diff
@@ -0,0 +1,89 @@
...
+
+		// TODO: github.RepositoryLicense{} only supports one license per repo
+		//       should that change to an array of licenses, the change would
+		//       be here to iterate over any such range.
+		handler.licenses = append(handler.licenses, clients.License{
+			Key:    bodyJSON.GetLicense().GetKey(),
+			Name:   bodyJSON.GetLicense().GetName(),
+			SPDXId: bodyJSON.GetLicense().GetSPDXID(),
+			Path:   bodyJSON.GetName(),
+			Type:   bodyJSON.GetType(),
```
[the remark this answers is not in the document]
shissam: This "feels" like another rabbit hole (like trying to find the definition for `Key` in the review comments. I'll dig more as a look into the `licensee` Ruby Gem site.

checks/raw/license.go
```diff
@@ -69,16 +69,49 @@ func License(c *checker.CheckRequest) (checker.LicenseData, error) {
...
-				Path: path,
-				Type: checker.FileTypeSource,
+		results.LicenseFiles = append(results.LicenseFiles,
+			checker.LicenseFile{
+				File: checker.File{
+					Path: path,
+					Type: checker.FileTypeSource,
+				},
+				License: checker.License{
+					Key:         "",
```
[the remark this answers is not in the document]
shissam: There is code that I saw in this check which appears to try to regex match license names, let me see what I can do. mapping into the `SpdxID` might get tricky - I just hate to re-invent someone else's code (i.e., Ruby Gem Licensee in this case).

checker/raw_result.go
shissam: On speed: I think the answer is absolutely yes. It appears that GH establishes the repo's license when a license file is added. In my trials, I kept changing my license and GH immediately gave me my license match in the right hand side of the webGUI and the `curl` would immediate (on the next invocation) show that same answer. So GH cooks the License at the time of the commit (it would appear).

My favorite test is with `torvalds/linux` (it breaks a lot of tools), when I use the GH API I got an answer back (using my changes in `scorecard` in < 5 seconds. When I permitted the legacy `scorecard` code to run, it took 1 min and 10'ish seconds. I am still building my regression suite for this code (to ensure I don't break scoring (unless we gradiate the score) - but when GH does not return a `404` it finishes quickly.

On **misses**: Now, one site I have been testing with is `github.com/twardoch/varfonts-ofl` - GH `404` this site, but the legacy code then exams ALL the files (**this is the time sink overall for the legacy code**) in the repo until it hits/misses and eventually come up with this (I think in this specific case this is a `false positive`):

```
time XXSCORECARD_EXPERIMENTAL=false SCORECARD_V6=true GITHUB_AUTH_TOKEN=${GITHUB_AUTH_TOKEN} /home/vagrant/src/scorecard/scorecard  --show-details --repo=github.com/twardoch/varfonts-ofl --checks=License 
Starting [License]
Finished [License]

RESULTS
-------
Aggregate score: 10.0 / 10

Check scores:
|---------|---------|-----------------------|----------------------------------------------------|--------------------------------------------------------------------------------------------------------|
|  SCORE  |  NAME   |        REASON         |                      DETAILS                       |                                       DOCUMENTATION/REMEDIATION                                        |
|---------|---------|-----------------------|----------------------------------------------------|--------------------------------------------------------------------------------------------------------|
| 10 / 10 | License | license file detected | Info: :                                            | https://github.com/ossf/scorecard/blob/6f84cb92bb88aef0db98813f127f3981f74775c4/docs/checks.md#license |
|         |         |                       | VotoSerifGX-OFL/Fonts/VotoSerifGX-VarTTF/OFL.txt:1 |                                                                                                        |
|---------|---------|-----------------------|----------------------------------------------------|--------------------------------------------------------------------------------------------------------|

real	0m16.701s
user	0m3.862s
sys	0m3.884s
```

By `false positive` I mean that in the reported/found file this text is there `This Font Software is licensed under the SIL Open Font License, Version 1.1.` but in context I don't this it applies to the entire repo `github.com/twardoch/varfonts-ofl`.

I agree with you... by trying the GH (or generally the repo's API (if they have one)) is the best to try first (it is _fast_ for sure), but leave the legacy code in the event that GH (or the repo) fails... the legacy _backup_ would be **no worse** than what it already does.. a best guess.

checks/raw/license.go
shissam: I believe the whole point of issue #1369 was to change the scoring. In this PR description above and my note to the reviewers is that I'd like to have a discussion about scoring (knowing now we can get more definitive answers, e.g., confidence, from GH/repo APIs and perhaps that scoring could be predicated on the "source" of the information.

My goal for this change to the license check is motivated by my need... I need (from scorecard, if I'm going to use it) more information about the license actually attached to a project. right now I have to go off and look to see just what that license is if scorecard say's "License Found". Since scorecard is "there" in the moment, just take a few and tell me which one it is.

For me, license check, would be more "usable" by using scorecard to retrieve the specific license detected as it would a) make this information more readily available to the end-user (me); and b) would support develop pipelines that need to sense the license (by name) as a risk management activity without having to move to other means to acquire such license information

I think we can make progress on the changes to the license check to use the repo APIs when possible (and still have the legacy), get that additional meta-data and report it (raw, default, json), and then change the scoring at a later date. As I change the code, I was **NOT** making the assumption that a change to the scoring evaluation would dictate merging this change--but I am happy to change the score evaluation if we can come to a decision.

clients/gitlabrepo/licenses.go
shissam: great - thanks, I'll do that for sure.

It was troublesome enough just to get this "dummy" component into the structure given that the build system was looking for it (that is for gitlab) and I could not figure out how to turn that off (in the build system). So now, if called, it will simply error as "not supported" -- but the legacy code will continue to do what it always did.

checks/raw/license.go
shissam: I'll look into that, I did not know that, I may circle back and ask questions if I have difficulty.

checker/raw_result.go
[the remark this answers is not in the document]
shissam: I believe I may have found the SpdxId sourced at Licensee (the believed backend of GH API):

https://github.com/licensee/licensee/tree/master/vendor/license-list-XML/src

Of the few I've experimented with, I can find them, string for string, in the `.xml` on that site:

```
GH API's SpdxID: GPL-3.0 --> src at Licensee: GPL-3.0.xml (licenseId="GPL-3.0")
GH API's SpdxID: MPL-2.0 --> src at Licensee: MPL-2.0.xml (licenseId="MPL-2.0")
GH API's SpdxID: Unlicense --> src at Licensee: Unlicense.xml (licenseId="Unlicense")
GH API's SpdxID: CC0-1.0 --> src at Licensee: CC0-1.0.xml (licenseId="CC0-1.0")
GH API's SpdxID: MIT --> src at Licensee: MIT.xml (licenseId="MIT")
```
that's 5 for 5, (oh, and +1 for Apache-2.0), that site has a total of 45 `.xml`s

checker/raw_result.go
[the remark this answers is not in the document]
laurentsimon: Let's file an issue on GitHub and ask them then? Realistically, only one of `SPDXId`, `Key` and `Name` seems to be needed, I think.

checker/raw_result.go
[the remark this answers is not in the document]
laurentsimon: > let me look... I tried to find the SpdxID specification (alluded to at https://docs.github.com/en/rest/licenses which simply points to https://spdx.org/) so more digging is needed. I can also look at the [Ruby Gem Licensee](https://github.com/benbalter/licensee) site which is reported to be on the backend -- their code may provide more insight.
> 
> Say I find it... when you say "_ add a link to the list of SPDX ID we may expect to see_" is the for the code (embedded comment and documentation) or for another purpose?

for documentation.

checker/raw_result.go
laurentsimon: follow-up question: do you think we should *only* use the GH API when running the check on GitHub? I wonder whether scorecard result should be consistent with what GitHub displays in their UI. I wonder whether if we ever provide a different result, some uses may get confused - I doubt this would happen very often though

checks/raw/license.go
laurentsimon: SGTM. I asked about the scoring because the issue you linked in the description mentioned a scoring change. I'm totally onboard updating the scoring later

checker/raw_result.go
shissam: On this I'd like to reserve judgement and make the call before this is merged. I have about 400 sites I was using for a previous PR and I believe I can use those results (going months back) to compare license results with these proposed changes -- I'd like to build a bit more confidence in the GH API use in SC to make a recommendation.

shissam: Here are results from 372 sites scanned on github using the github API (GH API) to retrieve the license for a repo as compared to using the existing scorecard (SC legacy) code for license detection:

Of the 372 sites scanned

- 87.4% of the sites, GH API and SC legacy were in `Agreement`, that is that both GH API and SC legacy code either found the same file or found NO license file
- 10.2% of the sites, GH API and SC legacy found `Different` files
-- example (1 of 38): for `github.com/eclipse/ditto` SC identified `.github/workflows/license-check.yml` as a license file while GH API identified `LICENSE` 
-- example (2 of 38): for `github.com/apache/iceberg` SC identified `.baseline/copyright/001_apache-2.0.txt` as a license file while GH API identified `LICENSE`
-- example (3 of 38): for `github.com/eclipse/kitalpha` SC identified `Docgen/business/features/org.polarsys.kitalpha.doc.gen.business.core.feature/license.html` as a license file while GH API identified `LICENSE.md`
- 2.4% of the sites, SC legacy identified a file as a license file (arguably a `false positive` on inspection of the identified file) where GH API returned NO license file
-- example (1 of 9): for `github.com/apache/doris-website` SC identified `static/fonts/OFL.txt` as a license file (probably is not for the entire repo)
-- example (2 of 9): for `github.com/eclipse/emf.egf` SC identified `deprecated/3.3/features/org.eclipse.egf.portfolio.genchain.ecoretools.feature/license.html` as a license file (probably is not for the entire repo)

The SC legacy approach to walk the depths of the repo in search of a file is likely the cause of a number of `false positives` (that 2.4%) (although it is often the case that when looking at actual source code, once often finds a license disclaimer at the top). But that same strategy is also the cause of the `Different` files but coming up with the same **license file detected** in SC--just good luck.

My conclusion is two-fold:

1. For repos that support a license API retrieval/query for a repo, like the GH API, use it... period. Don't walk the depths of the repo when the GH API returns **license file not detected**. The documentation and remedy should be clear and follow the site recommendations on establishing a license for the repo.
2. For repo sites where there is not license API like GH (e.g., localdir, and other "forges"), and extensive walk will continue to bring `false positives` if nothing is done to limit the depths of the search. I suspect that such may also be true for other SC checks the walk repos (like Security-Policy).

I can provide more details if necessary

checks/raw/license.go
shissam: done.

checks/raw/license.go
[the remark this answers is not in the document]
shissam: if the code is changed to use only the GitHub API (comment/design decision in another comment) - then the point is moot to populate these fields for the SC legacy. After seeing the behavior of the legacy code (and all the false positives) and the fact with all the regex's and many hits (where the contents of each file match would have to be loaded before the contents themselves would have to be regex'ed). This is exactly what Security-Policy does, but the pattern matching for files is very very specific (no regex there).

Running Security-Policy check on `torvalds/linux` (just now), SC took just under 3 minutes -- and there were no files that matched any one of the 13 specific names and therefore no content was loaded to scan for security policy information. Currently the License check (without the GH API) takes about a minute do its search and it is not loading any content. And that is **after** License check stops at the first regex match of the filename. If the contents were not to  garner a match - should License check keep looking for other possible file matches?

Now, with all that, given the content is available, the search of the text (if the blob is text) would not be 'on par' with `Licensee` (the Ruby Gem project that is on the backend of GH)--to be at that level--the right thing to do is engineer as way to use `Licensee` as a component (by some means).

checks/raw/license.go
shissam: Like Elmer Fudd chasing that “wascally wabbit”—I went down the _wabbit hole_ on the SC legacy code when I tried to update `checks.md` to match this PR.

In doing that review, I discovered a bunch of stuff which tells me that `checks.md` and the code simply don’t match.

I think the answer to **gradated scoring** is right in `checks.md` if the code actually **did** what `checks.md` seems to specify (and GH API is very close to doing this already, except looking inside a top-level dir `LICENSE/`).

**FIrst: scoring approach**

Naming Requirements (one or more) (6/10):
- A license file is in the top-level directory with any combination of the following names: LICENSE, LICENCE, COPYING, COPYRIGHT (regardless of extension)
> for GH API this would be limited to a file named `LICENSE`

- A folder named LICENSE, LICENCE, LICENSES or LICENCES and specific license(s) are named as files within that folder following the SPDX license identifier followed by an appropriate file extension, as described in the REUSE Specification.
> for GH API the SPDX identifier would be deduced from the `LICENSE` file content and be limited to the 13 supported by the GH

Location Requirements (3/10):
- License file is prominent for the repository, that is the license file or folder is in the top-level directory
> for GH API the `LICENSE` file would have to be at the top to work. So basically, if the GH API returns a SPDX identifier the repo would get `6+3=9` score at this point.

Identified license(s) (one or more) (1/10):
- is FSF (see “FSF Free/Libre?” column at https://spdx.org/licenses/)
- is OSI (see “OSI Approved?” column at https://spdx.org/licenses/)
> this can be done, but I have question about how to best implement that sensing. I elaborate on that closer to the bottom of this comment.

Remediation Steps

(pretty much stays the way it is now in `checks.md` without being specific about the extension.

As for FSF and OSI licenses and the list at SPDX.org
- 63 licenses are "recognized" (if that is the right way to say that) by both those organizations (FSF & OSI)
- 64 licenses are OSI only
- 46 licenses are FSF only
> for a total of 173 licenses that are eligible for the +1 up in the evaluation scoring

that would be some massive regex or large set of regexs to compile - but I need help, I have an idea which I think would work:

I took the table at `spdx.org/licenses` and converted it directly into a single json object (well, file, in my case for testing)

Assuming (legacy) found `LICENSE/MulanPSL-2.0.md`

- this would get `6 points` for being there (i.e., `LICENSE/` — a folder) and having a file matching an SPDX identifier (i.e., `MulanPSL-2.0`)
- this would get `3 more points` for being in a prominent location (top-level), e.g., `LICENSE/`
- this would get `1 more point` for being an OSI approved license, by removing the extension (**could be a bit tricky**) a json lookup would show this:
> GH API would not be able to get this extra `+1` has it would not find `MulanPSL-2.0` and the GH API will `404`
```
jq -r '.[] | select(."Identifier"=="MulanPSL-2.0") ' SPDX-license-IDs-202211131030CST.json
{
  "FSF Free/Libre?": "",
  "Full Name": "Mulan Permissive Software License, Version 2",
  "Identifier": "MulanPSL-2.0",
  "OSI Approved?": "Y"
}
```
an associative array/map is the way to go there. The map would have the 173 licences from the SPDX list (which has 466 SPDX ids in it—i’d eliminate any non-FSF/non-OSI licenses)—I'd would like to know if there would be a better way if this approach is even fathomable for SC.

Another example: simply `LICENSE`

- this would get `6 points` for being there (i.e., `LICENSE`)
- this would get `3 more points` for being in a prominent location (top-level), e.g., `LICENSE`
> I would stop there - no extra “point” for FSF/OSI, why?
-- 1. since LICENSE is a file, the SPDX identifer would have to be contained _within_ the file, hence SC have to read the contents and “regex”/“string search” the contents looking for a hit to one of the 173 identifiers
-- 2. if the “community” wanted the filename to be (for example) MulanPSL-2.0.md at the top level (not in a LICENSE/ folder), we’re back to regex searching file names in the repo for one of those 173 identifiers. this is a (n files x 173 liceences to regex search)
> GH API would, however, provide the info needed to assess the final `+1` points

Another example: `.absolutelynothingtoseehere/license` (I actually tried this one with current SC legacy code—I got 10 point score!)

- this would get `6 points` for being there (.i.e., `license`)
- this would NOT get `3 points`, the location is not at the top-level (it is obscured)
- this would NOT get `1 point` for FSF/OSI - same reason as before just too many regex’s and files to search across those regex's

checker/raw_result.go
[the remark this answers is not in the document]
shissam: I played a lot more with Github and editing licenses using the template editor, that editor seems to only support 13 specific licenses (and not the 45 at `Licensee`.

(BTW: I submitted a bug [report/issue](https://github.com/orgs/community/discussions/39026) regarding that editor and licenses.

Worse than that, is that SC's `checks.md` points to `spdx.org/licenses` having 466 licenses (with SPDX IDs) of which 173 are either recognized by the FSF, the OSI, or both.

Now, as for `checks.md` itself--it is really close to what the SC legacy license code __should__ be doing--and is really close to how the API behaves. The three big differences is

1. GH API seems to parse the contents--but I think it is limited to those 13 specific licenses (I tried a few from `spdx.org/licenses` and could not get GH API to acknowledge them as anything other than `Other`
2. GH API seems relegated to the top-level of the repo and does not go into any folder where as SC legacy looks until it finds some match.
3. SC legacy does nothing like `checks.md` states regarding doing anything with any filenames inside the top-level folder `LICENSE/` -- it "detects the presence of a license" only by the fact that the folder's name matches the regex for `(un)?licen[sc]e`

checks/raw/license.go
laurentsimon: Scoring you propose seems reasonable

checker/raw_result.go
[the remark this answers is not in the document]
shissam: no answer from the GitHub community. What i have discerned, on `SPDXId`, `Key`, and `Name`: Key is nearly always the lowercase of `SPDXId` except when the GH API cannot assert the `SPDXId` at which point, `SPDXId` becomes `NOASSERTION` and `Key` becomes the lowercase of `Name` which in this case is `Other`

checker/raw_result.go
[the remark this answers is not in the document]
shissam: I am following https://spdx.org/licenses for which licenses to detect which are approved licenses from FSF and OSI. Documentation updated accordingly.

checker/raw_result.go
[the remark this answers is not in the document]
shissam: Done.

clients/githubrepo/licenses.go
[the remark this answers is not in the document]
shissam: I removed `Type` from `License` struct. We can revisit if we hear / learn anything from GitHub

checks/raw/license.go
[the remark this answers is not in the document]
shissam: Done. I am following/parallel with what GH API returns, the SPDX Identifiers seems to make this possible.

checker/raw_result.go
shissam: I have made the code follow the two-fold conclusion above. It could be possible for the code to “fall” into the old way on GH if the GH API comes back with no license’s found, but that decision could be made at anytime.

checker/raw_result.go
shissam: I still do and I am returning this information to the evaluation for comprehension in the scoring.

checks/raw/license.go
shissam: Done.

clients/gitlabrepo/licenses.go
shissam: Still TBD. resolving this comment now and will issue the PR once this code is merged.

shissam: Still TBD. resolving this comment now and will issue the PR once this code is merged.

shissam: Still TBD. resolving this comment now and will issue the PR once this code is merged.

checker/raw_result.go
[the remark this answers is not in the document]
laurentsimon: Is there a way to simplify and use only `key` or `name` or `SPDXId`? Which should we pick?
SPDXid seems tied to the SPDX format, so maybe name?

checker/raw_result.go
```diff
@@ -125,10 +125,35 @@ type MaintainedData struct {
 	ArchivedStatus       ArchivedStatus
 }
 
+type LicenseAttributionType string
+
+const (
+	// sources of license information used to assert repo's license.
+	LicenseAttributionTypeOther     LicenseAttributionType = "otherAttribution"
+	LicenseAttributionTypeRepo      LicenseAttributionType = "repoAttribution"
```
laurentsimon, line 133: Can we make these more self-explanatory? Like `API`, `heuristics`. I'm sure we can find better than what I'm suggesting, though. Thoughts?

shissam: Two thoughts... I'd like your opinion:

1. see the new "checks" documentation, where it states that License using either a repo API (so I like `API`) in your suggestion/comment, and goes on to state "_Scorecard attempts to detect a published license file on its own_"-- to use `heuristics`, then (which I also like), I'd change the documentation to read something like "_Scorecard uses its own heuristics to detect a
published license file._"

2. `checks/evaluation/license.go` is the only reference to `checker.LicenseAttributionTypeOther` and since it is in a switch/case statement, I could remove it and eliminate the const, but I think the evaluation code is better off than falling into `default` in the event other const's are established later -- your call!

laurentsimon: For users who want to use the raw results to automate parsing the results, I was hoping the name could be more explanatory in the field of the results as well. Something like `repositoryAPI` and `builtinHeuristics` could work.

Doc + evaluation code LGTM

shissam: I'd go with this for the code overall:

```
const (
        // sources of license information used to assert repo's license.
        LicenseAttributionTypeOther      LicenseAttributionType = "other"
        LicenseAttributionTypeAPI        LicenseAttributionType = "repositoryAPI"
        LicenseAttributionTypeHeuristics LicenseAttributionType = "builtinHeuristics"
)
```

laurentsimon (approved):
LGTM. Thanks for explaining everything in such details, and the amount of work you've put in this PR!

PR #2445, opened 2022-11-09
🌱 cron: support reading prefix from file for controller input files (7/n)

spencerschrock:

The controller supports a `prefix` which is used to limit which files are used as input files within a bucket e.g. a specific file or directory. The value can be set via the config file or an environment variable (`SCORECARD_INPUT_BUCKET_PREFIX`)

* The `input-bucket` group of prefixes has been grouped and moved under `additional-params`.
  * This will change the corresponding env variables, but for now both are supported.
     * `SCORECARD_INPUT_BUCKET_URL` -> `INPUT_BUCKET_URL`
     * `SCORECARD_INPUT_BUCKET_PREFIX` -> `INPUT_BUCKET_PREFIX`
* Introduction of `prefix-file` (corresponding to env variable of `INPUT_BUCKET_PREFIX_FILE`) whose contents are used as a prefix if `prefix` is not set. This allows for easier changes in situations where the prefix changes frequently e.g. always using the most recent folder in a bucket.

NONE

the `release-note`

```release-note
NONE
```


spencerschrock:
@calebbrown FYI

cron/internal/controller/bucket.go
```diff
@@ -0,0 +1,92 @@
...
+	s := string(b)
+	return strings.TrimSpace(s), nil
+}
+
+func bucketFiles(ctx context.Context) data.Iterator {
+	var iters []data.Iterator
+
+	bucket, err := config.GetInputBucketURL()
+	if err != nil {
+		panic(err)
```
naveensrinivasan, line 63: Can we avoid panics? Is there a specific reason for not returning errors?

This makes it harder to write tests.

naveensrinivasan: This comment is applicable across panics in this func.

naveensrinivasan: Thanks. Appreciate it!

naveensrinivasan (approved):
Expect the panics; everything else is good. Thanks

calebbrown:
lgtm

PR #2451, opened 2022-11-11
:seedling: [cron] Controller should only write to the rawBucket if the bucket is set in the config.

calebbrown:
Signed-off-by: Caleb Brown <calebbrown@google.com>

This is a small bug fix to the cron controller

If the raw bucket is not specified in the controller the controller will crash (as is the case for criticality-score).

Previously the non-existent configuration setting would cause a panic, but since it was moved to the additional params section it is now only an empty value.

Writing to the raw bucket will only happen if the value exists in the config.

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cron/internal/controller/main.go
+++ b/cron/internal/controller/main.go
@@ -168,14 +168,16 @@ func main() {
 		panic(fmt.Errorf("error writing to BlobStore: %w", err))
 	}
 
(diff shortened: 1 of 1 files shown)


naveensrinivasan (approved):
Thanks

PR #2464, opened 2022-11-18
🌱 attestor: module -> subpackage

### LINKED ISSUE #2418: Keeping go.mod up-to-date for both `scorecard` and `scorecard-attestor` modules

raghavkaul:
* Make scorecard-attestor a subpackage instead of a module. 
* Fix lints. 
 
This will reduce errors with CI and with VSCode keeping multiple modules in sync.

Fixes #2418.

```
NONE
```


Makefile
```diff
@@ -334,9 +330,6 @@ e2e-gh-token: build-scorecard check-env | $(GINKGO)
 	# Run e2e tests. GITHUB_AUTH_TOKEN set to secrets.GITHUB_TOKEN must be used to run this.
 	TOKEN_TYPE="GITHUB_TOKEN" $(GINKGO) --race -p -v -cover -coverprofile=e2e-coverage.out --keep-separate-coverprofiles ./...
 
-e2e-attestor: ## Runs e2e tests for scorecard-attestor
```
spencerschrock, line 337: @naveensrinivasan Removing this target will cause the integration test to fail since the `integration.yml` file in `main` still calls `make e2e-attestor`. 

Do you think it's better to leave this as an empty target (and remove later) so the job passes? Or would you merge to main with a failing integration test since the new `integration.yml` wouldn't call this in the future? The integration check is still marked as not required, so it's an option.

Makefile
```diff
@@ -307,7 +306,7 @@ cron-worker-docker:
 
 ##@ Tests
 ################################# make test ###################################
-test-targets = unit-test unit-test-attestor e2e-pat e2e-gh-token ci-e2e
+test-targets = unit-test e2e-pat e2e-gh-token ci-e2e
```
spencerschrock, line 309: you would also need to remove `unit-test-attestor` from `all-targets`

spencerschrock (approved):

Makefile
naveensrinivasan: I think we can remove it later. Thanks

spencerschrock: Going to leave this unresolved until the target is added back.

raghavkaul: Added it back!

attestor/policy/attestation_policy.go
```diff
@@ -197,62 +200,52 @@ func toString(cs checker.Changeset) string {
...
 ) (PolicyResult, error) {
-	for _, changeset := range results.CodeReviewResults.DefaultBranchChangesets {
+	for i := range results.CodeReviewResults.DefaultBranchChangesets {
+		changeset := &results.CodeReviewResults.DefaultBranchChangesets[i]
 		numApprovers := 0
 		approvals := make(map[string]bool)
 		// CodeReview check is limited to github.com pull request reviews
 		// Log if a change isn't a github pr since it's a bit unintuitive
-		foundLinkedReviews := false
+		founloggerinkedReviews := false
```
spencerschrock, line 211: Looks like a find/replace error from `dl` -> `logger`.

spencerschrock (approved):

spencerschrock (approved):
Thanks

Issue #2493, opened 2022-11-28
Feature: Don't penalize Go projects for not "packaging" to a registry


laurentsimon:
Go projects don't need to upload their packages, so they should not be penalized.
I think we need an additional return value (-2?) to indicate NA for a check.

gabibguti:
Hey! I wanted to open a new PR to solve the packaging for Go projects. The proposal is to use [deps.dev API](https://docs.deps.dev/api/v3alpha/index.html) to get the repository related packages. If there are related Go packages, then it's a 10 for Packaging. The logic can be extended for other ecosystems too.

For example, https://github.com/bits-and-blooms/bitset, is a Go package https://pkg.go.dev/github.com/bits-and-blooms/bitset. deps.dev can identify for the repo, the related Go package https://deps.dev/project/github/bits-and-blooms%2Fbitset. We could also use pkg.go.dev API but I understand deps.dev API would be more flexible to solve NPM, CARGO, MAVEN, PYPI, NUGET packages too.

Kind of related to https://github.com/ossf/scorecard/issues/2494. Given XYZ packages in ABC ecossystems identified by deps.dev, we could output for the users the ecosystems that the repo publishes to.

laurentsimon:
Sounds like this should work! deps.dev is a single API that works for all packages, so +1 from me to use it instead of registry APIs if we can.

spencerschrock:
> Hey! I wanted to open a new PR to solve the packaging for Go projects. The proposal is to use [deps.dev API](https://docs.deps.dev/api/v3alpha/index.html) to get the repository related packages.

What's the need for any API? I thought all Go projects had this property?

Issue #2494, opened 2022-11-29
Feature: expose the ecosystem name in Packaging check


laurentsimon:
We currently don't expose the ecosystem name - raw results and in the logs.

spencerschrock:
copying from #2495

technically the name could be more complex than just the repo URI, we would need to read `go.mod`. 
e.g. scorecard would report `github.com/ossf/scorecard` even though it's now `github.com/ossf/scorecard/v4`.

running it on `https://github.com/golang/tools/` (which is a mirror) would report that instead of the module name `golang.org/x/tools`

laurentsimon:
Agreed that parsing go.mod is the right approach. Starting with the ecosystem name like `Go`, `PyPi` would be useful. Realistically, we won't be able to infer package names for other ecosystems unless the registry exposes it after verifying it themselves. Right?

azeemshaikh38:
What is the motivation for exposing the ecosystem name in Packaging check? Also, how do we handle monorepos?

laurentsimon:
the check currently looks for ecosystem-specific Actions, so we *already* surface this implcitely. If we find multiple Actions, we can identify multiples projects in monorepos.

PR #2495, opened 2022-11-29
feat: Don't penalize Go projects for not having a "packaging" workflow

laurentsimon:
closes https://github.com/ossf/scorecard/issues/2493

```release-note
Make Go project pass the packaging check automatically
```


docs/checks/internal/checks.yaml
```diff
@@ -424,7 +424,8 @@ checks:
       The check currently looks for
       [GitHub packaging workflows](https://docs.github.com/en/packages/learn-github-packages/publishing-a-package)
       and language-specific GitHub Actions that upload the package to a corresponding
-      hub, e.g., [Npm](https://www.npmjs.com/). We plan to add better support to query
+      hub, e.g., [Npm](https://www.npmjs.com/). Go projects automatically pass this check, since
```
spencerschrock, line 427: nit to stay consistent: npm

docs/checks/internal/checks.yaml
```diff
@@ -746,47 +747,29 @@ checks:
       Risk: `Low` (possible impediment to security review)
 
       This check tries to determine if the project has published a license. It
-      works by using either hosting APIs or by checking standard locations
-      for a file named according to common conventions for licenses.
+      works by checking standard locations for a file named according to common
```
spencerschrock, line 750: You've reverted some changes to the License check in multiple places in this file. Make sure you've rebased, merged those changes into this branch.

checks/evaluation/packaging.go
```diff
@@ -51,7 +51,7 @@ func Packaging(name string, dl checker.DetailLogger, r *checker.PackagingData) c
 
 	if pass {
 		return checker.CreateMaxScoreResult(name,
-			"publishing workflow detected")
+			"published package detected")
```
spencerschrock, line 54: "packaging detected"  since it's the packaging check?
WDUT?

checks/fileparser/github_workflow.go
```diff
@@ -484,6 +484,16 @@ func IsPackagingWorkflow(workflow *actionlint.Workflow, fp string) (JobMatchResu
...
+		{
+			// Dart packages.
+			Steps: []*JobMatcherStep{
+				{
+					Run: "dart.*pub.*publish",
+				},
+			},
+			LogText: "candidate dart publishing workflow",
+		},
+
```
spencerschrock, line 496: is this meant to be in this PR?

checks/raw/packaging.go
```diff
@@ -114,6 +118,41 @@ func Packaging(c *checker.CheckRequest) (checker.PackagingData, error) {
...
+		}
+		if fp == "go.sum" {
+			foundGoSum = true
+		}
+		if foundGoSum && foundGoMod {
+			pkg := checker.Package{
+				Ecosystem: ecosystemAsPointer(checker.PackageEcosystemGo),
+				Name:      stringPointer(c.Repo.URI()),
+			}
+			data.Packages = append(data.Packages, pkg)
```
spencerschrock, line 145: should this be in the `if ok` block of the caller?
I wouldn't expect `isGoProject` to need/modify data.

This could also be generalized to `identifyEcosystem` or something, if you're worried about too much logic in `func Packaging`

spencerschrock (commented):
Can you add a test? Would probably need a mockclient to test the go.mod/go.sum at root behavior.

docs/checks/internal/checks.yaml
laurentsimon: thought I had rebased. But apparently not! Thanks

checks/evaluation/packaging.go
laurentsimon: not sure :)

checks/fileparser/github_workflow.go
laurentsimon: not really, but I slipped it it as I stumbled upon it reading the code

checks/raw/packaging.go
laurentsimon: > should this be in the if ok block of the caller?

works for me. Thanks

> This could also be generalized to identifyEcosystem or something, if you're worried about too much logic in func Packaging

We can maybe do that as part of https://github.com/ossf/scorecard/issues/2494

laurentsimon:
> Can you add a test? Would probably need a mockclient to test the go.mod/go.sum at root behavior.

Done. PTAL

checks/raw/packaging.go
```diff
@@ -114,6 +127,28 @@ func Packaging(c *checker.CheckRequest) (checker.PackagingData, error) {
 	return data, nil
 }
 
+func isGoProject(c *checker.CheckRequest) (bool, error) {
+	matchedFiles, err := c.RepoClient.ListFiles(func(string) (bool, error) { return true, nil })
+	if err != nil {
+		return false, fmt.Errorf("%w", err)
+	}
```
spencerschrock, line 134: duplicate err checking?

checks/raw/packaging_test.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+func Test_Packaging(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name      string
+		filenames []string
+		err       error
+		score     int
```
spencerschrock, line 37: score seems weird to include here in `raw`. Maybe switch this to `want *string` or something? Can be done as part of #2501

spencerschrock (approved):
technically the package name could be more complex than just the URI, we would need to read `go.mod`. Not sure if this is 100% necessary to fix right now.

e.g. scorecard would report `github.com/ossf/scorecard` even though it's now `github.com/ossf/scorecard/v4`.

running it on `https://github.com/golang/tools/` (which is a mirror) would report that instead of the module name `golang.org/x/tools`

checks/raw/packaging_test.go
laurentsimon: include the score you mean?

laurentsimon: I updated the tests to check for the ecosystem name only.

azeemshaikh38:
2 high-level comments:

1) Can we generalize this to say don't penalize *any* project that doesn't have a packaging workflow? It seems to me that the right behavior should be - we found a known, trusted and accepted best practice packaging workflow, you get a 10. Else, -1.
2) Finding which ecosystem a project belongs to is a complex problem (a mono repo like https://github.com/ossf/scorecard-webapp is a good example) and we'd be creating lot of low-quality results by relying on filename and text parsing. Is there a strong reason to expose the ecosystem? If not, I'd prefer not saying anything at all.

azeemshaikh38 (changes requested):
Left a high-level comment about the motivation behind this PR.

laurentsimon:
> Let me know if I misunderstood something

quick follow-up: to display the project being Go, that info needs to be in the raw result, which is why it's populated.

I'd like to also send a follow-up PR to not penalize GitHub Actions either.

laurentsimon:
ping for reviews

olivekl (commented):
Doc additions LGTM

laurentsimon:
friendly ping

PR #2496, opened 2022-11-29
🐛  Fixes cache miss bug from commit-depth feature 

### LINKED ISSUE #2224: Feature: Support `--commit-depth`
The number of commits Scorecard analyzes is hardcoded to 30 (for GitHub) today. We can let users configure this by providing a `--commit-option`. The default can remain 30 so as we don't affect current behavior.

This will be useful to analyze developmental practices between 2 releases, for example.

@nadgowdas @lehors @leec94 fyi.

latortuga71:
Fixes cache miss issue from commit-depth feature #2224 

Currently when commits are greater than 99 and are paged, it causes the CI and SAST checks to throw cache miss informational print messages

New behavior changes the query that is used when running setupChecks. The new query takes each commit that is held in the handler, (after setup is run, commits should be populated in the handler) and runs the checkSuites query against it, this results in a 1 to 1 relationship with commits in the handler, and the checkRuns. So no cache misses happen.

Fixes #2224 

the `release-note`

```release-note

```

### DIFF
--- a/clients/githubrepo/graphql.go
+++ b/clients/githubrepo/graphql.go
@@ -161,9 +161,9 @@ type checkRunsGraphqlData struct {
 							}
 						} `graphql:"associatedPullRequests(first: $pullRequestsToAnalyze)"`
 					}
-				} `graphql:"history(first: $commitsToAnalyze)"`
+				} `graphql:"history(first:1)"`
 			} `graphql:"... on Commit"`
-		} `graphql:"object(expression: $commitExpression)"`
+		} `graphql:"object(oid: $commitExpression)"`
 	} `graphql:"repository(owner: $owner, name: $name)"`
 	RateLimit struct {
 		Cost *int
@@ -228,6 +228,21 @@ func populateCommits(handler *graphqlHandler, vars map[string]interface{}) ([]cl
 	return allCommits, nil
 }
 
+func populateCheckRuns(handler *graphqlHandler, vars map[string]interface{}) (checkRunCache, error) {
+	err := handler.client.Query(handler.ctx, handler.checkData, vars)
+	if err != nil {
+		if strings.Contains(err.Error(), "Resource not accessible by integration") {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to populate checkRuns: %w", err)
+	}
+	checkRunCacheVar := parseCheckRuns(handler.checkData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to populate checkruns: %w", err)
+	}
+	return checkRunCacheVar, nil
+}
+
 func (handler *graphqlHandler) setup() error {
 	handler.setupOnce.Do(func() {
 		commitExpression := handler.commitExpression()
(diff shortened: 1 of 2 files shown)


clients/githubrepo/graphql.go
```diff
@@ -161,9 +161,9 @@ type checkRunsGraphqlData struct {
 							}
 						} `graphql:"associatedPullRequests(first: $pullRequestsToAnalyze)"`
 					}
-				} `graphql:"history(first: $commitsToAnalyze)"`
+				} `graphql:"history(first:1)"`
```
azeemshaikh38, line 164: Won't this make the token usage far more expensive? Currently, we retrieve 30 commits in a single graphQL call, this would increase that to 30.

latortuga71: From what i have seen in the debugger no, the usage should stay as 1 because the query used for checkRuns, loops over each commit in the handler (After setup has run and populates the commits) and uses ` graphql:"object(oid: $commitExpression)"` to get only one commit. Let me know if this makes sense.

image

latortuga71: It does result in more individual calls each with a cost of 1, but i have tested with --commit-depth 300 and have had no issues so far.

azeemshaikh38: +1 to what @spencerschrock said. We are super conservative with our token usage since that helps us analyze 1M+ repos weekly - https://github.com/ossf/scorecard#public-data

spencerschrock:
> Currently when commits are greater than 99 and are paged, it causes the CI and SAST checks to throw cache miss informational print messages

The informational message was added to see how a graphql change would affect the cron. I'm not sure it's a bug to have the message displayed, although I can see how it's annoying from a user perspective. 

@azeemshaikh38 is it still worth having the cache miss message? Or hiding it behind a flag/env variable? It's still helpful when evaluating how something might impact the cronjob.

azeemshaikh38:
> The informational message was added to see how a graphql change would affect the cron. I'm not sure it's a bug to have the message displayed, although I can see how it's annoying from a user perspective.
> 
> @azeemshaikh38 is it still worth having the cache miss message? Or hiding it behind a flag/env variable? It's still helpful when evaluating how something might impact the cronjob.

We already have a [`--verbosity`](https://github.com/ossf/scorecard/blob/main/options/flags.go#L37) flag for this. Although, I suspect this isn't being used today. Ideally, the behavior should be such that `--verbosity` has a default value of `info` and these cache miss messages should only show up when the flag value is set to `debug`.

We can maybe file a P2 bug for this, but not worth trying to fix it right now.

clients/githubrepo/graphql.go
```diff
@@ -161,9 +161,9 @@ type checkRunsGraphqlData struct {
...
 			} `graphql:"... on Commit"`
-		} `graphql:"object(expression: $commitExpression)"`
+		} `graphql:"object(oid: $commitExpression)"`
```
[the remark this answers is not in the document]
azeemshaikh38: The CheckSuiteConnection and CheckSuiteEdge objects (https://docs.github.com/en/graphql/reference/objects#checksuiteconnection) return a pagination cursor. Could we use that? It would avoid requiring `handler.setup` to be called before `handler.setupCheckRuns`, can fetch multiple checkRuns in a single call saving extra GraphQL calls and we can continue supporting `expression` instead of `oid`. What do you think?

clients/githubrepo/graphql.go
latortuga71: oo yeah i didnt think about it regarding the cron. will look for alternatives

clients/githubrepo/graphql.go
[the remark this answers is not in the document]
latortuga71: i tried to get that working, but for some reason it was still throwing the misses. () Is it possible to have the query i wrote before only trigger when commit depth goes above 100? 
as far as i know cron is running with default (30) commits and that could stay the same. 

also the cache misses only trigger when going above 100 which is a very specific case and would only run when running locally using the users token. seems to work great even with 1000 commits. let me know what you think about this approach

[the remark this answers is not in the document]
azeemshaikh38: > i tried to get that working, but for some reason it was still throwing the misses.

Hmm, interesting. Do you mind sharing the code for this so that I can take a look and see if I can debug more into it?

> Is it possible to have the query i wrote before only trigger when commit depth goes above 100?

I don't feel comfortable leaving the code in that state. I'd rather we explicitly call out cache misses when there are >100 commits instead of silently switching over to making one HTTP call per commit. If we can only handle the case upto 100 commits elegantly, then maybe that's a good enough for now and we can re-visit >100 commit usecase in a future PR.

PR #2508, opened 2022-12-02
:book: Add correct contact to CODE_OF_CONDUCT.md

calebbrown:
Signed-off-by: Caleb Brown <calebbrown@google.com>

Docs Update

`[INSERT CONTACT METHOD]` is in the code of conduct.

Replace `[INSERT CONTACT METHOD]` with actual contact conduct@openssf.org (as verified in Slack).

NONE

the `release-note`

```release-note
NONE
```


azeemshaikh38 (approved):

olivekl (approved):
Sorry for the delay. Thanks for this, LGTM & approval.

spencerschrock:
The CI will never succeed for this, we'll need to fix the workflow:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

spencerschrock:
There doesn't seem to be any native way of fixing the skipped checks, for reasons that this blog sums up well.
https://blog.pantsbuild.org/skipping-github-actions-jobs-without-breaking-branch-protection/

laurentsimon:
> There doesn't seem to be any native way of fixing the skipped checks, for reasons that this blog sums up well. https://blog.pantsbuild.org/skipping-github-actions-jobs-without-breaking-branch-protection/

I've not read the link, but I think you can declare an additional job that depends on the skipped, and mark the new job `if: always()` and use that job in branch protection. We do something similar on the SLSA repos.

spencerschrock:
> > There doesn't seem to be any native way of fixing the skipped checks, for reasons that this blog sums up well. https://blog.pantsbuild.org/skipping-github-actions-jobs-without-breaking-branch-protection/
> 
> I've not read the link, but I think you can declare an additional job that depends on the skipped, and mark the new job `if: always()` and use that job in branch protection. We do something similar on the SLSA repos.

I think this is a little different. The jobs are never scheduled due to the path filtering at the top. I don't think `always()` would run as the job isn't skipped, it's never scheduled. But I can try an experiment.

spencerschrock (approved):

PR #2509, opened 2022-12-02
:warning: OSV scanner integration

### LINKED ISSUE #2162: Improve Score Reporting: OSV coverage is limited
We only query the OSV database for vulns on a commitSHA so we are extremely limited on the vulns we can report through Scorecard. There is a high possibility that we are rating a commitSHA as `10` on `Vulnerabilities` check but it actually has reported Vulnerabilities elsewhere. This will quickly erode users trust in Scorecard, need to fix this one ASAP.

another-rex:

Feature

Currently the vulnerability check only checks if the HEAD commit hash has any vulnerability specified in OSV.dev. Most of the time this will return 10/10.

This integrates the OSV-Scanner library, which will scan for manifest and SBOM files, retrieve the dependencies, and match those dependencies with OSV.dev database as well. (The commit hash check is still being done, though through the osvscanner library rather than directly setting up a client to use the OSV API).

Fixes #2162

Currently the e2e tests fail because the Scorecard score has changed since osvscanner found 3 vulnerabilities. I can change the expected results to match, but this might cause more issues in the future as new vulnerabilities are discovered, causing tests to break even though there are no code changes. Not sure what the best way to resolve this. 

Yes, TODO.

```release-note
TODO!
```

### DIFF
--- a/checker/client.go
+++ b/checker/client.go
@@ -45,7 +45,7 @@ func GetClients(ctx context.Context, repoURI, localURI string, logger *log.Logge
 			localdir.CreateLocalDirClient(ctx, logger), /*repoClient*/
 			nil, /*ossFuzzClient*/
 			nil, /*ciiClient*/
-			nil, /*vulnClient*/
+			clients.DefaultVulnerabilitiesClient(), /*vulnClient*/
 			retErr
 	}
 
--- a/checks/evaluation/vulnerabilities.go
+++ b/checks/evaluation/vulnerabilities.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/google/osv-scanner/pkg/grouper"
 	"github.com/ossf/scorecard/v4/checker"
 	sce "github.com/ossf/scorecard/v4/errors"
 )
@@ -31,20 +32,25 @@ func Vulnerabilities(name string, dl checker.DetailLogger,
 		return checker.CreateRuntimeErrorResult(name, e)
 	}
 
-	score := checker.MaxResultScore
-	IDs := []string{}
+	aliasVulnerabilities := []grouper.IDAliases{}
 	for _, vuln := range r.Vulnerabilities {
-		IDs = append(IDs, vuln.ID)
-		score--
+		aliasVulnerabilities = append(aliasVulnerabilities, grouper.IDAliases(vuln))
 	}
+
+	IDs := grouper.Group(aliasVulnerabilities)
+	score := checker.MaxResultScore - len(IDs)
 
 	if score < checker.MinResultScore {
 		score = checker.MinResultScore
 	}
 
 	if len(IDs) > 0 {
+		outputIDs := []string{}
+		for _, v := range IDs {
+			outputIDs = append(outputIDs, strings.Join(v.IDs, " / "))
+		}
 		dl.Warn(&checker.LogMessage{
-			Text: fmt.Sprintf("HEAD is vulnerable to %s", strings.Join(IDs, ", ")),
+			Text: fmt.Sprintf("HEAD is vulnerable to:\n%s", strings.Join(outputIDs, "\n")),
 		})
 		return checker.CreateResultWithScore(name,
 			fmt.Sprintf("%v existing vulnerabilities detected", len(IDs)), score)
--- a/checks/evaluation/vulnerabilities.go
+++ b/checks/evaluation/vulnerabilities.go
@@ -45,13 +45,12 @@ func Vulnerabilities(name string, dl checker.DetailLogger,
 	}
 
 	if len(IDs) > 0 {
-		outputIDs := []string{}
 		for _, v := range IDs {
-			outputIDs = append(outputIDs, strings.Join(v.IDs, " / "))
+			dl.Warn(&checker.LogMessage{
+				Text: fmt.Sprintf("HEAD is vulnerable to: %s", strings.Join(v.IDs, " / ")),
+			})
 		}
-		dl.Warn(&checker.LogMessage{
-			Text: fmt.Sprintf("HEAD is vulnerable to:\n%s", strings.Join(outputIDs, "\n")),
-		})
+
 		return checker.CreateResultWithScore(name,
 			fmt.Sprintf("%v existing vulnerabilities detected", len(IDs)), score)
 	}
--- a/checks/evaluation/vulnerabilities.go
+++ b/checks/evaluation/vulnerabilities.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	"github.com/google/osv-scanner/pkg/grouper"
+
 	"github.com/ossf/scorecard/v4/checker"
 	sce "github.com/ossf/scorecard/v4/errors"
 )
@@ -47,7 +48,7 @@ func Vulnerabilities(name string, dl checker.DetailLogger,
 	if len(IDs) > 0 {
 		for _, v := range IDs {
 			dl.Warn(&checker.LogMessage{
-				Text: fmt.Sprintf("HEAD is vulnerable to: %s", strings.Join(v.IDs, " / ")),
+				Text: fmt.Sprintf("Project is vulnerable to: %s", strings.Join(v.IDs, " / ")),
 			})
 		}
 
--- a/clients/osv.go
+++ b/clients/osv.go
@@ -53,23 +53,28 @@ func (v osvClient) HasUnfixedVulnerabilities(
 
 	for _, v := range res.Flatten() {
 		response.Vulnerabilities = append(response.Vulnerabilities, Vulnerability{
-			ID: v.Vulnerability.ID,
+			ID:      v.Vulnerability.ID,
+			Aliases: v.Vulnerability.Aliases,
 		})
 		// Remove duplicate vulnerability IDs for now as we don't report information
 		// on the source of each vulnerability yet, therefore having multiple identical
 		// vuln IDs might be confusing.
-		response.Vulnerabilities = removeDuplicate(response.Vulnerabilities)
+		response.Vulnerabilities = removeDuplicate(
+			response.Vulnerabilities,
+			func(key Vulnerability) string { return key.ID },
+		)
 	}
 	return response, nil
 }
 
-// RemoveDuplicate removes duplicate entries from a slice.
-func removeDuplicate[T comparable](sliceList []T) []T {
-	allKeys := make(map[T]bool)
+// RemoveDuplicate removes duplicate entries from a slice
+func removeDuplicate[T any, K comparable](sliceList []T, keyExtract func(T) K) []T {
+	allKeys := make(map[K]bool)
 	list := []T{}
 	for _, item := range sliceList {
-		if _, value := allKeys[item]; !value {
-			allKeys[item] = true
+		key := keyExtract(item)
+		if _, value := allKeys[key]; !value {
+			allKeys[key] = true
 			list = append(list, item)
 		}
 	}
--- a/clients/osv.go
+++ b/clients/osv.go
@@ -15,62 +15,63 @@
 package clients
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
-	"net/http"
+	"fmt"
 
-	"github.com/ossf/scorecard/v4/errors"
+	"github.com/google/osv-scanner/pkg/osvscanner"
 )
 
 var _ VulnerabilitiesClient = osvClient{}
 
 type osvClient struct{}
 
-const osvQueryEndpoint = "https://api.osv.dev/v1/query"
-
-type osvQuery struct {
-	Commit string `json:"commit"`
-}
-
-type osvResp struct {
-	Vulns []struct {
-		ID string `json:"id"`
-	} `json:"vulns"`
-}
-
 // HasUnfixedVulnerabilities implements VulnerabilityClient.HasUnfixedVulnerabilities.
-func (v osvClient) HasUnfixedVulnerabilities(ctx context.Context, commit string) (VulnerabilitiesResponse, error) {
-	query, err := json.Marshal(&osvQuery{
-		Commit: commit,
-	})
+func (v osvClient) HasUnfixedVulnerabilities(
+	ctx context.Context,
+	commit,
+	localPath string,
+) (VulnerabilitiesResponse, error) {
+	directoryPaths := []string{}
+	if localPath != "" {
+		directoryPaths = append(directoryPaths, localPath)
+	}
+	gitCommits := []string{}
+	if commit != "" {
+		gitCommits = append(gitCommits, commit)
+	}
+	res, err := osvscanner.DoScan(osvscanner.ScannerActions{
+		DirectoryPaths: directoryPaths,
+		SkipGit:        true,
+		Recursive:      true,
+		GitCommits:     gitCommits,
+	}, nil) // TODO: Do logging?
 	if err != nil {
-		return VulnerabilitiesResponse{}, errors.WithMessage(err, "failed to marshal query")
+		return VulnerabilitiesResponse{}, fmt.Errorf("osvscanner.DoScan: %w", err)
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, osvQueryEndpoint, bytes.NewReader(query))
-	if err != nil {
-		return VulnerabilitiesResponse{}, errors.WithMessage(err, "failed to create request")
-	}
+	response := VulnerabilitiesResponse{}
 
-	httpClient := &http.Client{}
-	resp, err := httpClient.Do(req)
-	if err != nil {
-		return VulnerabilitiesResponse{}, errors.WithMessage(err, "failed to send request")
-	}
-	defer resp.Body.Close()
-
-	var osvresp osvResp
-	decoder := json.NewDecoder(resp.Body)
-	if err := decoder.Decode(&osvresp); err != nil {
-		return VulnerabilitiesResponse{}, errors.WithMessage(err, "failed to decode response")
-	}
-
-	var ret VulnerabilitiesResponse
-	for _, vuln := range osvresp.Vulns {
-		ret.Vulnerabilities = append(ret.Vulnerabilities, Vulnerability{
-			ID: vuln.ID,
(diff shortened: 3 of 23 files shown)


checker/raw_result.go
```diff
@@ -186,7 +188,7 @@ type ContributorsData struct {
 // VulnerabilitiesData contains the raw results
 // for the Vulnerabilities check.
 type VulnerabilitiesData struct {
-	Vulnerabilities []clients.Vulnerability
+	OSVVulnerabilities osvmodels.VulnerabilityResults
```
laurentsimon, line 191: Generally, we want this structure to be agnostic to the database we query, so that if we ever add other databases (CVEs), we'd be able to do that. This means you'd need to translate some fields to the `clients.Vulnerability` structure.

An alternative - what you propose - is to expose OSV format directly. I think this also has benefits: users familiar with OSVs already know how to parse the results. But I'm not sure if we want to expose that level of granularity within Scorecard, since users who care about this can just query OSV directly or use a URL Scorecard can provide. 

One downside of exposing OSV directly is that if you bump the major version, Scorecard will need to bump its version too.

So I think I'm in favor of keeping the results database agnostic, ie keep `clients.Vulnerability`. The ID should already be unique `OSV-XXX` vs `CVE-XXX`. 

@azeemshaikh38 wdut?

laurentsimon (commented):
Thanks for this PR. I left one major comment about keeping `clients.Vulnerabilities` field.

checker/raw_result.go
another-rex: That makes sense to me, I have reverted the change here so that the interface is the same to avoid the major version bump. We can always come back to this later if we want to expose more details.

azeemshaikh38: +1. The way I visualized this is, we'd extend the `VulnerablitiesData` struct such that it includes a mapping of project name and its associated vulnerability. A list of such a mapping can easily encapsulate both the project's direct and indirect vulnerabilities. Then an underlying implementation (like OSV Scanner) would fill in this information to the best of it ability.

azeemshaikh38:
> a challenge to adding the unit tests for this is /that it will be picked up as a vulnerability, even though it's part of the test data in the repository. Any ideas for a quick fix? Eventually we want to have project level ignore files for osv-scanner.

unit tests should be hermetic and should not be making any network calls. You can use the mocks generated for the interfaces to write any unit tests and it shouldn't affect any vulnerability finding.

checker/client.go
```diff
@@ -45,7 +45,7 @@ func GetClients(ctx context.Context, repoURI, localURI string, logger *log.Logge
 			localdir.CreateLocalDirClient(ctx, logger), /*repoClient*/
 			nil, /*ossFuzzClient*/
 			nil, /*ciiClient*/
-			nil, /*vulnClient*/
+			clients.DefaultVulnerabilitiesClient(), /*vulnClient*/
```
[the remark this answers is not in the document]
another-rex: This is because I added the Vulnerabilities check to support FileBased as well as the original CommitBased. 
https://github.com/ossf/scorecard/blob/ca78e15ab8e75806e941104c049b517e21af4a81/checks/vulnerabilities.go#L29-L32
Since now we can scan manifests as well as the git commit hash.

clients/repo_client.go
```diff
@@ -32,6 +32,7 @@ type RepoClient interface {
 	URI() string
 	IsArchived() (bool, error)
 	ListFiles(predicate func(string) (bool, error)) ([]string, error)
+	LocalPath() (string, error)
```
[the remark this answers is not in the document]
another-rex: :+1: Done

another-rex:
> unit tests should be hermetic and should not be making any network calls. You can use the mocks generated for the interfaces to write any unit tests and it shouldn't affect any vulnerability finding.

That makes sense, what about e2e tests? It looks like a some repositories are set up in [ossf-tests](https://github.com/orgs/ossf-tests), can I set up a project with some lockfiles that contain known vulnerable dependencies?

azeemshaikh38:
> > unit tests should be hermetic and should not be making any network calls. You can use the mocks generated for the interfaces to write any unit tests and it shouldn't affect any vulnerability finding.
> 
> That makes sense, what about e2e tests? It looks like a some repositories are set up in [ossf-tests](https://github.com/orgs/ossf-tests), can I set up a project with some lockfiles that contain known vulnerable dependencies?

Yes that's possible. Curious, does OSV-Scanner look at direct dependencies only or transitive dependencies too?

laurentsimon:
> > > unit tests should be hermetic and should not be making any network calls. You can use the mocks generated for the interfaces to write any unit tests and it shouldn't affect any vulnerability finding.
> > 
> > 
> > That makes sense, what about e2e tests? It looks like a some repositories are set up in [ossf-tests](https://github.com/orgs/ossf-tests), can I set up a project with some lockfiles that contain known vulnerable dependencies?
> 
> Yes that's possible. Curious, does OSV-Scanner look at direct dependencies only or transitive dependencies too?

https://github.com/ossf-tests/scorecard-check-osv-e2e

Send a PR and cc me on it - I don't receive notifications automatically for these repos

inferno-chromium:
@another-rex - Please try to finish this soon, so a scorecard release can be cut by end of week or coming Monday.

another-rex:
> > > unit tests should be hermetic and should not be making any network calls. You can use the mocks generated for the interfaces to write any unit tests and it shouldn't affect any vulnerability finding.
> > 
> > 
> > That makes sense, what about e2e tests? It looks like a some repositories are set up in [ossf-tests](https://github.com/orgs/ossf-tests), can I set up a project with some lockfiles that contain known vulnerable dependencies?
> 
> Yes that's possible. Curious, does OSV-Scanner look at direct dependencies only or transitive dependencies too?

OSV-Scanner also looks at transitive dependencies, since it scans the lockfile, which generally lists the transitive dependencies.

naveensrinivasan:
> @another-rex - Please try to finish this soon, so a scorecard release can be cut by end of week or coming Monday.

Is there a specific reason we want to cut a release?

inferno-chromium:
> > @another-rex - Please try to finish this soon, so a scorecard release can be cut by end of week or coming Monday.
> 
> Is there a specific reason we want to cut a release?

I mean this is a very strong feature where scorecard was not looking at transitive deps before and can now can look at vulns transitively. By release, i just meant a minor release (not major announcement or anything for scorecard). OSV-scanner will go out next week, so it will be nice for people to try in scorecard too, not a big deal if you need more review time.

naveensrinivasan:
> > > @another-rex - Please try to finish this soon, so a scorecard release can be cut by end of week or coming Monday.
> > 
> > 
> > Is there a specific reason we want to cut a release?
> 
> I mean this is a very strong feature where scorecard was not looking at transitive deps before and can now can look at vulns transitively. By release, i just meant a minor release (not major announcement or anything for scorecard). OSV-scanner will go out next week, so it will be nice for people to try in scorecard too, not a big deal if you need more review time.

Yes, that makes sense! 👍 

Thanks for the update. We need to get better at releases. To be honest, we need to improve our release process by calling out new features and breaking changes in the release notes.

naveensrinivasan:
> Integration tests success for
> [[f405955](https://github.com/ossf/scorecard/commit/f4059551aa416f6cddbf938ac6facbd4d2175cf3)]
> (https://github.com/ossf/scorecard/actions/runs/3654448698)

Can you please rebase with the main? Also, DCO is missing. Thanks

another-rex:
> Can you please rebase with the main? Also, DCO is missing. Thanks

Done!

checks/evaluation/vulnerabilities.go
```diff
@@ -31,20 +32,25 @@ func Vulnerabilities(name string, dl checker.DetailLogger,
...
-	IDs := []string{}
+	aliasVulnerabilities := []grouper.IDAliases{}
 	for _, vuln := range r.Vulnerabilities {
-		IDs = append(IDs, vuln.ID)
-		score--
+		aliasVulnerabilities = append(aliasVulnerabilities, grouper.IDAliases(vuln))
 	}
 
+	IDs := grouper.Group(aliasVulnerabilities)
+	score := checker.MaxResultScore - len(IDs)
```
azeemshaikh38, line 41: Should we continue to keep the behavior as binary - if vuln found `0` else `10` ?

checks/evaluation/vulnerabilities.go
```diff
@@ -31,20 +32,25 @@ func Vulnerabilities(name string, dl checker.DetailLogger,
 		return checker.CreateRuntimeErrorResult(name, e)
 	}
 
-	score := checker.MaxResultScore
-	IDs := []string{}
+	aliasVulnerabilities := []grouper.IDAliases{}
 	for _, vuln := range r.Vulnerabilities {
-		IDs = append(IDs, vuln.ID)
-		score--
+		aliasVulnerabilities = append(aliasVulnerabilities, grouper.IDAliases(vuln))
```
azeemshaikh38, line 37: Out of curiosity, what does the `grouper` package do? Is it generic enough to be used for non-OSV usecases ?

azeemshaikh38 (approved):

checks/evaluation/vulnerabilities.go
```diff
@@ -31,20 +32,25 @@ func Vulnerabilities(name string, dl checker.DetailLogger,
...
+	score := checker.MaxResultScore - len(IDs)
+
 	if score < checker.MinResultScore {
 		score = checker.MinResultScore
 	}
 
 	if len(IDs) > 0 {
+		outputIDs := []string{}
+		for _, v := range IDs {
+			outputIDs = append(outputIDs, strings.Join(v.IDs, " / "))
```
laurentsimon, line 50: how about providing a `warn` message for each vuln? This would also make it nicer for the GitHub dashboard: each warn is reported as a standalone alert the user can fix independently.

checks/evaluation/vulnerabilities.go
laurentsimon: IIUC, we currently -1 for each vuln found https://github.com/ossf/scorecard/blob/main/checks/evaluation/vulnerabilities.go#L36-L38

docs/checks/internal/checks.yaml
```diff
@@ -691,14 +691,20 @@ checks:
     description: |
       Risk: `High`  (known vulnerabilities)
 
-      This check determines whether the project has open, unfixed vulnerabilities
-      using the [OSV (Open Source Vulnerabilities)](https://osv.dev/) service. An open
-      vulnerability is readily exploited by attackers and should be fixed as soon as
+      This check determines whether the project has open, unfixed vulnerabilities 
+      in it's dependencies using the [OSV (Open Source Vulnerabilities)](https://osv.dev/) service.
```
laurentsimon, line 695: maybe say `unfixed vulnerabilities in its own codebase or its dependencies`
/cc @olivekl

docs/checks/internal/checks.yaml
```diff
@@ -691,14 +691,20 @@ checks:
...
+      An open vulnerability is readily exploited by attackers and should be fixed as soon as
       possible.
     remediation:
       - >-
         Fix the vulnerabilities. The details of each vulnerability can be found
```
laurentsimon, line 700: maybe add or "update your dependencies to a non-vulnerable version"

olivekl: Yeah, need to include remediation tips for vulnerable dependencies. Something like: "If the vulnerability is in a dependency, update the dependency to a non-vulnerable version. If no update is available, consider whether to remove the dependency."

docs/checks/internal/checks.yaml
```diff
@@ -691,14 +691,20 @@ checks:
...
         on <https://osv.dev>.
+      - >-
+        If you believe the vulnerability does not actually affect your project, the
```
olivekl, line 703: nit: remove "actually" (not needed)

docs/checks/internal/checks.yaml
```diff
@@ -691,14 +691,20 @@ checks:
...
+      in it's dependencies using the [OSV (Open Source Vulnerabilities)](https://osv.dev/) service.
+      An open vulnerability is readily exploited by attackers and should be fixed as soon as
       possible.
     remediation:
       - >-
         Fix the vulnerabilities. The details of each vulnerability can be found
         on <https://osv.dev>.
+      - >-
+        If you believe the vulnerability does not actually affect your project, the 
+        vulnerability can be ignored by creating a `osv-scanner.toml` file next to the
```
olivekl, line 704: Consider breaking into two sentences for simplicity/easy reading: "the vulnerability can be ignored. To ignore, create an [...] and specify the..."

docs/checks/internal/checks.yaml
```diff
@@ -691,14 +691,20 @@ checks:
...
     remediation:
       - >-
         Fix the vulnerabilities. The details of each vulnerability can be found
         on <https://osv.dev>.
+      - >-
+        If you believe the vulnerability does not actually affect your project, the 
+        vulnerability can be ignored by creating a `osv-scanner.toml` file next to the 
+        dependency manifest (e.g. package-lock.json) and specifying the ID to ignore and reason
+        Details on the structure of `osv-scanner.toml` can be 
+        found on [OSV-Scanner repository](https://github.com/google/osv-scanner).
```
olivekl, line 707: Is there a more specific heading on the osv-scanner repo you can link to?

checks/evaluation/vulnerabilities.go
another-rex: As long as a list of vulnerabilities have ID fields and Alias fields, grouper will be able to group vulnerabilities together so that the same vulnerability published by multiple advisories won't be counted multiple times.

More details on the alias field here: https://ossf.github.io/osv-schema/#aliases-field

docs/checks/internal/checks.yaml
another-rex: Are you sure about adding `its own codebase`? While we do scan the commit hash, it might imply that we also scan the actual source code which we do not.

docs/checks/internal/checks.yaml
another-rex: Yep, updated the link to: https://github.com/google/osv-scanner#ignore-vulnerabilities-by-id

docs/checks/internal/checks.yaml
olivekl: The benefit there is to make the distinction between vulnerabilities in the project itself and vulnerabilities in its dependencies. Is there another way you'd say it so there's no confusion about what OSV is doing? (FWIW, I don't think saying "unfixed vulns in its own codebase" implies you're scanning the codebase itself, but I can see where you're coming from.)

clients/osv.go
```diff
@@ -15,62 +15,68 @@
...
-}
-
 // HasUnfixedVulnerabilities implements VulnerabilityClient.HasUnfixedVulnerabilities.
-func (v osvClient) HasUnfixedVulnerabilities(ctx context.Context, commit string) (VulnerabilitiesResponse, error) {
-	query, err := json.Marshal(&osvQuery{
-		Commit: commit,
-	})
-	if err != nil {
-		return VulnerabilitiesResponse{}, errors.WithMessage(err, "failed to marshal query")
+func (v osvClient) HasUnfixedVulnerabilities(
```
laurentsimon, line 29: maybe rename to `ListUnfixedVulnerabilities()`, since the function does not return a bool

docs/checks/internal/checks.yaml
olivekl: Isn't this addition expanding what was possible before, which was reporting a vulnerability in the project codebase? If the behavior there hasn't changed, I don't think we should remove the old wording or remediation suggestion. I might be misunderstanding, though?

checks/evaluation/vulnerabilities.go
laurentsimon: If they don't support it, it aliases field would be empty, and `IDAliases()` would return the same list. Correct?

checks/evaluation/vulnerabilities.go
```diff
@@ -31,21 +32,25 @@ func Vulnerabilities(name string, dl checker.DetailLogger,
...
 		score = checker.MinResultScore
 	}
 
 	if len(IDs) > 0 {
-		dl.Warn(&checker.LogMessage{
-			Text: fmt.Sprintf("HEAD is vulnerable to %s", strings.Join(IDs, ", ")),
-		})
+		for _, v := range IDs {
+			dl.Warn(&checker.LogMessage{
+				Text: fmt.Sprintf("HEAD is vulnerable to: %s", strings.Join(v.IDs, " / ")),
```
laurentsimon, line 50: Let's remove `HEAD`, since scorecard also works on commits, local repos, etc

docs/checks.md
```diff
@@ -644,14 +644,15 @@ Additionally, points are reduced if certain write permissions are defined for a
 
 Risk: `High`  (known vulnerabilities)
 
-This check determines whether the project has open, unfixed vulnerabilities
-using the [OSV (Open Source Vulnerabilities)](https://osv.dev/) service. An open
-vulnerability is readily exploited by attackers and should be fixed as soon as
+This check determines whether the project has open, unfixed vulnerabilities 
+in its dependencies using the [OSV (Open Source Vulnerabilities)](https://osv.dev/) service.
```
laurentsimon, line 648: do we still want to report the vuln in the project itself? I'd say that we want, no?

docs/checks/internal/checks.yaml
laurentsimon: same question from me.

another-rex: Yeah, that is correct, this is just expanding on the functionality from before, and this commit only adds scanning for dependencies on top of what was originally there. 

I just think the original functionality is so limited that while in theory it is possible for it to report vulnerabilities, it probably has never (or very rarely) reported any vulnerabilities. 

I can see that it is confusing to change the wording when the functionality hasn't been removed, so let's keep the `its own codebase` part.

docs/checks.md
another-rex: Added it back in.

checks/evaluation/vulnerabilities.go
another-rex: Yep

laurentsimon (approved):

laurentsimon (approved):

PR #2519, opened 2022-12-04
✨ Accept Go modules with semantic versions

### LINKED ISSUE #2491: Consider versioned go modules pinned
Currently, any `go install` and `go get` without using an SHA-256 hash is considered unpinned. I believe this is overkill and goes against the current practice.

I think versioned Go modules should be considered pinned because newer go tools (>= 1.13), by default, will verify the downloads against <https://sum.golang.org>. The checksum service <https://sum.golang.org/> has made it almost impossible for the original source to change a published release. Moreover, the design of the checksum service made it possible for others to audit the server itself. To change a versioned module, one would have to temper with the data on <https://sum.golang.org/>, and if that can be done without any auditor noticing it, we could very well doubt whether the go tools themselves are secure. It is perhaps more reasonable to assume Go module versions are practically as permanent as SHA-256 hashes.

**Additional context**
An empty version string or `latest` should still be considered unpinned.

favonia:

This PR accepts commands like `go install blah@v1.0.0`, considering fully versioned Go modules pinned.

`go install blah@v1.0.0` is currently considered unpinned, but it is effectively pinned in practice.

Fixes #2491.

* This also fixes the current parser in two ways:
   1. Do not detect `go INSTALL` but only `go install`.
   1. Skip flags between `get`/`install` and packages but record the flag `-insecure`.

* `goconst` is turned off for `"none"` because the string has very different meanings in different places in the package.

```release-note
Go modules with full semantic versions are now considered pinned.
```

### DIFF
--- a/checks/raw/permissions.go
+++ b/checks/raw/permissions.go
@@ -180,7 +180,7 @@ func typeOfPermission(val string) checker.PermissionLevel {
 	switch val {
 	case "read", "read-all":
 		return checker.PermissionLevelRead
-	case "none":
+	case "none": //nolint:goconst
 		return checker.PermissionLevelNone
 	}
 	return checker.PermissionLevelUnknown
--- a/checks/raw/pinned_dependencies_test.go
+++ b/checks/raw/pinned_dependencies_test.go
@@ -888,7 +888,7 @@ func TestShellScriptDownload(t *testing.T) {
 		{
 			name:     "pkg managers",
 			filename: "./testdata/script-pkg-managers",
-			warns:    38,
+			warns:    43,
 		},
 		{
 			name:     "invalid shell script",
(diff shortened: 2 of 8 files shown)


laurentsimon (commented):
Thank you for sending the PR.
Could you:
1. Add  `go install bla@latest` / `upgrade` / `patch` to the unit tests to match the description of the issue?
2. Add semver that don't comply with the rules, like `go install package@v1.2` and `go install package@v1`
3. Add the similar unit tests to the other files: Dockerfile-pkg-managers, .github/workflows/github-workflow-pkg-managers.yaml, just to be sure they behave consistently.

Then let's merge!

laurentsimon (approved):
Thank you!

favonia:
@iamamoose Oops I did `git rebase --signoff` to pass the DCO, which canceled the auto-merge :sweat_smile:

laurentsimon (approved):

PR #2532, opened 2022-12-09
🐛  dependency update tool + e2e tests

laurentsimon:

The dependency update tool check looks at commit https://github.com/ossf/scorecard/pull/2125. But the `registerCheck()` was not updated to indicate that the check no longer support local repos.

This PR:
1. Fix the above
2. Add e2e tests for the check and other checks that require API calls.

```release-note
Dependency update tool out-of-sync with local repo + e2e tests
```


naveensrinivasan (approved):
Thanks

Issue #2522, opened 2022-12-05
BUG: Regression in LICENSE check, outputs 22,000 lines for details


pnacht:
**Describe the bug**
The Gradle repo's JSON has 22,000 lines describing the License check.

```json
"details": [
    "Info: License file found in expected location: LICENSE:1",
    "...",
    "Info: License file found in expected location: COPYING.MIT:1",
    "Info: FSF or OSI recognized license: COPYING.MIT:1",
    "...",
    "Info: License file found in expected location: LICENSE.md:1",
    "Info: FSF or OSI recognized license: LICENSE.md:1",
    "...",
    "Info: License file found in expected location: LICENSE.txt:1",
    "Info: FSF or OSI recognized license: LICENSE.txt:1",
    "...",
    "Warn: Any licence detected not an FSF or OSI recognized license: LICENSE:1",
    "...",
    "Info: License file found in expected location: MIT-LICENSE:1",
    "Info: FSF or OSI recognized license: MIT-LICENSE:1",
    "...",
    "Info: License file found in expected location: UNLICENSE:1",
    "Warn: Any licence detected not an FSF or OSI recognized license: UNLICENSE:1",
    "...",
]
```

where each `"..."` is dozens of repetitions of these results (mostly the basic "Info: ... LICENSE:1").

Also worth mentioning the only file that actually exists in the [Gradle repo](https://github.com/gradle/gradle) is `LICENSE` on the root.

I also checked cpython's json (which I'd seen two weeks ago and was perfectly fine), and it also has the same issue.

**Reproduction steps**
- https://api.securityscorecards.dev/projects/github.com/gradle/gradle
- https://api.securityscorecards.dev/projects/github.com/python/cpython

Both runs are dated "2022-11-28".

**Expected behavior**
A single `"Info: License file found in expected location: LICENSE:1"` line in the `details` block (maybe a few other items?).

azeemshaikh38:
This is terrible! Can we please revert the change which caused this and make a minor release? @laurentsimon I believe you worked on this PR with a contributor?

pnacht:
I've also run this locally and the output is similar to what @laurentsimon shows above for both projects.

Not sure if this was a transitory bug or if there's a difference between what's happening in the CLI and the job. A good check will happen naturally in a few days when the projects are rerun by the job (the error happened in the 2022-11-28 run).

```shell
scorecard --repo=python/cpython --checks=License --show-details --format=json | jq
```

```json
{
[...]

Issue #2533, opened 2022-12-09
Unit tests for scorecard run multiple times


laurentsimon:
In https://github.com/ossf/scorecard/issues/2522, we found out that it's easy to make a mistake by not initializing client's variables. To avoid this in the future, we should have additional tests that run scorecard twice and verify the results are the same. We can start by comparing that the number of warn / info / debug details is identical.

spencerschrock:
> We can start by comparing that the number of warn / info / debug details is identical.

+1 on this. I tried a little to compare the results in more depth, but the comparison is difficult due to the non-deterministic ordering of many of the slices.

laurentsimon:
right, to go further you can order the slices. I think go-cmp package has support for this. Maybe too complicated for what you're trying to do?

spencerschrock:
It's somewhat feasible to order the details, it's very annoying to order all the slices and maps in all the raw data results

PR #2534, opened 2022-12-09
🐛  license check accumulates results

laurentsimon:
closes https://github.com/ossf/scorecard/issues/2522

This PR fixes https://github.com/ossf/scorecard/issues/2522.

To avoid this type of problems in the future, we need to add check that runs scorecard twice (with the same clients) and verify the results are all the same: https://github.com/ossf/scorecard/issues/2533

```release-note
fix: license check accumulates results
```


naveensrinivasan (approved):
Thanks

PR #2541, opened 2022-12-13
🐛  Remove 32bit platforms from GoReleaser architectures

laurentsimon:

OSV-Scanner depends on a library that cannot be built for 32bit platforms, so this PR removes them from released binaries.

```release-note
Remove 32bit platforms from GoReleaser architectures
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -31,8 +31,6 @@ builds:
   goarch:
     - amd64
     - arm64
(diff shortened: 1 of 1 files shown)


spencerschrock (approved):
Do we have any metrics about how much 32bit gets used? I assume this is broken for anyone building from source after #2509?

It would be good to get this back later if possible

Issue #987, opened 2021-09-09
Feature - Enable Codeql Check for GitHubAction


naveensrinivasan:
**Is your feature request related to a problem? Please describe.**
The scorecard repository has GitHub Actions that can run into script injections. https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#understanding-the-risk-of-script-injections.

To mitigate this there is CodeQL to analyze the code. https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-codeql-to-analyze-your-code

**Describe the solution you'd like**
To enable the CodeQL check for checking for insecure GitHubAction the repository needs a _one JavaScript file_ and CodeQL must be configured to make to analyze for Javascript.

PR #2547, opened 2022-12-15
🌱 Test infra: Workflow improvements

raghavkaul:
Simplify GitHub actions workflows.
* Remove unneeded actions
* Filter workflows by file paths affected on PR
* Consolidate some jobs into a single job to reduce overall step count

Today, we run into several rate limiting issues with GitHub action pulls. This PR intends to reduce the number of action pulls/API calls and improve overall test efficiency.

All tests still run on main:
See https://github.com/raghavkaul/scorecard/commits/main:
Screen Shot 2022-12-15 at 10 58 30 AM

Docs changes only re-trigger main and docs builds
* See: https://github.com/raghavkaul/scorecard/pull/5/checks

Cron changes re-trigger cron builds
* See: https://github.com/raghavkaul/scorecard/pull/7/checks

Consolidating build steps still results in individual steps being run:
https://github.com/raghavkaul/scorecard/actions/runs/3707777727/jobs/6284604744


.github/workflows/docs.yml
```diff
@@ -0,0 +1,121 @@
+# Copyright 2022 OpenSSF Scorecard Authors
```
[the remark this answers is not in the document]
laurentsimon: yeaah, scorecard on PR is working!

/cc @naveensrinivasan

laurentsimon (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'go','javascript' ]
+        language: [ 'go' ]
```
laurentsimon, line 52: thoughts on moving this one to daily run and have it create issues when it fails?

raghavkaul: Sgtm, changed it to run daily.

laurentsimon: do you know how if issues will be created? I think it's fine to do that in another PR entirely, since it's likely to require a bit more thoughts. Sorry about that.

@naveensrinivasan let us know if you'd be OK exploring codeQL daily only.

raghavkaul: I believe they should be available through the Security tab in Repo settings: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#viewing-the-alerts-for-a-repository

laurentsimon: You're right. I'm just worried we never check them. I certainly never do :)

naveensrinivasan: Why was the `javascript` removed? https://github.com/ossf/scorecard/issues/987

.github/workflows/docs.yml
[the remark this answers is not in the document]
naveensrinivasan: Cool!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,124 @@
...
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
```
naveensrinivasan, line 25: This is going to be hard to maintain. Now, this information is split between code and config.

.github/workflows/codeql-analysis.yml
```diff
@@ -27,11 +27,8 @@ name: "CodeQL"
 on:
   push:
     branches: [ main ]
-  pull_request:
```
naveensrinivasan, line 30: Why is the PR check removed? We will not catch if there are any issues in the PR.

naveensrinivasan (commented):
Thanks for all the changes! But I have few questions.

.github/workflows/codeql-analysis.yml
laurentsimon: Since codeQL realistically never finds any results and makes pre-submit slower, we've been thinking of running in only on daily schedules, and use an API like  https://docs.github.com/en/rest/code-scanning?apiVersion=2022-11-28 to retrieve result and alert us (create an issue or as pre-submit) if codeQL has found anything.

naveensrinivasan: OK, sounds good. Thanks

.github/workflows/codeql-analysis.yml
naveensrinivasan: 👍

.github/workflows/docs.yml
raghavkaul: What about changing this to `docs/*` which would be a little less brittle?

.github/workflows/codeql-analysis.yml
raghavkaul: Wasn't aware of that CodeQL could transitively scan Github actions, I can add Javascript back.

Though, isn't that something scorecard-action can also do via `actionlint`?

.github/workflows/docs.yml
naveensrinivasan: Yes, that works! Thanks

.github/workflows/codeql-analysis.yml
laurentsimon: we'd love to have Dangerous-Workflow check look into Actions (there's an issue for this), but not sure about the dependency Action of the project under scan.

raghavkaul: Added back `javascript` to CodeQL

laurentsimon:
@naveensrinivasan any last comment before we merge?

naveensrinivasan:
> @naveensrinivasan any last comment before we merge?

We are good.

laurentsimon (approved):

laurentsimon:
@spencerschrock @azeemsgoogle @naveensrinivasan @raghavkaul the pre-submit are not starting. Did we change anything recently?

spencerschrock:
> @spencerschrock @azeemsgoogle @naveensrinivasan @raghavkaul the pre-submit are not starting. Did we change anything recently?

is it from changes in the PR?

```
    paths:
      - "cron/**/*.docker"
      - "cron/**/Dockerfile"
      - "cron/**/*.go"
      - "clients/**/*.go"
```

none of these files were modified, so none of the cron checks will be skipped. this is the same issue as #2566

spencerschrock:
Since the filtering is different (positive instead of negative) I think this can be fixed by following [GitHub's answer](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks). Or we can use the same third-party action in the other PR #2566 (but with a different filter)

raghavkaul:
Having a problem with the changed_files filter, still debugging. Other than that, the CodeQL and build-* checks will need to be taken off the "required" set since this PR consolidates them all into a single check.

raghavkaul:
Side note: 
We could go further by consolidating all the cron Docker checks into a single multi-step job like we did with the build-* checks. It would take about ~12 minutes (added runtimes from the cron checks), but that's still shorter than the 16 minutes it takes to run integration tests, so it wouldn't hurt us overall.

PR #2565, opened 2022-12-22
📖 specify that classic personal access tokens are needed

mmaraya:
Signed-off-by: Mike Maraya <mmaraya@users.noreply.github.com>

#### What kind of change does this PR introduce? 

Doc update to clarify that classic, not fine-grained, personal access tokens are required for the scorecard CLI to work

(Is it a bug fix, feature, docs update, something else?) docs update

- [x] PR title follows the guidelines defined in our [pull request documentation](https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process)

#### What is the current behavior? documentation currently links to the creation of a fine-grained access token, which produces this error:
```
Error: RunScorecard: internal error: ListCommits:error during graphqlHandler.setup: internal error: githubv4.Query: non-200 OK status code: 401 Unauthorized body: "{\"message\":\"Personal access tokens with fine grained access do not support the GraphQL API\",\"documentation_url\":\"https://docs.github.com/graphql/guides/forming-calls-with-graphql#authenticating-with-graphql\"}"
```

#### What is the new behavior (if this is a feature change)?** n/a, doc change only

- [ ] Tests for the changes have been added (for bug fixes/features): n/a

#### Which issue(s) this PR fixes

NONE

#### Does this PR introduce a user-facing change?

For user-facing changes, please add a concise, human-readable release note to
the `release-note`

(In particular, describe what changes users might need to make in their
application as a result of this pull request.)
 

```release-note
NONE
```

PR #2566, opened 2022-12-22
🌱 Switch from paths-ignore to changed-files action to skip required checks.

spencerschrock:

CI fix

Due to path filtering, PRs which only change Markdown files will never run the steps in `docker.yml`:
https://github.com/ossf/scorecard/blob/c6d76807b76b7b72659525e103d6b1284c7b01b5/.github/workflows/docker.yml#L20-L30

Since these checks are (currently) required, PRs which only change Markdown files can never be merged. See #2565 and #2508. As long as these checks are required, there's no native way to solve this for reasons summarized well in [this blog post](https://blog.pantsbuild.org/skipping-github-actions-jobs-without-breaking-branch-protection/).

A third party action is used to detect doc only changes. Since skipped checks count as passing for required checks, this allows for doc only changes to merge.

NONE

I tested this with a few PRs in a fork here:
https://github.com/spencerschrock/scorecard-test/pull/3
https://github.com/spencerschrock/scorecard-test/pull/8

the `release-note`

```release-note
NONE
```

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,24 +21,45 @@ on:
  push:
   branches:
   - main
-  paths-ignore:
-  - "*.md"
  pull_request:
   branches:
   - main
-  paths-ignore:
-  - "*.md"
 
 env:
   PROTOC_VERSION: 3.17.3
-  GO_VERSION: 1.17
+  GO_VERSION: 1.19
 
 jobs:
+  docs_only_check:
+    name: Check for docs-only change
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      docs_only: ${{ steps.docs_only_check.outputs.docs_only }}
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+    - id: files
+      name: Get changed files
+      uses: tj-actions/changed-files@0626c3f94002c0a9d7491dd7fed7055bbdff6f92 #v35.1.0
+      with:
+        files_ignore: '**.md'
+    - id: docs_only_check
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -21,24 +21,45 @@ on:
  push:
   branches:
   - main
```
spencerschrock, line 23: Also meant to ask about this. Do we need to check the docker builds again after merging? We could save some API usage by removing it, and it wont affect publishing the docker containers which is done with Cloud Build.

.github/workflows/docker.yml
```diff
@@ -21,24 +21,45 @@
...
+  docs_only_check:
+    name: Check for docs-only change
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      docs_only: ${{ steps.docs_only_check.outputs.docs_only }}
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v3
```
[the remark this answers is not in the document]
spencerschrock: I'll fix this as well before merging.

.github/workflows/docker.yml
raghavkaul: My opinion is that we should remove it, to reduce overall steps (assuming the publish flow does a fresh build after merge).

raghavkaul:
Thanks for this change! Looks good to me.

.github/workflows/docker.yml
naveensrinivasan: I agree. 👍

naveensrinivasan (approved):
Thanks

laurentsimon (approved):

PR #2571, opened 2022-12-27
✨ Disable scorecard on PRs

laurentsimon:
We keep hit rate limits on PRs, so let's disable anything that's not necessary for now.

```release-note
Disable scorecard on PRs
```


naveensrinivasan (approved):

spencerschrock:
Note: `Scorecard analysis` will need to be marked as a non-required check. 
The changes in the PR mean the analysis will never run, so no PR can get merged so long as it's required.

spencerschrock (approved):

Issue #2189, opened 2022-08-23
BUG: Scorecard does not penalize unpinned third party actions in local composite / docker actions


stefreak:
**Describe the bug**
It's possible to call third party actions or docker containers in local GitHub actions, and call the local GitHub actions from a workflow.

scorecard does not seem to verify local GitHub actions in my tests

This might be related to https://github.com/ossf/scorecard/issues/2174 but I am not sure.

**Reproduction steps**
- [ ] Create a repo to reproduce this

**Expected behavior**
All GitHub action YAML files that can call third party actions should be considered.

**Additional context**

arianvp:
The fact that Github Actions doesn't have a lock-file that locks all recursive action dependencies of composite actions is kind of meh. It makes it very iffy to use from a supply-chain security perspective

PR #2594, opened 2023-01-11
🐛 Handle pinning by Docker URLs for GitHub actions workflows

### LINKED ISSUE #2580: BUG Pinned-Dependencies check does not handle direct docker GitHub Actions correctly
**Describe the bug**
Pinned-Dependency check does not detect a direct docker based action (uses statement is the docker container) correctly and fails to see that sha hash pinning is used, as advised. Therefor an alert is generated that should not be the case.

**Reproduction steps**
1. Use a docker image as an action with for example: 
`uses: docker://rhysd/actionlint@sha256:5f957b2a08d223e48133e1a914ed046bea12e578fe2f6ae4de47fdbe691a2468 # 1.6.22`
3. Run the `ossf/scorecard-action` action in a workflow
4. Result: a `Pinned-Dependencies` alert that you cannot resolve:
image

**Expected behavior**
Detect that this is proper use of a sha hash and not alert, or close an already open alert.

**Additional context**
Note: I am using the `ossf/scorecard-action` in a workflow to run the checks. Alert is integrated with the GitHub Advanced Security Code scanning alerts.

raghavkaul:
Handle Docker/other container registry URLs for GitHub actions workflows. `[jobs.<job_id>.steps[*].uses](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses)` in GitHub actions can take GitHub repos, docker containers, and local filepaths. This change supports repos and containers but not local filepaths.

Only public action pinning was detected.

Handle Docker image pinning. Stricter regex.

Fixes #2580.

```release-note
NONE
```


checks/raw/pinned_dependencies.go
```diff
@@ -498,3 +494,28 @@ var validateGitHubActionWorkflow fileparser.DoWhileTrueOnFileContent = func(
 
 	return true, nil
 }
+
+func isActionDependencyPinned(actionUses string) bool {
+	// Though GitHub Actions also allows actions to be specified as local filepaths
+	// (e.g. uses: <path/to/file/in/repo>), it's not yet practical to determine whether an
+	// arbitrary file that lives in the action context is a trustworthy, pinned workflow file
+	// So, those actions are treated as 'unpinned'
```
spencerschrock, line 502: I would argue pinning dependencies isn't about trust, it's about fixing dependencies you don't control. I'm not sure I would treat them as unpinned, although they would need to be analyzed for unpinned actions of their own?

checks/raw/pinned_dependencies.go
```diff
@@ -498,3 +494,28 @@ var validateGitHubActionWorkflow fileparser.DoWhileTrueOnFileContent = func(
...
+	publicActionRegex := regexp.MustCompile(`^([\w-]+)/([\w-]+)(/([\w-/]+))?(@([a-f\d]+)|([\w\.-]+))?`)
+
+	if m := publicActionRegex.FindStringSubmatch(actionUses); len(m) > 0 {
+		ref := m[6]
+		// Ensure a hash at least as large as SHA1 is used (40 hex characters).
+		return len(ref) >= 40
+	}
+
+	//nolint:lll
+	dockerhubActionRegex := regexp.MustCompile(`docker://([\w\.-]+/)?([\w\.-]+)(/[\w\.-:]+)?((@sha256:([a-f\d]+))|(:[\w\.\-]+))?`)
```
spencerschrock, line 513: My understanding is `_` is also valid in docker image names.

e.g. https://github.com/falconry/falcon/blob/0d808d99e83bc1aba6e062c4fa4433e32cb8c3b3/.github/workflows/create-wheels.yaml#L266

Docker hosts might have colons for a port? 
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-docker-public-registry-action

I noticed some `:` in a later group: `(/[\w\.-:]+)?`
That might allow multiple colons, not sure.

checks/raw/pinned_dependencies.go
```diff
@@ -498,3 +494,28 @@ var validateGitHubActionWorkflow fileparser.DoWhileTrueOnFileContent = func(
...
 	return true, nil
 }
+
+func isActionDependencyPinned(actionUses string) bool {
+	// Though GitHub Actions also allows actions to be specified as local filepaths
+	// (e.g. uses: <path/to/file/in/repo>), it's not yet practical to determine whether an
+	// arbitrary file that lives in the action context is a trustworthy, pinned workflow file
+	// So, those actions are treated as 'unpinned'
+
+	publicActionRegex := regexp.MustCompile(`^([\w-]+)/([\w-]+)(/([\w-/]+))?(@([a-f\d]+)|([\w\.-]+))?`)
```
spencerschrock, line 504: Should also double check if underscores are valid here.

checks/raw/pinned_dependencies.go
```diff
@@ -498,3 +494,28 @@ var validateGitHubActionWorkflow fileparser.DoWhileTrueOnFileContent = func(
...
+
+	if m := publicActionRegex.FindStringSubmatch(actionUses); len(m) > 0 {
+		ref := m[6]
```
spencerschrock, line 507: Does `6` correspond to the hash capture group? Maybe make that clear with a constant?

Is it possible for this to be out of bounds? You check there are capture groups, but not necessarily that there are more than 6.

I'm unfamiliar with the function, so it could be a guarantee provided by it
https://pkg.go.dev/regexp#Regexp.FindStringSubmatch

checks/raw/pinned_dependencies.go
```diff
@@ -498,3 +494,28 @@ var validateGitHubActionWorkflow fileparser.DoWhileTrueOnFileContent = func(
...
+		// Ensure a hash at least as large as SHA1 is used (40 hex characters).
+		return len(ref) >= 40
+	}
+
+	//nolint:lll
+	dockerhubActionRegex := regexp.MustCompile(`docker://([\w\.-]+/)?([\w\.-]+)(/[\w\.-:]+)?((@sha256:([a-f\d]+))|(:[\w\.\-]+))?`)
+
+	if m := dockerhubActionRegex.FindStringSubmatch(actionUses); len(m) > 0 {
+		tag := m[6]
```
spencerschrock, line 516: same question about the `6`, both in terms of a magic number and out of bounds.

checks/raw/pinned_dependencies.go
```diff
@@ -498,3 +494,28 @@ var validateGitHubActionWorkflow fileparser.DoWhileTrueOnFileContent = func(
...
+		return len(tag) >= 40
```
spencerschrock, line 517: if the regex looks for sha256, should this number be exactly 64?

checks/raw/pinned_dependencies.go
laurentsimon, line 507: verify the length of the `m` is at least 7.

checks/raw/pinned_dependencies.go
laurentsimon, line 516: verify the length of m is at least 6

spencerschrock (commented):
Long regexes can get tricky. Would it be simpler to use the current one but add a fall back or an `?` group for the cases that start with `@sha256:`?

checks/raw/testdata/.github/workflows/workflow-not-pinned.yaml
```diff
@@ -34,6 +34,10 @@ jobs:
         # a pull request then we can checkout the head.
         fetch-depth: 2
```
laurentsimon, line 36: can we add tests for images hosted from other registries (gcr, ghr)? (unless they are not supported)

laurentsimon (approved):
LGTM with some additional tests and length verification to the right size.

Thanks!

checks/raw/pinned_dependencies.go
raghavkaul: True, this comment is mostly referring to #2189. My hypothesis is that anything the action's local context can be used by `uses`, i.e. a step could write an arbitrary file and then the next step would use that. I'm just leaving a comment to revisit this behavior, this PR doesn't change how we handle those paths.

checks/raw/pinned_dependencies.go
raghavkaul: From what I observed, if that group isn't captured, FindStringSubmatch will return an empty string at that submatch. The array size is fixed relative to groups defined in the pattern.

raghavkaul:
Thanks both, I'll revisit/simplify the regex and add some tests.

checks/raw/pinned_dependencies.go
laurentsimon: I think we can treat local actions as pinned, since they are "vendored". I think https://github.com/ossf/scorecard/issues/2189 can be addressed by parsing the action.yml of the local action and looking for pinned dependencies in the yaml file.

raghavkaul: But isn't it possible for an action to modify or overwrite other vendored actions, effectively unpinning them?

laurentsimon: anytime you run an Action, it overwrites the $GITHUB_WORKSPACE directory, so yes. But that's how Actions work and their code is only available for the duration of the call.
```yaml
uses: checkout
uses: ./my/local/action
uses: some/external-action // This action overwrites the previous code
```

The goal of pinning Action is to ensure they don't change when the dependency has a new release, unless the user LGTM the changes. It also helps for inventory: you know exactly what code you ran (assuming the dependency also pins its dependency.. which may not be true :/). The code *may* be overwritten by malicious code, but that's a harder problem. At least we know what was the code of the direct dependency is .. ideally we'd carry on till the last further-away indirect dependency

raghavkaul: Gotcha. Yes, it's hard to capture all the ways that the local context may change before someone uses a local action, but it's a really common pattern, especially with `.github/workflows` and `.github/actions` - so we should allow it for now. But I think there's another improvement here, which is that if someone uses the `actions/checkout` action that's bundled with GitHub, we should make sure that the checkout is pinned to a SHA.

checks/raw/pinned_dependencies.go
raghavkaul: Changed this to `.`.

laurentsimon (approved):

spencerschrock (approved):
Looks good thanks

PR #2625, opened 2023-01-30
🌱  Add RepoClient re-use E2E tests.

### LINKED ISSUE #2533: Unit tests for scorecard run multiple times

spencerschrock:
e2e test

No e2e tests which test for accumulation of results.

Adds a simple test that the results for repository B are the same whether calculated:
* B in isolation
* A then B, reusing the repo client

This e2e test was backported to 28b116f1a79f548b3b3bd595d8e379d0b7cadeaf to test that it would have caught the bug.
It found issues with both `Contributors` and `License` accumulating results. 

##### Contributors
```
  "Contributors" reason mismatch: "55 different organizations found -- score normalized to 10" vs "21 different organizations found -- score normalized to 10"
  "Contributors" reason mismatch: "21 different organizations found -- score normalized to 10" vs "55 different organizations found -- score normalized to 10"
  "Contributors" reason mismatch: "21 different organizations found -- score normalized to 10" vs "55 different organizations found -- score normalized to 10"
  "Contributors" reason mismatch: "55 different organizations found -- score normalized to 10" vs "21 different organizations found -- score normalized to 10"
  "Contributors" reason mismatch: "21 different organizations found -- score normalized to 10" vs "55 different organizations found -- score normalized to 10"
  "Contributors" reason mismatch: "55 different organizations found -- score normalized to 10" vs "21 different organizations found -- score normalized to 10"
```
##### License
```
"License" details mismatch:
      debug: 0-0
      info: 4-2
      warn: 0-0
"License" details mismatch:
      debug: 0-0
      info: 2-4
      warn: 0-0
"License" details mismatch:
      debug: 0-0
      info: 2-4
      warn: 0-0
"License" details mismatch:
      debug: 0-0
      info: 4-2
      warn: 0-0
"License" details mismatch:
      debug: 0-0
      info: 2-4
      warn: 0-0
"License" details mismatch:
      debug: 0-0
      info: 4-2
      warn: 0-0
```

Fixes #2533 

I'm not 100% certain why the results are being outputted multiple times.
The E2E test currently runs all scorecard checks 3 times, which results in the following increase in token API usage:

* REST: 152 / 5000
* Search: 6/30
* graphql: 6/5000

the `release-note`

```release-note
NONE
```


spencerschrock:
We could also structure this in a way such that it only runs on a certain `make` command, but not part of the normal E2E suite if we're concerned about the increase in API usage.

spencerschrock (commented):
We could also structure this in a way such that it only runs on a certain make command, but not part of the normal E2E suite if we're concerned about the increase in API usage.

laurentsimon (approved):

laurentsimon:
> We could also structure this in a way such that it only runs on a certain make command, but not part of the normal E2E suite if we're concerned about the increase in API usage.

how would the make command avoid increasing API usage?

spencerschrock:
> > We could also structure this in a way such that it only runs on a certain make command, but not part of the normal E2E suite if we're concerned about the increase in API usage.
> 
> how would the make command avoid increasing API usage?

By making it available for contributors/maintainers to run on their own, without including it in the e2e suite that runs on every PR. Of course, the trade-off is remembering to run it when it is needed.

naveensrinivasan:
@spencerschrock A friendly ping.

spencerschrock:
This test exposed a data race that has been present for a while, which will need to be fixed before these tests can be merged.

PR #2280, opened 2022-09-20
:sparkles: Gitlab support (Part 2) - Tests

N8BWert:
## ----NOTE----

This PR should probably not be merged until the ossf-tests GitLab organization is made with the necessary tests.  This branch, however, can be used by those who want to perform experimental scorecard evaluations of their GitLab projects.

#### What kind of change does this PR introduce?

This PR creates the necessary GitLab support e2e tests and includes the experimental ability to conduct scorecard checks on GitLab repositories from the command line.

#### What is the current behavior?

I recently made a pull request laying down the foundations of how a gitlab client would be written, which is extended with this PR with the addition of tests and test frameworks.

#### What is the new behavior (if this is a feature change)?**

Most of the tests to demonstrate GitLab support for scorecard are demonstrated in this PR.  There are a few tests that I cannot verify (such as CI tests which require higher level GitLab accounts for testing).  For these tests the basic code outline is given meaning that anyone could quickly insert the url of a test repository for the test to work.
- [✔️] Tests for the changes have been added (for bug fixes/features)

#### Which issue(s) this PR fixes

Fixes #2266 

#### Special notes for your reviewer

A large number of these tests reference GitLab projects that are under my name.  I assume that Scorecard will likely create an organization like ossf-tests in GitLab where all of the test repositories will be hosted.  When this happens the tests in this PR can be modified to include the ossf-test project urls, allowing this to likely be merged into the main scorecard repository.

#### Does this PR introduce a user-facing change?
If a user is running scorecard on a GitHub repository that includes the string "gitlab." in the title scorecard will now think that repository is actually a GitLab project, however this is an incredibly rare edge case.

#### Release-note
Repositories that include "gitlab." in their title will be assumed to be gitlab projects.

#### Link to previous PR:
https://github.com/ossf/scorecard/pull/2265#issue-1374867978

PR #2265, opened 2022-09-15
:sparkles: Gitlab support

N8BWert:
#### What kind of change does this PR introduce?
This PR introduces GitLab functionality to the scorecard project.

#### What is the current behavior?
Currently scorecard only supports testing GitHub repositories.

#### What is the new behavior (if this is a feature change)?**
A client has been added to allow the command line execution of scorecard to include support for gitlab projects.

#### Which issue(s) this PR fixes
[#Support For GitLab Projects](https://github.com/ossf/scorecard/issues/2266#issue-1374875930)
https://github.com/ossf/scorecard/issues/2266#issue-1374875930

#### Special notes for your reviewer
It is important to note that projectID is not the same as the name of the project.  Scorecard will not find the GitLab project in the following situation:

Example: Project named test-project with projectID 1234 owned by steve who is a part of examplecompany
- This will not work:
```
scorecard gitlab.examplecompany.com/steve/test-project
```
+ However this will work:
```
scorecard gitlab.examplecompany.com/steve/1234
```

#### Does this PR introduce a user-facing change?
If a user is running scorecard on a GitHub repository that includes the string "gitlab." in the title scorecard will now think that repository is actually a GitLab project, however this is an incredibly rare edge case.

```release-note
Repositories that include "gitlab." in their title will be assumed to be gitlab projects.
```

PR #2655, opened 2023-02-13
✨ Gitlab support: RepoClient

### LINKED ISSUE #2266: Support for GitLab Projects
Scorecard only supports Github Repositories, but GitLab Projects would also benefit from scorecard checks.

A GitLab client to allow scorecard checks on GitLab Projects.

raghavkaul:
This PR adds tests for GitLab support in Scorecard. Some work remains: Some tests require a GitLab test organization, or test repos, or other scaffolding.

This work builds off @N8BWert's work in #2280, but updates the branch with main.

Fixes #2266.

If a user is running scorecard on a GitHub repository that includes the string "gitlab." in the title scorecard will now think that repository is actually a GitLab project, however this is an incredibly rare edge case.

#### Release Notes
Repositories that include "gitlab." in their title will be assumed to be gitlab projects.

Link to previous PR:
https://github.com/ossf/scorecard/pull/2265#issue-1374867978


checker/client.go
```diff
@@ -49,14 +54,28 @@ func GetClients(ctx context.Context, repoURI, localURI string, logger *log.Logge
...
-	if errGitHub != nil {
-		return githubRepo,
-			nil,
-			nil,
-			nil,
-			nil,
-			fmt.Errorf("getting local directory client: %w", errGitHub)
+	_, experimental := os.LookupEnv("SCORECARD_EXPERIMENTAL")
+
+	if strings.Contains(repoURI, "gitlab.") && experimental {
```
laurentsimon, line 59: maybe create a different function `makeRepo()`

laurentsimon (approved):

checker/client.go
raghavkaul: A GitLab self-hosted deployment could be named anything, not sure if this check is valid. If repoURI had to match exactly `gitlab.com`, we'd only support GitLab.com. 

Another approach that doesn't rely on domain names could be a `--gitlab-repo` flag.

kgangerlm: > A GitLab self-hosted deployment could be named anything, not sure if this check is valid. If repoURI had to match exactly `gitlab.com`, we'd only support GitLab.com.
> 
> Another approach that doesn't rely on domain names could be a `--gitlab-repo` flag.

What about checking if the repository host responds to the gitlab api? For example `https://gitlab.com/api/v4/version` that endpoint does require authentication, but there are some that don't.

raghavkaul: We can now detect self-hosted GitLab repos by checking if they respond to an unauthenticated API, also added tests to `clients/gitlabrepo/repo_test.go`.

azeemshaikh38 (changes requested):
Thanks @raghavkaul. Could you please split this PR into smaller ones? Maybe one PR per check? There are 50 changed files with 160 commits in this PR. That's a lot of code to review reasonably.

.github/workflows/integration.yml
```diff
@@ -71,3 +71,19 @@ jobs:
...
+          max_attempts: 3
+          retry_on: error
+          timeout_minutes: 30
+          command: make e2e-gitlab
+
+      - name: codecov
+        uses: codecov/codecov-action@81cd2dc8148241f03f5839d295e000b8f761e378 # 2.1.0
+        with:
+         files: ./e2e-coverage.out
+         verbose: true
```
azeemshaikh38, line 89: Not needed? Seems like GitLab e2e tests don't output this file.

Makefile
```diff
@@ -334,6 +334,9 @@ e2e-gh-token: build-scorecard check-env | $(GINKGO)
 	# Run e2e tests. GITHUB_AUTH_TOKEN set to secrets.GITHUB_TOKEN must be used to run this.
 	TOKEN_TYPE="GITHUB_TOKEN" $(GINKGO) --race -p -v -cover -coverprofile=e2e-coverage.out --keep-separate-coverprofiles ./...
 
+e2e-gitlab: ## Runs e2e tests for GitLab only. TOKEN_TYPE is arbitrary, but must be set to something
+	TOKEN_TYPE="GITLAB_PAT" $(GINKGO) --race -p -vv --focus '.*GitLab' ./...
```
azeemshaikh38, line 338: -> Why do we need `TOKEN_TYPE` for GitLab?
-> Can we run `./clients/gitlabrepo/...` and remove `--focus '.*GitLab'`?

checker/client.go
```diff
@@ -50,20 +53,55 @@ func GetClients(ctx context.Context, repoURI, localURI string, logger *log.Logge
...
+				fmt.Errorf("getting local directory client: %w", makeRepoError)
+		}
+		repoClient = ghrepo.CreateGithubRepoClient(ctx, logger)
+	}
+
+	//nolint:staticcheck
+	ossFuzzRepoClient, errOssFuzz := ghrepo.CreateOssFuzzRepoClient(ctx, logger)
+	var retErr error
+	if errOssFuzz != nil {
+		retErr = fmt.Errorf("getting OSS-Fuzz repo client: %w", errOssFuzz)
```
azeemshaikh38, line 98: @spencerschrock does this override your change on using specialized OssFuzz client?

clients/repo.go
```diff
@@ -17,6 +17,7 @@ package clients
 // Repo interface uniquely identifies a repo.
 type Repo interface {
 	URI() string
+	Host() string
```
azeemshaikh38, line 20: We don't seem to use this `Host()` API much apart from in tests. Any reason to add this?

clients/gitlabrepo/client.go
```diff
@@ -222,7 +221,7 @@ func (client *Client) Close() error {
 }
 
 func CreateGitlabClientWithToken(ctx context.Context, token string, repo clients.Repo) (clients.RepoClient, error) {
-	client, err := gitlab.NewClient(token, gitlab.WithBaseURL(repo.URI()))
+	client, err := gitlab.NewClient(token, gitlab.WithBaseURL(repo.Host()))
```
azeemshaikh38, line 224: Since we have access to internal members of `repoURL` struct, we can just access `repo.host` instead ?

clients/gitlabrepo/client.go
```diff
@@ -269,10 +268,22 @@ func CreateGitlabClientWithToken(ctx context.Context, token string, repo clients
...
 }
 
 // TODO(#2266): implement CreateOssFuzzRepoClient.
 func CreateOssFuzzRepoClient(ctx context.Context, logger *log.Logger) (clients.RepoClient, error) {
 	return nil, fmt.Errorf("%w, oss fuzz currently only supported for github repos", clients.ErrUnsupportedFeature)
 }
+
+// DetectGitLab: check whether the repoURI is a GitLab URI
+// Makes HTTP request to GitLab API.
+func DetectGitLab(repoURI string) bool {
```
azeemshaikh38, line 282: Add tests for this func?

azeemshaikh38 (approved):
Looks good overall to submit. Left some minor comments.

checker/client.go
spencerschrock: Yes, `CreateOssFuzzRepoClient` is marked deprecated. I'm surprised the linter hasn't complained about this.

Makefile
spencerschrock: it may prevent the gitlab tests from running during `make e2e-pat` or `make e2e-gh-token` ?

azeemshaikh38: Correct, IIUC `e2e-pat` and `e2e-gh-token` should be running the GitLab tests already give scope of `./...`. So either we remove the extra `e2e-gitlab` or narrow the scopes for it to only run on GitLab.

clients/repo.go
raghavkaul: The main reason is that GitLab enterprise hosts (and GitHub enterprise hosts) may vary from just `gitlab|github.com`, so exposing the parameter may help for logging, etc.

clients/gitlabrepo/client.go
raghavkaul: Do you mean by exporting the variable as `Host` instead of using a getter? I don't think we can access the private field from here.

Makefile
raghavkaul: Updated this:
* e2e-gitlab-token will be a make target that can be used from the CLI. The user can pass in their GitLab token to run this test.
* e2e-gitlab will be a make target for unauthenticated tests that can run from CI. The rate limit is 500 requests/minute, which we should be well under for now. If this is an issue, it can be split into its own workflow file and marked optional.
* We can't just run ./clients/gitlabrepo since some future tests will be in `e2e/`

.github/workflows/integration.yml
raghavkaul: Subsequent PRs will add tests for this.

clients/gitlabrepo/client.go
azeemshaikh38: You should be able to access the private field of `repoURL` struct since its in the same package.

PR #2678, opened 2023-02-17
:seedling: Update Go version to 1.20

naveensrinivasan:
- Update Go version to `1.20` across all .yml and .yaml files

[.github/workflows/publishimage.yml]
- Change the Go version from `1.17` to `1.20` [go.mod]
- Update the Go version from `1.19` to `1.20` [.github/workflows/integration.yml]
- Update Go version from `1.19` to `1.20` [.github/workflows/main.yml]
- Change Go version from `1.19` to `1.20` [.github/workflows/slsa-goreleaser.yml]
- Update Go version from `1.19` to `1.20` [.github/workflows/goreleaser.yaml]
- Update Go version from `1.19` to `1.20`


spencerschrock (commented):
Places that are still 1.19:
* `.github/workflows/docker.yml`

There's also a 8 Dockerfiles that reference the `golang` image. At the very least the main Dockerfile should bump here.
```
# golang:1.19
FROM golang@sha256:25de7b6b28219279a409961158c547aadd0960cf2dcbc533780224afa1157fd4 AS base
```

Dependabot is apparently bugged when pinned without the tag, so we need to include the tag when pinning:

```
FROM golang:1.20@sha256:2edf6aab2d57644f3fe7407132a0d1770846867465a39c2083770cf62734b05d AS base
```

Places that still use/reference 1.17
* CONTRIBUTING.md says 1.17 minimum (this may not need to go to 1.20, but we use `strings.Cut` so at least 1.18)

naveensrinivasan:
> * CONTRIBUTING.md says 1.17 minimum (this may not need to go to 1.20, but we use `strings.Cut` so at least 1.18)

Hi there,

I'm able to build scorecards locally, but I don't understand what you mean by `strings.Cut`. Could you please explain it to me?

Thanks!

spencerschrock:
> added in go1.18

https://pkg.go.dev/strings#Cut 

I thought we used it (I think I tried to include it in the past but ended up using a different implementation), but apparently our dependencies do. Which is why publishimage was failing
https://github.com/ossf/scorecard/actions/runs/4199180538/jobs/7283722674#step:7:190

naveensrinivasan:
> > added in go1.18
> 
> https://pkg.go.dev/strings#Cut
> 
> I thought we used it (I think I tried to include it in the past but ended up using a different implementation), but apparently our dependencies do. Which is one reason why publishimage was failing https://github.com/ossf/scorecard/actions/runs/4199180538/jobs/7283722674#step:7:190

Looks like we need an upgrade that https://github.com/Abirdcfly/dupword/blob/main/go.mod, which is moved to 1.19 and shouldn't have an issue.

naveensrinivasan:
> There's also a 8 Dockerfiles that reference the `golang` image. At the very least the main Dockerfile should bump here.
> 
> ```
> # golang:1.19
> FROM golang@sha256:25de7b6b28219279a409961158c547aadd0960cf2dcbc533780224afa1157fd4 AS base
> ```
> 
> Dependabot is apparently bugged when pinned without the tag, so we need to include the tag when pinning:
> 
> ```
> FROM golang:1.20@sha256:2edf6aab2d57644f3fe7407132a0d1770846867465a39c2083770cf62734b05d AS base
> ```
> 
> Places that still use/reference 1.17
> 
> * CONTRIBUTING.md says 1.17 minimum (this may not need to go to 1.20, but we use `strings.Cut` so at least 1.18)

For Dockerfiles lets do a separate PR if dependabot doesn't update it.

naveensrinivasan:
Need lot more work. Closing it for now

PR #2704, opened 2023-03-01
:bug: Updates osv-scanner dependency to 1.2.0.

another-rex:

 Updates osv-scanner dependency to 1.2.0 from 0.0.0.
The 1.0 release changed the return value for osv-scanner to output an error when vulnerabilities are found, modified to handle this error correctly. This will fix https://github.com/ossf/scorecard-action/issues/1092 once it is also updated.

https://github.com/ossf/scorecard-action/issues/1092

No

```release-note
NONE
```


clients/osv.go
```diff
@@ -45,26 +46,32 @@ func (v osvClient) ListUnfixedVulnerabilities(
...
-		// vuln IDs might be confusing.
-		response.Vulnerabilities = removeDuplicate(
-			response.Vulnerabilities,
-			func(key Vulnerability) string { return key.ID },
-		)
+	if err == nil {
+		return response, nil
+	}
+
+	if errors.Is(err, osvscanner.VulnerabilitiesFoundErr) {
```
laurentsimon, line 55: should this be an `!errors.Is()`? I may be mis-understanding what `osvscanner.VulnerabilitiesFoundErr`. Can you add a comment? Does it mean that *some* results were populated, but not all?

spencerschrock: I don't believe so based on the code for `DoScan` https://github.com/google/osv-scanner/blob/6eab4504663395b4f5983ebe4570e8b816844ab5/pkg/osvscanner/osvscanner.go#L469-L479

another-rex: Added some additional comments to clarify it, Spencer is correct, if vulnerabilities are found VulnerabilitiesFoundErr will be returned.

laurentsimon (approved):

another-rex:
Updated the osv-scanner version to a pseudo version 1.2.1-0.20230302232134-592acbc2539b, which contains some extra fixes, notably :
- https://github.com/google/osv-scanner/pull/258 
- https://github.com/google/osv-scanner/pull/250 

@spencerschrock fyi

spencerschrock:
@another-rex Thanks for the pseudo update. If you could just DCO your last commit I can get this merged in.

PR #2716, opened 2023-03-02
🌱 Add projects to cronjob

pnacht:

Adds 7 popular projects to cronjob:

- bufbuild/protoc-gen-validate
- rust-lang/cfg-if
- codemirror/codemirror5
- firebase/flutterfire
- google/zerocopy
- openjdk/jdk19u
- tukaani-project/xz

Note that two of these repositories already exist in projects.csv, but with old names:

- [bufbuild/protoc-gen-validate](https://github.com/bufbuild/protoc-gen-validate) exists under [envoyproxy/protoc-gen-validate](https://github.com/envoyproxy/protoc-gen-validate)
- [rust-lang/cfg-if](https://github.com/rust-lang/cfg-if) exists under [alexcrichton/cfg-if](github.com/alexcrichton/cfg-if)

Given that the Scorecard API doesn't perform redirects from old repo names to new ones, we need to update projects occasionally. I wasn't sure whether it's best to replace them or add them as duplicates (in case someone relies on the old name). Let me know if I should replace them instead.


spencerschrock (approved):
Thanks.

> Given that the Scorecard API doesn't perform redirects from old repo names to new ones, we need to update projects occasionally.

This will probably need some discussion that is beyond the scope of this PR. In my opinion, what you have now works. There may be some duplicate work done, but both names will work on the scorecard side of things. Since each result is generated separately, there is a possibility for some inconsistent results between old/new name.

PR #2719, opened 2023-03-03
🐛 Check OSS Fuzz build file for Fuzzing check

### LINKED ISSUE #2670: BUG: Fuzzing check: OSS-Fuzz detection intermittent
**Describe the bug**
The Fuzzing check intermittently fails to search the `google/oss-fuzz` repo. This is causing the check to fail at the CLI as well as the integration tests to fail.

* https://github.com/ossf/scorecard/actions/runs/4177522887/jobs/7235192440#step:7:12242
* https://github.com/ossf/scorecard/actions/runs/4172074986/jobs/7222747174#step:7:12515
* etc

**Reproduction steps**
1. `scorecard --repo ossf/scorecard --checks Fuzzing --show-details --format json | jq`
2.  The reason string no longer mentions the project being fuzzed with `OSSFuzz` despite the project being fuzzed: [link](https://github.com/google/oss-fuzz/blob/9f8d8b8a61a00862bdc70d0c3dcb31be886d3c5f/projects/ossf-scorecard/project.yaml#L4)

**Expected behavior**

The project should detect OSSFuzz

**Additional context**

I think it's a problem with the code search API. The Fuzzing check makes a code search API request like the following:
```
curl \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer $GITHUB_AUTH_TOKEN"\
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/search/code?q=github.com+ossf+scorecard+repo:google/oss-fuzz
```

Using the search box on the website shows valid results.

spencerschrock:

bug fix

The fuzzing check uses a GitHub repo client to search `google/oss-fuzz` for a project to determine if it's fuzzed.
Searching the `google/oss-fuzz` hasn't been working for a while.

There's a new `ossfuzz` client which only implements the `Search` method.  
This method fetches the OSS Fuzz status file once and parses the projects.

Using a client involves stubbing all of the unneeded methods, but allows the change to be made in a backwards compatible manner for everyone that calls `pkg.RunScorecard`.

Fixes #2670 

the `release-note`

```release-note
"githubrepo.CreateOssFuzzRepoClient" has been deprecated. If you directly call this function, please switch to 
"ossfuzz.CreateOSSFuzzClient" or "ossfuzz.CreateOSSFuzzClientEager". If you use `checker.GetClients`, no change is needed.
```


naveensrinivasan (approved):
This cool! Thanks

spencerschrock:
Hmm, the unit tests work locally for me when run with `make unit-test`.
When I emulate the GitHub action with `act -r -j unit-test -W ./.github/workflows/main.yml` it's failing like here. Some of our other tests do an `os.ReadFile` without problem. Will need to debug on Monday.

laurentsimon (approved):

azeemshaikh38 (changes requested):
Do we really need a OssFuzzClient at all? How bad would it be if we always made a HTTP call and parsed the JSON?

spencerschrock:
> Do we really need a OssFuzzClient at all? How bad would it be if we always made a HTTP call and parsed the JSON?

It's very simple to do it just with a HTTP call, the initial commit did just that (ignore the byte search instead of JSON parsing which is very much needed)
https://github.com/ossf/scorecard/blob/5335f4be3cdd8ad2d86873b9a1b9dc0589031192/checks/raw/fuzzing.go#L173-L185

But there are numerous problems with this.
- Each `pkg.RunScorecard()` call would redownload a nearly 1MB file for each repo in the cron job. Which would be O(TB) each week.
- It's not easily testable with a hardcoded URL. We could include the URL in the CheckRequest struct or as a global var, but then the existing `OssFuzzRepo` is unused. Removing it would break other code that depends on it, and there is external code that depends on it.

In my opinion the `OSSFuzzClient` approach results in minimal changes within the code base and outside of it. What are your concerns with the approach?

azeemshaikh38:
LGTM. For context, I'm hoping to reduce the number of dependencies we inject into `RunScorecards` fn. These were added for improving cron performance, but its causing spaghetti dependencies in the code. If keeping the structure is simpler in terms of code change, let's stick with that for now.

azeemshaikh38 (approved):

Issue #1709, opened 2022-03-04
Feature: use `go-git` 


azeemshaikh38:
Proposing that we use [`go-git`](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.4.2#section-readme) for powering `RepoClient` APIs that require git-like features. Some advantages I see of using this:

- A common implementation for `ListFiles`, `GetFileContent`, `ListCommits` and `Search` that can be used across any Git-like system (GitHub, GitLab or even local git repos).
- Enables us to add support for things like tags (e.g run scorecard on `--commit=v4.1.0`)
- Might provide better performance since we won't be uncompressing a tarball and some added API token efficiency.

Considering to start work on this (not a top priority for me, will work on it on the side). Just wanted to run it by folks to make sure no one has concerns. @laurentsimon @naveensrinivasan @justaugustus

naveensrinivasan:
👍

justaugustus:
@azeemshaikh38 -- What would you say is the primary motivation for `go-git` vs `go-github`?

Is it:
> A common implementation ... that can be used across any Git-like system (GitHub, GitLab or even local git repos).

Are you thinking something like this?

```go
// git/clients.go
type Git interface{
	ListFiles()
	GetFileContent()
	ListCommits()
	Search()
}

// git/gitlab.go
type Gitlab struct { ... }

// git/local.go
type Local struct { ... }

// git/github.go
type Github struct {
	http.RoundTripper
}
func NewGithub(rt *http.RoundTripper) *Github {
	return (go-github).github.NewClient(&http.Client{Transport: rt})
}

func (gh *Github) ListFiles() error {
	// go-github implementation of ListFiles()
}
```

---

Suggested in https://github.com/ossf/scorecard/issues/1491#issuecomment-1018192893: 

> @naveensrinivasan -- my team in Kubernetes Release Engineering maintains a GitHub package, which is pretty well-tested and would allow you to offload this.
> 
> Could you look around and see if some of this fits your needs?
> 
> https://github.com/kubernetes-sigs/release-sdk/blob/main/github/github.go

azeemshaikh38:
> What would you say is the primary motivation for `go-git` vs `go-github`?

Not really `go-git` vs. `go-github`, but rather using `go-git` vs. our custom logic for handling git data. Something along these lines: 

```
// clients/git/git.go
type Git struct {
  CloneRemote(urlToRemote string)
  OpenFrom(pathToLocalDir string)
  ListFiles()
[...]

Issue #2016, opened 2022-07-01
BUG: invalid number of commits reported


laurentsimon:
Running scorecard on https://github.com/danielaparker/jsoncons/commits/master reports
```
"Info: all commits (3) are checked with a SAST tool
```

Thera are a lot more commits in the repo, though. I would hav expected Scorecard to look at 30 commits (at least that's how it used to work)

laurentsimon:
I'm running at HEAD. This seems to be a problem on most repos I'm testing

spencerschrock:
Scorecard still looks at 30 commits (or whatever `commitsToAnalyze` is set to). The number being reported refers to the number of commits in the 30 pulled which have an associated PR. If there is no associated PR, the commit is skipped which leads to the undercounting.

```go
if pr.MergedAt.IsZero() {
	continue
}
totalMerged++
```

The exact numbers depend on the repo when run. When I run against the repo you mentioned today, 30 commits are analyzed and 14 don't have an associated PR, so I get the following:
> all commits (16) are checked with a SAST tool

Looking back in time to when you made the issue, there were a lot of commits which didn't have associated PR's:
[one](https://github.com/danielaparker/jsoncons/commit/b00070885c58cf595b3f8581274b9ff82ee669e9)
[two](https://github.com/danielaparker/jsoncons/commit/40757fa5dbebcacc79dcfb96bf9485180f13242a)
 etc ...
 

I think clarifying expected behavior is good here. Should we expect Scorecard to look at 30 commits overall, or 30 commits which get included in the analyses?

laurentsimon:
30 commits overall is fine, and that's what we've been doing.
I think I got thrown off by the message. It should say `X out of Y commits are checked with a SAST tool`

I recall there were some problems with graphQL where we may be looking at fewer (than 30 ) commits if there are not enough commits in the lest 3 (?) months or so.

@azeemsgoogle do you remember the exact problem and whether it has been resolved?

azeemshaikh38:
IIUC, the problem here is that some repos do not squash when they merge a PR. So when a PR with 5 commits gets merged, each of those commits are returned in `ListCommits`. But the number of PRs associated with these commits is just 1. So checks like SAST, CI-Tests which care about PRs rather than commits seem off.

[...]

PR #2720, opened 2023-03-04
🌱 Initial implementation of go-git client

azeemshaikh38:

Initial implementation of `RepoClient` APIs using `go-git`. Note that this is not an implementation of RepoClient interface, but rather a handler to replace tarballHandler. Part of #1709.

the `release-note`

```release-note
NONE
```


clients/git/coverage.out
```diff
@@ -0,0 +1,51 @@
+mode: set
```
naveensrinivasan, line 1: Do we need this file?

azeemshaikh38: Removed

clients/git/client.go
```diff
@@ -0,0 +1,208 @@
...
+	// path/filename
+	var pathExpr string
+	switch {
+	case request.Path != "" && request.Filename != "":
+		pathExpr = filepath.Join(fmt.Sprintf("^%s", request.Path),
+			fmt.Sprintf(".*%s", request.Filename))
+	case request.Path != "":
+		pathExpr = fmt.Sprintf("^%s", request.Path)
+	case request.Filename != "":
+		pathExpr = filepath.Join(".*", request.Filename)
```
spencerschrock, line 164: similar to path searching at the start, should there be a `$` for the filename search (here and a few lines above)?

clients/git/client.go
```diff
@@ -0,0 +1,208 @@
...
+		return clients.SearchResponse{}, fmt.Errorf("git.Grep: %w", err)
+	}
+
+	ret := clients.SearchResponse{}
+	for _, grepResult := range grepResults {
+		ret.Results = append(ret.Results, clients.SearchResult{
+			Path: grepResult.FileName,
+		})
+	}
+	ret.Hits = len(grepResults)
```
spencerschrock, line 186: if yes to the comment above, this would need changed

clients/git/client.go
```diff
@@ -0,0 +1,208 @@
...
+	c.commitDepth = commitDepth
+	tempDir, err := os.MkdirTemp("", repoDir)
+	if err != nil {
+		return fmt.Errorf("os.MkdirTemp: %w", err)
+	}
+
+	// git clone
+	const filePrefix = "file://"
+	if strings.HasPrefix(uri, filePrefix) {
+		if err := cp.Copy(strings.TrimPrefix(uri, filePrefix), tempDir); err != nil {
```
spencerschrock, line 68: if this is being called on a temp directory already (from another scorecard check perhaps), does it make sense to have a way to avoid this extra copy?

azeemshaikh38: The expectation here is that this will be called when `--local` CLI flag is used by the client (never from another Scorecard check). In that case, its easiest to maintain state of the client's workspace and instead create a copy and delete it when done.

clients/git/client.go
```diff
@@ -0,0 +1,208 @@
...
+			return fmt.Errorf("cp.Copy: %w", err)
+		}
+		c.gitRepo, err = git.PlainOpen(tempDir)
```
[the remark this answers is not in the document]
azeemshaikh38: There isn't a concept of default branch on a local repo AFAIK? Also, the usecase for running Scorecard on a local repo is to test out local changes made by the client. Changing branch would defeat the purpose.

clients/git/client.go
azeemshaikh38: Done.

clients/git/client.go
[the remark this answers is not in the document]
spencerschrock: I was confusing it with `origin/HEAD`. But it is indeed something associated with a remote. 
https://git-scm.com/docs/git-remote#Documentation/git-remote.txt-emset-headem

clients/git/client.go
spencerschrock: I think I was imagining it could also be used to augment the `githubrepo` client. Although that downloads via the tarball, so the `.git` folder wouldn't be there.

clients/git/client.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// Grep
+	grepResults, err := c.worktree.Grep(grepOpts)
+	if err != nil {
+		return clients.SearchResponse{}, fmt.Errorf("git.Grep: %w", err)
+	}
+
+	ret := clients.SearchResponse{}
+	for _, grepResult := range grepResults {
+		ret.Results = append(ret.Results, clients.SearchResult{
```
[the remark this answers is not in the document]
azeemshaikh38: I would be surprised if it is across trees since the API is `Worktree.Grep()`. Also, I did some basic testing and it seems to return results from the single commit SHA that was checked out.

clients/git/client.go
azeemshaikh38: See above.

clients/git/client.go
spencerschrock: Resolved after discussion of which role the git client plays. Not a top-level RepoClient implementation, but rather to potentially replace `tarballHandler`

clients/git/client.go
```diff
@@ -0,0 +1,208 @@
...
+				break
+			}
+
+			if commit == nil {
+				// Not sure in what case a nil commit is returned. Fail explicitly.
+				c.errListCommits = fmt.Errorf("%w", errNilCommitFound)
+				return
+			}
+
+			c.commits = append(c.commits, clients.Commit{
```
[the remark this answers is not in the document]
azeemshaikh38: I prefer dynamic allocation here since there is a possibility that there are not enough commits for `commitDepth`. In which case `len([]commit) == commitDepth` with lot of empty slots. This then puts the burden on caller code to handle this.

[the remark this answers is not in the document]
spencerschrock: `len([]commit) == commitDepth` isn't true if you use explicity set the length to 0: `make([]commit, 0, commitDepth)`

In this case: `cap([]commit) == commitDepth`.

There is a chance to over allocate space, but the caller code won't need to handle empty slots.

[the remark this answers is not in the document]
azeemshaikh38: Do you mean something like this?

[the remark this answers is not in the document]
spencerschrock: Yeah, thanks

clients/git/gitrepo_suite_test.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021 OpenSSF Scorecard Authors
```
spencerschrock, line 1: optional: year

clients/git/client.go
```diff
@@ -0,0 +1,209 @@
...
+	if strings.HasPrefix(uri, filePrefix) {
+		if err := cp.Copy(strings.TrimPrefix(uri, filePrefix), tempDir); err != nil {
+			return fmt.Errorf("cp.Copy: %w", err)
+		}
+		c.gitRepo, err = git.PlainOpen(tempDir)
+		if err != nil {
+			return fmt.Errorf("git.PlainOpen: %w", err)
+		}
+	} else {
+		c.gitRepo, err = git.PlainClone(tempDir, false /*isBare*/, &git.CloneOptions{
```
spencerschrock, line 76: I see comments that the e2e tests take a while, does adding a clone depth help?

https://pkg.go.dev/github.com/go-git/go-git/v5#CloneOptions

clients/git/client.go
```diff
@@ -0,0 +1,209 @@
...
+				c.errListCommits = fmt.Errorf("%w", errNilCommitFound)
+				return
+			}
+
+			c.commits = append(c.commits, clients.Commit{
+				SHA:           commit.Hash.String(),
+				Message:       commit.Message,
+				CommittedDate: commit.Committer.When,
+				Committer: clients.User{
+					Login: commit.Committer.Email,
```
spencerschrock, line 135: probably outside the scope of this PR, but changes may be needed here.`DependencyUpdateTool` looks for `Author: "dependabot[bot]"`

clients/git/client.go
```diff
@@ -0,0 +1,209 @@
...
+			})
+		}
+	})
+	return c.commits, c.errListCommits
+}
+
+func (c *Client) Search(request clients.SearchRequest) (clients.SearchResponse, error) {
+	// Pattern
+	if request.Query == "" {
+		return clients.SearchResponse{}, fmt.Errorf("%w", errEmptyQuery)
```
spencerschrock, line 146: don't think we need a wrap here

clients/git/client.go
```diff
@@ -0,0 +1,209 @@
+// Copyright 2021 OpenSSF Scorecard Authors
```
spencerschrock, line 1: optional: year

clients/git/e2e_test.go
```diff
@@ -0,0 +1,155 @@
+// Copyright 2021 OpenSSF Scorecard Authors
```
spencerschrock, line 1: optional: year

spencerschrock (approved):
Looks good for initial implementation. Some small nits

clients/git/client.go
azeemshaikh38: Well problem with clone depth here is that we won't be able to support `--commit` option.

clients/git/client.go
azeemshaikh38: Yeah, some checks may still not be able to function with `go-git`. We probably need to experiment a bit with `Name` and `Email` fields [here](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.6.0/plumbing/object#Signature) to see if we can find some reasonable approximations.

clients/git/client.go
spencerschrock: True. We could always optimize for the common case though and add a clone depth if `commitSHA == clients.HeadSHA`

clients/git/gitrepo_suite_test.go
azeemshaikh38: I'll skip this if that's ok.

clients/git/client.go
azeemshaikh38: I would prefer to leave this as is for now without over optimizing for performance. One reason being, I want to try and test out the proposal [here](https://github.com/ossf/scorecard/issues/2016#issuecomment-1468381435) which assumes Scorecard has access to all commit history. If clone depth is really a performance blocker, we can reconsider. Let me know if that works.

spencerschrock: SGTM

PR #2721, opened 2023-03-06
🌱 Setup cron for running as GitHub App

spencerschrock:

cron update

The cron used personal access tokens

* Dropped the GITHUB AUTH server to 0 replicas since it doesn't support the GitHub App tokens
* Added logic for `retry-after` headers and some metrics when/if we see it.
* Disabled 4 checks which don't work well at the moment
  * Search based checks hit secondary rate limits frequently with the GitHub App tokens
    * `Dependency-Update-Tool`
    * `Fuzzing`, which should be re-enabled again soon now that #2719 is merged
    * `SAST`
  * `Vulnerabilities` due to an inefficiency in osv-scanner that has been fixed upstream but needs a new release cut
* Runtime errors are ignored in the release worker due to known errors with the `Branch-Protection` check while using a GitHub App token

 [] Tests for the changes have been added (for bug fixes/features)

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/clients/githubrepo/roundtripper/transport.go
+++ b/clients/githubrepo/roundtripper/transport.go
@@ -65,10 +65,5 @@ func (gt *githubTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 		stats.Record(ctx, githubstats.RemainingTokens.M(int64(remaining)))
 	}
 
-	retryAfter, err := strconv.Atoi(resp.Header.Get("Retry-After"))
-	if err == nil {
-		stats.Record(r.Context(), githubstats.RetryAfter.M(int64(retryAfter)))
-	}
-
 	return resp, nil
 }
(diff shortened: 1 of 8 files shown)


laurentsimon (approved):

clients/githubrepo/roundtripper/transport.go
```diff
@@ -64,5 +64,11 @@ func (gt *githubTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 	if err == nil {
 		stats.Record(ctx, githubstats.RemainingTokens.M(int64(remaining)))
 	}
+
+	retryAfter, err := strconv.Atoi(resp.Header.Get("Retry-After"))
+	if err == nil {
+		stats.Record(r.Context(), githubstats.RetryAfter.M(int64(retryAfter)))
+	}
+
```
azeemshaikh38, line 72: Shouldn't this be in the `rate_limit.go` file where we handle `Retry-After`? Otherwise we'll always record `0` since `Retry-After` would be handled already.

azeemshaikh38 (approved):
LGTM with comment/question.

clients/githubrepo/roundtripper/transport.go
spencerschrock: Good catch. I haven't observed any `retry-after` headers yet, so had no indication this was working/broken.

Issue #2228, opened 2022-09-02
BUG: `pip install -e .` dinged for not using hashes, but it can't


nedbat:
**Describe the bug**

My project (https://deps.dev/pypi/coverage) gets a warning about pinned dependencies:

```
Warn: pipCommand not pinned by hash: .github/workflows/coverage.yml:114
```

[The line](https://github.com/nedbat/coveragepy/blob/master/.github/workflows/coverage.yml#L114) in question is: 

```
python -m pip install -e .
```
I think scorecard is looking for the `--require-hashes` option, but it's not allowed with `-e`.  This is [what happens when I try](https://github.com/nedbat/coveragepy/runs/7996337904?check_suite_focus=true#step:4:34):

```
+ python -m pip install --require-hashes -e .
Obtaining file:///home/runner/work/coveragepy/coveragepy
ERROR: The editable requirement file:///home/runner/work/coveragepy/coveragepy cannot be installed when requiring hashes, because there is no single file to hash.
```

laurentsimon:
Thanks for the report.  So I think the fix is for scorecard to verify that no `-e local/path` is used in the command. Is that correct?

`-e` also seems to accept a remote URL as input, which is harder to verify.

Do you think ignoring `-e local/path` is enough?

nedbat:
Definitely ignoring `-e local/path` will help.  Why not also allow a URL?

laurentsimon:
I don't know how the URL works. Does pip fetch the URL and search for a requirement.txt? Or does it simply take the URL as being the source code? Is the source code guaranteed to be immutable (which is the property we're looking for)?

nedbat:
The local file path or the URL are the same: they are a place to get an installable directory of files.

A difference is that the URL could include a SHA that would make it immutable.  So perhaps the rule should be a local file path, or a URL with a SHA?

nedbat:
This seems to be the best docs: https://pip.pypa.io/en/stable/topics/vcs-support/

Some syntaxes to support (shell quoting may or may not be needed):
```
pip install "MyProject @ git+https://git.example.com/MyProject.git@da39a3ee5e6b4b0d3255bfef95601890afd80709"
pip install "MyProject[extra1,extra2]@git+https://git.example.com/MyProject.git@da39a3ee5e6b4b0d3255bfef95601890afd80709"
pip install git+https://git.example.com/MyProject.git@da39a3ee5e6b4b0d3255bfef95601890afd80709#egg=MyProject
[...]

PR #2731, opened 2023-03-08
:bug: Handle editable pip installs

gabibguti:

Handle editable pip installs in Pinned-Dependencies check.

##### Description

When performing `pip install -e <pkg>`, the install can be considered secure if the package and it's dependencies are hash-pinned. The package can come from a local or remote source (VCS install). If the package is local, it is considered secure. If the package is remote, it needs to pinned by commit hash to be secure. We have no way to identify if the dependencies installed from requirements are hash-pinned. The flag `--require-hashes` does not work with `-e` flag. To guarantee we are not installing malicious dependencies, we can use `--no-deps` flag to not install the dependencies. That way the install can be considered secure.

##### Examples
| | |
| -- | -- |
| secure | pip install --no-deps -e git+https://github.com/username/repo.git@0123456789abcdef0123456789abcdef01234567#egg=package" |
| secure | pip install --no-deps -e . |
| not secure | pip install -e git+https://github.com/username/repo.git@0123456789abcdef0123456789abcdef01234567#egg=package" |
| not secure | pip install --no-deps -e git+https://github.com/username/repo.git#egg=package" |

##### References
- pip install syntax: https://pip.pypa.io/en/stable/cli/pip_install/#pip-install
- secure pip install: https://pip.pypa.io/en/stable/topics/secure-installs/#do-not-use-setuptools-directly
- what is editable pip install: https://setuptools.pypa.io/en/latest/userguide/development_mode.html
- pip VCS install syntax: https://pip.pypa.io/en/latest/topics/vcs-support/#vcs-support
- https://github.com/pypa/pip/issues/4995

Currently, we don't handle editable pip installs. We suggest using `--require-hashes` as remediation but it's not possible.

Closes https://github.com/ossf/scorecard/issues/2228

the `release-note`

```release-note
Accept pip installs with `-e` flag along with `--no-deps` flag in Pinned-Dependencies check. Editable pip installs will be accepted if coming from a local source or a remote Git source pinned by hash, as long as it's not installing dependencies.
```

### DIFF
--- a/checks/raw/shell_download_validate.go
+++ b/checks/raw/shell_download_validate.go
@@ -495,13 +495,10 @@ func isPinnedEditableSource(pkgSource string) bool {
 	}
 	regexGitSource := regexp.MustCompile(`^git(\+(https?|ssh|git))?\:\/\/.*(.git)?@[a-fA-F0-9]{40}(#egg=.*)?$`)
 	// Is VCS install from Git and it's pinned
-	if regexGitSource.MatchString(pkgSource) {
-		return true
-	}
+	return regexGitSource.MatchString(pkgSource)
 	// Disclaimer: We are not handling if Subversion (svn),
 	// Mercurial (hg) or Bazaar (bzr) remote sources are pinned
 	// because they are not common on GitHub repos
-	return false
 }
 
 func isUnpinnedPipInstall(cmd []string) bool {
--- a/checks/raw/shell_download_validate.go
+++ b/checks/raw/shell_download_validate.go
@@ -493,8 +493,9 @@ func isPinnedEditableSource(pkgSource string) bool {
 	if !regexRemoteSource.MatchString(pkgSource) {
 		return true
 	}
-	regexGitSource := regexp.MustCompile(`^git(\+(https?|ssh|git))?\:\/\/.*(.git)?@[a-fA-F0-9]{40}(#egg=.*)?$`)
 	// Is VCS install from Git and it's pinned
+	// https://pip.pypa.io/en/latest/topics/vcs-support/#vcs-support
+	regexGitSource := regexp.MustCompile(`^git(\+(https?|ssh|git))?\:\/\/.*(.git)?@[a-fA-F0-9]{40}(#egg=.*)?$`)
 	return regexGitSource.MatchString(pkgSource)
 	// Disclaimer: We are not handling if Subversion (svn),
 	// Mercurial (hg) or Bazaar (bzr) remote sources are pinned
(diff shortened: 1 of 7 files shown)


di (approved):
LGTM! Nice job!

checks/raw/shell_download_validate.go
```diff
@@ -507,6 +524,22 @@ func isUnpinnedPipInstall(cmd []string) bool {
 			break
 		}
 
+		if strings.EqualFold(cmd[i], "--no-deps") {
```
laurentsimon, line 527: do you mind adding as comment some of the PR description, including the links to the relevant doc?

laurentsimon (approved):
Thank you!

checks/raw/shell_download_validate.go
gabibguti: Done. Can you take a look?

checks/raw/shell_download_validate.go
```diff
@@ -525,6 +564,14 @@ func isUnpinnedPipInstall(cmd []string) bool {
 		hasAdditionalArgs = true
 	}
 
+	// If is editable install, it's secure if package is from local source
+	// or from remote (VCS install) pinned by hash, and if dependencies are
+	// not installed.
+	// Example: `pip install --no-deps -e git+https://git.repo/some_pkg.git@da39a3ee5e6b4b0d3255bfef95601890afd80709`
+	if isEditableInstall {
+		return !hasNoDeps || !isPinnedEditableInstall
```
laurentsimon, line 572: hoe do the new options work with existing `hasAdditionalArgs`, `hasWheels`, etc? Can we return safely here or do we need to return after checking the other options first (code below)?

checks/raw/shell_download_validate.go
laurentsimon: LG, thanks

checks/raw/shell_download_validate.go
gabibguti: I will double check the test cases to see if we need to rearrange the return line.

gabibguti: The cases with `--require-hashes` and `<pkg>.whl` are working fine. Since `--require-hashes` and `-e` is not possible by pip install syntax, I'm not testing this case. Same thing for `-e` with `<pkg>.whl`.
Only case I caught which is NOT working is support installing 2 packages with editable `-e` install, should we cover that?

laurentsimon: Let's cover it, yes. Otherwise someone may tell us we have false positives.
Can you also add comments about the reasoning you explain in your comment? That will also help document the code.

Thanks again for the hard work figuring out all these subtle differences.

gabibguti: @laurentsimon Done. Please check if you agree.

laurentsimon: lgtm. I think the unit tests speak for themselves. If you've added some of these corner cases, we're good to go.

laurentsimon (approved):

laurentsimon:
Please rebase and we'll be able to merge.

PR #2734, opened 2023-03-08
✨  Add simple accessor to avoid atomics in a single token scenario.

spencerschrock:

feature

When provided a single token, the round robin accessor still manages state and uses expensive atomic operations and resource contention.

A new simple accessor is provided for when only one token is provided.
This speeds up running scorecard

before:
```
time ./scorecard --repo ossf/scorecard --format json
real	0m38.171s
user	1m15.050s
sys	0m1.269s
```

after
```
real	0m23.237s
user	0m2.837s
sys	0m1.145s
```

Aside from the expensive atomic operations, there is a design decision here which I don't have context for:

https://github.com/ossf/scorecard/blob/5f13a66c271ed810560c1fd96a04630f67afed2e/clients/githubrepo/roundtripper/transport.go#L43-L45

Based on this code, every HTTP request will rotate through a different token. In the round robin case with a single token, this meant rotating through the same token, but there was some resource contention.  Each graphQL call and REST call is essentially serialized. 

Is this an intentional decision to limit concurrent usages of the same token (essentially one at a time with a 30s timeout)?
If so, this PR probably should be closed instead of merged.

@azeemshaikh38 FYI

the `release-note`

```release-note
NONE
```

### DIFF
--- a/clients/githubrepo/roundtripper/tokens/accessor.go
+++ b/clients/githubrepo/roundtripper/tokens/accessor.go
@@ -43,7 +43,11 @@ func readGitHubTokens() (string, bool) {
 // MakeTokenAccessor is a factory function of TokenAccessor.
 func MakeTokenAccessor() TokenAccessor {
 	if value, exists := readGitHubTokens(); exists {
-		return makeRoundRobinAccessor(strings.Split(value, ","))
+		tokens := strings.Split(value, ",")
+		if len(tokens) == 1 {
+			return makeSimpleAccessor(tokens[0])
+		}
+		return makeRoundRobinAccessor(tokens)
 	}
 	if value, exists := os.LookupEnv(githubAuthServer); exists {
 		return makeRPCAccessor(value)
--- a/clients/githubrepo/roundtripper/tokens/simple.go
+++ b/clients/githubrepo/roundtripper/tokens/simple.go
@@ -1,0 +1,33 @@
+// Copyright 2023 OpenSSF Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


azeemshaikh38:
Great find :) Yes, the code was added to reduce concurrent usage of tokens. Probably explains why we're not seeing issues with a single GHApp token. Could we run a nightly cron job with a locally built binary of this change? I'm curious to see if removing this constraint would cause the GHApp token to hit rate limits more frequently.

spencerschrock:
> Yes, the code was added to reduce concurrent usage of tokens. Probably explains why we're not seeing issues with a single GHApp token

The GH App doesn't use this token accessor. It makes uses the default `http.DefaultTransport`:
https://github.com/ossf/scorecard/blob/b5254fea7c6edef0674c7ea25fdc0f627438aa0b/clients/githubrepo/roundtripper/roundtripper.go#L47-L57

azeemshaikh38:
> The GH App doesn't use this token accessor. It uses the default `http.DefaultTransport`:

I see. In that case, LGTM.

azeemshaikh38 (approved):

spencerschrock:
Quick note that this could be an issue for anyone running Scorecard from the CLI.

Something like this
```
for i in {1..15}; do ./scorecard --repo ossf/scorecard --format json; done
```

results in Retry-After headers after ~5 iterations.
```
INFO[0006] listCheckRunsForRef cache miss: ossf/scorecard:5ea8cc2f364284e59571378dd7347accdb11793b 
INFO[0006] listCheckRunsForRef cache miss: ossf/scorecard:5ea8cc2f364284e59571378dd7347accdb11793b 
INFO[0006] listCheckRunsForRef cache miss: ossf/scorecard:e1c93eb234f230b7b5cbb5aa6fb89a925dd4aaa2 
INFO[0006] Retry-After header set. Waiting 1m0s to retry... 
INFO[0007] listCheckRunsForRef cache miss: ossf/scorecard:5b7acc117e4849b9bb4c1def16996bfcc390169d 
INFO[0007] Retry-After header set. Waiting 1m0s to retry... 
INFO[0007] Retry-After header set. Waiting 1m0s to retry...
```

Where as the existing single token round robin accessor doesn't run into these headers.

PR #2764, opened 2023-03-16
✨ GitLab: Code Review check

raghavkaul:
Adds GitLab support for the Code-Review check.

Code reviews in GitLab, including implicit reviews and comments, are now detected by Scorecard. This check is still slow because of the number of API calls per commit.

```release-note
GitLab support for Code-Review check
```


clients/gitlabrepo/commits.go
```diff
@@ -92,13 +68,6 @@ func (handler *commitsHandler) setup() error {
 						mergeRequest = mergeRequests[i]
```
azeemshaikh38, line 68: This line will seg fault. `mergeRequest` is guaranteed to be `nil` when it reaches this line for the first time.

azeemshaikh38, line 68: Remove unecessary `if len(mergeRequests) > 0`. Can simply do:

```
var mergeRequest *gitlab.MergeRequest
for i := range mergeRequests {
  ...
}
```

clients/gitlabrepo/commits.go
```diff
@@ -107,15 +76,42 @@ func (handler *commitsHandler) setup() error {
...
+				if reviewer.State == "unreviewed" {
+					continue
+				}
+				reviews[reviewer.ID] = clients.Review{
+					Author: &clients.User{Login: reviewer.Username, ID: int64(reviewer.ID)},
+					State:  "COMMENTED",
+				}
+			}
+
+			approvals, _, err := handler.glClient.MergeRequests.GetMergeRequestApprovals(
```
azeemshaikh38, line 95: Can we avoid this extra API call by using GitLab's GraphQL API - https://docs.gitlab.com/ee/api/graphql/reference/#mergerequest ?

azeemshaikh38 (commented):
Overall looks good. Left a comment about optimizing the API calls a bit more with GitLab graphQL.

clients/gitlabrepo/commits.go
raghavkaul: I believe it's set on L62: `mergeRequest = mergeRequests[0]`. Or do you mean that's always nil?

azeemshaikh38: The first time this line is reached when all `mergeRequests[0:i-1]` are `nil`. So `mergeRequests[0] == mergeRequest == nil`.

azeemshaikh38 (approved):

Issue #1578, opened 2022-02-01
Add support for NuGet


JonDouglas:
The scorecard project currently only supports npm, golang, and pip as far as I could tell. I'm a PM on the NuGet team at Microsoft and would love to help contribute adding support for NuGet in this tool or providing the right guidance to implement support for NuGet. This closely aligns with a proposal I had last year and would love to experiment with this scorecard in .NET:

https://github.com/dotnet/designs/pull/216

Please feel free to reach out to us over at NuGet/Home on GitHub or in this issue. Any steps on how to best contribute adding this support would be greatly appreciated!

naveensrinivasan:
It would be great to add support for NuGet. 

Here is an example of how we are looking for `go` 
https://github.com/ossf/scorecard/blob/696553be2dd358a1b177cf28a1c8e321ee9e779f/checks/shell_download_validate.go#L591

PR's are welcome!

laurentsimon:
@JonDouglas that's great, thanks for reaching out!

@naveensrinivasan pointed you to the file where we parse commands and look for unpinned go, pip and npm commands.
This is the place to add it.

Shameless plug: as part of the OSSF Best Practice WG, we have a work stream on package manager's best security practices. We've started with npm and pip. If you're interested in collaborating - or even lead! - a similar effort for NuGet, please let me know.

laurentsimon:
an additional place to add support for NuGet is in the Packaging check https://github.com/ossf/scorecard/blob/main/checks/packaging.go, which tries to infer if the project published a package. We typically look for known GitHub actions and commands to detect this.

azeemshaikh38:
Also feel free to add an agenda item to the upcoming meet in this [doc](https://docs.google.com/document/d/1dB2U7_qZpNW96vtuoG7ShmgKXzIg6R5XT5Tc-0yz6kE/edit#heading=h.4k8ml0qkh7tl). You'll need to join ossf-scorecard-announce@googlegroups.com for being able to modify the doc.

laurentsimon:
Looking forward to meeting you @[JonDouglas](https://github.com/JonDouglas)!
Thanks @azeemshaikh38 for the prompt info!

JonDouglas:
Hi all,

I had originally planned to join the meeting later this afternoon, but unfortunately I have a conflict I cannot clear this month. I'll be looking to join the next one on the 10th instead.

laurentsimon:
Thanks for letting us know. See you in 2 weeks then!

[...]

PR #2779, opened 2023-03-22
✨ add support for Nuget ad-hoc commands (add/install) in Pinned Dependency checks 

balteravishay:

Add support for [Nuget](https://www.nuget.org/) ad-hoc commands (add/install) in pinned dependency checks.

- [v] PR title follows the guidelines defined in our [pull request documentation](https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process)

Scorecard does not detect unpinned dependencies when using nuget and dotnet clis to add ad-hoc libraries.

Scorecard detects and warns when using dotnet and nuget cli to add ad-hoc libraries packages without declaring their specific version.

- [v] Tests for the changes have been added (for bug fixes/features)

This PR relates to but does not fix issue #1578

the `release-note`

```release-note
Scorecard checks for unpinned dependencies that are retrieved ad-hoc using nuget and dotnet CLIs ("nuget install" and "dotnet add").
```


checks/raw/shell_download_validate.go
```diff
@@ -696,6 +697,83 @@ func isChocoUnpinnedDownload(cmd []string) bool {
...
+	}
+
+	// Search for install commands.
+	if !strings.EqualFold(cmd[1], "install") {
+		return false
+	}
+
+	// package.config schema has required version field
+	// https://learn.microsoft.com/en-us/nuget/reference/packages-config#schema
+	if strings.HasSuffix(cmd[2], "packages.config") {
```
laurentsimon, line 713: so here the assumption is that users use a file named `<any>packages.config`, correct? Maybe add a comment about the assumption.

laurentsimon, line 713: does this file ensure that packages are declared immutably? In these examples https://learn.microsoft.com/en-us/nuget/reference/packages-config#schema, it seems like they are declared using a version. Are versions immutable in nuget?

checks/raw/shell_download_validate.go
```diff
@@ -696,6 +697,83 @@ func isChocoUnpinnedDownload(cmd []string) bool {
...
+	for i := 2; i < len(cmd); i++ {
+		// look for version flag
+		if strings.EqualFold(cmd[i], "-Version") {
+			hasVersion = true
+			break
+		}
+	}
+
+	if hasVersion {
+		return !hasVersion
```
laurentsimon, line 727: nit: `return false` is easier to read

checks/raw/shell_download_validate.go
```diff
@@ -696,6 +697,83 @@ func isChocoUnpinnedDownload(cmd []string) bool {
...
+	}
+
+	return true
+}
+
+func isUnpinnedDotNetCliInstall(cmd []string) bool {
+	// Search for dotnet commands.
+	if !isBinaryName("dotnet", cmd[0]) {
```
laurentsimon, line 735: nit: add a note that length of `cmd` is verified by the caller. I don't think we've added these commands in the rest of the code, we / I should have :)

checks/raw/shell_download_validate.go
```diff
@@ -696,6 +697,83 @@ func isChocoUnpinnedDownload(cmd []string) bool {
...
+		// look for version flag
+		// https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-add-package
+		if strings.EqualFold(cmd[i], "-v") || strings.EqualFold(cmd[i], "--version") {
+			hasVersion = true
+			break
+		}
+	}
+
+	if hasVersion {
+		return !hasVersion
```
laurentsimon, line 755: nit: `return false` is easier to read

laurentsimon (commented):
Thanks for the PR. I left comments about immutability of versions in nuget ecosystem because I'm unfamiliar with it.

checks/raw/shell_download_validate.go
balteravishay: added, thanks

balteravishay: yes, checked with @JonDouglas a Nuget PM that Nuget follows semantic versioning 2.0.0, and versions cannot be changed. documented in the code. 
https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#semantic-versioning-200

checks/raw/shell_download_validate.go
balteravishay: fixed

checks/raw/shell_download_validate.go
balteravishay: changed where the check is performed

checks/raw/shell_download_validate.go
balteravishay: fixed

checks/raw/shell_download_validate.go
laurentsimon: so the NuGet registry enforces immutability on the content of a package? Say, a package maintainer publishes benign package `hello-v1.2.3`. Can they later push the package / version with a different hash (this time containing malicious content)?

Several ecosystems have a concept of "lock file", which "pins" packages to the cryptographic hash of the package content. In the case of NuGet, we trust the registry to enforce immutability?

JonDouglas: The central registry (NuGet.org) enforces immutability on package metadata such as the name, version, dependencies, etc. Once a package is published, its metadata cannot be changed. The package contents are generally seen as immutable once packaged and published. But like other ecosystems, it is not impossible for the contents to be overwritten if following bad practices on a private feed they control. [The central registry has a policy on deletion and unlisting for this purpose.](https://learn.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages)

In the case you outline, the maintainer would need to increment a new package version to then push with new contents. Usually through [semantic versioning](https://semver.org/).

We also support lock files, dependency pinning, and much more for cases where you may require more determinism.

laurentsimon: Got it, thanks for the information. Do you know if we should encourage lock files for applications? We do this for other ecosystems. I don't know what the best practices are for NuGet. But iiuc, recommending users to use lock files would provide similar guarantees while reducing trust on the registry. In terms of usability for users, is it the same amount of work?

laurentsimon: In OpenSSF we encourage folks to use lock files. But I want to be sure this is not going against NuGet's best practices.
Do you think we should do the same for NuGet, ie lock files? If it's as easy for users to use, we could recommend it instead of version pinning.

Wdut?

balteravishay: is this meant as a documentation recommendation or in the scorecard checks? if in code, could you please link to where that is implemented fro other ecosystems? @laurentsimon

laurentsimon: sure, you're actually in the right file.
pip: https://github.com/ossf/scorecard/blob/main/checks/raw/shell_download_validate.go#L549. `--require-hashes` means there must be a lock file that is complete.
npm: https://github.com/ossf/scorecard/blob/main/checks/raw/shell_download_validate.go#L411-L415. `npm ci` command fails if there is no lock file or if it's incomplete.

We also check for similar immutable declaration for Dockerfile's BASE and GitHub Action `@<ref>` https://github.com/ossf/scorecard/blob/main/checks/raw/pinned_dependencies.go#L37-L42

laurentsimon: `RestorePackagesWithLockFile` alone does not seem enough, iiuc. `If NuGet detects a change in the defined dependencies as enumerated in the project file(s), it re-evaluates the package graph and updates the lock file`

Looks like there are various ways then
1. check the relevant file for `RestoreLockedMode` property being set. This may be dynamically set, so we'd need to also check for the propoerty it refers to MSBuild property set (`ContinuousIntegrationBuild ` in the article). On GitHub, there is an env variable `CI=true` which is set, so I suppose we'd verify the property is set with it?
2. for donet, check for `restore --locked-mode`. If another command is used (`install` or `restore` alone), we would flag it
3. for msbuild, same approach as (2) but with `/t:restore /p:RestoreLockedMode=true`
4. I'm not sure about nugget. Do you know if it has similar support?

Since there are many possibilities, let's split the PR to make it easier to merge one-by-one. 

Maybe we could start with (2) and (3) if you agree they are the right approaches to recommend?

Wdut?

checks/raw/shell_download_validate.go
```diff
@@ -696,6 +697,93 @@ func isChocoUnpinnedDownload(cmd []string) bool {
...
 }
 
+func isUnpinnedNugetCliInstall(cmd []string) bool {
+	// looking for command of type nuget install ...
+	if len(cmd) < 2 {
+		return false
+	}
+
+	// Search for nuget commands.
+	if !isBinaryName("nuget", cmd[0]) {
```
laurentsimon, line 707: do we need to check for `nuget.exe` as well?

checks/raw/shell_download_validate.go
laurentsimon: > I'm all in favor for breaking up the issue to a number of prs, thanks for suggesting that! I can open the open points as g.h issues after we lock down the plan.

+1, thanks!

> Should there be a lower score that is not zero for using pinned versions without lock file (I.e. As the current implementation) and the highest score for repos that do use lock file? 

Definitely worth considering. If there's an easy path for upgrade for users, I think it would be OK. We just want to avoid a situation where we ask users to do a bunch of work for version, and the work is not "re-usable" to further improve for hashes.

checks/raw/shell_download_validate.go
nkolev92: it's not uncommon that people rename nuget.exe occassionally as well, usually by appending the version

checks/raw/shell_download_validate.go
nkolev92: NuGet would only automatically discover `projectName.packages.config`

nkolev92: msbuild & dotnet restore are exactly the same, just different commands. 
NuGet.exe also has matching commands for lock files. 

The expectation is that users run `--locked-mode` on their CI when using lock files.

Note that lock files are not really possible for what seems to be referred as pinned dependencies, as for example with nuget.exe install, there's simply no project file in play.

nkolev92 (commented):
A few comments with the limited context that I have. 

BTW, how commonly used phrase is the `pinned dependency` concept? :)

NuGet has a CPM feature that has a similar name that'd confuse some people: https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#transitive-pinning

checks/raw/shell_download_validate.go
nkolev92: Another note is that NuGet's resolution (in PackageReference) explicitly prefers lowest acceptable version, so it's not all that likely that restore today is different from restore 1 month from now even without a lock file, given the policy of nuget.org.

checks/raw/shell_download_validate.go
balteravishay: added thanks!

checks/raw/shell_download_validate.go
laurentsimon: LGTM. So we will keep this PR and tweak the score not be lower than the full score?

balteravishay: yes, i guess that will be more relevant once #2690 is merged and more dependency types can be scored differently

laurentsimon: @gabibguti please take this thread into consideration in your PR https://github.com/ossf/scorecard/pull/2690

checks/raw/shell_download_validate.go
```diff
@@ -696,6 +697,93 @@ func isChocoUnpinnedDownload(cmd []string) bool {
...
+	if !isBinaryName("nuget", cmd[0]) && !isBinaryName("nuget.exe", cmd[0]) {
+		return false
+	}
+
+	// Search for install commands.
+	if !strings.EqualFold(cmd[1], "install") {
+		return false
+	}
+
+	// Asseume installing a project with PackageReference (with versions)
```
laurentsimon, line 716: nit: Assume

laurentsimon (approved):
Thank you

checks/raw/shell_download_validate.go
laurentsimon: made the changes myself.

checks/raw/shell_download_validate.go
laurentsimon: Thanks everyone. Resolving comment.

checks/raw/shell_download_validate.go
laurentsimon: Thanks everyone. Resolving comment.

laurentsimon:
Can you fix the DCO pre-submit? Click the link and it will tell you the commands to run to sign your commits.

PR #2781, opened 2023-03-22
🐛 Add tie breaker when sorting changesets by RevisionID in tests.

spencerschrock:

bug fix

`phabricatorCommitD` and `commitB` both have `RevisionID` of 2.
Both of these commits are used in the `mixed: phabricator + gh` test

The map iteration order isn't deterministic and sorting the slices isn't good enough when the revision IDs are equal.

This is causing the unit-tests to fail intermittently on unrelated PRs. e.g. #2780 
https://github.com/ossf/scorecard/actions/runs/4494760078/jobs/7907638544?pr=2780#step:7:138

review platform is used as a tie breaker when sorting in the test code

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/checks/raw/code_review_test.go
+++ b/checks/raw/code_review_test.go
@@ -93,11 +93,11 @@ func Test_getChangesets(t *testing.T) {
 
 		gerritCommitB = clients.Commit{
 			Message: "first change\nReviewed-on: server.url \nReviewed-by:user-123",
-			SHA:     "abc",
+			SHA:     "abcg",
 		}
 		gerritCommitA = clients.Commit{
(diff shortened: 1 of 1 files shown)


azeemshaikh38 (commented):
The revision IDs being same is intentional. The test is checking whether revision ID collision across different platforms is handled correctly.

Consider updating the comparator we pass to `cmp.Equal`?

spencerschrock:
> The revision IDs being same is intentional. The test is checking whether revision ID collision across different platforms is handled correctly.
> 
> Consider updating the comparator we pass to `cmp.Equal`?

Ack. Added a tie breaker via lexicographical comparison of the review platform.

azeemshaikh38 (approved):

PR #2795, opened 2023-03-27
🐛  Restore upload of existing raw result Big Query data

spencerschrock:

bug fix

The raw result BigQuery transfer has been failing for a while due to schema changes. The [raw results outputted by the cron](
https://github.com/ossf/scorecard/blob/dde9aa1aefb9e26196c03ef7401287ff2fd35b01/cron/internal/format/json_raw_results.go#L101-L115) is kept separate from the [raw results in the rest of Scorecard](https://github.com/ossf/scorecard/blob/dde9aa1aefb9e26196c03ef7401287ff2fd35b01/pkg/json_raw_results.go#L252). This was done intentionally to avoid changes breaking the cron, although the two have diverged quite a bit.

The schema has been updated to include the new fields introduced to the cron version of the struct. Once applied to `openssf:scorecardcron.scorecard-rawdata`, the upload of raw results should resume, although there is further work to be done to get the cron copy to match the scorecard copy.

NONE

This has been tested in the release test version of the table `openssf:scorecardcron.scorecard-rawdata-releasetest` and the transfers have resumed successfully.

the `release-note`

```release-note
The schema for `openssf:scorecardcron.scorecard-rawdata` has been fixed to upload the new Code-Review Changeset data. As a reminder, the raw data table is experimental.
```

### DIFF
--- a/cron/internal/format/bq.raw.schema
+++ b/cron/internal/format/bq.raw.schema
@@ -331,6 +331,102 @@
         "mode": "REPEATED",
         "name": "defaultBranchCommits",
         "type": "RECORD"
+      },
+      {
+        "description": "",
+        "fields": [
+          {
+            "description": "",
+            "mode": "NULLABLE",
+            "name": "number",
+            "type": "STRING"
+          },
+          {
+            "description": "",
+            "mode": "NULLABLE",
+            "name": "platform",
+            "type": "STRING"
+          },
+          {
+            "description": "",
+            "mode": "REPEATED",
+            "name": "reviews",
+            "type": "RECORD",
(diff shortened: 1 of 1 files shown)


azeemshaikh38 (approved):
Thanks. Is this updated on prod BQ too? Also, could we test such divergence using unit tests?

spencerschrock:
> Thanks. Is this updated on prod BQ too? 
Will submit shortly. Wanted to get an approval/second set of eyes here before applying to prod.

> Also, could we test such divergence using unit tests?

I'll open an issue for the raw results in general and add this to the list of things

PR #2821, opened 2023-04-04
✨ GitLab: Documentation and cleaner errors

raghavkaul:
Add documentation for GitLab support and show cleaner error messages (i.e., InconclusiveResult) if a check isn't supported. 

Also, some panics are fixed and logging improvements are made.

Currently, we give 10/10 scores to GitLab in certain checks due to lack of data.

```release-note
* GitLab: For checks that aren't possible on GitLab, return Inconclusive Result
* GitHub: If no Workflows are detected, Dangerous-Workflows check returns Inconclusive instead of max score
```


checks/evaluation/dangerous_workflow.go
```diff
@@ -30,6 +30,10 @@ func DangerousWorkflow(name string, dl checker.DetailLogger,
 		return checker.CreateRuntimeErrorResult(name, e)
 	}
 
+	if r.NumWorkflows == 0 {
```
azeemshaikh38, line 33: Use `len(Workflows)`? Do we need this extra field in the struct?

checks/evaluation/permissions/permissions.go
```diff
@@ -40,6 +40,10 @@ func TokenPermissions(name string, c *checker.CheckRequest, r *checker.TokenPerm
 		return checker.CreateRuntimeErrorResult(name, e)
 	}
 
+	if r.NumTokens == 0 {
```
azeemshaikh38, line 43: Use `len(Tokens)` and remove this field.

checks/raw/dangerous_workflow.go
```diff
@@ -102,6 +102,8 @@ var validateGitHubActionWorkflowPatterns fileparser.DoWhileTrueOnFileContent = f
 		return true, nil
 	}
 
+	pdata.NumWorkflows += 1
```
azeemshaikh38, line 105: Remove.

checks/raw/permissions.go
```diff
@@ -84,6 +84,8 @@ var validateGitHubActionTokenPermissions fileparser.DoWhileTrueOnFileContent = f
 		return true, nil
 	}
 
+	pdata.results.NumTokens += 1
```
azeemshaikh38, line 87: Remove.

azeemshaikh38 (approved):
Approved with comments.

checks/evaluation/dangerous_workflow.go
raghavkaul: This is intentional, `Workflows` doesn't have all workflows, just ones we've found with risky security patterns.

clients/gitlabrepo/branches.go
```diff
@@ -73,7 +73,6 @@ func (handler *branchesHandler) setup() error {
 				handler.repourl.project, &gitlab.ListOptions{})
 			if err != nil && resp.StatusCode != 404 {
 				handler.errSetup = fmt.Errorf("request for external status checks failed with error %w", err)
-				return
```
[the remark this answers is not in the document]
raghavkaul: Yes, this was intentional - this API requires a more privileged API token. It typically returns 401 when the token isn't authorized. I can change this to ensure that `resp.StatusCode == 401` if that makes it clearer?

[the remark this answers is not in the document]
raghavkaul: What's the best way to silently pass a potential permission error here?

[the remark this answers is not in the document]
azeemshaikh38: If we want `401` to not have an error, we should be setting `handler.errSetup` to be `nil`.

PR #2839, opened 2023-04-07
✨ Use local files instead of search for SAST CodeQL check

spencerschrock:

feature

The SAST check uses the GitHub Search endpoint to look for workflow files that call CodeQL via `github/codeql-action/analyze`

Looks for steps which call `github/codeql-action/analyze` in the local files. This saves us a Search API call 

NONE

This is a precursor for re-enabling the SAST check in the weekly cron.

the `release-note`

```release-note
The SAST check will look for workflows with the "github/codeql-action/analyze" action locally instead of the GitHub Search API endpoint.
```


e2e/sast_test.go
```diff
@@ -44,10 +44,10 @@ var _ = Describe("E2E TEST:"+checks.CheckSAST, func() {
...
 			expected := scut.TestReturn{
 				Error:         nil,
-				Score:         0,
-				NumberOfWarn:  2,
-				NumberOfInfo:  0,
-				NumberOfDebug: 0,
+				Score:         10,
+				NumberOfWarn:  1,
+				NumberOfInfo:  1,
+				NumberOfDebug: 1,
```
azeemshaikh38, line 50: why did e2e test score change?

spencerschrock: The test is labeled "Should return use of SAST tools", so I suspect the old score of 0 was incorrect.

Looking at the history, it used to be a 7, but was changed to a 0 here #1152.

If you check the repo with scorecard, you'll see it also is currently returning a 0 for SAST:
```
go run main.go --repo ossf-tests/airflow --checks SAST --format json --show-details | jq
```

But there is a workflow which uses it:
https://github.com/ossf-tests/airflow/blob/0a6850647e531b08f68118ff8ca20577a5b4062c/.github/workflows/codeql-analysis.yml#L106

If you try to manually run the search API, no results are returned.
```
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer $GITHUB_AUTH_TOKEN"\
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/search/code?q=github%20codeql-action%20analyze%20repo%3Aossf-tests%2Fairflow%20path%3A%2F.github%2Fworkflows
{
  "total_count": 0,
  "incomplete_results": false,
  "items": [

  ]
}
```

The same query works for the `ossf/scorecard` repo. So I think this is similar to #2670 where the search API is broken for some repos.

azeemshaikh38 (approved):

PR #2891, opened 2023-04-20
:seedling: Unit tests for checks/evaluation/packaging.go

naveensrinivasan:
- Include unit tests for checks/evaluation/packaging.go

the `release-note`

```release-note

```

### DIFF
--- a/checks/evaluation/packaging_test.go
+++ b/checks/evaluation/packaging_test.go
@@ -1,0 +1,139 @@
+// Copyright 2023 OpenSSF Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


checks/evaluation/packaging_test.go
```diff
@@ -0,0 +1,139 @@
...
+	type args struct { //nolint:govet
+		name string
+		dl   checker.DetailLogger
+		r    *checker.PackagingData
+	}
+	tests := []struct {
+		name string
+		args args
+		want checker.CheckResult
+	}{
```
spencerschrock, line 114: Lets add two tests, as this is only checking the error pathway at the start of the func:
* empty packaging data (but not nil)
* a valid packaging data, you can probably use the same package from the `runs are not zero` test above.

spencerschrock (approved):

PR #2917, opened 2023-04-26
✨ Add a new format for details

laurentsimon:
This PR adds a new format `finding` to improve on the existing `json` format. This new format will be the same as `json`, but will display the details in a more structured format.

Note that this is *not* the final "structured" format. This is a simpler version of it.

I have named the format "finding" because it's the underlying structure that will be used to populate it, but please suggest a better name.

```release-note
Add a new format for details
```

### DIFF
--- a/pkg/json.go
+++ b/pkg/json.go
@@ -22,7 +22,6 @@ import (
 
 	docs "github.com/ossf/scorecard/v4/docs/checks"
 	sce "github.com/ossf/scorecard/v4/errors"
-
 	"github.com/ossf/scorecard/v4/log"
 )
 
--- a/pkg/json.go
+++ b/pkg/json.go
@@ -20,12 +20,10 @@ import (
 	"io"
 	"time"
 
-	"github.com/ossf/scorecard/v4/checker"
 	docs "github.com/ossf/scorecard/v4/docs/checks"
 	sce "github.com/ossf/scorecard/v4/errors"
-	"github.com/ossf/scorecard/v4/finding"
+
 	"github.com/ossf/scorecard/v4/log"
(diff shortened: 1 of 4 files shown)


pkg/json.go
```diff
@@ -196,80 +168,16 @@ func (r *ScorecardResult) AsJSON2(showDetails bool,
...
 	}
 
 	return nil
 }
 
-func (r *ScorecardResult) AsSJSON(showDetails bool,
+func (r *ScorecardResult) AsFJSON(showDetails bool,
 	logLevel log.Level, checkDocs docs.Doc, writer io.Writer,
 ) error {
-	score, err := r.GetAggregateScore(checkDocs)
```
laurentsimon, line 209: removing this function for now as we focus on the "finding" format first instead of "structured" format.

pkg/json.go
```diff
@@ -196,80 +168,16 @@ func (r *ScorecardResult) AsJSON2(showDetails bool,
...
-			}
-		}
-		out.Checks = append(out.Checks, tmpResult)
-	}
-	if err := encoder.Encode(out); err != nil {
-		return sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("encoder.Encode: %v", err))
-	}
-
-	return nil
+	return sce.WithMessage(sce.ErrScorecardInternal, "WIP: not supported")
```
laurentsimon, line 182: this will be implemented in follow-up PRs

spencerschrock (approved):
Looking forward to what the difference between the three experimental formats will be. 

I think its fine for the names to be a WIP if this is experimental.

PR #2919, opened 2023-04-26
✨ [experimental] Create probes within findings

laurentsimon:
This PR introduces the concept of probes, which will have a unique ID. These used to be called "rules".

```release-note
Create probes within findings
```


finding/finding.go
```diff
@@ -52,45 +53,120 @@ type Location struct {
 type Outcome int
 
 const (
+	// NOTE: The additional '-' are intended for future use.
+	// This allows adding outcomes without breaking the values
+	// of existing outcomes.
 	// OutcomeNegative indicates a negative outcome.
 	OutcomeNegative Outcome = iota
+	_
```
laurentsimon, line 61: note: I have added some empty value to ensure we have a possibility to add outcomes, without changing existing values for backward compatibility.
These values are ordered from worst (OutcomeNegative =0) to best OutcomeNotSupported, and we can use comparison operators.
The ordering between OutcomeNotAvailable, OutcomeError, etc is a bit handwavy / subjective at this point. The ability to compare is useful for "structured" results (not this PR).
We may want to re-visit this before GA.

checks/evaluation/permissions/permissions.go
```diff
@@ -77,14 +77,12 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
 		if r.File != nil {
 			loc = &finding.Location{
 				Type:      r.File.Type,
-				Value:     r.File.Path,
+				Path:      r.File.Path,
```
laurentsimon, line 80: note: this code will be re-arranged in a follow-up PR.

checks/evaluation/permissions/permissions.go
```diff
@@ -132,38 +130,38 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
...
-		if err := reportFinding("GitHubWorkflowPermissionsStepsNoWrite",
-			text, finding.OutcomeNotApplicable, dl); err != nil {
+		if err := reportFinding("gitHubWorkflowPermissionsStepsNoWrite",
+			text, finding.OutcomeNotAvailable, dl); err != nil {
 			return err
 		}
-		if err := reportFinding("GitHubWorkflowPermissionsTopNoWrite",
-			text, finding.OutcomeNotApplicable, dl); err != nil {
+		if err := reportFinding("gitHubWorkflowPermissionsTopNoWrite",
+			text, finding.OutcomeNotAvailable, dl); err != nil {
```
laurentsimon, line 151: may want to have a `NotApplicable` outcome. Trying to see if it's actually needed when we migrate other checks. Will re-visit.

finding/probe/probe_test.go
```diff
@@ -0,0 +1,109 @@
...
+			name: "invalid risk",
+			id:   "invalid-risk",
+			path: "testdata/invalid-risk.yml",
+			err:  errInvalid,
+		},
+		{
+			name: "invalid effort",
+			id:   "invalid-effort",
+			path: "testdata/invalid-effort.yml",
+			err:  errInvalid,
```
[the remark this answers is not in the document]
spencerschrock: (i think it lacks an id in the yaml file)

finding/finding.go
```diff
@@ -131,16 +240,37 @@ func (f *Finding) WithRemediationMetadata(values map[string]string) *Finding {
 	if f.Remediation != nil {
 		// Replace all dynamic values.
 		for k, v := range values {
+			// Replace metadata.
 			f.Remediation.Text = strings.Replace(f.Remediation.Text,
-				fmt.Sprintf("${{ %s }}", k), v, -1)
+				fmt.Sprintf("${{ metadata.%s }}", k), v, -1)
```
spencerschrock, line 245: was this change reflected everywhere? I still see the non metadata values used in the gitHubWorkflowPermissionsTopNoWrite yaml

finding/finding_test.go
```diff
@@ -184,7 +167,12 @@ func Test_New(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			r, err := New(testfs, tt.id)
+			content, err := os.ReadFile(tt.path)
+			if err != nil {
+				panic(err)
```
spencerschrock, line 172: Lets avoid panic'ing, how about `t.Fatalf`

finding/finding.go
```diff
@@ -52,45 +55,142 @@ type Location struct {
 type Outcome int
 
 const (
+	// NOTE: The additional '-' are intended for future use.
```
spencerschrock, line 58: nit: you have a `-` instead of a `_`

finding/finding.go
```diff
@@ -52,45 +55,142 @@ type Location struct {
...
+
+	f = f.WithMessage(text).WithLocation(loc)
 	return f, nil
 }
 
+// Anonimize removes the probe ID and outcome
+// from the finding. It is a temporary solution
+// to integrate the code in the details without exposing
+// too much information.
+func (f *Finding) Anonimize() *AnonymousFinding {
```
spencerschrock, line 190: spelling?: "Anonymize"

finding/finding.go
```diff
@@ -52,45 +55,142 @@ type Location struct {
...
+func NewNegative(efs embed.FS, probeID, text string, loc *Location,
+) (*Finding, error) {
+	f, err := NewWith(efs, probeID, text, loc, OutcomeNegative)
+	if err != nil {
+		return nil, fmt.Errorf("finding.NewWith: %w", err)
 	}
-	if r.Remediation != nil {
-		f.Risk = r.Risk
+
+	f = f.WithMessage(text).WithLocation(loc)
```
spencerschrock, line 158: what does this extra line do? isn't location set by `NewWith`?

do you need anything more in these functions than 
```
return NewWith(efs, probeID, text, loc, OutcomeNegative)
```

spencerschrock: same question applies for all of these NewOutcome functions

spencerschrock (commented):
I had to skim over some parts but wanted to get the partial review out before I travel a bit.

finding/finding.go
spencerschrock: It may also be worth explicitly setting the values of each outcome explicitly instead of using `iota`.

We could let us have gaps still, if we wanted to pursue having everything be ordered.

checks/evaluation/permissions/permissions.go
```diff
@@ -67,24 +67,22 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
 	dl := c.Dlogger
 	//nolint:errcheck
 	remediationMetadata, _ := remediation.New(c)
-	negativeRuleResults := map[string]bool{
-		"GitHubWorkflowPermissionsStepsNoWrite": false,
-		"GitHubWorkflowPermissionsTopNoWrite":   false,
+	negativeProbeResults := map[string]bool{
+		"gitHubWorkflowPermissionsStepsNoWrite": false,
+		"gitHubWorkflowPermissionsTopNoWrite":   false,
```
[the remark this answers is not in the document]
laurentsimon: i would like to do that in the future. Just need to find a good way to avoid redundancy. I plan to re-visit in follow-up PRs

finding/finding.go
laurentsimon: added to https://github.com/ossf/scorecard/issues/2928.

finding/finding.go
laurentsimon: +1, must be a copy-paste mistake. Fixed.

finding/finding.go
laurentsimon: +1, forgot to update. Thanks!

finding/finding_test.go
laurentsimon: fine with me. What's the difference?

finding/probe/probe_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+func errCmp(e1, e2 error) bool {
+	return errors.Is(e1, e2) || errors.Is(e2, e1)
+}
+
+func Test_New(t *testing.T) {
```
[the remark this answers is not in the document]
laurentsimon: renamed. Last-minute renaming to make tests works :-)

finding/probe/probe_test.go
```diff
@@ -0,0 +1,109 @@
...
+					Effort:   RemediationEffortLow,
+				},
+			},
+			err: errInvalid,
+		},
+		{
+			name: "invalid risk",
+			id:   "invalid-risk",
+			path: "testdata/invalid-risk.yml",
+			err:  errInvalid,
```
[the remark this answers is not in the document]
laurentsimon: renamed the file. Was supposed to be missing ID test.

finding/probe/probe_test.go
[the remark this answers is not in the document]
laurentsimon: updated the error messages to indicate what's invalid (effort, ID, etc)

finding/probe/testdata/invalid-risk.yml
```diff
@@ -6,7 +6,6 @@ motivation: >
 implementation: >
   line1
   line2
-risk: Low
```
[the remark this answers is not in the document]
laurentsimon: renamed to missing-id.

checks/evaluation/permissions/permissions.go
```diff
@@ -185,11 +187,15 @@ func reportFinding(rule, text string, o finding.Outcome, dl checker.DetailLogger
...
 	if loct == nil || *loct == checker.PermissionLocationTop {
-		Rule = "GitHubWorkflowPermissionsTopNoWrite"
+		probe = "gitHubWorkflowPermissionsTopNoWrite"
+	}
+	content, err := probes.ReadFile(probe + ".yml")
+	if err != nil {
+		return nil, fmt.Errorf("%w", err)
 	}
-	f, err := finding.New(rules, Rule)
+	f, err := finding.FromBytes(content, probe)
```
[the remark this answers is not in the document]
laurentsimon: same answer as above. it's a temporary solution until I crate a `probe` folder with all the probes (next PR)

laurentsimon:
Thanks for the review! Made the changes, PTAL

finding/finding.go
```diff
@@ -102,13 +203,22 @@ func (f *Finding) WithMessage(text string) *Finding {
 // No copy is made.
 func (f *Finding) WithLocation(loc *Location) *Finding {
 	f.Location = loc
+	if f.Remediation != nil && loc != nil {
```
spencerschrock, line 206: is there any ordering to be worried about here?

In terms of when remediation info is set vs when location is?

finding/probe/probe_test.go
```diff
@@ -0,0 +1,109 @@
...
+		},
+	}
+	for _, tt := range tests {
+		tt := tt // Re-initializing variable so it is not changed while executing the closure below
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			content, err := os.ReadFile(tt.path)
+			if err != nil {
+				t.Fatalf(err.Error())
```
spencerschrock, line 93: should use t.fatal if you're not using a format string

finding/finding_test.go
spencerschrock: they both cause the test to fail with a message, but panic clogs up the output with the goroutine traces.

finding/finding.go
spencerschrock: i dont think this is resolved. 

We don't need `f = f.WithMessage(text)` either as NewWith takes care of the text. The same applies to all NewFoo functions.

spencerschrock (approved):
overall LGTM as experimental changes.

I think we should do a thorough review before GA, as there will likely be a lot of TODOs left. Thanks for opening #2928 as it helps track things.

finding/finding.go
laurentsimon: ho you're right! Fixed.

finding/finding_test.go
laurentsimon: thanks. will keep in mind.

finding/finding.go
laurentsimon: I've updated the code to use `f.Location` instead of `loc` in the part that sets the remediation. Thanks

laurentsimon:
> Still think the const approach is a good step for now.

I've made the changes.

checks/evaluation/permissions/permissions.go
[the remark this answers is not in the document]
laurentsimon: done.

PR #2921, opened 2023-04-27
:seedling: Unit tests checks/evaluation/signedrelease.go

naveensrinivasan:
- Add a test file to the Signed Releases check provenance

the `release-note`

```release-note

```

### DIFF
--- a/checks/evaluation/signed_releases_test.go
+++ b/checks/evaluation/signed_releases_test.go
@@ -1,0 +1,108 @@
+//  Copyright 2023 OpenSSF Scorecard Authors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


checks/evaluation/signed_releases_test.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+func TestSignedReleases(t *testing.T) {
+	tests := []struct {
+		name           string
+		releases       []clients.Release
+		expectedResult checker.CheckResult
+	}{
+		{
+			name: "Full score",
```
spencerschrock, line 35: optional: the test names would be clearer to me if the name included the behavior we're looking for. 
So for this test we're looking for the provenance, in the other test we're looking for signature, etc.

spencerschrock (approved):
Thanks

checks/evaluation/signed_releases_test.go
naveensrinivasan: I am going to skip this for now. Thanks for the feedback.

PR #2936, opened 2023-05-02
✨ Gitlab: Add projects to cron

raghavkaul:

Add cron worker support for GitLab projects. Switch worker's RepoClient when using a GitLab repo.


cron/k8s/controller.yaml
```diff
@@ -52,7 +52,11 @@ spec:
           containers:
             - name: controller
               image: gcr.io/openssf/scorecard-batch-controller:stable
-              args: ["--config=/etc/scorecard/config.yaml", "cron/internal/data/projects.csv"]
+              args: [
+                "--config=/etc/scorecard/config.yaml", 
+                "cron/internal/data/projects.csv", 
+                "cron/internal/data/gitlab-projects-selected.csv"
```
spencerschrock, line 58: what are your thoughts on doing this in the release controller config instead?

raghavkaul: What is the difference between this controller and the release controller?

spencerschrock: The controller enumerates the weekly 1.2M repos for the production workers.
The release controller is the nightly test that does 5000 repos for the staging worker that moves commits from `latest` to `stable`.

My opinion is that the experimental features should stay in the release/nightly worker to not affect the cron.

checker/client.go
```diff
@@ -58,7 +58,7 @@ func GetClients(ctx context.Context, repoURI, localURI string, logger *log.Logge
 	var repoClient clients.RepoClient
 
 	//nolint:nestif
-	if experimental && glrepo.DetectGitLab(repoURI) {
```
spencerschrock, line 61: There are reasons to use `SCORECARD_EXPERIMENTAL` for GitHub repos as well. This change forces the gitlab repo client.

clients/gitlabrepo/client.go
```diff
@@ -228,6 +229,22 @@ func (client *Client) Close() error {
 	return nil
 }
 
+func CreateGitlabDotComClient(ctx context.Context) (clients.RepoClient, error) {
+	var err error
+	var gitlabComRepo clients.Repo
+	if gitlabComRepo, err = MakeGitlabRepo("https://gitlab.com/gitlab-org/gitlab"); err != nil {
```
spencerschrock, line 235: Is there significance to this repo? Or is this just an arbitrary repo on the gitlab.com host?

clients/gitlabrepo/client.go
```diff
@@ -285,14 +302,3 @@ func CreateGitlabClientWithToken(ctx context.Context, token string, repo clients
 func CreateOssFuzzRepoClient(ctx context.Context, logger *log.Logger) (clients.RepoClient, error) {
 	return nil, fmt.Errorf("%w, oss fuzz currently only supported for github repos", clients.ErrUnsupportedFeature)
 }
-
-// DetectGitLab: check whether the repoURI is a GitLab URI
-// Makes HTTP request to GitLab API.
-func DetectGitLab(repoURI string) bool {
```
spencerschrock, line 291: regarding the comment in `checker/client.go`, does it make sense to have some sort of gitlab detection still?

clients/gitlabrepo/repo.go
```diff
@@ -46,6 +47,9 @@ type repoURL struct {
 The following input format is not supported:
 	* https://gitlab.<companyDomain:string>.com/projects/<projectID:int>
 */
+
+var errInvalidGitlabRepoURL = errors.New("repo is not a gitlab repo")
+
```
spencerschrock, line 52: can we move this higher so we don't get between the godoc and function for `parse`?

clients/gitlabrepo/repo_test.go
```diff
@@ -157,9 +157,13 @@ func TestRepoURL_DetectGitlab(t *testing.T) {
 		if tt.flagRequired && os.Getenv("TEST_GITLAB_EXTERNAL") == "" {
 			continue
 		}
-		g := DetectGitLab(tt.repouri)
-		if g != tt.expected {
-			t.Errorf("got %s isgitlab: %t expected %t", tt.repouri, g, tt.expected)
+		g, err := MakeGitlabRepo(tt.repouri)
```
spencerschrock, line 160: If we're switching what the test is testing, lets change the name of the function above as well (wouldn't let me comment up there.)

clients/gitlabrepo/repo_test.go
```diff
@@ -157,9 +157,13 @@ func TestRepoURL_DetectGitlab(t *testing.T) {
...
+		if (g != nil) != (err == nil) {
```
spencerschrock, line 161: should this just be `if err != nil` ? or are we expecting some calls to error?

cron/data/iterator.go
```diff
@@ -83,9 +84,14 @@ func (reader *csvIterator) Next() (RepoFormat, error) {
...
-	if _, err := githubrepo.MakeGithubRepo(reader.next.Repo); err != nil {
-		return reader.next, fmt.Errorf("invalid GitHub URL: %w", err)
+
+	repoURI := reader.next.Repo
+
+	// validate gitlab or github url
+	if _, err := gitlabrepo.MakeGitlabRepo(repoURI); err != nil {
+		if _, err := githubrepo.MakeGithubRepo(repoURI); err != nil {
+			return reader.next, fmt.Errorf("invalid URL, neither github nor gitlab: %w", err)
+		}
```
spencerschrock, line 94: this makes a call to the gitlab API correct?

Can we at least switch the ordering around so the cron isn't making 1.2M gitlab requests each week?

raghavkaul: It shouldn't do that; the solution is a little ugly but in `gitlabrepo.isvalid` we exit early if the `repo.host` is `github.com`.

cron/internal/data/gitlab-projects-selected.csv
```diff
@@ -0,0 +1,100 @@
+https://gitlab.com/gitlab-org/gitlab-foss,
```
spencerschrock, line 1: while we support not having the csv header, lets keep it here for consistency with `projects.csv`

cron/internal/data/projects.release.csv
```diff
@@ -0,0 +1,2501 @@
+repo,metadata
```
spencerschrock, line 1: this file shouldn't be committed.

cron/internal/worker/main.go
```diff
@@ -131,22 +161,27 @@ func processRequest(ctx context.Context,
 	batchRequest *data.ScorecardBatchRequest,
 	blacklistedChecks []string, bucketURL, rawBucketURL, apiBucketURL string,
 	checkDocs docs.Doc,
-	repoClient clients.RepoClient, ossFuzzRepoClient clients.RepoClient,
+	githubClient, gitlabClient clients.RepoClient, ossFuzzRepoClient clients.RepoClient,
 	ciiClient clients.CIIBestPracticesClient,
 	vulnsClient clients.VulnerabilitiesClient,
 	logger *log.Logger,
 ) error {
+	fmt.Println("reached")
```
spencerschrock, line 169: this got left in

clients/gitlabrepo/repo_test.go
raghavkaul: There should be an error whenever the result is nil, and there should be no error when the result is non-nil. For every test where expected=false, we expect an error & nil return. It's probably overkill, but this is my pattern to check that the conditions on L164 are either both true or both false.

clients/gitlabrepo/client.go
raghavkaul: I think there's a way to do without detection. `MakeGitlabRepo` and `DetectGitlab` both called `parse` and `isvalid`, so it didn't feel like they needed to be separate functions.

cron/worker/worker_test.go
```diff
@@ -55,3 +55,49 @@ func TestResultFilename(t *testing.T) {
 		})
 	}
 }
+
+func asptr(s string) *string {
+	return &s
+}
+
+func TestGitlabRepo(t *testing.T) {
```
spencerschrock, line 63: what is this test trying to do?

spencerschrock, line 63: what is this test trying to do?
it seems like a copy/paste of the test above it with a minimal change that doesnt affect output?

spencerschrock, line 63: what is this test trying to do?
it seems like a copy/paste of the test above it with a minimal change that doesnt affect output?

cron/internal/worker/main.go
```diff
@@ -143,10 +174,14 @@ func processRequest(ctx context.Context,
 	// TODO: run Scorecard for each repo in a separate thread.
 	for _, repoReq := range batchRequest.GetRepos() {
 		logger.Info(fmt.Sprintf("Running Scorecard for repo: %s", *repoReq.Url))
-		repo, err := githubrepo.MakeGithubRepo(*repoReq.Url)
-		if err != nil {
+		var repo clients.Repo
+		var err error
+		repoClient := githubClient
+		if repo, err = gitlabrepo.MakeGitlabRepo(*repoReq.Url); err != nil {
+			repoClient = gitlabClient
```
spencerschrock, line 181: same question about this making a gitlab request in the cron, which will result in millions of queries.

cron/internal/worker/main.go
```diff
@@ -161,7 +196,16 @@ func processRequest(ctx context.Context,
 		if err != nil {
 			return fmt.Errorf("error during policy.GetEnabled: %w", err)
 		}
-		for _, check := range blacklistedChecks {
+
+		disabledChecks := blacklistedChecks
+		if _, err := gitlabrepo.MakeGitlabRepo(*repoReq.Url); err != nil {
```
spencerschrock, line 201: can we save the results from above where we set the gitlab repos? or make this logic be in the same place?

spencerschrock:
Sorry for all the spam, this was supposed to go out as one review.

cron/internal/worker/main.go
raghavkaul: API queries for client detection shouldn't happen unless the URL is not `github.com|gitlab.com`.

clients/gitlabrepo/client.go
raghavkaul: It's arbitrary; changed this to use `gitlab.WithBaseURL`

cron/worker/worker_test.go
raghavkaul: Yes, it's spurious. Removed

checker/client.go
spencerschrock: Marking unresolved, as this still looks like it's an issue. Try running this to see my point:
```
SCORECARD_EXPERIMENTAL=1 go run main.go --repo https://github.com/ossf/scorecard
```

cron/internal/worker/main.go
```diff
@@ -91,7 +118,11 @@ func newScorecardWorker() (*ScorecardWorker, error) {
 
 	sw.ctx = context.Background()
 	sw.logger = log.NewLogger(log.InfoLevel)
-	sw.repoClient = githubrepo.CreateGithubRepoClient(sw.ctx, sw.logger)
+	sw.githubClient = githubrepo.CreateGithubRepoClient(sw.ctx, sw.logger)
+	// TODO(raghavkaul): Read GitLab auth token from environment
+	if sw.gitlabClient, err = gitlabrepo.CreateGitlabClient(sw.ctx, "gitlab.com"); err != nil {
```
spencerschrock, line 123: the baseurls are expected to include the scheme, or you'll get repo unreachable errors for every repo

can we make this a constant with value `https://gitlab.com/`?

cron/k8s/controller.release.yaml
```diff
@@ -52,7 +52,11 @@ spec:
           containers:
             - name: controller
               image: gcr.io/openssf/scorecard-batch-controller:latest
-              args: ["--config=/etc/scorecard/config.yaml", "cron/internal/data/projects.release.csv"]
+              args: [
+                "--config=/etc/scorecard/config.yaml",
+                "cron/internal/data/projects.release.csv",
+                "cron/internal/data/gitlab-projects-selected.csv"
```
spencerschrock, line 58: I still get panics running this against repos in this list. Lets fix these up before merging in.

spencerschrock (commented):
Overall the logic is good, but there are a few blockers to getting this merged. Let's aim for having this run in the cron emulator without errors before merging it in for our nightly release tests.

Also, I personally think we shouldn't check in `gitlab-projects.csv` yet, but not a blocker on my side.

spencerschrock (approved):
Looks good overall

cron/internal/worker/main.go
```diff
@@ -49,14 +50,40 @@ const (
...
+		"Fuzzing",
+		"License",
+		"Maintained",
+		"Packaging",
+		"Pinned-Dependencies",
+		"SAST",
+		// "Security-Policy",
+		"Signed-Releases",
+		"Token-Permissions",
+		"Vulnerabilities",
```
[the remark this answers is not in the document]
raghavkaul: Yes, added

cron/k8s/controller.release.yaml
```diff
@@ -66,6 +70,8 @@ spec:
                   value: "gs://ossf-scorecard-rawdata-releasetest"
                 - name: SCORECARD_SHARD_SIZE
                   value: "5"
+                - name: "SCORECARD_EXPERIMENTAL"
```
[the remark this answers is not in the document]
raghavkaul: Yes, this flag should be set in the worker instead, changed

clients/gitlabrepo/contributors_test.go
```diff
@@ -48,7 +48,7 @@ func Test_ContributorsSetup(t *testing.T) {
 				t.Error("couldn't make gitlab repo", err)
 			}
 
-			client, err := CreateGitlabClientWithToken(context.Background(), "", repo)
+			client, err := CreateGitlabClientWithToken(context.Background(), "", repo.Host())
```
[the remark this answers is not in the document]
raghavkaul: Will create a follow up PR to move this testing to `e2e/`.

PR #2951, opened 2023-05-04
:seedling: Included e2e tests for push to main

naveensrinivasan:
- Update trigger for integration tests to enable running on `push` and `pull_request` on the `main` branch

the `release-note`

```release-note

```


raghavkaul (approved):

.github/workflows/integration.yml
```diff
@@ -14,7 +14,14 @@
...
 name: Integration tests
-on: pull_request
+
+on:
+  push:
+    branches:
+      - main # The e2e coverage is required to be run on main branch to get the coverage report
+  pull_request:
+    branches:
+      - main
```
spencerschrock, line 24: I had thought we disabled this at one point due to token usage from running integration tests before/after PR merge?

I wonder if there's anyway to use artifacts so we can upload it from `main.yml` without re-running it?

naveensrinivasan: I examined the file for blame and found that it had never been pushed to the main section. Can we enable it and see if it is causing any problems?

Maintaining the state from a pull request to a main branch merge can be complicated. I'm not sure how to do it.

Issue #2582, opened 2023-01-05
Feature: Improve Pinned-Dependencies score calculation


gabibguti:
**Is your feature request related to a problem? Please describe.**

I'm frustrated when I have multiple Pinned-Dependencies warnings, I fix multiple warnings and my score does not change for that check.

After some code analysis I got to the following conclusions:

Main problem:
The pinning results are not proportional to the number of dependency warnings of each type.

E.g.: If I get 1 or 100 `containerImage` warnings, the `containerImage` will be `notPinned` in pinning results, the `dockerFromScore` will be 0, cause it's 10 or 0, and this will affect 1/4 of my final score.

Side problems:
- The dependencies of types `goCommand`, `chocoCommand`, `npmCommand` and `pipCommand` scores do not affect the final score.
- The dependencies of types `goCommand`, `chocoCommand`, `npmCommand` and `pipCommand` generate warnings on JSON and on the Security Dashboard. The Security Dashboard should not be populated with warnings that do not affect the score.
- The dependencies of type `downloadThenRun` affect both `dockerDownloadScore` and `scriptScore` in a way that the score is penalized twice for each `downloadThenRun` warning.

**Describe the solution you'd like**

I would like the pinning results to be proportional to the number os dependency warnings of each type, and the final score to be composed by all dependency types and not grouped into `actionScore` and etc.

**Describe alternatives you've considered**

None.

**Additional context**

None.

PR #2960, opened 2023-05-05
🐛  Add npm installs to Pinned-Dependencies score

gabibguti:

This is a bug fix. Just like pip installs, npm installs were identified but did not count for Pinned-Dependencies score. Now, they do.

Npm installs are identified as pinned or unpinned and reported by Scorecard, but do not count for Pinned-Dependencies score.

Npm installs now count in the aggregated score for Pinned-Dependencies.

It's addressing one of the problems reported at https://github.com/ossf/scorecard/issues/2582

the `release-note`

```release-note
Npm installs count for Pinned-Dependencies score.
```


checks/evaluation/pinned_dependencies.go
```diff
@@ -127,15 +127,22 @@ func PinningDependencies(name string, c *checker.CheckRequest,
 		return checker.CreateRuntimeErrorResult(name, err)
 	}
 
+	// Npm installs.
+	npmScore, err := createReturnForIsNpmInstallPinned(pr, dl)
+	if err != nil {
+		return checker.CreateRuntimeErrorResult(name, err)
+	}
```
naveensrinivasan, line 134: Can you please include tests for this, also?

naveensrinivasan (approved):
Thank you!

checks/evaluation/pinned_dependencies.go
gabibguti: I have done some validations and I'm pretty sure this code is unreachable.

Further explanation:
`PinningDependencies` creates `pr`. `pr` is a map of dependency types (e.g. npm install) and their classification (one of `pinned`, `notPinned`, `pinnedUndefined`). Then, for each dependency type in `pr` it calls `createReturnValues` to generate a score based on the classification. The score is given as follows:
- 10, if that dependency type is `pinned` or `pinnedUndefined`
- 0, if that dependency type is `notPinned`
- Error, as default case if does not meet any cases above

But, it will never happen that the classification will not meet any of the cases above. That's because `pr` is initialized with all dependency types with default classification as `pinnedUndefined`. Also, all changes to the classification will result in either `pinned` or `notPinned`. So, when the classification is scored in `createReturnValues`, it will always be one of those 3 values, and the default case is never met.
This could happen if new values were added to the classification, but that's unlikely.

naveensrinivasan: Sounds good. Thanks

naveensrinivasan:
Can you please resolve the conflicts? We should be good!

checks/evaluation/pinned_dependencies.go
```diff
@@ -260,6 +267,15 @@ func createReturnForIsPipInstallPinned(pr map[checker.DependencyUseType]pinnedRe
 		dl)
 }
 
+// Create the result for npm install commands.
+func createReturnForIsNpmInstallPinned(pr map[checker.DependencyUseType]pinnedResult,
+	dl checker.DetailLogger,
+) (int, error) {
+	return createReturnValues(pr, checker.DependencyUseTypeNpmCommand,
```
jadekler, line 274: i poked through the CL per your request. nothing stands out, but here and elsewhere: the google style guide would suggest a single line function invocation, rather than putting params on other lines

(though, I can see what this codebase does not generally do that, so your call on whether you want to start now)

https://google.github.io/styleguide/go/decisions#function-formatting

everything else looks fine (there's not too much code, so easy to review!)

gabibguti: Thanks for reviewing, Jean!

> I can see what this codebase does not generally do that, so your call on whether you want to start now

Yeah, I will make the call to not address this in this PR. But will suggest this change for this file in a new PR! I do see the benefits.

gabibguti:
Hey @naveensrinivasan! Are we good to merge this?

balteravishay:
I'd like to add the same functionality for nuget, but do not want to create conflicts with this PR. 
Is this PR about to be merged? @gabibguti  @spencerschrock

spencerschrock:
> I'd like to add the same functionality for nuget, but do not want to create conflicts with this PR. Is this PR about to be merged? @gabibguti @spencerschrock

I will work on getting this merged.

checks/evaluation/pinned_dependencies.go
```diff
@@ -260,6 +267,15 @@ func createReturnForIsPipInstallPinned(pr map[checker.DependencyUseType]pinnedRe
...
+		"Npm installs are pinned",
```
spencerschrock, line 275: optional: `npm`

checks/evaluation/pinned_dependencies_test.go
```diff
@@ -229,9 +229,42 @@ func Test_PinningDependencies(t *testing.T) {
...
+			expected: scut.TestReturn{
+				Error:         nil,
+				Score:         8,
+				NumberOfWarn:  1,
+				NumberOfInfo:  6,
+				NumberOfDebug: 0,
+			},
+		},
+		{
+			name: "undefined npm install",
```
spencerschrock, line 255: what makes a npm install undefined? I don't know if there's a way to get this scenario, so let's get rid of this test?

https://github.com/ossf/scorecard/blob/397c97f766bb6c2cf6740c698e52e5c74de40dbb/checks/raw/shell_download_validate.go#L839-L855

spencerschrock (approved):

checks/evaluation/pinned_dependencies_test.go
gabibguti: > what makes a npm install undefined?

By this test I meant to cover the case when an error happens to parse a npm install dependency (or really any dependency). Here's the commit explanation:

test: Undefined npm install score
When an error happens to parse a npm install dependency, the error/debug message is saved in "Msg" field. In this case, we were not able to define if the npm install is pinned or not. This dependency is classified as pinned undefined. We treat such cases as pinned cases, so it logs as Info that npm installs are all pinned and counts the score as 10. Then, the final score makes it to 10 as well. Since it logs the error/debug message, the Debug log goes to 1.

spencerschrock: > (or really any dependency)

I would recommend renaming this test then. I don't see anyway for the npm specific logic to throw a parse error?

gabibguti: > I don't know if there's a way to get this scenario, so let's get rid of this test?

Great question. In my validations, there's no way a dependency can end up in `pinnedUndefined` case:
https://github.com/gabibguti/scorecard/blob/397c97f766bb6c2cf6740c698e52e5c74de40dbb/checks/evaluation/pinned_dependencies.go#L290
As I recall, that's because all dependencies go through `addPinnedResult` evaluation and cannot get out of there as `pinnedUndefined`. 
https://github.com/gabibguti/scorecard/blob/397c97f766bb6c2cf6740c698e52e5c74de40dbb/checks/evaluation/pinned_dependencies.go#L211C6-L211C21
Anyways, since `createReturnValues` can fall in this case I felt the need to cover it, but I see no way for it to happen, **so I agree on removing this test :)**

gabibguti: I removed the test, please check if you agree.

PR #2978, opened 2023-05-09
🌱 Remove hard-coded resources to allow auto-scaling

azeemshaikh38:
The cron/k8s workers were initialized with hardcoded resources. Due to this GKE engine wasn't optimizing the worker load properly. Removed this hardcoding and confirmed that GKE engine optimization works as expected.

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cron/k8s/worker.yaml
+++ b/cron/k8s/worker.yaml
@@ -44,13 +44,6 @@ spec:
                 secretKeyRef:
                   name: github
                   key: installation_id
-          resources:
-            requests:
-              memory: 5Gi
-              ephemeral-storage: 100Gi
-            limits:
-              memory: 12Gi
-              ephemeral-storage: 500Gi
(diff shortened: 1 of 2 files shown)


cron/k8s/worker.yaml
```diff
@@ -44,13 +44,6 @@ spec:
                 secretKeyRef:
                   name: github
                   key: installation_id
-          resources:
-            requests:
-              memory: 5Gi
-              ephemeral-storage: 100Gi
-            limits:
-              memory: 12Gi
-              ephemeral-storage: 500Gi
```
spencerschrock, line 53: Does autoscaling work with `requests` deleted, but `limits` still there? If we remove `limits` do we risk consuming crazy resources if we have some resource leak?

spencerschrock (approved):

PR #2964, opened 2023-05-07
:sparkles: Add packaging workflow for semantic-release

travi:
#### What kind of change does this PR introduce?

Adds detection for the recommended use of semantic-release in a GitHub Actions workflow.

- [x] PR title follows the guidelines defined in our [pull request documentation](https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process)

#### What is the current behavior?

`semantic-release` is not detected as a packaging tool. detection for `python-semantic-release` was added in https://github.com/ossf/scorecard/pull/1654, but that is a separate project from `semantic-release`

#### What is the new behavior (if this is a feature change)?**

`semantic-release` is now detected when used with npx in the workflow. the following variations are detected in the `run` property of a step:
* `npx semantic-release`
* `npx semantic-release@21`
* `npx semantic-release@21.0.2`
* `npx semantic-release@beta`
* `npx -p @semantic-release/git semantic-release`

- [x] Tests for the changes have been added (for bug fixes/features)

I verified that the variations above are detected when set in the testdata workflow file, but the PR currently only contains one of those variations. I did not find examples of other situations that defined multiple workflow variations for a single packaging tool detection, but I am open to feedback if additional tests would be desired.

#### Which issue(s) this PR fixes

Fixes #2929 

#### Special notes for your reviewer

i personally use [a reusable workflow](https://github.com/form8ion/.github/blob/master/.github/workflows/release-package.yml) ([example reference](https://github.com/travi/cli/blob/c41f5ea2e4fcd11fd00492708ebffef09ba0f803/.github/workflows/node-ci.yml#L62)) to capture the details of running `semantic-release`. (Since I am a `semantic-release` maintainer, I mostly do this to enable easy switching from use of the latest version to a pre-release in order to test further across my own projects.) I assume reusable workflows would prevent detection of the `semantic-release` usage, is that correct or does the scanning also evaluate those?

even if reusable workflows are not scanned, adding this detection would be valuable for much of our community. i would be interested to add this support and follow up with further discussion afterward about reusable workflows if not already scanned.

#### Does this PR introduce a user-facing change?

For user-facing changes, please add a concise, human-readable release note to
the `release-note`

(In particular, describe what changes users might need to make in their
application as a result of this pull request.)

```release-note
✨ Add packaging workflow for semantic-release
```

Issue #2929, opened 2023-04-29
Feature: Support semantic-release for packaging


travi:
**Is your feature request related to a problem? Please describe.**
As a maintainer of semantic-release, I am also a heavy user. Since semantic-release is not detected as a packaging tool, both "Packaging" and "Token-Permissions" checks are impacted negatively, even though fully automated publishing workflows are implemented

**Describe the solution you'd like**
It would be great if semantic-release could be detected as a packaging tool and result in improved scores for the related checks

**Describe alternatives you've considered**
https://github.com/ossf/scorecard/blob/27cfe92ed356fdb5a398c919ad480817ea907808/docs/checks.md#packaging mentions future improved querying of the npm registry, so waiting for that could cover a large amount of semantic-release usage. however, semantic-release is used for additional package ecosystems than npm, so this could involve waiting for support for each ecosystem registry to be supported. in addition, i imagine this would not account for "Token-Permissions"

**Additional context**
* https://github.com/semantic-release/semantic-release/
* https://github.com/semantic-release/semantic-release/issues/2779

I am not a Go developer, so I would need guidance in order to contribute the change myself. At minimum, I am interested in coordinating with members of our community to get this implemented or at least help clarify usage details around semantic-release.

PR #2989, opened 2023-05-11
:sparkles: Add releasing workflow for semantic-release

travi:

this is a follow up to https://github.com/ossf/scorecard/pull/2964 to support `semantic-release` as a releasing tool in addition to a packaging tool. i assumed incorrectly that both were covered by identifying `semantic-release` as a packaging tool

`semantic-release` is recognized as a packaging tool, but not as a releasing tool

`semantic-release` is recognized as both a packaging tool and a releasing tool

i was unable to find tests for other releasing tools, even by removing existing tools and checking for failing tests. i'm happy to add a test for this case, if desired, but would appreciate guidance from maintainers of a desired approach

related to #2929 

the `release-note`

```release-note
:sparkles: Add releasing workflow for semantic-release
```


spencerschrock (commented):
For tests, we sometimes have tests at the check level, instead of `raw` or `evaluation`. We're working on getting more tests broken down into the various packages where the code is.

Based on existing tests, it should be relatively painless to add a "release workflow contents write semantic-release" (or similarly named) entry into the table here:
https://github.com/ossf/scorecard/blob/3d1c472c2d87af163eedb5c90422a8eb4dbee7cb/checks/permissions_test.go#L260-L282

travi:
> Based on existing tests, it should be relatively painless to add a "release workflow contents write semantic-release" (or similarly named) entry into the table here:

thanks for the quick feedback. i got a test added there. open to feedback if you'd like to see more than that added

spencerschrock (approved):
Thanks for adding the test

PR #2992, opened 2023-05-11
🌱 Gitlab: e2e test fixes in main

raghavkaul:
Fixes E2E test issues on main. Environment secrets are only passed to Workflows on the `main` branch when the merge is made by a contributor. E.g.:

- PR merged by contributor (secrets available): https://github.com/ossf/scorecard/actions/runs/4949906433/jobs/8852756512
- PR merged by non-contributor (secrets not available): https://github.com/ossf/scorecard/actions/runs/4949721009/jobs/8852291352

Also fixes a test failure.


.github/workflows/gitlab.yml
```diff
@@ -21,6 +21,9 @@ on:
     branches:
       - main
 
+env:
+    GITLAB_AUTH_TOKEN: ${{ secrets.GITLAB_TOKEN }}
```
spencerschrock, line 25: lets keep this back at the job level

.github/workflows/gitlab.yml
```diff
@@ -46,10 +49,9 @@ jobs:
         run: |
             go mod download
 
-      - name: Run GitLab PAT E2E  #using retry because the GitHub token is being throttled.
+      - name: Run GitLab PAT E2E  # skip if auth token is not available
+        if: ${{ env.GITLAB_AUTH_TOKEN != '' }}
```
spencerschrock, line 53: since this is related to a dependabot feature, would it make more sense to check for `github.actor != "dependabot[bot]"`
https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/

e2e/ci_tests_test.go
```diff
@@ -151,7 +151,7 @@ var _ = Describe("E2E TEST:"+checks.CheckCITests, func() {
 			}
 			expected := scut.TestReturn{
 				Error:         nil,
-				Score:         2,
+				Score:         3,
```
spencerschrock, line 154: Do we know why this value changed?

raghavkaul: It was always that way, but it was merged because the test doesn't run on merge (token stuff).

.github/workflows/gitlab.yml
raghavkaul: Do you mean at the step level?

spencerschrock (approved):
We should still look into what caused the score change as it doesn't appear to be anything on our end, but this will at least fix the broken tests.

PR #3012, opened 2023-05-15
:book: Clarify AI/ML not human code review - in .yml file

david-a-wheeler:
This clarifies that AI/ML doesn't count as human code review. This was earlier done in #2953 but that didn't modify the relevant .yml file - this does.


olivekl (approved):
Thanks for this, LGTM

PR #3020, opened 2023-05-16
✨ add --nuget package manager flag

balteravishay:

Add --nuget ([nuget](https://www.nuget.org/)) switch to scorecard flags so that users have the option to run Scorecard using the Nuget package manager.
 
- [v] PR title follows the guidelines defined in our [pull request documentation](https://github.com/ossf/scorecard/blob/main/CONTRIBUTING.md#pr-process)

Users of the npm, pypi and rubygems ecosystems can run Scorecard on packages that are hosted on the respective registries, by running scorecard with the ecosystem flag, for example: --npm=angular

Users of the nuget ecosystem can also run Scorecard on packages that are hosted on Nuget.org, by running scorecard with the ecosystem flag, for example: --nuget=Newtonsoft.Json

- [v] Tests for the changes have been added (for bug fixes/features)

Addresses but does not closes this issue: https://github.com/ossf/scorecard/issues/1578
-->

the `release-note`

```release-note
Run Scorecard on packages hosted at Nuget.org using --npm. For instance: --nuget=Newtonsoft.Json
```


cmd/package_managers.go
```diff
@@ -138,3 +176,60 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+	}
+	defer respPackageIndex.Body.Close()
+	packageIndexResults := &nugetpackageIndexResults{}
+	err = json.NewDecoder(respPackageIndex.Body).Decode(packageIndexResults)
+
+	if err != nil {
+		return "", sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("failed to parse nuget package index json: %v", err))
+	}
+
+	lastVersion := packageIndexResults.Versions[len(packageIndexResults.Versions)-1]
```
joelverhagen, line 214: The ordering of versions in the index.json document here is not well defined (note there is not reference ordering in the protocol docs: https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource#enumerate-package-versions). NuGet.org may put them in SemVer order today (and will perhaps always do this to avoid unnecessary breaking changes) but it's not something that should be depended on if there's a reasonable alternative. Also, the latest (or any version in this document) version may be unlisted. In NuGet. the standard definition of "latest" is "highest SemVer version that is not unlisted from search and is available (not deleted)". 

I don't know there full intent here, but it may be necessary to go to the package metadata endpoint so you can filter out latest unlisted versions.

balteravishay: @joelverhagen thanks for the review and correcting the ordering of versions proposed at the first iteration.
please review the new implementation and the tests that show how this works with the nuget protocol.
leaving this un-resolved until your review.

balteravishay:
I'm not sure how to get past the codecov check. 
currently it seems as if only code in flags.g and root.go is not covered. 
any advice?

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("failed to parse nuget index json: %v", err))
+	}
+	nugetPackageBaseURL, err := getFieldFromIndexResults(nugetIndexResults.Resources,
+		"PackageBaseAddress/3.0.0")
+	if err != nil {
+		return "", err
+	}
+	nugetRegistrationBaseURL, err := getFieldFromIndexResults(nugetIndexResults.Resources,
+		"RegistrationsBaseUrl/3.4.0")
```
joelverhagen, line 227:
```suggestion
		"RegistrationsBaseUrl/3.6.0")
```

You'll very likely want to include SemVer 2.0.0 packages. All the main NuGet experiences today include this.

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+		}
+	}
+	return "", sce.WithMessage(sce.ErrScorecardInternal,
+		"failed to get a listed version for package")
+}
+
+func getFieldFromIndexResults(resources []nugetIndexResult, resultType string) (string, error) {
+	packageBaseAddressIndex := slices.IndexFunc(resources,
+		func(n nugetIndexResult) bool { return n.Type == resultType })
+	if packageBaseAddressIndex == -1 {
```
joelverhagen, line 310:
```suggestion
	if resourceIndex == -1 {
```

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("failed to find %v URI at nuget index json", resultType))
+	}
+
+	return resources[packageBaseAddressIndex].ID, nil
```
joelverhagen, line 315:
```suggestion
	return resources[resourceIndex].ID, nil
```

This appears to be a generic resource index, not specific to package base address

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
 	}
 	return v.SourceCodeURI, nil
 }
+
+// Gets the GitHub repository URL for the nuget package.
+func fetchGitRepositoryFromNuget(packageName string, manager packageManagerClient) (string, error) {
```
joelverhagen, line 207: Is `packageName` guaranteed to be lowercase at this point? The URLs always contain lowercase ID and version strings for the V3 API.

joelverhagen: Lowercase is done via .NET's `ToLowerInvariant()`. It might be a nightmare trying to get the same lowercase behavior in Go. I've haven't studies lowercase folding behavior in Go at all but it's surprisingly nuanced in .NET.

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+			defer respPage.Body.Close()
+			err = json.NewDecoder(respPage.Body).Decode(&page)
+
+			if err != nil {
+				return "", sce.WithMessage(sce.ErrScorecardInternal,
+					fmt.Sprintf("failed to parse nuget package registration page: %v", err))
+			}
+		}
+		for packageIndex := len(page.Packages) - 1; packageIndex >= 0; packageIndex-- {
+			if page.Packages[packageIndex].Entry.Listed {
```
joelverhagen, line 298: This may be a prerelease version. Is this desired? Here's some examples:
https://api.nuget.org/v3/registration5-gz-semver2/basetestpackage.searchfilters/index.json - latest version is prerelease

Also, there is something in SemVer 2.0.0 that is called metadata. It's the part after a `+` sign in the version which is used for informational purposes and not comparison purposes. Example:
https://api.nuget.org/v3/registration5-gz-semver2/serilog.exceptions/index.json
The latest version is `8.4.0+build.694`
The `+build.694` part is ignored for comparison purposes (per SemVer 2.0.0 spec) and is therefore not included when building URLs.

Does Go have a SemVer parsing library? You may able to get away with these tricks:
- First: If there is a `+` in the version string, remove it and everything after it. This takes a "full version string" in NuGet parlance to become a "normalized version version string".
- Second: If there is a `-` in the version string, it is a prerelease string.
- Third: If you are building a URL with this version, lowercase the string first. Thankfully a naive lowercase method will be fine here since package versions must only contain ASCII and all language I've seen have consistent lowercase foldering for ASCII characters.

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+	}
+
+	lastPackageVersion, err := getLatestListedVersion(nugetRegistrationBaseURL,
+		packageName, manager)
+	if err != nil {
+		return "", err
+	}
+
+	respPackageSpec, err := manager.Get(
+		nugetPackageBaseURL+"%[1]v/"+lastPackageVersion+"/%[1]v.nuspec", packageName)
```
joelverhagen, line 239: Similar comment from above -- this should lowercase the version when building the URL.

Example package with uppercase version (only matters if you want to include prerelease versions -- stable versions are only dot and digits).
https://api.nuget.org/v3/registration5-gz-semver2/linqtoexcel/index.json

Note that some of our CDN providers accept uppercase IDs or versions. This is an unfortunate inconsistency we have since we have multiple CDN providers, but clients should assume the ID and version must be lowercase.

cmd/package_managers.go
```diff
@@ -77,6 +87,60 @@ type rubyGemsSearchResults struct {
...
+	return nil
+}
+
+type nugetNuspec struct {
+	XMLName  xml.Name       `xml:"package"`
+	Metadata nuspecMetadata `xml:"metadata"`
+}
+
+type nuspecMetadata struct {
+	XMLName    xml.Name         `xml:"metadata"`
```
joelverhagen, line 135: Wow, I like this JSON and xml mapping syntax! It's very clean!

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("failed to parse nuget nuspec xml: %v", err))
+	}
+	if packageSpecResults.Metadata == (nuspecMetadata{}) ||
+		packageSpecResults.Metadata.Repository == (nuspecRepository{}) ||
+		packageSpecResults.Metadata.Repository.URL == "" {
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("source repo is not defined for nuget package %v", packageName))
+	}
+	return packageSpecResults.Metadata.Repository.URL, nil
```
joelverhagen, line 259: Note that NuGet.org does not validate this repository URL. Theoretically this could be anything. New package X could point to any repo. It need not be a matching repo or even GitHub.com. This is a theoretical problem, not one we've seen cause pain in practice. I just wanted to raise awareness but, given my limited understanding of the goals here, it's probably fine to just log the URL to the user so they can know what's being validated against and assume the URL is good.

joelverhagen (changes requested):
IMO the only blocking issues are lowercase handling.

The rest is nitpicking or non-blocking issues (IMHO).

cmd/package_managers.go
joelverhagen: Examples: `CodeProject.ObjectPool` uses GitLab.

joelverhagen: Also, in .NET, it's common practice to have multiple packages coming from one repo (mono-repo style). Example: 17 packages (top ID = OpenMod.NuGet) have this shared repository URL: https://github.com/openmod/openmod

joelverhagen: I tried to find obvious mismatch examples (where repo owner != package owner) with some hueristics. There's only one case I've found so it indeed doesn't seem like a big problem right now.

This repo URL has the following associated package IDs:
https://github.com/emgarten/Sleet

DotNet.SleetLib - by some other person, linking to the original repo. Probably a fork without properly updated metadata.
SleetLib - by the main owner, a mono-repo
Sleet - by the main owner, a mono-repo
SleetExe - by the main owner, a mono-repo

joelverhagen, line 259: Do you think it's reasonable to fall back to the project URL metadata if there is no repository URL and the repository URL is GitHub?

Example package: `Stateless`, 10 M downloads, no repo URL, but its project URL is a GitHub repo.

cmd/package_managers.go
balteravishay: fixed

cmd/package_managers.go
balteravishay: fixed

cmd/package_managers.go
balteravishay: gitlab would work as well, scorecard supports that. 
as for repository URL not matching the origin of the package, this seems like a general question for the package managers ecosystem, which we can bring and collaborate with the OSSF "Security Software Repositories"  WG.

cmd/package_managers.go
```diff
@@ -138,3 +206,134 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+	if len(strings.TrimSpace(packageSpecResults.Metadata.ProjectURL)) != 0 &&
+		isSupportedProjectURL(packageSpecResults.Metadata.ProjectURL) {
+		return packageSpecResults.Metadata.ProjectURL, nil
+	}
+	return "", sce.WithMessage(sce.ErrScorecardInternal,
+		fmt.Sprintf("source repo is not defined for nuget package %v", packageName))
+}
+
+func isSupportedProjectURL(projectURL string) bool {
+	pattern := `^(?:https?://)?(?:www\.)?(?:github|gitlab)\.com/([A-Za-z0-9_\.-]+)/([A-Za-z0-9_\.-]+)$`
```
joelverhagen, line 272: nit: trailing slash should probably be acceptable. This URL appears to work fine:
https://github.com/ossf/scorecard/

Example: https://www.nuget.org/packages/awssdk.core has `https://github.com/aws/aws-sdk-net/`

cmd/package_managers.go
```diff
@@ -138,3 +206,134 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+			defer respPage.Body.Close()
+			err = json.NewDecoder(respPage.Body).Decode(&page)
+
+			if err != nil {
+				return "", sce.WithMessage(sce.ErrScorecardInternal,
+					fmt.Sprintf("failed to parse nuget package registration page: %v", err))
+			}
+		}
+		for packageIndex := len(page.Packages) - 1; packageIndex >= 0; packageIndex-- {
+			semVersion, err := semver.NewVersion(page.Packages[packageIndex].Entry.Version)
```
joelverhagen, line 315: I don't know the behavior of `semver` on a package version with 4 digits. NuGet diverged slightly from SemVer to allow compat with a popular .NET versioning scheme (System.Version) that has 4 digits. 

Example: Google.Apis.Storage.v1 has this version `1.60.0.2981`.

It's a valid version to NuGet but not a valid string per the strict SemVer spec. Here's an interesting doc we have about how NuGet and SemVer diverge in a couple nuanced places:
https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#where-nugetversion-diverges-from-semantic-versioning

cmd/package_managers.go
```diff
@@ -138,3 +206,134 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+			if err != nil {
+				return "", sce.WithMessage(sce.ErrScorecardInternal,
+					fmt.Sprintf("failed to parse nuget version as semver: %v", err))
+			}
+			// skipping non listed and pre-releases
+			if page.Packages[packageIndex].Entry.Listed && len(strings.TrimSpace(semVersion.Prerelease())) == 0 {
+				return fmt.Sprintf("%[1]d.%[2]d.%[3]d", semVersion.Major(), semVersion.Minor(), semVersion.Patch()), nil
```
joelverhagen, line 322: this string building is to avoid SemVer metadata `+foo` right? (instead of using `page.Packages[packageIndex].Entry.Version` directly)

joelverhagen (approved):
Nits about 4th digit in the version and the regex not accepting trailing slashes. But looks good otherwise!

joelverhagen (approved):
Nits about 4th digit in the version and the regex not accepting trailing slashes. But looks good otherwise!

cmd/package_managers.go
balteravishay: that's right!

cmd/package_managers.go
balteravishay: this one actually broke the golang semver. switched to string splitting.

cmd/package_managers.go
balteravishay: changed to splitting strings

balteravishay:
A reminder to start the review process for this PR :)
cc. @azeemsgoogle @laurentsimon @naveensrinivasan

laurentsimon (approved):
Thanks. LGTM. Can we separate the logic into a nuget.go?

@spencerschrock additional feedback on this PR?

cmd/package_managers.go
```diff
@@ -77,6 +89,61 @@ type rubyGemsSearchResults struct {
...
+}
+
+type nugetPackageRegistrationCatalogEntry struct {
+	Version string `json:"version"`
+	Listed  bool   `json:"listed"`
+}
+
+func (entry *nugetPackageRegistrationCatalogEntry) UnmarshalJSON(text []byte) error {
+	type Alias nugetPackageRegistrationCatalogEntry
+	aux := Alias{
```
spencerschrock, line 121: Is the type alias needed?

cmd/package_managers.go
spencerschrock: Seeing as the functionality impacts only users using `--nuget` I'm fine with taking the simple approach for now, and someone filing an issue later if it's an issue.

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+func fetchGitRepositoryFromNuget(packageName string, manager packageManagerClient) (string, error) {
+	nugetIndexURL := "https://api.nuget.org/v3/index.json"
+	respIndex, err := manager.GetURI(nugetIndexURL)
+	if err != nil {
+		return "", sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("failed to get nuget index json: %v", err))
+	}
+	defer respIndex.Body.Close()
+	nugetIndexResults := &nugetIndexResults{}
+	err = json.NewDecoder(respIndex.Body).Decode(nugetIndexResults)
+
```
spencerschrock, line 219: minor nit, don't like the linebreak between the err and the err check

cmd/package_managers.go
```diff
@@ -77,6 +89,61 @@ type rubyGemsSearchResults struct {
...
+type nugetPackageRegistrationPackage struct {
+	Entry nugetPackageRegistrationCatalogEntry `json:"catalogEntry"`
+}
+
+type nugetPackageRegistrationCatalogEntry struct {
+	Version string `json:"version"`
+	Listed  bool   `json:"listed"`
+}
+
+func (entry *nugetPackageRegistrationCatalogEntry) UnmarshalJSON(text []byte) error {
```
spencerschrock, line 119: minor nit: receivers are normally 1 or 2 letters.

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+func getNugetSemVer(versionString string) (string, string) {
+	metadataAndVersion := strings.Split(versionString, "+")
+	prereleaseAndVersions := strings.Split(metadataAndVersion[0], "-")
+	if len(prereleaseAndVersions) == 1 {
+		return prereleaseAndVersions[0], ""
+	}
+	return prereleaseAndVersions[0], prereleaseAndVersions[1]
+}
+
+func getFieldFromIndexResults(resources []nugetIndexResult, resultType string) (string, error) {
```
spencerschrock, line 340: can this just be a method on the `nugetIndexResults` type?

maybe something like
```
func (n nugetIndexResults) Field(name string) (string, error) {
...
}
```

So we could just call:
```
nugetRegistrationBaseURL, err := nugetIndexResults.Resources.Field("RegistrationsBaseUrl/3.6.0")
```

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+	if err != nil {
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("failed to get nuget package spec json: %v", err))
+	}
+	defer respPackageSpec.Body.Close()
+
+	packageSpecResults := &nugetNuspec{}
+	err = xml.NewDecoder(respPackageSpec.Body).Decode(packageSpecResults)
+
+	if err != nil || packageSpecResults.Metadata == (nuspecMetadata{}) {
```
spencerschrock, line 253: optional, no parens (and other occurrences)

spencerschrock: also, err can be nil here, which can make the error message below confusing if the problem is really an empty response. it may be worth breaking into 2 checks

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("failed to parse nuget nuspec xml: %v", err))
+	}
+	projectURL := ""
```
spencerschrock, line 257: `var projectURL string` is the preferred initialization for zero values

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+	if packageSpecResults.Metadata.Repository != (nuspecRepository{}) &&
+		len(strings.TrimSpace(packageSpecResults.Metadata.Repository.URL)) != 0 {
```
spencerschrock, line 259: Can both conditions can just be simplified to :

```
if strings.TrimSpace(packageSpecResults.Metadata.Repository.URL) != ""
```
?

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+		projectURL = packageSpecResults.Metadata.Repository.URL
+	}
+	if len(strings.TrimSpace(packageSpecResults.Metadata.ProjectURL)) != 0 {
```
spencerschrock, line 262: same comment about just checking against "" after the trim

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+			fmt.Sprintf("failed to get nuget package registration index json: %v", err))
+	}
+	defer resPackageRegistrationIndex.Body.Close()
+	packageRegistrationCatalogRoot := &nugetPackageRegistrationCatalogRoot{}
+	err = json.NewDecoder(resPackageRegistrationIndex.Body).Decode(packageRegistrationCatalogRoot)
+	if err != nil {
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("failed to parse package registration index json: %v", err))
+	}
+	return getLatestListedVersionFromPackageRegistrationPages(packageRegistrationCatalogRoot.Pages, manager)
```
spencerschrock, line 293: same comment about being able to define this as a method on `nugetPackageRegistrationCatalogRoot`

Maybe something like ?
```
func (n nugetPackageRegistrationCatalogRoot) LatestVersion( manager packageManagerClient)

File: cmd/package_managers.go [hunk 22]
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+	}
+	return "", sce.WithMessage(sce.ErrScorecardInternal,
+		"failed to get a listed version for package")
+}
+
+// Nuget semver diverges from Semantic Versioning.
+// This method returns the Nuget represntation of version and pre release strings.
+// nolint
+// more info: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#where-nugetversion-diverges-from-semantic-versioning
+func getNugetSemVer(versionString string) (string, string) {
```
spencerschrock, line 331: does something like this work, to better explain what the two return values are?
```
func getNugetSemVer(version string) (base string, suffix string)
```

cmd/package_managers.go
```diff
@@ -138,3 +205,145 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+			}
+		}
+	}
+	return "", sce.WithMessage(sce.ErrScorecardInternal,
+		"failed to get a listed version for package")
+}
+
+// Nuget semver diverges from Semantic Versioning.
+// This method returns the Nuget represntation of version and pre release strings.
+// nolint
```
spencerschrock, line 329: can you be specific about which linter you're ignoring and why?

```
//nolint:lintername // some comment why
```

cmd/package_managers_test.go
```diff
@@ -706,3 +709,2268 @@ func Test_fetchGitRepositoryFromRubyGems(t *testing.T) {
...
+	args    nugetTestArgs
+	wantErr bool
+}
+
+func Test_fetchGitRepositoryFromNuget(t *testing.T) {
+	t.Parallel()
+
+	tests := []nugetTest{
+		{
+			name: "fetchGitRepositoryFromNuget_find_latest_version_in_single_page",
```
spencerschrock, line 738: names only need to include the subtest. and spaces are fine.

go automatically adds the base func when you do
```
t.Run(tt.name
```

so the name would be reported as:

Test_fetchGitRepositoryFromNuget.fetchGitRepositoryFromNuget_find_latest_version_in_single_page

(or something like that)

cmd/package_managers_test.go
```diff
@@ -706,3 +709,2268 @@ func Test_fetchGitRepositoryFromRubyGems(t *testing.T) {
...
+		if test.wantErr && (test.args.resultIndex == "") {
+			//nolint
+			return nil, errors.New("error")
+		}
+
+		return &http.Response{
+			StatusCode: 200,
+			Body:       io.NopCloser(bytes.NewBufferString(test.args.resultIndex)),
+		}, nil
+	} else {
```
spencerschrock, line 2930: probably doesn't need to be an else block if everything in the `if` statement causes a `return`

cmd/packagemanager_client_test.go
```diff
@@ -0,0 +1,128 @@
...
+				// nolint
+				w.WriteHeader(http.StatusOK)
+				// nolint
+				w.Write([]byte(tt.wantResponse))
+			}))
+			defer server.Close()
+			client := packageManager{}
+			got, err := client.GetURI(server.URL + "/" + tt.args.inputURL)
+			if err != nil {
+				log.Fatalln(err)
```
spencerschrock, line 64: lets avoid `log.Fatal` functions. Did you mean `t. Fatal` or `t. Fatalf`?

cmd/packagemanager_client_test.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+			body, err := io.ReadAll(got.Body)
+			if err != nil {
+				log.Fatalln(err)
```
spencerschrock, line 68: same comment

cmd/packagemanager_client_test.go
```diff
@@ -0,0 +1,128 @@
...
+			}))
+			defer server.Close()
+			client := packageManager{}
+			got, err := client.Get(server.URL+"/"+tt.args.inputURL, tt.args.packageName)
+			if err != nil {
+				log.Fatalln(err)
+			}
+			body, err := io.ReadAll(got.Body)
+			if err != nil {
+				log.Fatalln(err)
```
spencerschrock, line 121: same comments

spencerschrock (commented):
Left some style comments. I am not familiar with nuget, so am deferring to others reviews on that

cmd/package_managers.go
balteravishay: changed, thanks

cmd/package_managers.go
balteravishay: it is, otherwise the UnmarshalJSON method goes into infinite loop.
see here: https://stackoverflow.com/questions/30445479/how-to-specify-default-values-when-parsing-json-in-go 
and here: https://stackoverflow.com/questions/59190078/how-to-set-a-default-value-when-unmarshal-a-json-to-struct

cmd/package_managers.go
balteravishay: fixed

cmd/package_managers.go
balteravishay: fixed the second comment. not sure what you mean by no parens in this case. since this doesn't compile:
packageSpecResults.Metadata == nuspecMetadata{} {
...
}

cmd/package_managers.go
balteravishay: fixed

cmd/package_managers.go
balteravishay: fixed

cmd/package_managers.go
balteravishay: fixed

cmd/package_managers.go
balteravishay: makes sense. fixed

cmd/package_managers.go
balteravishay: sure, fixed

cmd/package_managers.go
balteravishay: fixed

cmd/package_managers.go
balteravishay: done

cmd/package_managers.go
```diff
@@ -138,3 +202,115 @@ func fetchGitRepositoryFromRubyGems(packageName string, manager packageManagerCl
...
+	respPackageSpec, err := manager.Get(
+		nugetPackageBaseURL+"%[1]v/"+lastPackageVersion+"/%[1]v.nuspec", packageName)
+	if err != nil {
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("failed to get nuget package spec json: %v", err))
+	}
+	defer respPackageSpec.Body.Close()
+
+	packageSpecResults := &nugetNuspec{}
+	err = xml.NewDecoder(respPackageSpec.Body).Decode(packageSpecResults)
```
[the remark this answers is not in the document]
balteravishay: @spencerschrock do you think this is relevant?

cmd/package_managers_test.go
balteravishay: fixed

cmd/package_managers_test.go
balteravishay: fixed

cmd/packagemanager_client_test.go
balteravishay: fixed all

cmd/package_managers.go
spencerschrock: Yes, you're right. ignore that.

spencerschrock (commented):
I did try to spot-check some nuget repos from https://www.nuget.org/packages (e.g. `go run main.go --nuget AWSSDK.Core`), and some seem to be not working. Is this a matter of missing data for the packages? Or alternative ways of searching for the link to the source repo? All of them seem to have `Source repository` links on the web UI, but perhaps not in the API page we're checking:

`error: source repo is not defined for nuget package <packagename>`.
* https://www.nuget.org/packages/Newtonsoft.Json
* https://www.nuget.org/packages/Serilog/3.0.0-dev-02022
* https://www.nuget.org/packages/AutoMapper

Sometimes repos link to a specific file instead of the top-level link (`github.com/org/repo`). the logic on this can differ on this between github and gitlab sadly.
`repo unreachable: GET https://api.github.com/repos/Azure/azure-sdk-for-net/blob/Azure.Core_1.32.0/sdk/core/Azure.Core/README.md`. 
* https://api.nuget.org/v3/registration5-semver1/azure.core/index.json#page/1


Other than that, I have some broad style comments, but not going to block on any more as it can always be refactored later, but wanted to mention it as a resource.
*`package_managers_test.go` has a huge diff. Which I assume comes from the examples you unmarshall for your tests. We tend to place those in a `testdata` folder as files, and read the contents during the tests. It helps separate the test logic from the data.
* getters with `get` in the name: https://go.dev/doc/effective_go#Getters
* a lot of types starting with `nuget`, which to me means some of this might better be in it a `nuget` package somewhere (maybe `internal/packagemanager/nuget`?).

cmd/package_managers.go
balteravishay: @spencerschrock @laurentsimon any thoughts regarding this comment or can I resolve that?

cmd/package_managers.go
balteravishay: fixed and tested

balteravishay:
Thanks for the thorough review @spencerschrock ! I sure appreciate it !
I fixed your comments, moving both packagemanager and nuget clients under "clients" folder. that added some changes to the makefile and the initialization, which I hope works for you.

Makefile
```diff
@@ -152,9 +153,12 @@ clients/mockclients/cii_client.go: clients/cii_client.go | $(MOCKGEN)
 checks/mockclients/vulnerabilities.go: clients/vulnerabilities.go | $(MOCKGEN)
 	# Generating MockCIIClient
 	$(MOCKGEN) -source=clients/vulnerabilities.go -destination=clients/mockclients/vulnerabilities.go -package=mockrepo -copyright_file=clients/mockclients/license.txt
-cmd/packagemanager_mockclient.go: cmd/packagemanager_client.go | $(MOCKGEN)
+clients/packagemanager/packagemanager_mockclient.go: clients/packagemanager/client.go | $(MOCKGEN)
```
spencerschrock, line 156: I'm seeing changes if I force a re-generation. We should check in the auto generated version. 

```
touch clients/packagemanager/client.go
make generate-mocks
```

Makefile
```diff
@@ -139,7 +139,8 @@ generate-mocks: clients/mockclients/repo_client.go \
 	clients/mockclients/repo.go \
 	clients/mockclients/cii_client.go \
 	checks/mockclients/vulnerabilities.go \
-	cmd/packagemanager_mockclient.go
+	clients/packagemanager/packagemanager_mockclient.go \
+	clients/nuget/nuget_mockclient.go
```
spencerschrock, line 143: Do you mind if we keep these under an internal folder?

spencerschrock (approved):
Thanks for addressing all the feedback! I know it meant a lot of re-writing, but I think there was a huge readability benefit from it.

Only a small thing. Can we keep these two new packages private by moving them under an `internal` folder (and update the `Makefile` rules)? The functionality was previously unexported inside of `cmd`. Now that they're in their own packages we had to exported some functionality, but we can still stop people from relying on it.

I think either option works fine:
`internal/packagemanager`
`internal/nuget`

or 

`cmd/internal/packagemanager`
`cmd/internal/nuget`

https://go.dev/doc/go1.4#internalpackages
https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

cmd/package_managers.go
spencerschrock: > as for repository URL not matching the origin of the package, this seems like a general question for the package managers ecosystem

+1. As long as scorecard makes it clear which repo is being analyzed (which it does when using `--format json`, but not the default format), I don't see this as a scorecard problem.

Things like [npm's package provenance](https://github.blog/2023-04-19-introducing-npm-package-provenance/) could help, but that's outside of the scope of this PR in my opinion.

PR #3027, opened 2023-05-17
🐛 Gitlab: test fixes

raghavkaul:
Fix e2e test failures in gitlab client and make e2e pat tests required


spencerschrock (approved):
Confirmed with a local `make e2e-gitlab-token`.

Just to confirm, the increase in score for citests and code review is from fixing the typo?

PR #3048, opened 2023-05-23
✨ [experimental] Support for new `--format probe`

laurentsimon:
Based upon https://github.com/ossf/scorecard/pull/2944 which needs to be merged first.

```release-note
Support for new `--format probe`
```


pkg/scorecard.go
```diff
@@ -115,11 +139,35 @@ func RunScorecard(ctx context.Context,
...
+	// Run the probes.
+	var findings []finding.Finding
+	// TODO(#3049): only run the probes for checks.
+	// NOTE: We will need separate functions to support:
+	// - `--probes X,Y`
+	// - `--check-definitions-file path/to/config.yml`
+	findings, err = zrunner.Run(&ret.RawResults, probes.All)
+	if err != nil {
+		return ScorecardResult{}, err
+	}
```
spencerschrock, line 170: If the probes have an error this will also affect non-probe users.

pkg/scorecard_result.go
```diff
@@ -45,12 +46,13 @@ type RepoInfo struct {
...
-	Scorecard  ScorecardInfo
-	Checks     []checker.CheckResult
-	RawResults checker.RawResults
-	Metadata   []string
+	Repo         RepoInfo
+	Date         time.Time
+	Scorecard    ScorecardInfo
+	Checks       []checker.CheckResult
+	RawResults   checker.RawResults
+	ProbeResults []finding.Finding
```
spencerschrock, line 54: Should we stick with `Findings` for consistency?

spencerschrock (changes requested):

pkg/json_probe_results.go
```diff
@@ -0,0 +1,59 @@
...
+			Name:   r.Repo.Name,
+			Commit: r.Repo.CommitSHA,
+		},
+		Scorecard: jsonScorecardV2{
+			Version: r.Scorecard.Version,
+			Commit:  r.Scorecard.CommitSHA,
+		},
+		Date:     r.Date.Format("2006-01-02"),
+		Findings: r.ProbeResults,
+	}
```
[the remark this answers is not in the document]
laurentsimon: I did not just because 
- I wanted some feedback o nest vs flat structure (one of your earlier comment). 
- I don't know if this is needed for practical purposes, so I did not want to over-commit to it right away

I added it back. It will show as:
```json
"metadata": {
    "repository.defaultBranch": "main",
    "repository.host": "github.com",
    "repository.name": "ossf/scorecard-action",
    "repository.sha1": "7fe8d54f4acf26efc826129c396173562e2c766c",
    "repository.uri": "github.com/ossf/scorecard-action"
  }
```

Let me know if you'd like me to keep it or not. If not, I'll create a PR for tracking so that we can re-visit

pkg/scorecard.go
```diff
@@ -90,6 +94,21 @@ func RunScorecard(ctx context.Context,
 	ossFuzzRepoClient clients.RepoClient,
 	ciiClient clients.CIIBestPracticesClient,
 	vulnsClient clients.VulnerabilitiesClient,
+) (ScorecardResult, error) {
+	return RunScorecardChecksV5(ctx, repo, commitSHA, commitDepth,
+		checksToRun, repoClient, ossFuzzRepoClient,
+		ciiClient, vulnsClient)
+}
+
+func RunScorecardChecksV5(ctx context.Context,
```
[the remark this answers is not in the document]
laurentsimon: +1, I will not export it for now. I think I'll need this function to run scorecard in "probe mode" vs "check mode". For now I'll make it private.

pkg/scorecard.go
```diff
@@ -102,6 +121,11 @@ func RunScorecard(ctx context.Context,
 	if err != nil || commitSHA == "" {
 		return ScorecardResult{}, err
 	}
+	defaultBranch, err := repoClient.GetDefaultBranchName()
+	if err != nil {
+		return ScorecardResult{}, err
+	}
```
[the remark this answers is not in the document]
laurentsimon: Good catch. I will return an error only if the retuned error is not `clients.ErrUnsupportedFeature`. If the errors is `clients.ErrUnsupportedFeature`, I'll set the defaultBranch to "unknown". It's no ideal because the fix will be incorrect. This has some impact on Scorecard run on PR (which we don't official support). Created https://github.com/ossf/scorecard/issues/3069 for tracking.

pkg/scorecard.go
laurentsimon: Good point. I think we can discard the error. The errors will all be part of the findings so we don't need the returned err, see https://github.com/ossf/scorecard/blob/main/probes/zrunner/runner.go#L34-L45

Ill add a comment to explain why we discard the err.

Let me know if this is enough.

pkg/scorecard_result.go
laurentsimon: Done.

checker/raw_result.go
```diff
@@ -43,6 +43,7 @@ type RawResults struct {
 	LicenseResults              LicenseData
 	TokenPermissionsResults     TokenPermissionsData
 	CITestResults               CITestData
+	Metadata                    map[string]string
```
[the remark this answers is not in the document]
spencerschrock: I just meant sticking it in a struct like all the other fields in `RawResults`. So we can add more to it later in a non-breaking way if we had to. Do you ever envision any metadata that would be non-string?

```
// If you have a better name use it haha.
type MetadataData {
	Metadata map[string]string
}
```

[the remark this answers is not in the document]
spencerschrock: Although stepping back, does this metadata really fit with RawResults?

There's already `Metadata []string` in the `ScorecardResult`struct, which is a bit different since it's user supplied.
```
type ScorecardResult struct {
	Repo       RepoInfo
	Date       time.Time
	Scorecard  ScorecardInfo
	Checks     []checker.CheckResult
	RawResults checker.RawResults
	Metadata   []string
}
```

pkg/json_probe_results.go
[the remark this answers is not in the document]
spencerschrock: If you don't want to over-commit right now I would remove the `Metadata` field from `JSONScorecardProbeResult`

pkg/scorecard.go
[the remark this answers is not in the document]
spencerschrock: Is this just in preparation for future argument changes? Does this need to be added this PR?

pkg/scorecard.go
```diff
@@ -115,11 +144,36 @@ func RunScorecard(ctx context.Context,
...
+	var findings []finding.Finding
+	// TODO(#3049): only run the probes for checks.
+	// NOTE: We will need separate functions to support:
+	// - `--probes X,Y`
+	// - `--check-definitions-file path/to/config.yml
+	// NOTE: we discard the returned error because the errors are
+	// already cotained in the findings and we want to return the findings
+	// to users.
+	// See https://github.com/ossf/scorecard/blob/main/probes/zrunner/runner.go#L34-L45.
+	findings, _ = zrunner.Run(&ret.RawResults, probes.All)
```
spencerschrock, line 176: linter is failing you here for ignoring err, which you should be checking. 

My original feedback meant more along the lines of dont run the probes if you're not on experimental

checker/raw_result.go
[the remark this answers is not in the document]
spencerschrock: Sorry to nitpick, just trying to plan in a way that avoids potential breaking changes in the future to the top level struct

spencerschrock (commented):
Once we answer some of these questions, it would be good to get some tests written

checker/raw_result.go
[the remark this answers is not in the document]
laurentsimon: no that's fine. I could make it a `Metadata map[string]any`? I would like to keep all values strings in practice, but defining it as a map to `any` would give us flexibility if we want to. Wdut?

pkg/json_probe_results.go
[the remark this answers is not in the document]
laurentsimon: ok, rolling back then.

pkg/scorecard.go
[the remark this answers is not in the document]
laurentsimon: does not have to, but it's private now.

checker/raw_result.go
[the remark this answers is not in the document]
spencerschrock: I still think putting it in a struct gives us the most flexibility to change things later:
```go
// If you have a better name use it haha.
type MetadataData {
	Metadata map[string]string
}
```

pkg/json_probe_results.go
```diff
@@ -0,0 +1,58 @@
...
+
+// JSONScorecardProbeResult exports results as JSON for flat findings without checks.
+//
+//nolint:govet
+type JSONScorecardProbeResult struct {
+	Date      string            `json:"date"`
+	Repo      jsonRepoV2        `json:"repo"`
+	Scorecard jsonScorecardV2   `json:"scorecard"`
+	Findings  []finding.Finding `json:"findings"`
+	Metadata  map[string]any    `json:"metadata"`
```
spencerschrock, line 34: if we're not populating the `Metadata` field yet, let's drop it from the struct. Can always add it later.

pkg/scorecard.go
```diff
@@ -100,8 +121,18 @@ func RunScorecard(ctx context.Context,
 
 	commitSHA, err := getRepoCommitHash(repoClient)
 	if err != nil || commitSHA == "" {
+		//nolint:wrapcheck
 		return ScorecardResult{}, err
 	}
+	defaultBranch, err := repoClient.GetDefaultBranchName()
+	if err != nil {
+		if !errors.Is(err, clients.ErrUnsupportedFeature) {
+			//nolint:wrapcheck
```
spencerschrock, line 130: can you wrap instead of ignoring?

pkg/scorecard.go
```diff
@@ -100,8 +121,18 @@ func RunScorecard(ctx context.Context,
 
 	commitSHA, err := getRepoCommitHash(repoClient)
 	if err != nil || commitSHA == "" {
+		//nolint:wrapcheck
```
spencerschrock, line 124: my understanding was the wrapcheck only complained about external errors?

pkg/scorecard.go
```diff
@@ -90,6 +96,21 @@ func RunScorecard(ctx context.Context,
 	ossFuzzRepoClient clients.RepoClient,
 	ciiClient clients.CIIBestPracticesClient,
 	vulnsClient clients.VulnerabilitiesClient,
+) (ScorecardResult, error) {
+	return runScorecardChecksV5(ctx, repo, commitSHA, commitDepth,
+		checksToRun, repoClient, ossFuzzRepoClient,
+		ciiClient, vulnsClient)
+}
+
+func runScorecardChecksV5(ctx context.Context,
```
spencerschrock, line 105: this still seems like an unneeded change to me, but not going to block on it.

pkg/json_probe_results.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// TODO: finsinds should enventually be part of the scorecard structure.
+func (r *ScorecardResult) AsPJSON(writer io.Writer) error {
```
spencerschrock, line 38: please add tests

pkg/json_probe_results.go
laurentsimon: done, thanks.

pkg/scorecard.go
laurentsimon: done.

pkg/scorecard.go
laurentsimon: done

pkg/scorecard.go
laurentsimon: removed.

pkg/json_probe_results.go
laurentsimon: created https://github.com/ossf/scorecard/issues/3110 for tracking. Can you comment on the issue what kind of test would be useful? (This PJSON function just prints to an io)

spencerschrock (approved):
Thanks. There's a linter warning about lll

PR #3055, opened 2023-05-23
:seedling: Refactor checks to use interface and update tests

naveensrinivasan:
- Refactor BinaryArtifacts check to use an interface
- Update tests to reflect the new BinaryArtifactsCheck interface
- Replace `checks.BinaryArtifacts` calls with `b.RunCheck` calls
- Add a `RunCheck` function to the `Check` interface

Eventually, move all of them to the same interface. This is the first PR.

the `release-note`

```release-note

```


checks/binary_artifact.go
```diff
@@ -24,19 +24,22 @@ import (
...
-	if err := registerCheck(CheckBinaryArtifacts, BinaryArtifacts, supportedRequestTypes); err != nil {
+	b := BinaryArtifactsCheck{}
+	if err := registerCheckInterface(CheckBinaryArtifacts, b, supportedRequestTypes); err != nil {
 		// this should never happen
 		panic(err)
 	}
 }
 
-// BinaryArtifacts  will check the repository contains binary artifacts.
-func BinaryArtifacts(c *checker.CheckRequest) checker.CheckResult {
```
spencerschrock, line 39: Can we do this in a way that doesn't remove this signature (or at least yet)? I know Allstar calls the BinaryArtifacts function, and wouldn't be surprised if others do as well.

Then when we bump the major version we can make the change?

checks/checks.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checks
+
+import "github.com/ossf/scorecard/v4/checker"
+
+type Check interface {
+	RunCheck(c *checker.CheckRequest) checker.CheckResult
```
spencerschrock, line 20: Thoughts on calling it `Run`?

spencerschrock (commented):
Can you speak to the motivation about moving checks to an interface?

naveensrinivasan:
> Can you speak to the motivation about moving checks to an interface?

The goal is to have a plugin interface for checks so that we can allow users to run their own checks.

Provide a framework for running custom checks are to be run within the Scorecard.

Then we could use https://github.com/hashicorp/go-plugin; we would need an interface. By enabling this, we could let users write their custom checks and integrate them with the scorecard.

checks/checks.go
naveensrinivasan: I agree. We could do that.

checks/binary_artifact.go
naveensrinivasan: Ok, we can do that.

naveensrinivasan: 👍

checks/binary_artifact.go
```diff
@@ -24,18 +24,39 @@ import (
...
+	if c.RawResults != nil {
+		c.RawResults.BinaryArtifactResults = rawData
+	}
+
+	// Return the score evaluation.
+	return evaluation.BinaryArtifacts(CheckBinaryArtifacts, c.Dlogger, &rawData)
+}
+
 // BinaryArtifacts  will check the repository contains binary artifacts.
+// This is being used by AllStar and that is the reason to keep it.
```
spencerschrock, line 59: Sorry for any confusion on my initial comment, but I don't think we need two implementations, just maintaining the function signature

This could just be simplified as 
```
func BinaryArtifacts(c *checker.CheckRequest) checker.CheckResult {
    b := checks.BinaryArtifactsCheck{}
    result := b.Run(&req)
    return result
}
```

attestor/command/check.go
```diff
@@ -84,10 +84,10 @@ func RunCheckWithParams(repoURL, commitSHA, policyPath string) (policy.PolicyRes
 	}
 
 	requiredChecks := attestationPolicy.GetRequiredChecksForPolicy()
-
+	b := checks.BinaryArtifactsCheck{}
 	enabledChecks := map[string]checker.Check{
 		checks.CheckBinaryArtifacts: {
-			Fn: checks.BinaryArtifacts,
+			Fn: b.Run,
```
spencerschrock, line 90: As we're leaving `checks.BinaryArtifacts`, might as well leave it here for readability.

checks/checks.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checks
+
+import "github.com/ossf/scorecard/v4/checker"
+
+type Check interface {
+	Run(c *checker.CheckRequest) checker.CheckResult
+}
```
spencerschrock, line 21: Some other questions that popped up regarding things associated with checks:
* Should supported request types be part of the interface?
* What about `Name`?
* Risk?

spencerschrock (commented):
I think the contents of the PR are in a good state, but I think this is something big enough that there should be some thought put in before any code. At the very least around the interface.

naveensrinivasan:
> I think the contents of the PR are in a good state, but I think this is something big enough that there should be some thought put in before any code. At the very least around the interface.

Do you think we need additional reviewers? Can you be specific?

checks/checks.go
naveensrinivasan: I am maintaining the current check function design. We should review it when we make alterations.

checks/binary_artifact.go
naveensrinivasan: 👍

attestor/command/check.go
naveensrinivasan: 👍

checks/checks.go
naveensrinivasan: 👍

naveensrinivasan:
> > I think the contents of the PR are in a good state, but I think this is something big enough that there should be some thought put in before any code. At the very least around the interface.
> 
> Do you think we need additional reviewers? Can you be specific?

Moved it to an internal package till it is ready for primetime.

spencerschrock:
> > > I think the contents of the PR are in a good state, but I think this is something big enough that there should be some thought put in before any code. At the very least around the interface.
> > 
> > 
> > Do you think we need additional reviewers? Can you be specific?
> 
> Moved it to an internal package till it is ready for primetime.

Should we close this given the approach in #3095?

naveensrinivasan:
> > > > I think the contents of the PR are in a good state, but I think this is something big enough that there should be some thought put in before any code. At the very least around the interface.
> > > 
> > > 
> > > Do you think we need additional reviewers? Can you be specific?
> > 
> > 
> > Moved it to an internal package till it is ready for primetime.
> 
> Should we close this given the approach in #3095?

Agreed. Thanks

PR #3090, opened 2023-05-30
:seedling: Speed up Security-Policy e2e tests

spencerschrock:

e2e test speedup

tensorflow/tensorflow is huge and was slowing down tests because the whole repo was downloaded to scan for a security policy file. 

* The e2e tests now use https://github.com/ossf-tests/scorecard-check-security-policy-e2e, which is a much smaller repo and takes less time to clone.
* Also removed the rust e2e tests because they're already present as unit tests.

- [] Tests for the changes have been added (for bug fixes/features)

the `release-note`

```release-note

```


naveensrinivasan (approved):
Thanks

PR #3096, opened 2023-05-31
:seedling: Improve dependency analysis workflow

naveensrinivasan:
- Add GitHub Action for dependency analysis
- Harden GitHub Action runner
- Run OSSF Scorecard for dependency analysis

[.github/workflows/dependency-analysis.yml]
- Add a GitHub Action for dependency analysis
- Harden the GitHub Action runner
- Checkout code for dependency analysis
- Run the dependency analysis using the OSSF Scorecard

the `release-note`

```release-note

```

### DIFF
--- a/.github/workflows/dependency-analysis.yml
+++ b/.github/workflows/dependency-analysis.yml
@@ -1,0 +1,52 @@
+# Copyright 2023 OpenSSF Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+name: scorecard-dependency-analysis
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+permissions:
+  pull-requests: write
+
+jobs:
+  dependency-analysis:
+    name: Scorecards dependency analysis
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+      GITHUB_ACTOR: ${{ github.actor }}
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@128a63446a954579617e875aaab7d2978154e969 # v2.4.0
+        with:
+          egress-policy: audit
+
+      - name: Checkout code
+        uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2.7.0
```
spencerschrock, line 47: Do we need to checkout anything here? My understanding is `ossf/scorecard-dependencyanalysis` just does API calls?

If we do need it, lets make sure these are up-to-date
`actions/checkout@f095bcc56b7c2baf48f3ac70d6d6782f4f553222 # v3.5.2`

.github/workflows/dependency-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Run analysis
+        uses: ossf/scorecard-dependencyanalysis@dependency-analysis
```
spencerschrock, line 52: I'm assuming it's pinned to the branch while the action is in development?
We should comeback and pin this later.

.github/workflows/dependency-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+permissions:
+  pull-requests: write
```
spencerschrock, line 26: comment on why we need the permission?

spencerschrock (approved):
I assume this check will not be `Required` for now?

.github/workflows/dependency-analysis.yml
naveensrinivasan: We need this permission to write comments on the PR.

spencerschrock: I know, I'm asking for that to be added as a comment in the file

naveensrinivasan: OK, I will do that.

PR #3119, opened 2023-06-02
🌱 CI: Don't run PAT E2E on dependabot merges

raghavkaul:
Don't run GitHub PAT E2E tests on dependabot merges to the `main` branch since those merges don't have access to environment secrets. 

`github.actor` is dependabot when using `dependabot merge`; when the PR approver also merges the PR, they are the github.actor. 

See: https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/


raghavkaul:
This PR is intended to fix build failures like: https://github.com/ossf/scorecard/actions/runs/5147547179/jobs/9268078821

Where the github.actor is `dependabot` and the secret for `GH_AUTH_TOKEN` isn't available in the environement.

naveensrinivasan (approved):

PR #3142, opened 2023-06-08
:seedling: Increase test coverage for finding outcomes

naveensrinivasan:
- Add tests for Outcome UnmarshalYAML function in `finding/finding_test.go`

the `release-note`

```release-note

```

### DIFF
--- a/finding/finding_test.go
+++ b/finding/finding_test.go
@@ -201,18 +201,19 @@ func Test_FromBytes(t *testing.T) {
 }
 
 func TestOutcome_UnmarshalYAML(t *testing.T) {
+	t.Parallel()
 	type args struct {
 		n *yaml.Node
 	}
 	tests := []struct { //nolint:govet
-		name    string
-		o       Outcome
-		args    args
-		wantErr bool
+		name        string
+		wantOutcome Outcome
+		args        args
+		wantErr     bool
 	}{
 		{
-			name: "positive outcome",
-			o:    OutcomePositive,
+			name:        "positive outcome",
+			wantOutcome: OutcomePositive,
 			args: args{
 				n: &yaml.Node{
 					Kind:  yaml.ScalarNode,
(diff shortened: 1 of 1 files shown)


finding/finding_test.go
```diff
@@ -198,3 +199,88 @@ func Test_FromBytes(t *testing.T) {
...
 	}
 }
+
+func TestOutcome_UnmarshalYAML(t *testing.T) {
+	type args struct {
+		n *yaml.Node
+	}
+	tests := []struct { //nolint:govet
+		name    string
+		o       Outcome
```
spencerschrock, line 209: wouldn't `o Outcome` be `want Outcome` in this case?

finding/finding_test.go
```diff
@@ -198,3 +199,88 @@ func Test_FromBytes(t *testing.T) {
...
+			args: args{
+				n: &yaml.Node{
+					Kind:  yaml.ScalarNode,
+					Value: "Unknown",
+				},
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
```
spencerschrock, line 279: Any reason not to parallelize? and therefore include a `tt := tt`?

finding/finding_test.go
```diff
@@ -198,3 +199,88 @@ func Test_FromBytes(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			if err := tt.o.UnmarshalYAML(tt.args.n); (err != nil) != tt.wantErr {
```
spencerschrock, line 281: we don't want to unmarshall until the value we're looking for.
 
I think you need a local var here to unmarshall into, and then a comparison between that value and what we wanted?

right now we're only testing for an error.

spencerschrock (approved):

finding/finding_test.go
naveensrinivasan: Agreed! Fixed it

finding/finding_test.go
naveensrinivasan: 👍

finding/finding_test.go
naveensrinivasan: 👍

PR #3162, opened 2023-06-13
🐛 Forgive job-level permissions

### LINKED ISSUE #3045: BUG: High-risk write permissions only punished at top-level, low-risk permissions punished everywhere
**Describe the bug**
Scorecard currently only punishes the high-risk permissions `contents/packages/actions: write` if they are at the top-level. At the job-level, a warning is shown but no score penalty is applied.

However, all the other lower-risk permissions (i.e. `checks`) apply a score penalty wherever they are.

**Reproduction steps**
1. See https://github.com/ossf/scorecard/blob/main/checks/evaluation/permissions/permissions.go#L325-L374
2. Note that all low-risk permissions are checked with `permissionIsPresent` while high-risk permissions use `permissionIsPresentInTopLevel`.

**Expected behavior**
The current situation is the opposite of what'd be expected: high-risk permissions should be punished more severely than low-risk permissions (or they should be punished equally).

So, I'd expect either:

1. all write permissions are always punished, regardless of where they are defined. However, this goes against the conclusion in #2338.
2. all write permissions should be forgiven at the job-level. This seems to be closest to the spirit of #2338.
3. all permissions have two levels of punishment: a high value if set at the top-level, a low value if set at the job-level. (The values don't have to be the same for all permissions).

(2) seems closest to the spirit of #2338, but I'd also see value in (3), with something like:

| permission | Penalty<br>@ top-level | Penalty<br>@ job-level |
| - | - | - |
| High-risk (contents,<br>packages, actions) | -10 | -1 |
| Low-risk (other) | -1 | 0 |

This would somewhat capture the risk of adding high-risk permissions even if only at the job-level (which could be further reduced to no loss if hash-pinned, as suggested by #3022).

pnacht:

Job-level `write` permissions are no longer penalized.

Currently, high-risk permissions (`contents/packages/actions: write`) are not penalized when set at job-level. However, low-risk permissions (everything else) are penalized wherever they appear (top- or job-level).

As per the discussion in #2338, all permissions are allowed if set at job-level.

Fixes #3045 

the `release-note`

```release-note
Job-level permissions are no longer penalized.
```


spencerschrock (commented):
Thanks for the PR.

1. Can you update the documentation along with this scoring change?
2. Linter is complaining about dead code, so `permissionIsPresent` will need removed.
3. Can we add a test to evaluate "job level write permissions not penalized" or something like that? Something that has:

```
    permissions:
      statuses: write
      checks: write
      security-events: write
      deployments: write
      contents: write
      packages: write
      actions: write
```

pnacht:
Done!

Gby56:
Thanks a lot for this, noticed the bug/noise today, definitely the `checks: write` is a chicken and egg scenario, maybe if the granularity was at a step-level we could consider it, but for the whole job, that's too much.

Maybe this should be brought to Github's attention, the fundamental problem is that the action is reporting itself the check OK or not OK, so we give it the token with the permission, meaning a malicious actor could force all checks to be OK and merge ? Shouldn't the runner or some wrapper around that, check the step/job final output and decide on the check status ?

pnacht:
@spencerschrock, is there anything else I can do on this PR? The failed test doesn't seem related, as far as I can tell.

checks/evaluation/permissions/permissions.go
```diff
@@ -325,21 +325,21 @@ func calculateScore(result map[string]permissions) int {
 		// status: https://docs.github.com/en/rest/reference/repos#statuses.
 		// May allow an attacker to change the result of pre-submit and get a PR merged.
 		// Low risk: -0.5.
-		if permissionIsPresent(perms, "statuses") {
+		if permissionIsPresentInTopLevel(perms, "statuses") {
```
spencerschrock, line 328: Thanks. I can't comment on these lines, so I just picked a random line, but can we clarify the messages here. Something short mentioning least privilege working to match the checks.md file?

https://github.com/ossf/scorecard/blob/066a4f5848cfdf0680879be027f8879b0ea93e22/checks/evaluation/permissions/permissions.go#L57-L64

When I run this PR against `ossf/scorecard` I get the following, even though we have some write permissions.
```
      "score": 10,
      "reason": "tokens are read-only in GitHub workflows",
      "name": "Token-Permissions",
```

spencerschrock (approved):

checks/evaluation/permissions/permissions.go
pnacht: Done, PTAL.

docs/checks/internal/checks.yaml
```diff
@@ -673,23 +673,13 @@ checks:
...
-      * `statuses` - May allow an attacker to change the result of pre-submit checks and get a PR merged.
-      * `checks` - May allow an attacker to remove pre-submit checks and introduce a bug.
-      * `security-events` - May allow an attacker to read vulnerability reports before a patch is available. However, points are not reduced if the job utilizes a recognized action for uploading SARIF results.
-      * `deployments` - May allow an attacker to charge repo owner by triggering VM runs, and tiny chance an attacker can trigger a remote service with code they own if server accepts code/location variables unsanitized.
-
-      ### Write permissions causing a large reduction
-      * `contents` - Allows an attacker to commit unreviewed code. However, points are not reduced if the job utilizes a recognized packaging action or command.
-      * `packages` - Allows an attacker to publish packages. However, points are not reduced if the job utilizes a recognized packaging action or command.
-      * `actions` - May allow an attacker to steal GitHub secrets by approving to run an action that needs approval.
-
```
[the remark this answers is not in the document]
pnacht: Mentioned the warning without penalty along with an explanation. PTAL.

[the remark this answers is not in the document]
laurentsimon: some context: warnings have been used in scorecard to indicate "something that can be remediated"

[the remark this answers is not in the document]
pnacht: > some context: warnings have been used in scorecard to indicate "something that can be remediated"

Not always:
- Pinned-Dependencies has warnings if you don't pin your language dependencies (`go get`, `npm install`, etc), but doesn't penalize the score (https://github.com/ossf/scorecard/pull/2922 recently started scoring `pip install`).
- Token-Permissions already currently throws a warning but doesn't penalize job-level `contents: write` (and a few other high-risk permissions).

docs/checks/internal/checks.yaml
```diff
@@ -670,6 +670,11 @@ checks:
       This configuration is secure, but there is a chance that when a new job is added to the workflow, its job permissions could be
       left undefined because of human error.
 
+      Though a project's score won't be penalized, the check's details will include
+      warnings for all run-level permissions. This compromise makes it clear the
```
spencerschrock, line 674: It's not all run-level permissions though. Just the ones of "interest" that previously led to point deductions. The README explained why they resulted in point deduction, and I still think there's value in that section about why we WARN for them.

https://github.com/ossf/scorecard/blob/ebf967241d5b7d8548fab04399826c17e390be4a/checks/raw/permissions.go#L32-L47

pnacht: Oh, I knew only a few permissions cost points, but thought all permissions got warnings. Fixed, PTAL.

PR #3192, opened 2023-06-19
✨ Enable ResultsFile Switch for cli

### LINKED ISSUE #3191: Feature: Enable ResultsFile switch
Would like to store off the test results into a file.

There's a TODO of "Add logic for writing results to file" that I'd like implemented

I've tee'd output and stdout

**Additional context**
None

jimrobison:
`--resultsfile <string filename>`

feature

Switch has a TODO but hasn't been enabled

#3191 

Enables use of the resultsfile switch that outputs the stdout and stderr to the file specified in the switch `--resultsfile`
the `release-note`

```release-note
`--resultsfile` switch is used to populate a file with the stdout and stderr for the cli operation.
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -16,6 +16,7 @@
 package cmd
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -90,6 +91,21 @@ func rootCmd(o *options.Options) error {
...
+	if o.ResultsFile != "" {
+		outFile, err = os.Create(o.ResultsFile)
+		if err != nil {
+			panic(err)
+		}
+		defer outFile.Close()
+		os.Stdout = outFile
+		os.Stderr = outFile
+
+		fmt.Fprintf(os.Stderr, "Repo Queried: %s\n\n", o.Repo)
```
spencerschrock, line 106: this seems unrelated to the change.

cmd/root.go
```diff
@@ -90,6 +91,21 @@ func rootCmd(o *options.Options) error {
...
+	var outFile *os.File
+
+	if o.ResultsFile != "" {
+		outFile, err = os.Create(o.ResultsFile)
+		if err != nil {
+			panic(err)
+		}
+		defer outFile.Close()
+		os.Stdout = outFile
+		os.Stderr = outFile
```
spencerschrock, line 104: Is there anyway to do this without taking over stdout/stderr?

spencerschrock (commented):
I would expect this to just capture the normal stdout output, not the logging?

Do you have a strong preference on `--resultsfile` vs `--output` ? I think the latter is more standard.

PR #3237, opened 2023-06-30
:seedling: Unit tests  rate limit handling in roundtripper

naveensrinivasan:
- Add rate limit testing and handling functionality
- Add tests for successful response and Retry-After header set scenarios

the `release-note`

```release-note

```

### DIFF
--- a/clients/githubrepo/roundtripper/rate_limit_test.go
+++ b/clients/githubrepo/roundtripper/rate_limit_test.go
@@ -1,0 +1,91 @@
+// Copyright 2023 OpenSSF Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package roundtripper
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/ossf/scorecard/v4/log"
+)
+
+func TestRoundTrip(t *testing.T) {
+	var requestCount int
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Customize the response headers and body based on the test scenario
+		switch r.URL.Path {
+		case "/error":
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte("Internal Server Error")) // nolint: errcheck
+		case "/retry":
+			requestCount++
+			if requestCount == 2 {
+				// Second request: Return successful response
+				w.Header().Set("X-RateLimit-Remaining", "10")
+				w.WriteHeader(http.StatusOK)
+				w.Write([]byte("Success")) // nolint: errcheck
+			} else {
+				// First request: Return Retry-After header
+				w.Header().Set("Retry-After", "5")
+				w.WriteHeader(http.StatusTooManyRequests)
+				w.Write([]byte("Rate Limit Exceeded")) // nolint: errcheck
+			}
+		case "/success":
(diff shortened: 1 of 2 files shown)


clients/githubrepo/roundtripper/rate_limit_test.go
```diff
@@ -0,0 +1,91 @@
...
+				w.WriteHeader(http.StatusTooManyRequests)
+				w.Write([]byte("Rate Limit Exceeded")) // nolint: errcheck
+			}
+		case "/success":
+			w.Header().Set("X-RateLimit-Remaining", "10")
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("Success")) // nolint: errcheck
+		}
+	}))
+	defer ts.Close()
```
spencerschrock, line 51: use `t.Cleanup` to avoid timing issues, as defer will close the server when this function returns. Meanwhile, the various t.Run subtests are still running in another goroutine and might break.

clients/githubrepo/roundtripper/rate_limit_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/ossf/scorecard/v4/log"
+)
+
+func TestRoundTrip(t *testing.T) {
```
spencerschrock, line 24: this (and the subtests should be able to `t.Parallel`)

clients/githubrepo/roundtripper/rate_limit_test.go
```diff
@@ -0,0 +1,91 @@
...
+	var requestCount int
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Customize the response headers and body based on the test scenario
+		switch r.URL.Path {
+		case "/error":
```
spencerschrock, line 29: is this supposed to have a t.Run?

clients/githubrepo/roundtripper/rate_limit_test.go
```diff
@@ -0,0 +1,91 @@
...
+		case "/retry":
+			requestCount++
+			if requestCount == 2 {
+				// Second request: Return successful response
+				w.Header().Set("X-RateLimit-Remaining", "10")
+				w.WriteHeader(http.StatusOK)
+				w.Write([]byte("Success")) // nolint: errcheck
+			} else {
+				// First request: Return Retry-After header
+				w.Header().Set("Retry-After", "5")
```
spencerschrock, line 41: since this incurs a sleep, can we make it 1 to keep the tests as fast as we can?

spencerschrock (approved):
Thanks, another `net/http/httptest` victory :)

A table driven test would reduce some duplication here between the t.Run calls (especially if you add the t.Run for the `/error` endpoint.

Can we rename the PR to clarify this is adding tests. The current title makes me think the roundtripper logic is changing, which it's not.

clients/githubrepo/roundtripper/rate_limit_test.go
spencerschrock: Alternatively, the roundtripper could be (internally) extended in a way to pass a mock sleeper, similar to how "Learn Go with tests" does it [here](https://quii.gitbook.io/learn-go-with-tests/go-fundamentals/mocking#mocking).

`MakeRateLimitedTransport` could return a roundtripper with a sleeper that uses the default `time.Sleep`, so no behavior changes. And the test could use a fake one.

clients/githubrepo/roundtripper/rate_limit_test.go
naveensrinivasan: 👍

clients/githubrepo/roundtripper/rate_limit_test.go
naveensrinivasan: 👍

clients/githubrepo/roundtripper/rate_limit_test.go
naveensrinivasan: For now, I am going to keep it as is. Thanks

clients/githubrepo/roundtripper/rate_limit_test.go
naveensrinivasan: Yes, for later. I ran into a few issues while testing it.

Issue #3250, opened 2023-07-07
BUG: Branch-Protection scoring


gabibguti:
**Describe the bug**
The Branch-Protection score should count 1 point for each setting you enable, of course, considering the Tiers. However, the scoring is wrong for the following cases:

- Using non-admin token, when you reach tier 3 you have 8/10 and should be 7/10.
- Using non-admin token, when you reach tier 3 you have 8/10. When enabling "Require review from code owners" you still get an 8/10.
- Using admin token, when you reach tier 1 you have 4/10 and should be 3/10.
- Using admin token, when you reach tier 2 and enable "Require at least 1 reviewer for approval before merging " you have 5/10. When enabling "For administrators: Require branch to be up to date before merging" you still get a 5/10.
- Using admin token, when you reach tier 3 you have 8/10 and should be 7/10.
- Using admin token, when you reach tier 3 you have 8/10. When enabling "Require at least 2 reviewers for approval before merging" you still get an 8/10.
- Using admin token, when you reach tier 3 you have 8/10. When enabling "Require review from code owners" you still get an 8/10.

**Reproduction steps**

To reproduce, get an non-admin token, enable the branch protection settings until you reach Tier 2, run Scorecard with non-admin token and then do it again until you reach Tier 3 and compare the scores.
Same for admin-token, but comparing the Tiers 1, 2 and 3 scores.

**Expected behavior**

The Branch-Protection score should count 1 point for each setting you enable but considering the Tiers.

**Additional context**
None.

gabibguti:
Spencer confirmed me that in Tier 3 you should gain 2 points for enabling one setting, "Require status checks to pass before merging > Status Checks". So, updating the wrong scoring cases:

- Using admin token, when you reach tier 1 you have 4/10 and should be 3/10.
- Using admin token, when you reach tier 2 and enable "Require at least 1 reviewer for approval before merging " you have 5/10. When enabling "For administrators: Require branch to be up to date before merging" you still get a 5/10.

gabibguti:
The bug about GitHub responding wrongly if "branches are required to be up to date before merge" is already reported https://github.com/orgs/community/discussions/59471

PR #3251, opened 2023-07-07
:bug: Fix Branch-Protection scoring

gabibguti:

Fix Branch-Protection scoring.

We score 1 wrong point verifying if branch is required to be up to date before merge. We retrieve this info from GitHub GraphQL API checking if `RequiresStrictStatusChecks` setting is enabled. However, this setting comes wrong due a bug https://github.com/orgs/community/discussions/59471, messing up the Tiers scoring.

We have a workaround that verifies first if both `RequiresStatusChecks` and `RequiresStatusChecks` settings are enabled. Meaning we verify if the branch requires status checks to merge, to then verify if between those checks it requires branches to be up to date. If the branch does not require status checks to merge, then it cannot require branches to be up to date.

Fixes https://github.com/ossf/scorecard/issues/3250

the `release-note`

```release-note
Fix verification if branch is required to be up to date before merging, impacting Branch-Protection score.
```


clients/githubrepo/branches.go
```diff
@@ -190,7 +190,13 @@ func copyAdminSettings(src *branchProtectionRule, dst *clients.BranchProtectionR
 	copyBoolPtr(src.DismissesStaleReviews, &dst.RequiredPullRequestReviews.DismissStaleReviews)
 	if src.RequiresStatusChecks != nil {
 		copyBoolPtr(src.RequiresStatusChecks, &dst.CheckRules.RequiresStatusChecks)
-		copyBoolPtr(src.RequiresStrictStatusChecks, &dst.CheckRules.UpToDateBeforeMerge)
+		// Workaround for GitHub GraphQL bug https://github.com/orgs/community/discussions/59471
+		// The setting RequiresStrictStatusChecks should tell if the branch is required
+		// to be up to date before merge, but it only returns the correct value if
+		// RequiresStatusChecks is true. If RequiresStatusChecks is false, RequiresStrictStatusChecks
+		// is wrongly retrieved as true.
+		upToDateBeforeMerge := *src.RequiresStatusChecks && *src.RequiresStrictStatusChecks
```
laurentsimon, line 198: is pointer `src.RequiresStrictStatusChecks` guaranteed to be non nil?

clients/githubrepo/branches.go
```diff
@@ -190,7 +190,13 @@ func copyAdminSettings(src *branchProtectionRule, dst *clients.BranchProtectionR
 	copyBoolPtr(src.DismissesStaleReviews, &dst.RequiredPullRequestReviews.DismissStaleReviews)
 	if src.RequiresStatusChecks != nil {
 		copyBoolPtr(src.RequiresStatusChecks, &dst.CheckRules.RequiresStatusChecks)
-		copyBoolPtr(src.RequiresStrictStatusChecks, &dst.CheckRules.UpToDateBeforeMerge)
+		// Workaround for GitHub GraphQL bug https://github.com/orgs/community/discussions/59471
```
laurentsimon, line 193: Can you create a tracking bug on scorecard repo and update the comment `TODO(number): Update when grahpQL big is fixed`

clients/githubrepo/branches.go
gabibguti: Yes, both fields `RequiresStatusChecks` and `RequiresStrictStatusChecks` according to GitHub GraphQL API documentation cannot be nil. Also, in my local tests, with admin token, the fields didn't come as nil.

https://docs.github.com/en/graphql/reference/objects#branchprotectionrule
image

clients/githubrepo/branches.go
gabibguti: Yes!

clients/githubrepo/branches.go
spencerschrock: I think Laurent was talking about `src` being guaranteed to be non-nil, which would be based on logic.

gabibguti: Thanks for the clarification, Spencer! I will take a look.

gabibguti: Would this better guarantee I'm not doing an invalid access to `src` pointer?

```suggestion
		if src.RequiresStrictStatusChecks != nil {
			dst.CheckRules.UpToDateBeforeMerge = new(bool)
			*dst.CheckRules.UpToDateBeforeMerge = *src.RequiresStatusChecks && *src.RequiresStrictStatusChecks
		}
```

In the code we have a father validation of `src.RequiresStatusChecks != nil` and I'm adding a validation if `src.RequiresStrictStatusChecks != nil` to then access both values. I believe this would cover what's done in `copyBoolPtr` but performing the AND operation.

spencerschrock: I personally prefer a mix of what you have now and what you suggested, and I think it resolves @laurentsimon 's question.

```go
if src.RequiresStrictStatusChecks != nil {
    upToDateBeforeMerge := *src.RequiresStatusChecks && *src.RequiresStrictStatusChecks
    copyBoolPtr(&upToDateBeforeMerge, &dst.CheckRules.UpToDateBeforeMerge)
}
```

gabi.log
```diff
@@ -0,0 +1,24 @@
+2023/07/10 12:01:24 RequiresStatusChecks: false
```
spencerschrock, line 1: I don't think you meant to commit this file?

spencerschrock (approved):

gabi.log
gabibguti: You're correct

PR #3295, opened 2023-07-21
:seedling: Improve search commit e2e tests

naveensrinivasan:
- Add 2 tests for searching commits in e2e/searchCommits_test.go
- Fix errors in e2e/searchCommits_test.go when not using HEAD or when user does not exist

[e2e/searchCommits_test.go]
- Add 2 tests for searching commits
- Fix error when not using HEAD
- Fix error when user does not exist

the `release-note`

```release-note

```


e2e/searchCommits_test.go
```diff
@@ -36,5 +36,21 @@ var _ = Describe("E2E TEST:SearchCommits", func() {
 			Expect(err).Should(BeNil())
 			Expect(len(commits)).Should(BeNumerically(">", 0))
 		})
+		It("Should return error as it is not using HEAD", func() {
+			repo, err := githubrepo.MakeGithubRepo("ossf/scorecard")
```
spencerschrock, line 40: Can we use a smaller repo from `ossf-tests` perhaps? Every now and then I go back and prune these references since downloading `ossf/scorecard` takes some time.

e2e/searchCommits_test.go
```diff
@@ -36,5 +36,21 @@ var _ = Describe("E2E TEST:SearchCommits", func() {
...
+			Expect(err).Should(BeNil())
+			repoClient := githubrepo.CreateGithubRepoClient(context.Background(), logger)
+			err = repoClient.InitRepo(repo, "123456789", 0)
+			Expect(err).ShouldNot(Not(BeNil()))
+		})
+		It("Should return error as the user does not exists", func() {
+			repo, err := githubrepo.MakeGithubRepo("ossf/scorecard")
```
spencerschrock, line 47: same comment

spencerschrock (approved):

spencerschrock:
Alternatively, these could probably be hooked up to a stub http.RoundTripper and avoid the need for downloading altogether.
Similar to what we do in the webhooks test:
https://github.com/ossf/scorecard/blob/cc55b364b90bee78cd816a781b79b74cac5969e5/clients/githubrepo/webhook_test.go#L71-L89

e2e/searchCommits_test.go
naveensrinivasan: The previous tests require a valid repo with dependabot, and so these tests would cache the existing repo and shouldn't be an issue.

e2e/searchCommits_test.go
naveensrinivasan: The same comment as above.

e2e/searchCommits_test.go
spencerschrock: Caching is done at the repoclient level right? And there's no repo client shared between these tests (and it can't be since one expects an err and the other doesnt)

spencerschrock: I'm actually wrong here for another reason: these tests don't trigger the tarball download.

Issue #3193, opened 2023-06-19
BUG: Gitlab Commits Before Date Needs More Logic


jimrobison:
**Describe the bug**
Using the first commit returned has proven to not be the best source for the retrieval of merge requests details.

**Expected behavior**
The before field is intended to help limit the number of merge request details based on the last commit date for the default branch. By using the first record, this may not always include all of the related merge requests, throwing the scoring calculation off.

**Additional context**
I believe it would be better to calculate the date by cycling through the commits to determine the best date to use. I have a this is a working branch that I'm working through while validating the `Code-Review` check's scores

raghavkaul:
By default, GitLab GraphQL returns the most recent merges on a GitLab repo. But we don't guarantee the merge request that brought in a commit on a GitLab repo's main branch is captured by our check - we may undercount - or miss - merge requests in zip if GraphQL and the REST API don't line up. (We use 1 GraphQL query + 1 REST API call instead of 1 REST API call per commit to find the linked Merge Request, which wouldn't scale).

The purpose of using `mergedBefore` in [getMergeRequestDetail](https://github.com/raghavkaul/scorecard/blob/5fced752bf954141d22fa9d0f5c04b5533d41dea/clients/gitlabrepo/graphql.go#L130) is to improve this - "don't give me any merge requests newer than the newest commit I've listed". But, I think you're right that `before := commitsRaw[0].CommittedDate` may not reflect the commit date of the newest commit (it might be much older, therefore leading to a lower score). We might fix this is to find a better bound commit date in `commitsRaw`, or by making a single API query to [`ListMergeRequestsByCommit`](https://docs.gitlab.com/ee/api/commits.html#list-merge-requests-associated-with-a-commit) to get the right date for `before`.

Devansh7006:
Hi, I would like to work on this issue. I’ll analyze the current logic for retrieving commit dates and propose a fix to improve accuracy.

Thanks!

swetalin-10:
Hi, I would like to work on this issue.
I have experience working with APIs and Go, and I’d like to improve the commit filtering logic for GitLab.
Could you please confirm if I can proceed?

PR #3319, opened 2023-07-26
🌱 GitLab: Add log message

raghavkaul:
Adds log message when GitLab is running on non-HEAD that scores may be inaccurate.

Temporary warning #3193 to unblock release.


clients/repo_client.go
```diff
@@ -38,6 +45,8 @@ type RepoClient interface {
 	LocalPath() (string, error)
 	GetFileContent(filename string) ([]byte, error)
 	GetBranch(branch string) (*BranchRef, error)
+	GetCommitSHA() string
```
spencerschrock, line 48: I don't think this is a property of the repo client, but rather the repo right?

clients/repo_client.go
```diff
@@ -38,6 +45,8 @@ type RepoClient interface {
...
+	GetPlatform() string
```
spencerschrock, line 49: this is a breaking change, do we know if allstar is still the only implementer of custom clients? What about legitify?

spencerschrock (commented):
Is this something we want in the results? Or just when running the CLI?

clients/repo_client.go
raghavkaul: Oh right... what do you think of just a bare `fmt.Printf` in `clients/gitlabrepo/commits.go`? I had that originally but it felt janky so I did this whole thing

clients/repo_client.go
raghavkaul: Yeah but repos aren't accessible in `checks`

clients/repo_client.go
spencerschrock: It's a little hacky, yeah. At the minimum, we need to ensure it's printed to stderr. Printing to stdout breaks scorecard action at the very least

spencerschrock (commented):
Can we add a TODO comment with tracking issue?

Note: this can run multiple times, but not sure pushing the error deeper can be done as easily as right here

```console
SCORECARD_EXPERIMENTAL=1 go run main.go --repo gitlab.com/ossf-test/scorecard-archived-project --commit 240e1fbc4c4ea614527c59f453bf54107d8833f2
...
Scorecard may be missing merge requests when running on non-HEAD of a GitLab repo. Code-Review scores may be lower.
Scorecard may be missing merge requests when running on non-HEAD of a GitLab repo. Code-Review scores may be lower.
Scorecard may be missing merge requests when running on non-HEAD of a GitLab repo. Code-Review scores may be lower.
Scorecard may be missing merge requests when running on non-HEAD of a GitLab repo. Code-Review scores may be lower.
...
```

spencerschrock (approved):

PR #3320, opened 2023-07-26
🌱 GitLab: Add RepoHost tag to OpenCensus view

raghavkaul:
This PR adds the RepoHost tag to the opencensus view for metrics. This should allow us to break down error rates/runtimes in cron monitoring by github.com/gitlab.com repos.


checker/check_runner.go
```diff
@@ -100,6 +100,12 @@ func (r *Runner) Run(ctx context.Context, c Check) CheckResult {
 	if err != nil {
 		panic(err)
 	}
+
+	ctx, err = tag.New(ctx, tag.Upsert(stats.RepoHost, r.CheckRequest.Repo.Host()))
+	if err != nil {
+		panic(err)
```
spencerschrock, line 106: Lets avoid `panic`ing. Can you change the one above too? You can find a non-`panic` approach in `logStats`

checker/check_runner.go
```diff
@@ -98,8 +98,14 @@ func (r *Runner) Run(ctx context.Context, c Check) CheckResult {
 
 	ctx, err := tag.New(ctx, tag.Upsert(stats.CheckName, r.CheckName))
 	if err != nil {
-		panic(err)
+		r.CheckRequest.Dlogger.Debug(&LogMessage{Text: fmt.Sprintf("tag.New: %v", err)})
 	}
+
+	ctx, err = tag.New(ctx, tag.Upsert(stats.RepoHost, r.CheckRequest.Repo.Host()))
+	if err != nil {
+		r.CheckRequest.Dlogger.Debug(&LogMessage{Text: fmt.Sprintf("tag.New: %v", err)})
```
spencerschrock, line 106: I'm not sure the cron will see debug statements, do we just want to return these as error like we do in `logStats` above?

raghavkaul: I could change that to a warning, but I don't see logging failure as a reason the entire checkrun should fail.

spencerschrock: can we log in a way that doesn't associate it with the the check request? it's not really a detail of the check either

spencerschrock (approved):
Thanks

PR #3359, opened 2023-08-07
🌱 e2e tests: revert

raghavkaul:
Fix E2E test that expected `nil`. This was due to the GitLab service user ([user](https://gitlab.com/gossts)) not being set up correctly for the repo ([repo](https://gitlab.com/ossf-test/e2e-issues))., leading to E2E test failures ([link](https://github.com/ossf/scorecard/actions/runs/5788395634/job/15687305862)).


naveensrinivasan (approved):
Thank you!

spencerschrock:
I think the problem is that the error depends on if we're running on a PR (no GitLab token) vs after pushing to main (GitLab token).

PR #3361, opened 2023-08-08
:seedling: Increase test coverage for GitLab repository associations

naveensrinivasan:
- Add test to map GitLab access levels to Scorecard repository associations

[clients/gitlabrepo/issues_test.go]
- Add test for mapping GitLab access levels to Scorecard repository associations

the `release-note`

```release-note

```

### DIFF
--- a/clients/gitlabrepo/issues_test.go
+++ b/clients/gitlabrepo/issues_test.go
@@ -1,0 +1,49 @@
+// Copyright 2023 OpenSSF Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


naveensrinivasan:
> Hi, I know this increases test coverage, but doesn't it just check whether the switch case in that function corresponds to the actual enum values?

Yes, Do you have any other suggestions?

spencerschrock:
> > Hi, I know this increases test coverage, but doesn't it just check whether the switch case in that function corresponds to the actual enum values?
> 
> Yes, Do you have any other suggestions?

Now that #3393 is merged, would it be better to get this coverage through this test?
https://github.com/ossf/scorecard/blob/e82d15dc02d66320f3650515226c143b8027b694/clients/gitlabrepo/issues_test.go#L65

That way we test more of the feature and less of the implementation detail?

Issue #2462, opened 2022-11-18
CI: Report scoring differences for PRs that modify scorecard checks


raghavkaul:
**Is your feature request related to a problem? Please describe.**
As discussed during the 11/17 OpenSSF scorecard meeting, contributor who make changes to scorecard could potentially drastically alter scorecard scores. It should be easier to determine how a change impacts scorecard scores on top projects.

**Describe the solution you'd like**
Contributions to scorecard should have an easy way to analyze how their branch changes would affect scorecard scoring on a set of repos, and self-reporting this should be part of the pull request guidelines.

**Describe alternatives you've considered**
One alternative was raised to run this analysis as part of any CI jobs, however, this may exhaust the GitHub API token quota and cause us to get rate limited.

**Additional context**
@shissam hints that they may have done prior work automating the process of inspecting score diffs for their PRs in [this discussion](https://github.com/ossf/scorecard/pull/2442#discussion_r1020580754).

PR #2882, opened 2023-04-17
🐛 Code Review: Use proportional scoring

raghavkaul:
#### What kind of change does this PR introduce?
Revert Code-Review check to proportional scoring.

#### What is the current behavior?
Currently, Code-Review works like this:
If the only commit activity is by bots, we return inconclusively
If any number of bot code changes are unreviewed, the score drops 3 points
If ==1 human change is unreviewed, the score drops by 7 points
If >1 human change is unreviewed, the score drops to 0

#### What is the new behavior (if this is a feature change)?**
This takes Code-Review back to proportional scoring
If the only commit activity is by bots, we return inconclusively 
If there is human and bot commit activity, but only bot commits are unreviewed, the score drops by 3 points
If all bot commits are reviewed, the score is based on the proportion of human commits that are unreviewed

#### Which issue(s) this PR fixes
Fixes #2812 .

#### Does this PR introduce a user-facing change?
```release-note
* Code-Review: Use proportional instead of leveled scoring.
```

Issue #2812, opened 2023-03-31
BUG (?): Large changes to Code-Review scores between 4.10.2 and 4.10.5


pnacht:
**Describe the bug**
4.10.5 gives significantly lower scores for Code-Review than 4.10.2. 

I am running Scorecard frequently for some 200 projects. I just upgraded from 4.10.2 to 4.10.5 and noticed that average Code-Review scores are drastically lower: of the 200 projects, 137 saw their score drop. And the average drop (looking only at those projects that dropped) was of ~5 points.

| Δ score | # projects |
| - | - |
| -10	| 3 |
| -9	| 6 |
| -8	| 17 |
| -7	| 14 |
| -6	| 17 |
| -5	| 17 |
| -4	| 8 |
| -3	| 21 |
| -2	| 14 |
| -1	| 20 |
| 0	| 61 |
| 1	| 2 |

**Reproduction steps**
Run Scorecard for the following sample projects with 4.10.2 and 4.10.5:
- https://github.com/pandas-dev/pandas
- https://github.com/systemd/systemd
- http://github.com/microsoft/typescript

Results (trimmed):
--

```json
# pandas @ 4.10.2
    {
      "score": 9,
      "reason": "28 out of last 30 changesets reviewed before merge -- score normalized to 9",
      "name": "Code-Review",
    },
# pandas @ 4.10.5
    {
      "score": 0,
      "reason": "found 2 unreviewed human changesets",
      "name": "Code-Review",
    },
```

```json
# systemd/systemd @ 4.10.2
    {
      "score": 8,
      "reason": "18 out of last 21 changesets reviewed before merge -- score normalized to 8",
      "name": "Code-Review",
    },

# systemd/systemd @ 4.10.5
    {
      "score": 0,
      "reason": "found 3 unreviewed human changesets",
      "name": "Code-Review",
    },
```

```json
# microsoft/typescript @ 4.10.2
    {
      "score": 7,
      "reason": "22 out of last 30 changesets reviewed before merge -- score normalized to 7",
      "name": "Code-Review",
    },

# microsoft/typescript @ 4.10.5
    {
      "score": 0,
      "reason": "found 8 unreviewed human changesets",
      "name": "Code-Review",
    },
```

I intentionally picked these three projects to display here since they are very popular, active and have mature communities. It intuitively seems more likely that 80% of PRs are reviewed than that none are.

Diving in
--

[...]

PR #3363, opened 2023-08-08
:sparkles: scdiff: add basic compare functionality

spencerschrock:

feature

scdiff output is generated but not used

result output is read back into the results format
Note: this support is preliminary still

```console
go run cmd/internal/scdiff/main.go generate --repos repos --output=foo
go run cmd/internal/scdiff/main.go generate --repos repos --output=bar
# modify bar to introduce a diff
go run cmd/internal/scdiff/main.go compare foo bar
  pkg.ScorecardResult{
        Repo: pkg.RepoInfo{
-               Name:      "github.com/ossf/scorecard",
+               Name:      "github.com/ossf/scorecardy",
                CommitSHA: "",
        },
        Date:      s"0001-01-01 00:00:00 +0000 UTC",
        Scorecard: {},
        ... // 4 identical fields
  }

results differ
exit status 1
```

Related to https://github.com/ossf/scorecard/issues/2462, 4/n

From the [go-cmp documentation](https://pkg.go.dev/github.com/google/go-cmp/cmp):

> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal.

I think this tool is a valid use-case, but curious to hear other opinions. 

There's also the issue of the function: `ExperimentalFromJSON2`, which I'm taking the naming approach from `https://github.com/golang/go/issues/34409`

the `release-note`

```release-note

```


spencerschrock:
Here's a real example from #2812, simulating if #2882 was reverted:

```
  pkg.ScorecardResult{
        Repo:      {Name: "github.com/microsoft/typescript"},
        Date:      s"0001-01-01 00:00:00 +0000 UTC",
        Scorecard: {},
        Checks: []checker.CheckResult{
                ... // 2 identical elements
                {Name: "CI-Tests", Score: 10, Reason: "22 out of 22 merged PRs checked by a CI test -- score normalized"..., Details: {}, ...},
                {Name: "CII-Best-Practices", Reason: "no effort to earn an OpenSSF best practices badge detected", Details: {}},
                {
                        Name:    "Code-Review",
                        Version: 0,
                        Error:   nil,
-                       Score:   7,
+                       Score:   0,
                        Reason: strings.Join({
                                "found 9 unreviewed ",
-                               "changesets out of 30 -- score normalized to 7",
+                               "human changesets (30 total)",
                        }, ""),
                        Details: {},
                        Rules:   nil,
                },
                {Name: "Contributors", Score: 10, Reason: "34 different organizations found -- score normalized to 10", Details: {{Msg: {Text: " contributors work for DefinitelyTyped,HearTao,SwitchaBLE,babel,"...}}}, ...},
                {Name: "Dangerous-Workflow", Score: 10, Reason: "no dangerous workflow patterns detected", Details: {}, ...},
                ... // 11 identical elements
        },
        RawResults: {},
        Findings:   nil,
        Metadata:   nil,
  }

results differ
exit status 1
```

cmd/internal/scdiff/app/compare.go
```diff
@@ -0,0 +1,119 @@
...
+		if shouldContinue, err := advanceScanners(xs, ys); err != nil {
+			return err
+		} else if !shouldContinue {
+			break
+		}
+		xResult, yResult, err := loadResults(xs, ys)
+		if err != nil {
+			return err
+		}
+		if !compare.Results(&xResult, &yResult) {
```
raghavkaul, line 82: We know exactly where in the JSON there's a diff between `r1`, `r2` but the return value of this function is a boolean and then prints the entire contents of the JSON, instead of giving the first location where there's a diff. Trying to understand why we use `cmp.Diff` and then implement `compare.Results` separately?

raghavkaul (commented):
I assume the endgame for `scdiff` is to run `scorecard` and then the diff module, eliminating the need to compare structure/format between two JSON outputs, right? As in, we'd expect consistent result format, modulo runtime/API errors?

cmd/internal/scdiff/app/compare.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/ossf/scorecard/v4/pkg"
+)
+
+//nolint:gochecknoinits // common for cobra apps
+func init() {
+	rootCmd.AddCommand(compareCmd)
+}
+
+var (
+	errMissingInputFiles = errors.New("must provide at least two files")
```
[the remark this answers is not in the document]
spencerschrock: files produced by the `generate` command.

cmd/internal/scdiff/app/compare.go
spencerschrock: My intention was using our own compare func to ignore certain fields, but then fall back on cmp.Diff to pretty print.

spencerschrock:
> I assume the endgame for `scdiff` is to run `scorecard` and then the diff module, eliminating the need to compare structure/format between two JSON outputs, right? As in, we'd expect consistent result format, modulo runtime/API errors?

1. run `scdiff generate` on a set to produce the golden.
2. commit golden to repo.
3. run `scdiff generate` on proposed changes
4. run `scdiff compare` to evaluate changes.
5. if intentional, run `scdiff update` to merge changes to golden, which will be committed

cmd/internal/scdiff/app/compare.go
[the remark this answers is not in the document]
raghavkaul: Could that flow be made more explicit, either in code documentation or in the error message displayed to the user?

cmd/internal/scdiff/app/compare.go
raghavkaul: Gotcha. I think it'd be cleaner if we committed to a single comparison function that included diff location, because it seems like repeated work to diff twice, and also that `go-cmp` might point out more differences than we care between two scorecard runs, but that's up to whatever seems to you to be more maintainable.

spencerschrock:
@raghavkaul addressed feedback, please take a look at the changes from the last round:
https://github.com/ossf/scorecard/pull/3363/files/16e11f2df84e77feedd1b208d0eb4578acebb4d8..9c4d6d2b187599376d4c934dca6f089e84a63228

raghavkaul (approved):

PR #3387, opened 2023-08-14
:seedling: Migrate to go 1.21

### LINKED ISSUE #3386: Migrate to go1.21
With the release of go1.21, go1.19 becomes unsupported. We should start testing with go1.21 and transition.

Note: we don't need to necessarily bump the version in our go.mod until we use newer features. This is one downside to the `go-version-file` approach we were taking in some of the workflow files.
```
       uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
       with:
         go-version-file: ${{ env.GO_VERSION_FILE }}
```

I had hoped it would reduce the number of places we need to update when changing Go versions, but now it creates coupling between our module minimum version and the toolchain version. So perhaps its best to revert back to using `go-version` instead.

spencerschrock:

dependency update

- use go1.19, which is no longer supported

- use go1.21 for our CI and builds/releases
- keep go1.19 in our go.mod as we don't currently use any 1.20 or 1.21 features. this is for our dependents.

#3386 

the `release-note`

```release-note
Scorecard releases will now use go1.21
```

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,7 +23,7 @@ on:
   - main
 
 env:
-  GO_VERSION_FILE: go.mod # no good way of getting a mutual version between go.mod and tools/go.mod
+  GO_VERSION: 1.21
 
 jobs:
   docs_only_check:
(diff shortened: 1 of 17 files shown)


justaugustus (approved):
Thanks @spencerschrock!
(Filed a separate issue for the code coverage hit: https://github.com/ossf/scorecard/issues/3388)

PR #3393, opened 2023-08-15
:seedling: Fix failing GitLab e2e tests

spencerschrock:

test fix

- two e2e tests fail

- one e2e test is replaced with a unit test
- one the type mismatch was corrected

[Failing PR run](https://github.com/ossf/scorecard/actions/runs/5869856344/job/15915698490#step:10:13107)

the `release-note`

```release-note

```


naveensrinivasan (approved):
Thanks

clients/gitlabrepo/issues_test.go
```diff
@@ -0,0 +1,131 @@
...
+			repoURL := repoURL{
+				owner:     "ossf-tests",
+				commitSHA: clients.HeadSHA,
+			}
+			handler.init(&repoURL)
+			got, err := handler.listIssues()
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("listIssues error: %v, wantedErr: %t", err, tt.wantErr)
+			}
+			if !cmp.Equal(got, tt.want) {
```
raghavkaul, line 126: This test seems kind of brittle because we expect the entire JSON response from GitLab to match our expected payload, can we narrow it to only test the specific things scorecard needs, like number of issues?

spencerschrock: I don't think that's what this is doing. It's only checking what we parse from their response?
```
want: []clients.Issue{
				{
					URI:       strptr("131356518"),
					CreatedAt: timeptr(time.Date(2023, time.July, 26, 14, 22, 52, 0, time.UTC)),
					Author: &clients.User{
						ID: 1355794,
					},
					AuthorAssociation: associationptr(clients.RepoAssociationMaintainer),
				},
			},
```

raghavkaul: Ah that makes sense, I misread `tt.want` and `got`. `got` is pulled from testdata.

PR #3394, opened 2023-08-16
:seedling: Add Tests for GitLab CheckRuns API

naveensrinivasan:
- Add tests for the `CheckRuns` function

the `release-note`

```release-note

```


clients/gitlabrepo/checkruns_test.go
```diff
@@ -0,0 +1,90 @@
...
+			handler := &checkrunsHandler{
+				glClient: client,
+			}
+
+			repoURL := repoURL{
+				owner:     "ossf-tests",
+				commitSHA: clients.HeadSHA,
+			}
+			handler.init(&repoURL)
+			got, err := handler.listCheckRunsForRef("main")
```
spencerschrock, line 80: ref is intended to be a SHA in I think (not that it really matters since the response is hardcoded. I can file an issue for this, as GitLab seems to distinguish between `ref` and `sha` in ways that GitHub doesn't.

clients/gitlabrepo/checkruns_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}{
+		{
+			name:         "valid checkruns",
+			responsePath: "./testdata/valid-checkruns",
+			want: []clients.CheckRun{
+				{
+					Status:     "pending",
+					URL:        "https://example.com/foo/bar/pipelines/48",
+					Conclusion: "",
+				},
```
spencerschrock, line 43: If we try this with a completed checkrun, such as the one listed in the example [here](https://docs.gitlab.com/ee/api/pipelines.html#get-a-single-pipeline), I don't think the code, as written, would set the `Conclusion` field.

Might be worth filing an issue and expanding the tests. GitLab doesn't have a `conclusion` field, but I think we can parse it to emulate the GitHub equivalent. Might require some discussion in the issue.

clients/gitlabrepo/checkruns_test.go
spencerschrock: #3395

clients/gitlabrepo/checkruns_test.go
spencerschrock: #3396

PR #3415, opened 2023-08-22
:sparkles: Support Binary-Artifacts check again for local repos

### LINKED ISSUE #3258: Issue: scorecard to detect Binary Artifacts from local directory in command line (terminal)
Hi Team,

Note: Editing my originally raised issue:

We are trying to use scorecard for Binary Detection. Our aim is to detect binary in local folder. I was able to detect binaries in my personal github repo, however wasn't able to get results while executing on local folder using --local argument. I then built the code locally, added **FileBased** option in checks/binary_artifact.go init() and generated the build. The Binary Detection was successful. I was able to generate json output with score 9, due to 1 binary in the local folder. 

I did added Environment variable for SARIF format as well as "ENABLE_SARIF":"1", however, I am getting below error while generating SARIF format:
"error during command execution: failed to format results: failed to output results: internal error: missing policy for check: Binary-Artifacts"

**I have below 2 questions:**
1. Is it fine to change checks/binary_artifact.go by adding FileBased in init(), this indirectly fails the tests in policy_test.go? Can we update it? So that the scorecard tool can be used for local folder Binary Artifact Detection
2. Any inputs regarding when will the SARIF format be supported in command line? I am trying to add environment variables and run the tool, however i get policy missing error as mentioned above. Is there any way to achieve this pelase?

Thanks in advance,

Shilpa

spencerschrock:

feature

- local repos can't use the `Binary-Artifacts` check.

- local repos can use `Binary-Artifacts`, but there are no gradle-wrapper exceptions because we can't verify workflow runs locally (see #2039)

Fixes #3258

the `release-note`

```release-note

```

### DIFF
--- a/checks/binary_artifact.go
+++ b/checks/binary_artifact.go
@@ -24,10 +24,11 @@ import (
 // CheckBinaryArtifacts is the exported name for Binary-Artifacts check.
 const CheckBinaryArtifacts string = "Binary-Artifacts"
 
-//nolint
+//nolint:gochecknoinits
 func init() {
 	supportedRequestTypes := []checker.RequestType{
 		checker.CommitBased,
+		checker.FileBased,
 	}
 	if err := registerCheck(CheckBinaryArtifacts, BinaryArtifacts, supportedRequestTypes); err != nil {
 		// this should never happen
--- a/checks/raw/binary_artifact.go
+++ b/checks/raw/binary_artifact.go
@@ -15,6 +15,7 @@
 package raw
 
 import (
+	"errors"
 	"fmt"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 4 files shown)


checks/binary_artifact.go
```diff
@@ -24,10 +24,11 @@ import (
 // CheckBinaryArtifacts is the exported name for Binary-Artifacts check.
 const CheckBinaryArtifacts string = "Binary-Artifacts"
 
-//nolint
+//nolint:gochecknoinits
 func init() {
 	supportedRequestTypes := []checker.RequestType{
 		checker.CommitBased,
+		checker.FileBased,
```
laurentsimon, line 31: there's the remaining problem that this test would be enabled on pull_request in the Action, but should not because it needs API access? (Different problem, I suppose.. I recall there's an existing issue for it)

laurentsimon (approved):

checks/binary_artifact.go
spencerschrock: Yes, that did come to mind. If run locally on PRs, then we would warn about something that might be ok after merging.
Since PRs are experimental, I'm not sure I want to block on this currently. I can track down the relevant issue and leave a comment.

PR #3423, opened 2023-08-24
:sparkles: Increase PyPI parsing flexibility

### LINKED ISSUE #3249: Can I contribute to "Repo lookup from PyPI too conservative to use"
**Describe the bug**
Many PyPI packages specify their GitHub source repository in ways that `scorecard` cannot parse. The issue seems to be that PyPI places no restriction on the names in the `project_urls` map. scorecard however only looks at exactly one key and then bails if it doesn't find a github repo there.

Consider the structure of these very well supported projects, which *all use different structures*

```
$ curl --silent https://pypi.org/pypi/scipy/json | jq '.info.project_urls'
{
  "Documentation": "https://docs.scipy.org/doc/scipy/",
  "Download": "https://github.com/scipy/scipy/releases",
  "Homepage": "https://scipy.org/",
  "Source": "https://github.com/scipy/scipy",
  "Tracker": "https://github.com/scipy/scipy/issues"
}

$ curl --silent https://pypi.org/pypi/numpy/json | jq '.info.project_urls'
{
  "Bug Tracker": "https://github.com/numpy/numpy/issues",
  "Documentation": "https://numpy.org/doc/1.25",
  "Download": "https://pypi.python.org/pypi/numpy",
  "Homepage": "https://www.numpy.org",
  "Source Code": "https://github.com/numpy/numpy"
}

$ curl --silent https://pypi.org/pypi/tensorflow/json | jq '.info.project_urls'
{
  "Download": "https://github.com/tensorflow/tensorflow/tags",
  "Homepage": "https://www.tensorflow.org/"
}
```

**Reproduction steps**
scorecard --pypi numpy # This fails 

**Expected behavior**
I think `scorecard` ought to be able to handle this non-uniformity to find a github repo. 

**Additional context**
I'm happy to put in a PR to fix this (already started) if the reviewers agree with the approach (and will help me with my golang =D).

Proposal 1: Construct and ordered list of case & whitespace insensitive keys to check in project urls. If any look like a github repo use the first one that matches.
Proposal 2: Find all the github repos listed in project_urls. Fail if there are any number of repos besides exactly 1 found.

Thank you so much for this tool!

joshgc:

This is a likely a non-breaking change, but due to PyPI diversity I can't prove it doesn't break on any real-world package. I analyzed 396 packages (those I'm using in my current project). Of those **only 63** could be correctly parsed by current parsing (before this PR) and I break none of them with this PR.

However after this PR, of the 396 packages **329** get parsed correctly..

Improvement opportunities:

1. Some sphinx packages list a central repo for issues but a different, valid, repo for the source code. I don't think our approach of "parse everything" can handle these.
2. Similar to sphinx, many of the types-* have the same problem. See types-protobuf

Only use ["Info"]["project_urls"]["Source"] to find a github URL

Look through all the ["Info"]["project_urls"] and ["Info"]["project_url"] to find all possible repos. Fail if that number is anything but 1.

Fixes #3249 

None

```release-note
Changes to the PyPI parsing will make scorecard fail on a small number of previously working packages 
```


cmd/package_managers.go
```diff
@@ -108,6 +113,51 @@ func fetchGitRepositoryFromNPM(packageName string, packageManager pmc.Client) (s
...
+	}
+	return ""
+}
+
+func findGitRepositoryInPYPIResponse(packageName string, response io.Reader) (string, error) {
+	v := &pypiSearchResults{}
+	err := json.NewDecoder(response).Decode(v)
+	if err != nil {
+		return "", sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("failed to parse pypi package json: %v", err))
+	}
```
ssmall, line 139: consider using `fmt.Errorf` and the `%w` directive

cmd/package_managers.go
```diff
@@ -108,6 +113,51 @@ func fetchGitRepositoryFromNPM(packageName string, packageManager pmc.Client) (s
...
+
+	if len(validURLs) > 1 {
+		return "", sce.WithMessage(sce.ErrScorecardInternal,
+			fmt.Sprintf("found too many possible source repos for pypi package: %s", packageName))
+	}
+
+	for url, _ := range validURLs {
+		return url, nil
+	}
```
ssmall, line 156: If I understand this code correctly, the `validURLs` map functions as a set to de-duplicate repo URLs. However, it doesn't appear that the full set contents ever matter since `len(validURLs) > 1` is an error case. Is the set even really necessary, then? Could you just have a single string variable and bail out the first time you find a URL that does not match the one you've already seen?

spencerschrock:
Note, you'll need to fix DCO on the commits you have already. Another repo has written a nice guide on fixing it here:
https://github.com/src-d/guide/blob/7bb18c0577b779a1d80c3f6cd6e3a825200e7dad/developer-community/fix-DCO.md#how-to-add-sign-offs-retroactively

I've enabled the rest of the CI to run, where the linter may pickup some things. I'll review the rest of the change when I have some time.

cmd/package_managers.go
spencerschrock: Note: the author is matching the idioms in the rest of the file.

So probably something we (the maintainers) would need to fix more broadly. `ErrScorecardInternal` is something that should really be a wrapper error type which supports inner errors, but we're stuck with what we have now for backwards compatibility, and we're not using go1.20 yet for multi-errors.

joshgc: Ack leaving as is for now

cmd/package_managers.go
joshgc: Good call! Done

cmd/package_managers.go
```diff
@@ -18,12 +18,46 @@ package cmd
...
+	"regexp"
 
 	ngt "github.com/ossf/scorecard/v4/cmd/internal/nuget"
 	pmc "github.com/ossf/scorecard/v4/cmd/internal/packagemanager"
 	sce "github.com/ossf/scorecard/v4/errors"
 )
 
+var _GITHUB_DOMAIN_REGEXP = regexp.MustCompile(`^https?://github[.]com/([^/]+)/([^/.]+)`)
+var _GITHUB_SUBDOMAIN_REGEXP = regexp.MustCompile(`^https?://([^.]+)[.]github[.]io/([^/.]+).*`)
+var _GITLAB_DOMAIN_REGEXP = regexp.MustCompile(`^https?://gitlab[.]com/([^/]+)/([^/.]+)`)
```
spencerschrock, line 31: The linter doesn't like multiple things here. It's expecting something more of the form:
```
var (
    githubDomainRegexp = ...
    githubSubdomainRegexp = ...
    gitlabSubdomainRegexp = ...
)
```

cmd/package_managers.go
```diff
@@ -18,12 +18,46 @@ package cmd
...
+	if len(urlAndPathParts) < 3 {
+		return ""
+	}
+	if urlAndPathParts[1] == "sponsors" {
+		return ""
+	}
+	return fmt.Sprintf("https://github.com/%s/%s", urlAndPathParts[1], urlAndPathParts[2])
+}
+
+var _PYPI_MATCHERS = []func(string) string{
```
spencerschrock, line 43: Can we just make this a func? Are we expecting a project URL to match more than 1 pattern?

```
func repoMatch(url string) string {
  // ...
}
```

At the very least, the name needs to be camel case

cmd/package_managers.go
```diff
@@ -108,6 +141,36 @@ func fetchGitRepositoryFromNPM(packageName string, packageManager pmc.Client) (s
...
 }
 
+func findGitRepositoryInPYPIResponse(packageName string, response io.Reader) (string, error) {
+	v := &pypiSearchResults{}
+	err := json.NewDecoder(response).Decode(v)
+	if err != nil {
+		return "", sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("failed to parse pypi package json: %v", err))
+	}
+
+	v.Info.ProjectURLs["key_not_used"] = v.Info.ProjectURL
```
spencerschrock, line 151: Technically this could collide with a key from the repos `project_urls` field (though unlikely)? Not sure if the empty string is a better alternative? Not sure there's any one best answer here.

cmd/package_managers.go
```diff
@@ -108,6 +141,36 @@ func fetchGitRepositoryFromNPM(packageName string, packageManager pmc.Client) (s
...
+	validURL := ""
```
spencerschrock, line 152: nit: prefer `var` syntax for zero value inits
```
var validURL string
```

cmd/package_managers.go
```diff
@@ -18,12 +18,46 @@ package cmd
...
+	"io"
+	"regexp"
 
 	ngt "github.com/ossf/scorecard/v4/cmd/internal/nuget"
 	pmc "github.com/ossf/scorecard/v4/cmd/internal/packagemanager"
 	sce "github.com/ossf/scorecard/v4/errors"
 )
 
+var _GITHUB_DOMAIN_REGEXP = regexp.MustCompile(`^https?://github[.]com/([^/]+)/([^/.]+)`)
+var _GITHUB_SUBDOMAIN_REGEXP = regexp.MustCompile(`^https?://([^.]+)[.]github[.]io/([^/.]+).*`)
```
spencerschrock, line 30: `.` is a valid character in a github repo name. (e.g. https://github.com/0k/kids.cmd). I assume this is to strip the `.git` suffix? Perhaps do that at the end of `makeGithubRepo`?

spencerschrock:
`make check-linter` should help flag some of these linter warnings locally.

spencerschrock:
> inform,maturin,quantiphy lists the same github url twice but with different URLs. Github redirects these to the same repo but I don't think thats documented behavior

These just look like capitalization issues. I think it's safe to normalize for your deduplication logic with `strings.ToLower` or something, we do it elsewhere in the code as well.

> Some sphinx packages list a central repo for issues but a different, valid, repo for the source code. I don't think our approach of "parse everything" can handle these.

agree. unless we have a list of tags ("source code", "repository", etc)  we prefer, before falling back to the "parse everything" approach.

> Similar to sphinx, many of the types-* have the same problem. See types-protobuf

Some sort of counting approach could help? Of course this may introduce issues for other repos.
```
{
  "Changes": "https://github.com/typeshed-internal/stub_uploader/blob/main/data/changelogs/protobuf.md",
  "Chat": "https://gitter.im/python/typing",
  "GitHub": "https://github.com/python/typeshed",
  "Homepage": "https://github.com/python/typeshed",
  "Issue tracker": "https://github.com/python/typeshed/issues"
}
```

cmd/package_managers.go
joshgc: Good call, done

cmd/package_managers.go
joshgc: Done

cmd/package_managers.go
joshgc: Yea the empty quotes felt worse, so I made the key longer =\

Happy to do whatever

cmd/package_managers.go
joshgc: Upto you! I originally had it as a single function and felt I started to repeat myself. I think it depends on how sophisticated we want to get with Gitlab or other host parsing in the future

cmd/package_managers.go
spencerschrock: Let's leave it as is. Collision is unlikely, and worst case someone files an issue one day and we fix it down the line.

cmd/package_managers.go
joshgc: Ah I just realized i put in CapitalCamelCase instead of camelCase. The linter didn't complain, happy to do whatever you prefer

joshgc:
> > inform,maturin,quantiphy lists the same github url twice but with different URLs. Github redirects these to the same repo but I don't think thats documented behavior
> 
> These just look like capitalization issues. I think it's safe to normalize for your deduplication logic with `strings.ToLower` or something, we do it elsewhere in the code as well.
> 
> ```
> {
>   "changelog": "https://github.com/KenKundert/inform/blob/master/doc/releases.rst",
>   "documentation": "https://inform.readthedocs.io",
>   "homepage": "https://inform.readthedocs.io",
>   "repository": "https://github.com/kenkundert/inform"
> }
> ```
> 
> > Some sphinx packages list a central repo for issues but a different, valid, repo for the source code. I don't think our approach of "parse everything" can handle these.
> 
> agree. unless we have a list of tags ("source code", "repository", etc) we prefer, before falling back to the "parse everything" approach.
> 
> > Similar to sphinx, many of the types-* have the same problem. See types-protobuf
> 
> Some sort of counting approach could help? Of course this may introduce issues for other repos.
> 
> ```
> {
>   "Changes": "https://github.com/typeshed-internal/stub_uploader/blob/main/data/changelogs/protobuf.md",
>   "Chat": "https://gitter.im/python/typing",
>   "GitHub": "https://github.com/python/typeshed",
>   "Homepage": "https://github.com/python/typeshed",
>   "Issue tracker": "https://github.com/python/typeshed/issues"
> }
> ```

Yes I made this case insensitive. Even without the ranked key look up we have x5 the hit rate! I can also add the list look up method too. I ended up having ~10 items in it before

cmd/package_managers.go
spencerschrock: Capitalization causes the symbol to be exported in Go, so would prefer keeping these regexes private.

spencerschrock: Same comment for the PYPI matcher func array

joshgc: Done!

cmd/package_managers.go
```diff
@@ -108,6 +146,38 @@ func fetchGitRepositoryFromNPM(packageName string, packageManager pmc.Client) (s
...
+	var validURL string
+	for _, url := range v.Info.ProjectURLs {
+		for _, matcher := range pypiMatchers {
+			repo := matcher(url)
+			if repo == "" {
+				continue
+			}
+			if validURL == "" {
+				validURL = repo
+			} else if validURL != repo {
```
spencerschrock, line 166: Currently, the `strings.ToLower` is in `makeGithubRepo`, so the GitLab pattern wouldn't benefit from it. What do you think about moving the `ToLower` call out of `makeGithubRepo` to here?

```go
for _, url := range v.Info.ProjectURLs {
		url = strings.ToLower(url)
		for _, matcher := range pypiMatchers {
```

spencerschrock (approved):
LGTM thanks, just the small issue. And the PR will need to be marked ready for review.

cmd/package_managers.go
joshgc: @spencerschrock sorry just seeing this. I'm happy to move the ToLower to cover gitlab as well. I am not sure if gitlab is case insensitive, but I'll assume you are and make that change now

spencerschrock: GitLab is case insensitive as well. At least for the owner/repo strings

joshgc: Added to the gitlab parsing bit with a comment...because I'm paranoid about one day there being githubknockoff.com that is case sensitive. Also added a test

spencerschrock:
Just need DCO fixed again for your very last commit

joshgc:
> Just need DCO fixed again for your very last commit

Done. Sorry/Thanks!!

cmd/package_managers_test.go
```diff
@@ -161,6 +162,144 @@ func Test_fetchGitRepositoryFromNPM(t *testing.T) {
...
+    "project_url": "foo",
+    "project_urls": {
+      "RandomKey": "https://gitlab.com/htaslan/color/",
+      "raNdoMkEY": "https://gitlab.com/hTASLan/color"
+    }
+  }
+}
+`,
+			want:       "https://github.com/htaslan/color",
+			wantErrStr: "",
```
spencerschrock, line 258: apologies for all the back and forth, the CI requiring approval for first-time contributors sometimes lets copy/paste bugs fall through the cracks.

joshgc: Sigh. No that was my fault. confirm from cmd/ 

go test
go fmt
gofumpt -d package* 

all succeed

spencerschrock (approved):
Thanks for the PR!

Issue #3254, opened 2023-07-10
BUG: Dependency-Pinning Scoring


balteravishay:
**Describe the bug**

The Dependency-Pinning Scoring is taking into account all ecosystems, even if there is no evidence for them in a repository.

For instance, if a repo only contains a single shell script with a single line "pip install -r requirements.txt", the pinned-dependency score for that repo will by 8/10. 
To get to a score of 0/10, means that all ecosystems that are known by scorecard exist in the repo, at the same time, and all checks for unpinned dependencies for those ecosystems report for unpinned dependencies. That seems like an edge case, at best.

**Reproduction steps**
1. Have a repo with only one ecosystem that is checked and scored by the unpinned dependency check (at the time of opening this issue the supported checks are: docker files, docker downloads, pip or scripts download, PR #2960 by @gabibguti adds npm support)
2. run Scorecard with --checks=Pinned-Dependencies and the repo name
3. The result is 8/10.

**Expected behavior**
It is expected that the scoring is normalized only for the ecosystems that are actually found in the repository. so that a repository is only measured on the technology stacks that the repo actually implements.

Examples:
* A repository that uses python. pip install commands are unpinned: expected result is 0/10
* A repository that uses python. pip install commands are pinned: expected result is 10/10
* A repository that uses python and docker. pip install commands are pinned (i.e. use --require-hashes) while docker base images are not pinned: expected result is 5/10

**Additional context**
The root cause seems to be that raw dependency checks do not differentiate between not finding unpinned dependencies for an ecosystem (score 10), and not finding any evidence of using dependencies of the ecosystem the check is for.

spencerschrock:
+1 I think this could be converted to some proportional logic (e.g. x / y dependencies are pinned), or keep the equal split approach we have now, but only consider repos with dependencies present.

spencerschrock:
Similarly I don't think we need an `Info` statement that dependencies are pinned for an ecosystem that isn't used.

gabibguti:
I will work on this next.

As commented:
> The root cause seems to be that raw dependency checks do not differentiate between not finding unpinned dependencies for an ecosystem (score 10), and not finding any evidence of using dependencies of the ecosystem the check is for.

First, I will work on getting the dependencies count for each ecossystem.
e.g. we have 20 pip installs, 20 go installs and 0 npm installs in this repository

Second, use this counting to know if the project depends on not on that ecosystem and only count ecossystems it depends on.
[...]

PR #3424, opened 2023-08-24
🐛 Add go installs to Pinned-Dependencies score

gabibguti:

This is a bug fix. Just like pip and npm installs, go installs were identified but did not count for Pinned-Dependencies score. Now, they do.

go installs are identified as pinned or unpinned and reported by Scorecard, but do not count for Pinned-Dependencies score.

go installs now count in the aggregated score for Pinned-Dependencies.

It's addressing one of the problems reported at https://github.com/ossf/scorecard/issues/2582

the `release-note`

```release-note
go installs count for Pinned-Dependencies score.
```


checks/evaluation/pinned_dependencies_test.go
```diff
@@ -247,7 +247,23 @@ func Test_PinningDependencies(t *testing.T) {
...
+			name: "unpinned go install",
+			dependencies: []checker.Dependency{
+				{
+					Location: &checker.File{},
+					Type:     checker.DependencyUseTypeGoCommand,
+				},
+			},
+			expected: scut.TestReturn{
+				Error:         nil,
+				Score:         8,
```
spencerschrock, line 264: Just an FYI: #3254 seems very relevant here.

checks/evaluation/pinned_dependencies_test.go
```diff
@@ -111,7 +111,7 @@ func Test_PinningDependencies(t *testing.T) {
 				Error:         nil,
 				Score:         checker.MaxResultScore,
 				NumberOfWarn:  0,
-				NumberOfInfo:  7,
+				NumberOfInfo:  8,
```
spencerschrock, line 114: similar fyi: this is also something that #3254 should probably fix.

spencerschrock (approved):
conversations are good to resolve, mainly leaving them as FYI comments. I think we should prioritize #3254 before any more forgotten ecosystems.

checks/evaluation/pinned_dependencies_test.go
gabibguti: Ack, I will prioritize https://github.com/ossf/scorecard/issues/3254 next!

checks/evaluation/pinned_dependencies_test.go
gabibguti: Ack!

PR #3426, opened 2023-08-25
:bug: Print Info in Empty Repo Scans

### LINKED ISSUE #2157: Improving Score Reporting: Empty repository

leec94:

For empty repos, still return repository information, scorecard version, and time. This feature resolves #2157

For empty repos, empty json is returned. 

For empty repos, still return repository information, scorecard version, time, and score of -1. Local scans have not changed. 

Fixes #2157

Not sure if pkg/scorecard_test.go needs to be updated. How do I run this test to check? 

the `release-note`

```release-note
prints repository, scorecard version, and time data for scans of empty repos. no user action change required. 
```


pkg/scorecard.go
```diff
@@ -105,9 +105,25 @@ func RunScorecard(ctx context.Context,
...
+				Name:      repo.URI(),
+				CommitSHA: commitSHA,
+			},
+			Scorecard: ScorecardInfo{
+				Version:   versionInfo.GitVersion,
+				CommitSHA: versionInfo.GitCommit,
+			},
+			Date: time.Now(),
+		}
+		return ret, nil
```
spencerschrock, line 122: If we're not surfacing the error, let's not add it to the public error list. A private error at the top of this file should work too?

I was curious, so if we return the error we get something like this instead:
```
Error: RunScorecard: repository empty
2023/08/25 10:42:40 error during command execution: RunScorecard: repository empty
exit status 1
```

pkg/scorecard.go
```diff
@@ -105,9 +105,25 @@ func RunScorecard(ctx context.Context,
...
+		ret := ScorecardResult{
+			Repo: RepoInfo{
+				Name:      repo.URI(),
+				CommitSHA: commitSHA,
+			},
+			Scorecard: ScorecardInfo{
+				Version:   versionInfo.GitVersion,
+				CommitSHA: versionInfo.GitCommit,
+			},
+			Date: time.Now(),
```
spencerschrock, line 120: we have some code below that does the same thing. What do you think about moving it higher so we only define it in one spot?

pkg/scorecard.go
```diff
@@ -105,9 +105,25 @@ func RunScorecard(ctx context.Context,
...
+		}
+		return ret, nil
+	} else if err != nil || commitSHA == "" {
```
spencerschrock, line 123: do we still need the commitSHA comparison?

spencerschrock (commented):
Some of the unit tests need fixed. You can check locally with `make unit-test`. Same with the linter: `make check-linter`

pkg/scorecard.go
leec94: changed this to a local error

pkg/scorecard.go
leec94: good idea. moved this higher too

pkg/scorecard.go
leec94: i removed this. hopefully its not needed in some random other edge case

leec94:
@spencerschrock thanks for the feedback so far. i ran the tests and looks like there's still 1 where it's not passing, but i'm not sure how to get the `want` to match up with what's being returned. any advice?

pkg/scorecard.go
spencerschrock: can you make it private? e.g `errEmptyRepository`

spencerschrock:
Repasting some of our Slack conversation here for others / so it doesn't go away:


So the test currently does:
```go
			if !reflect.DeepEqual(got, tt.want) {
				t.Errorf("RunScorecard() got = %v, want %v", got, tt.want)
			}
```
`reflect.DeepEqual` will compare every field. The timing issue is a common one in testing, and is a common example when dependency injection is talked about.

We use the `go-cmp` library elsewhere in our test code, which lets you do a better comparison:
https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal
Here's an example from our code where we explicitly ignore a field. You will probably want to do that with the date field:
https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreFields
https://github.com/ossf/scorecard/blob/63789835132192f60929f7a4511815536e7ec7b4/checks/evaluation/contributors_test.go#L149-L152

leec94:
updated the test to ignore the date, since the seconds will always be off, and to check for "unknown" commitSHA. should pass unit-tests now! i can also remove the commented out function if youd like @spencerschrock

spencerschrock:
> updated the test to ignore the date, since the seconds will always be off, and to check for "unknown" commitSHA. should pass unit-tests now! i can also remove the commented out function if youd like @spencerschrock

Which function?

This is what I'm seeing for an empty repo:
```console
go run main.go --repo=github.com/azeemshaikh38/empty-repo-test --format=json  --show-details | jq
{
  "date": "2023-09-14T14:40:01-07:00",
  "repo": {
    "name": "github.com/azeemshaikh38/empty-repo-test",
    "commit": "HEAD"
  },
  "scorecard": {
    "version": "",
    "commit": "unknown"
  },
  "score": -1,
  "checks": null,
  "metadata": null
}
```
And local seems to work still:
```console
go run main.go --local=. --format json --show-details | jq
{
  "date": "2023-09-14T14:44:21-07:00",
  "repo": {
    "name": "file://.",
    "commit": "HEAD"
  },
  "scorecard": {
    "version": "",
    "commit": "unknown"
  },
  "score": 7.9,
 ...
}
```

can review more in a bit

leec94:
sorry i misspoke. it was just the previous way of doing the check i initially commented out, but then just removed. let me know if there's anything i need to change

pkg/scorecard.go
```diff
@@ -83,7 +86,7 @@ func getRepoCommitHash(r clients.RepoClient) (string, error) {
 	if len(commits) > 0 {
 		return commits[0].SHA, nil
 	}
-	return "", nil
+	return "", errEmptyRepository
```
spencerschrock, line 89: optional: can we flip this existing code around for "happy path" readability?

```go
if len(commits) == 0 {
	return "", errEmptyRepository
}
return commits[0].SHA, nil
```

pkg/scorecard_test.go
```diff
@@ -168,8 +180,8 @@ func TestRunScorecard(t *testing.T) {
 				t.Errorf("RunScorecard() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("RunScorecard() got = %v, want %v", got, tt.want)
+			if !cmp.Equal(got, tt.want, cmpopts.IgnoreFields(ScorecardResult{}, "Date")) { //nolint:govet
+				t.Errorf("expected %v, got %v", got, cmp.Diff(tt.want, got, cmpopts.IgnoreFields(ScorecardResult{}, "Date"))) //nolint:lll
```
spencerschrock, line 184: one thing you can do to reduce duplication and the line length is saving the `opt`.
```go
ignoreDate := cmpopts.IgnoreFields(ScorecardResult{}, "Date")
if !cmp.Equal(got, tt.want, ignoreDate) {
    t.Errorf("expected %v, got %v", got, cmp.Diff(tt.want, got, ignoreDate))
}
```

ps. not sure what the govet linter was for, doesn't seem to be triggering anymore? can probably be removed

spencerschrock (approved):

pkg/scorecard.go
leec94: done

pkg/scorecard_test.go
leec94: done. thanks for these suggestions and approving the PR!

leec94:
is the code coverage check a blocker to getting this merged?

spencerschrock:
> is the code coverage check a blocker to getting this merged?

No. It's an unfortunate side-effect of our current coverage uploading process. Which we should probably spend some time fixing, as it can be confusing.

leec94:
if you think it's suitable id be happy to help with the code coverage improvements too

spencerschrock:
> if you think it's suitable id be happy to help with the code coverage improvements too

Ah sorry, I had already started working on this yesterday. Just saw this now when I went to link to your comment from #3492

leec94:
no problem! i'd like to keep up helping out tho so if you have suggestions on other issues to look at, that would be great

Issue #2578, opened 2023-01-03
Feature: Improve documentation to save maintainer's time on dealing with pinned dependencies alerts


joycebrum:
**Is your feature request related to a problem? Please describe.**
It is not clear to the user how to handle the security alert on dockerfile or bash files regarding some package managers. 

I've done some research myself and most of them I wasn't able to even determine if they allow a pinned by hash install or not. 

Some examples I've look into were: 

- yum - it is **impossible** to pin by hash (AFAIK).
- wget - it is **impossible** to pin by hash (AFAIK).
- curl -  it **can be possible** to pin by hash (AFAIK).

**Describe the solution you'd like**
To make it easier for the user to remediate Pinned Dependencies alert, we could cover in the documentation the most common package managers and whether it is possible to pin them by hash or not.

It would save maintainers time on useless trying to hash pin a package manager that does not allow it.

And, in these cases when it is impossible to pin by hash, we should suggest they dismiss the alert on github since it cannot be fixed. 

**Additional context**

Although I didn't find a way to pin by hash yum installs, it seems there is some kind of digest verification at least when it is installing local rpm packages (that's what I could get from [this link](https://joelitechlife.ca/2021/12/08/yum-gpg-check-failed-or-package-no-digest/))

Curl seems to have the [--pinnedpubkey option](https://everything.curl.dev/usingcurl/tls/pinning) but I'm not sure if it would really be considered pinned by hash.

pnacht:
A (clunky) idea would be to check for something like

``` shell
curl https://www.foo.com -o my_artifact.jar  # or equivalent for wget/yum
echo "1234abcd my_artifact.jar" | sha256sum -c
# or
sha256sum -c hashes.txt
# we'd need to check that hashes.txt actually includes all the downloaded files
```

That is: make sure they download the file and then check its hash.

PR #3436, opened 2023-08-30
:bug: Dependency-Pinning Scoring

gabibguti:

This is a bug fix. Pinned-Dependencies score used to score a 10 for ecossystems not used by the project. Now, it scores -1 when the ecossystem is not used, and this score is not included in the final score for Pinned-Dependencies.

For example, if project A uses only Python and has only Python or pip dependencies, the project would score 10 for Dockerfiles, npm, go and other ecossystems dependencies it does not use.

In this PR, if no dependencies are found for an ecossystem, this ecossystem score is not included in the final score.

Fixes https://github.com/ossf/scorecard/issues/3254

the `release-note`

```release-note
Pinned-Dependencies score includes only used ecossystems (npm, go, pip, etc.)
```

### DIFF
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -122,6 +122,7 @@ type Dependency struct {
 	Location *File
 	Msg      *string // Only for debug messages.
 	Type     DependencyUseType
+	Pinned   bool
 }
 
 // MaintainedData contains the raw results
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -122,7 +122,7 @@ type Dependency struct {
 	Location *File
 	Msg      *string // Only for debug messages.
 	Type     DependencyUseType
-	Pinned   bool
+	Pinned   *bool
 }
 
 // MaintainedData contains the raw results
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -121,8 +121,8 @@ type Dependency struct {
 	PinnedAt *string
 	Location *File
 	Msg      *string // Only for debug messages.
-	Type     DependencyUseType
 	Pinned   *bool
+	Type     DependencyUseType
 }
 
 // MaintainedData contains the raw results
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -337,20 +337,24 @@ func createReturnValuesForGitHubActionsWorkflowPinned(r worklowPinningResult, ma
 
 	if r.gitHubOwned.total == r.gitHubOwned.pinned {
 		score += 2
-		dl.Info(&checker.LogMessage{
-			Type:   finding.FileTypeSource,
-			Offset: checker.OffsetDefault,
-			Text:   fmt.Sprintf("%s %s", "GitHub-owned", maxResultMsg),
-		})
+		if r.gitHubOwned.total != 0 {
+			dl.Info(&checker.LogMessage{
+				Type:   finding.FileTypeSource,
+				Offset: checker.OffsetDefault,
+				Text:   fmt.Sprintf("%s %s", "GitHub-owned", maxResultMsg),
+			})
+		}
 	}
 
 	if r.thirdParties.total == r.thirdParties.pinned {
 		score += 8
-		dl.Info(&checker.LogMessage{
-			Type:   finding.FileTypeSource,
-			Offset: checker.OffsetDefault,
-			Text:   fmt.Sprintf("%s %s", "Third-party", maxResultMsg),
-		})
+		if r.thirdParties.total != 0 {
+			dl.Info(&checker.LogMessage{
+				Type:   finding.FileTypeSource,
+				Offset: checker.OffsetDefault,
+				Text:   fmt.Sprintf("%s %s", "Third-party", maxResultMsg),
+			})
+		}
 	}
 
 	return score, nil
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -181,7 +181,7 @@ func updatePinningResults(rr *checker.Dependency,
 	}
 
 	// Update other result types.
-	var p pinnedResult = pr[rr.Type]
+	p := pr[rr.Type]
 	addPinnedResult(rr, &p)
 	pr[rr.Type] = p
 }
@@ -290,17 +290,18 @@ func createReturnValues(pr map[checker.DependencyUseType]pinnedResult,
 	//nolint
 	r := pr[t]
 
-	if r.total == 0 {
+	switch r.total {
+	case 0:
 		dl.Info(&checker.LogMessage{
 			Text: inconclusiveResultMsg,
 		})
 		return checker.InconclusiveResultScore, nil
-	} else if r.total == r.pinned {
+	case r.pinned:
 		dl.Info(&checker.LogMessage{
 			Text: maxResultMsg,
 		})
 		return checker.MaxResultScore, nil
-	} else {
+	default:
 		return checker.MinResultScore, nil
 	}
 }
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -133,17 +133,18 @@ func PinningDependencies(name string, c *checker.CheckRequest,
 		return checker.CreateRuntimeErrorResult(name, err)
 	}
 
-	// Scores may be inconclusive.
-	actionScore = maxScore(0, actionScore)
-	dockerFromScore = maxScore(0, dockerFromScore)
-	dockerDownloadScore = maxScore(0, dockerDownloadScore)
-	scriptScore = maxScore(0, scriptScore)
-	pipScore = maxScore(0, pipScore)
-	npmScore = maxScore(0, npmScore)
-	goScore = maxScore(0, goScore)
+	// If no dependencies of an ecossystem are found, it results in an inconclusive score.
+	// We filter out inconclusive scores so only applicable ecossystems are considered in
+	// the final score.
+	scores := []int{actionScore, dockerFromScore, dockerDownloadScore, scriptScore, pipScore, npmScore, goScore}
+	conclusiveScores := []int{}
+	for i := range scores {
+		if scores[i] != checker.InconclusiveResultScore {
+			conclusiveScores = append(conclusiveScores, scores[i])
+		}
+	}
 
-	score := checker.AggregateScores(actionScore, dockerFromScore,
-		dockerDownloadScore, scriptScore, pipScore, npmScore, goScore)
+	score := checker.AggregateScores(conclusiveScores...)
 
 	if score == checker.MaxResultScore {
 		return checker.CreateMaxScoreResult(name, "all dependencies are pinned")
@@ -197,14 +198,6 @@ func generateOwnerToDisplay(gitHubOwned bool) string {
 		return "GitHub-owned"
 	}
 	return "third-party"
-}
-
-// TODO(laurent): need to support GCB pinning.
-func maxScore(s1, s2 int) int {
-	if s1 > s2 {
-		return s1
-	}
-	return s2
 }
 
 func addPinnedResult(rr *checker.Dependency, r *pinnedResult) {
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -28,13 +28,10 @@ import (
 
 var errInvalidValue = errors.New("invalid value")
 
-type pinnedResult int
-
-const (
-	pinnedUndefined pinnedResult = iota
-	pinned
-	notPinned
-)
+type pinnedResult struct {
+	pinned int
+	total  int
+}
 
 // Structure to host information about pinned github
 // or third party dependencies.
@@ -80,7 +77,7 @@ func PinningDependencies(name string, c *checker.CheckRequest,
 				Text:      *rr.Msg,
 				Snippet:   rr.Location.Snippet,
 			})
-		} else {
+		} else if !*rr.Pinned {
 			dl.Warn(&checker.LogMessage{
 				Path:        rr.Location.Path,
 				Type:        rr.Location.Type,
@@ -90,11 +87,11 @@ func PinningDependencies(name string, c *checker.CheckRequest,
 				Snippet:     rr.Location.Snippet,
 				Remediation: generateRemediation(remediationMetadata, &rr),
 			})
+		}
 
 			// Update the pinning status.
 			updatePinningResults(&rr, &wp, pr)
 		}
-	}
 
 	// Generate scores and Info results.
 	// GitHub actions.
@@ -177,13 +174,13 @@ func updatePinningResults(rr *checker.Dependency,
 		// Note: `Snippet` contains `action/name@xxx`, so we cna use it to infer
 		// if it's a GitHub-owned action or not.
 		gitHubOwned := fileparser.IsGitHubOwnedAction(rr.Location.Snippet)
-		addWorkflowPinnedResult(wp, false, gitHubOwned)
+		addWorkflowPinnedResult(rr, wp, gitHubOwned)
 		return
 	}
 
 	// Update other result types.
-	var p pinnedResult
-	addPinnedResult(&p, false)
+	var p pinnedResult = pr[rr.Type]
+	addPinnedResult(rr, &p)
 	pr[rr.Type] = p
 }
 
@@ -213,28 +210,18 @@ func maxScore(s1, s2 int) int {
 	return s2
 }
 
-// For the 'to' param, true means the file is pinning dependencies (or there are no dependencies),
-// false means there are unpinned dependencies.
-func addPinnedResult(r *pinnedResult, to bool) {
-	// If the result is `notPinned`, we keep it.
-	// In other cases, we always update the result.
-	if *r == notPinned {
-		return
+func addPinnedResult(rr *checker.Dependency, r *pinnedResult) {
+	if *rr.Pinned {
+		r.pinned += 1
 	}
-
-	switch to {
-	case true:
-		*r = pinned
-	case false:
-		*r = notPinned
-	}
+	r.total += 1
 }
 
-func addWorkflowPinnedResult(w *worklowPinningResult, to, isGitHub bool) {
+func addWorkflowPinnedResult(rr *checker.Dependency, w *worklowPinningResult, isGitHub bool) {
 	if isGitHub {
-		addPinnedResult(&w.gitHubOwned, to)
+		addPinnedResult(rr, &w.gitHubOwned)
 	} else {
-		addPinnedResult(&w.thirdParties, to)
+		addPinnedResult(rr, &w.thirdParties)
 	}
 }
 
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -15,7 +15,6 @@
 package evaluation
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/ossf/scorecard/v4/checker"
@@ -25,8 +24,6 @@ import (
 	"github.com/ossf/scorecard/v4/remediation"
 	"github.com/ossf/scorecard/v4/rule"
 )
-
-var errInvalidValue = errors.New("invalid value")
 
 type pinnedResult struct {
 	pinned int
@@ -89,9 +86,9 @@ func PinningDependencies(name string, c *checker.CheckRequest,
 			})
 		}
 
-			// Update the pinning status.
-			updatePinningResults(&rr, &wp, pr)
-		}
+		// Update the pinning status.
+		updatePinningResults(&rr, &wp, pr)
+	}
 
 	// Generate scores and Info results.
 	// GitHub actions.
@@ -231,6 +228,7 @@ func createReturnForIsShellScriptFreeOfInsecureDownloads(pr map[checker.Dependen
 ) (int, error) {
 	return createReturnValues(pr, checker.DependencyUseTypeDownloadThenRun,
 		"no insecure (not pinned by hash) dependency downloads found in shell scripts",
+		"no dependency downloads found in shell scripts",
 		dl)
 }
 
@@ -240,6 +238,7 @@ func createReturnForIsDockerfilePinned(pr map[checker.DependencyUseType]pinnedRe
 ) (int, error) {
 	return createReturnValues(pr, checker.DependencyUseTypeDockerfileContainerImage,
 		"Dockerfile dependencies are pinned",
+		"no Dockerfile dependencies found",
 		dl)
 }
 
@@ -249,6 +248,7 @@ func createReturnForIsDockerfileFreeOfInsecureDownloads(pr map[checker.Dependenc
 ) (int, error) {
 	return createReturnValues(pr, checker.DependencyUseTypeDownloadThenRun,
 		"no insecure (not pinned by hash) dependency downloads found in Dockerfiles",
+		"no dependency downloads found in Dockerfiles",
 		dl)
 }
 
@@ -257,7 +257,8 @@ func createReturnForIsPipInstallPinned(pr map[checker.DependencyUseType]pinnedRe
 	dl checker.DetailLogger,
 ) (int, error) {
 	return createReturnValues(pr, checker.DependencyUseTypePipCommand,
-		"Pip installs are pinned",
+		"pip installs are pinned",
+		"no pip installs found",
 		dl)
 }
 
@@ -267,6 +268,7 @@ func createReturnForIsNpmInstallPinned(pr map[checker.DependencyUseType]pinnedRe
 ) (int, error) {
 	return createReturnValues(pr, checker.DependencyUseTypeNpmCommand,
 		"npm installs are pinned",
+		"no npm installs found",
 		dl)
 }
 
@@ -276,27 +278,32 @@ func createReturnForIsGoInstallPinned(pr map[checker.DependencyUseType]pinnedRes
 ) (int, error) {
 	return createReturnValues(pr, checker.DependencyUseTypeGoCommand,
 		"go installs are pinned",
+		"no go installs found",
 		dl)
 }
 
 func createReturnValues(pr map[checker.DependencyUseType]pinnedResult,
-	t checker.DependencyUseType, infoMsg string,
+	t checker.DependencyUseType, maxResultMsg string, inconclusiveResultMsg string,
 	dl checker.DetailLogger,
 ) (int, error) {
-	// Note: we don't check if the entry exists,
-	// as it will have the default value which is handled in the switch statement.
+	// If there are zero dependencies of this type, it will generate an inconclusive score,
+	// so we can disconsider it in the aggregated score.
+	// If all dependencies of this type are pinned, it will get a maximum score.
+	// If 1 or more dependencies of this type are unpinned, it will get a minimum score.
 	//nolint
-	r, _ := pr[t]
-	switch r {
-	default:
-		return checker.InconclusiveResultScore, fmt.Errorf("%w: %v", errInvalidValue, r)
-	case pinned, pinnedUndefined:
+	r := pr[t]
+
+	if r.total == 0 {
 		dl.Info(&checker.LogMessage{
-			Text: infoMsg,
+			Text: inconclusiveResultMsg,
+		})
+		return checker.InconclusiveResultScore, nil
+	} else if r.total == r.pinned {
+		dl.Info(&checker.LogMessage{
+			Text: maxResultMsg,
 		})
 		return checker.MaxResultScore, nil
-	case notPinned:
-		// No logging needed as it's done by the checks.
+	} else {
 		return checker.MinResultScore, nil
 	}
 }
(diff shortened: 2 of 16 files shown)


dependencydiff/mapping.go
```diff
@@ -22,7 +22,7 @@ import (
 type ecosystem string
 
 // OSV ecosystem naming data source: https://ossf.github.io/osv-schema/#affectedpackage-field
-//nolint
+// nolint
```
spencerschrock, line 25: can the PR focus on the change at hand? It makes it easier to track changes in the future, as these lines PR will get added as another layer in `git blame`

renaming `asPointer` -> `asStringPointer`?

checks/evaluation/pinned_dependencies.go
```diff
@@ -90,10 +84,10 @@ func PinningDependencies(name string, c *checker.CheckRequest,
 				Snippet:     rr.Location.Snippet,
 				Remediation: generateRemediation(remediationMetadata, &rr),
 			})
-
-			// Update the pinning status.
-			updatePinningResults(&rr, &wp, pr)
 		}
+
+		// Update the pinning status.
+		updatePinningResults(&rr, &wp, pr)
```
spencerschrock, line 90: so we previously didn't update results when rr.Msg != nil, which was used as an indicator of a parsing error:
https://github.com/ossf/scorecard/blob/08e7fb89f49f8d65edacca3190283e43ba566fb6/checks/raw/pinned_dependencies.go#L395-L400

Is this intentional moving this out?

checks/evaluation/pinned_dependencies.go
```diff
@@ -139,17 +133,22 @@ func PinningDependencies(name string, c *checker.CheckRequest,
...
-	dockerFromScore = maxScore(0, dockerFromScore)
-	dockerDownloadScore = maxScore(0, dockerDownloadScore)
-	scriptScore = maxScore(0, scriptScore)
-	pipScore = maxScore(0, pipScore)
-	npmScore = maxScore(0, npmScore)
-	goScore = maxScore(0, goScore)
+	// If no dependencies of an ecossystem are found, it results in an inconclusive score.
+	// We filter out inconclusive scores so only applicable ecossystems are considered in
+	// the final score.
+	scores := []int{actionScore, dockerFromScore, dockerDownloadScore, scriptScore, pipScore, npmScore, goScore}
```
spencerschrock, line 139: before this, it was common to forget ecosystems. what do you think about refactoring the logic in a way to avoid needing to remember to update this line in the future?

checks/evaluation/pinned_dependencies.go
```diff
@@ -289,27 +275,33 @@ func createReturnForIsGoInstallPinned(pr map[checker.DependencyUseType]pinnedRes
...
-	r, _ := pr[t]
-	switch r {
-	default:
-		return checker.InconclusiveResultScore, fmt.Errorf("%w: %v", errInvalidValue, r)
-	case pinned, pinnedUndefined:
+	r := pr[t]
+
+	switch r.total {
+	case 0:
+		dl.Info(&checker.LogMessage{
```
spencerschrock, line 295: let's make this a debug. we don't need a detail for every ecosystem a project isn't using

checks/evaluation/pinned_dependencies.go
```diff
@@ -318,31 +310,52 @@ func createReturnValues(pr map[checker.DependencyUseType]pinnedResult,
...
-	dl checker.DetailLogger,
+func createReturnValuesForGitHubActionsWorkflowPinned(r worklowPinningResult, maxResultMsg string,
+	inconclusiveResultMsg string, dl checker.DetailLogger,
 ) (int, error) {
-	score := checker.MinResultScore
-
-	if r.gitHubOwned != notPinned {
-		score += 2
+	if r.gitHubOwned.total == 0 {
 		dl.Info(&checker.LogMessage{
```
spencerschrock, line 321: debug

checks/evaluation/pinned_dependencies.go
```diff
@@ -318,31 +310,52 @@ func createReturnValues(pr map[checker.DependencyUseType]pinnedResult,
...
-			Offset: checker.OffsetDefault,
-			Text:   fmt.Sprintf("%s %s", "GitHub-owned", infoMsg),
+			Text: fmt.Sprintf("%s %s", "no GitHub-owned", inconclusiveResultMsg),
 		})
 	}
 
-	if r.thirdParties != notPinned {
-		score += 8
+	if r.thirdParties.total == 0 {
 		dl.Info(&checker.LogMessage{
```
spencerschrock, line 327: debug

checks/raw/pinned_dependencies.go
```diff
@@ -273,8 +272,9 @@ var validateDockerfilesPinning fileparser.DoWhileTrueOnFileContent = func(
 						EndOffset: uint(child.EndLine),
 						Snippet:   child.Original,
 					},
-					Name:     asPointer(name),
-					PinnedAt: asPointer(asName),
+					Name:     asStringPointer(name),
+					PinnedAt: asStringPointer(asName),
+					Pinned:   asBoolPointer(pinned || regex.MatchString(name)),
```
spencerschrock, line 277: this duplicates some of the logic above. can we just use the value of `pinnedAsNames[asName]`?

checks/raw/pinned_dependencies.go
```diff
@@ -283,27 +283,26 @@ var validateDockerfilesPinning fileparser.DoWhileTrueOnFileContent = func(
...
+
+			dep := checker.Dependency{
+				Location: &checker.File{
+					Path:      pathfn,
+					Type:      finding.FileTypeSource,
+					Offset:    uint(child.StartLine),
+					EndOffset: uint(child.EndLine),
+					Snippet:   child.Original,
+				},
+				Pinned: asBoolPointer(pinned || regex.MatchString(name)),
```
spencerschrock, line 295: possibly same comment about duplicating code

checks/raw/shell_download_validate.go
```diff
@@ -397,11 +399,7 @@ func collectExecuteFiles(startLine, endLine uint, node syntax.Node, cmd, pathfn
 
 // Npm install docs are here.
 // https://docs.npmjs.com/cli/v7/commands/npm-install
-func isNpmUnpinnedDownload(cmd []string) bool {
-	if len(cmd) == 0 {
```
spencerschrock, line 401: without the length check, this code can panic when accessing `cmd[0]`

checks/raw/shell_download_validate.go
```diff
@@ -419,50 +417,39 @@ func isNpmUnpinnedDownload(cmd []string) bool {
 	return false
 }
 
-func isGoUnpinnedDownload(cmd []string) bool {
-	if len(cmd) == 0 {
```
spencerschrock, line 423: same comment

checks/evaluation/pinned_dependencies.go
```diff
@@ -139,17 +133,22 @@ func PinningDependencies(name string, c *checker.CheckRequest,
...
+		}
+	}
+
+	if len(conclusiveScores) == 0 {
+		return checker.CreateInconclusiveResult(name, "no dependencies found")
+	}
 
-	score := checker.AggregateScores(actionScore, dockerFromScore,
-		dockerDownloadScore, scriptScore, pipScore, npmScore, goScore)
+	score := checker.AggregateScores(conclusiveScores...)
```
spencerschrock, line 151: does this contradict the approach from the issue?
https://github.com/ossf/scorecard/issues/3254#issuecomment-1692422889

fixing this could fix the issue above about forgetting ecosystems

spencerschrock (commented):
Left a partial review, first thoughts:
1. Touching asPointer in unrelated files makes the change bigger than it needs to be
2. The scoring approach is different than discussed.

dependencydiff/mapping.go
gabibguti: sure!
about the renaming I was striving for consistency in the naming across the files, but I agree with the git blame argument. I will remove the renaming `asPointer` to `asStringPointer` and add/use the method `asBoolPointer` only where it's necessary. So we'll have `asPointer` and `asBoolPointer` methods in the same file, but less diff changes.

checker/raw_result.go
```diff
@@ -121,6 +121,7 @@ type Dependency struct {
 	PinnedAt *string
 	Location *File
 	Msg      *string // Only for debug messages.
+	Pinned   *bool
```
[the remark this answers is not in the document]
gabibguti: Yes, I changed the logic when calculating "raw" pinned dependencies to flag pinned and unpinned cases and store in this variable.

checks/evaluation/pinned_dependencies.go
gabibguti: Definetely not intentional. The behavior we want is "if dependency has a parsing error, log error and don't count it on score", is that it?

gabibguti: I'll add some validation in the evaluation for when a dependency does not have a `Pinned` state, like in the case of a parsing error, so we have a decent error message. And also make sure we don't count dependencies when we already know they had parsing problems. Otherwise it will break the code in runtime, thanks Spencer!

checks/evaluation/pinned_dependencies.go
gabibguti: Yes, I totally agree. I just think we would make lots of changes for pinned dependencies evaluation, since we would be needing to replace scoring for individual ecossystems by an iteration of all ecossystems and applying a score for them. So I was thinking of changing the evaluaiton in a second PR, if this one works out, but we can do it all at once if you think the changes need to come together.

checks/evaluation/pinned_dependencies.go
gabibguti: Yes as replied above, but we can make the evaluation changes in this PR instead of in a second PR if you'd like.

checks/raw/pinned_dependencies.go
gabibguti: Yes, sure, nice catch!

checks/raw/pinned_dependencies.go
gabibguti: Yes, sure, nice catch!

checks/raw/pinned_dependencies.go
gabibguti: Doing a better analysis, here yes, we can replace `pinned || regex.MatchString(name)` by `pinnedAsNames[asName]`.

We have 3 pinned cases:
- `FROM something@sha256` - Pinned directly
- `FROM something@sha256 as base` - Pinned directly but defining AS NAME
- `FROM base` - Pinned indirectly using already defined AS NAME

In the code, we hold the info if the image is pinned directly or indirectly. `pinned` tells us if this image was already pinned AS NAME before, and `regex.MatchString(name)` tells us if the image is pinned right now.

In this part of the code, we handle AS NAME cases, so the if the image is pinned right now or before AS NAME, this info will be present in `pinnedAsNames[asName]`.

checks/raw/shell_download_validate.go
gabibguti: I moved this validation to a step before.
In `collectUnpinnedPakageManagerDownload` we evaluate if the command is a go, pip, pip, npm, choco or nuget download. In this function I added this validation to do an early return if the command is empty. Otherwise, it will not be empty when it reaches `isNpmDownload` and similar isXYXDownload functions. Since isXYXDownload functions are only used in `collectUnpinnedPakageManagerDownload`, I believe the code will not panic.
I can readd this validation to the isXYXDownload methods for precaution if you'd like.

checker/raw_result.go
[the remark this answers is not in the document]
spencerschrock: Yeah, this was one of my first comments sorry. Thought I got them all once I read the later files.

checks/evaluation/pinned_dependencies.go
spencerschrock: > Definitely not intentional. The behavior we want is "if dependency has a parsing error, log error and don't count it on score", is that it?

This is what the existing code did, but only in this one spot

checks/evaluation/pinned_dependencies.go
spencerschrock: It seems like one change to me. If we're going to make major changes to how we score pinned dependencies, might as well do it in one change, not two.

gabibguti: Ack, I'm working on these changes

gabibguti:
@spencerschrock I think I addressed all suggested changes. It's ready for a second review.

gabibguti:
The code may be lacking Info and Debug logs now. Can I get your opinion if it would be good to add:
- debug logs to inform which dependencies are pinned e.g. pipCommand dependency pinned at {file}
- debug logs to inform the proportion of each ecossystem e.g. 5 out of 6 pipCommands pinned

checker/check_result.go
```diff
@@ -88,6 +88,14 @@ type LogMessage struct {
 	Remediation *rule.Remediation // Remediation information, if any.
 }
 
+// ProportionalScoreWithWeight is a structure that contains
+// the fields to calculate weighted proportional scores.
+type ProportionalScoreWithWeight struct {
```
spencerschrock, line 93: What do you think about `ProportionalScoreWeighted`?

checker/check_result.go
```diff
@@ -97,6 +105,20 @@ func CreateProportionalScore(success, total int) int {
 	return int(math.Min(float64(MaxResultScore*success/total), float64(MaxResultScore)))
 }
 
+// CreateProportionalScoreWithWeight creates the proportional score
+// between multiple successes over the total, but some proportions
+// are worth more.
+func CreateProportionalScoreWithWeight(scores ...*ProportionalScoreWithWeight) int {
+	ws := 0
+	wt := 0
```
spencerschrock, line 113: can we prefer this for zero values? let's us know the type easier:
```
var ws, wt int
```

checker/check_result.go
```diff
@@ -97,6 +105,20 @@ func CreateProportionalScore(success, total int) int {
 	return int(math.Min(float64(MaxResultScore*success/total), float64(MaxResultScore)))
 }
 
+// CreateProportionalScoreWithWeight creates the proportional score
+// between multiple successes over the total, but some proportions
+// are worth more.
+func CreateProportionalScoreWithWeight(scores ...*ProportionalScoreWithWeight) int {
```
spencerschrock, line 111: does this need to be a pointer arg?

checks/evaluation/pinned_dependencies.go
```diff
@@ -90,66 +95,32 @@ func PinningDependencies(name string, c *checker.CheckRequest,
...
-	}
-
-	// Npm installs.
-	npmScore, err := createReturnForIsNpmInstallPinned(pr, dl)
-	if err != nil {
-		return checker.CreateRuntimeErrorResult(name, err)
+	var scores []*checker.ProportionalScoreWithWeight
+	// Go through all dependency types
+	// GitHub Actions need to be handled separately since they are not in pr
+	scores = append(scores, createScoreForGitHubActionsWorkflow(&wp)...)
```
spencerschrock, line 107: Is there a reason why they're not in `pr`, would it be easier to just add them to `pr`?
I do see how this might simplify the weighted 20/80 split though.

checks/evaluation/pinned_dependencies.go
```diff
@@ -90,66 +95,32 @@ func PinningDependencies(name string, c *checker.CheckRequest,
...
+	// Only exisiting dependencies will be found in pr
+	// We will only score the ecossystem if there are dependencies
+	// This results in only existing ecossystems being included in the final score
+	for t := range pr {
+		scores = append(scores, &checker.ProportionalScoreWithWeight{
+			Success: pr[t].pinned,
+			Total:   pr[t].total,
+			Weight:  10,
```
spencerschrock, line 115: can we make this weight a `const` (perhaps right before this loop?), with a comment about why 10? I assume it's because github actions are split 2 and 8?

spencerschrock: if we make 2 and 8 consts, we can also define this const in terms of those const, that way we ensure GitHub action will always be weighted equally to other types

spencerschrock: Although thinking about this now, shouldn't the weight be dependent on total? If all ecosystems/types have weight 10, then we're still averaging the ecosystems unweighted.

checks/evaluation/pinned_dependencies.go
```diff
@@ -205,144 +176,53 @@ func generateOwnerToDisplay(gitHubOwned bool) string {
...
+			{
+				Success: wp.gitHubOwned.pinned,
+				Total:   wp.gitHubOwned.total,
+				Weight:  2,
+			},
+			{
+				Success: wp.thirdParties.pinned,
+				Total:   wp.thirdParties.total,
+				Weight:  8,
+			},
```
spencerschrock, line 209: same comment about the weights not reflecting the number of occurrences.

checks/evaluation/pinned_dependencies.go
gabibguti: Yes. I believe the reason is that `pr = map[checker.DependencyUseType]pinnedResult`. The `DependencyUseType` for GHA is `DependencyUseTypeGHAction`, but it does not distinct between `GitHub-owned` and `third-party` GHA. This comes like this from the `raw` part. So, we make the distinction at the `evaluation` using a separate structure `wp = worklowPinningResult`

To add GHAs into `pr`, I see the following options:
- branch `DependencyUseTypeGHAction` into `DependencyUseTypeGHAGitHubOwnedAction` and `DependencyUseTypeGHAThirdPartyAction`. This would move the distinction to `raw` instead of `evaluation`.
- modify `pr` to be `map[checker.DependencyUseType] (pinnedResult OR worklowPinningResult)`. Not sure if this is possible in Go.
- modify `pr` to be `map[string]pinnedResult`.

gabibguti: > I do see how this might simplify the weighted 20/80 split though.
Actually, depending on the approach I think this weighting could be simplified.

checks/evaluation/pinned_dependencies.go
gabibguti: > Although thinking about this now, shouldn't the weight be dependent on total?

No, I don't think so. 

> If all ecosystems/types have weight 10, then we're still averaging the ecosystems unweighted.

Yes, that's true! The only reason for the weighting is to make GHA GitHub-owned and third-party have different weights.

e.g. 1 / 20 GHA GitHub-owned + 2 / 20 GHA third-party + 1 / 1 pipCommand
```
(1 * 2) + (2 * 8) + (1 * 10)
-------------------------------
(20 * 2) + (20 * 8) + (1 * 10)

= 28 / 210 = 0.13
```

e.g. 1 / 20 GHA GitHub-owned + 20 / 20 GHA third-party + 1 / 1 pipCommand
```
(1 * 2) + (20 * 8) + (1 * 10)
-------------------------------
(20 * 2) + (20 * 8) + (1 * 10)

= 172 / 210 = 0.81
```

e.g. 20 / 20 GHA GitHub-owned + 2 / 20 GHA third-party + 1 / 1 pipCommand
```
(20 * 2) + (2 * 8) + (1 * 10)
-------------------------------
(20 * 2) + (20 * 8) + (1 * 10)

= 66 / 210 = 0.31
```

This makes Third-party count more than GitHub-owned.

spencerschrock: Yes, I think I was confused about where the weights was being taken, so ignore that part. Please take a look at this part still though:
> if we make 2 and 8 consts, we can also define this const in terms of those const, that way we ensure GitHub action will always be weighted equally to other types

something like
```go
const (
    githubOwnedActionWeight = 2
    thirdPartyActionWeight = 8
    nonActionWeight = githubOwnedActionWeight + thirdPartyActionWeight
)
```

gabibguti: Yes! We can do that.

checker/check_result.go
gabibguti: No, it doesn't! Fixing it.

gabibguti:
@spencerschrock I think I addressed the new block of requests.

checks/raw/shell_download_validate.go
```diff
@@ -419,50 +417,39 @@ func isNpmUnpinnedDownload(cmd []string) bool {
...
-		if slices.Contains([]string{"get", "install"}, cmd[i]) {
-			found = true
-		}
 
-		if !found {
-			continue
+	for i := 2; i < len(cmd); i++ {
+		// The -d flag instructs get to download but not install the packages
+		if strings.EqualFold(cmd[i], "-d") {
+			return false
```
spencerschrock, line 438: this change seems slipped in?

checks/raw/shell_download_validate.go
```diff
@@ -609,7 +589,7 @@ func isUnpinnedPipInstall(cmd []string) bool {
 
 	// Any other form of install is unpinned,
 	// e.g. `pip install`.
```
spencerschrock, line 591: these comments can probably be deleted now?

checks/raw/shell_download_validate.go
```diff
@@ -846,15 +827,16 @@ func collectUnpinnedPakageManagerDownload(startLine, endLine uint, node syntax.N
 					EndOffset: endLine,
 					Snippet:   cmd,
 				},
-				Type: checker.DependencyUseTypeNpmCommand,
+				Pinned: asBoolPointer(false),
```
spencerschrock, line 830: i think the hardcoded false here is a bug? Or rather we dont reward npm commands which do call "ci"

checks/raw/shell_download_validate.go
```diff
@@ -977,7 +961,8 @@ func collectFetchProcSubsExecute(startLine, endLine uint, node syntax.Node, cmd,
 				EndOffset: endLine,
 				Snippet:   cmd,
 			},
-			Type: checker.DependencyUseTypeDownloadThenRun,
+			Pinned: asBoolPointer(false),
```
spencerschrock, line 964: this one i'm not sure about. can we detect a pinned version of this one? 
probably fine as hard coded false?

e2e/pinned_dependencies_test.go
```diff
@@ -49,9 +49,9 @@ var _ = Describe("E2E TEST:"+checks.CheckPinnedDependencies, func() {
 			}
 			expected := scut.TestReturn{
 				Error:         nil,
-				Score:         4,
+				Score:         2,
 				NumberOfWarn:  139,
-				NumberOfInfo:  3,
+				NumberOfInfo:  0,
```
spencerschrock, line 54: This is a placeholder for me to figure out the 3 info messages that went away.

spencerschrock (commented):
Overall, I like the approach. I'm curious if we need to scale the pinned/total for the pip/npm commands by how many dependencies are in the lockfiles. Right now a command just counts as 1? @laurentsimon thoughts?

checker/check_result.go
```diff
@@ -97,6 +105,19 @@ func CreateProportionalScore(success, total int) int {
 	return int(math.Min(float64(MaxResultScore*success/total), float64(MaxResultScore)))
 }
 
+// CreateProportionalScoreWeighted creates the proportional score
+// between multiple successes over the total, but some proportions
+// are worth more.
+func CreateProportionalScoreWeighted(scores ...ProportionalScoreWeighted) int {
```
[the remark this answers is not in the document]
spencerschrock: I don't think it would get called like that, but will default to you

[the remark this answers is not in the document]
gabibguti: Ah, no, we don't. It shouldn't get called if total is 0. I will review the test cases to make sure this is caught in case we have and unweighted version, and perhaps throw an error at the moment if total or weight is 0.

checks/raw/shell_download_validate.go
gabibguti: Yes, kind of. I will try to remove it and leave it for another PR.

checks/raw/shell_download_validate.go
gabibguti: IMO they are still useful when understanding the function.

e2e/pinned_dependencies_test.go
gabibguti: The repo being tested, `ossf-tests/scorecard-check-pinned-dependencies-e2e`, has:
- 0 GHA Third-party actions
- 1+ GHA GitHub-owned actions, all unpinned
- 0 npm installs
- 1+ go installs, all pinned
- and 1+ other dependencies types, all unpinned

The total Info logs decreased because previously we had logs for:
- "all Third-party actions are pinned"
- "all npm installs are pinned"
- "all go installs are pinned"
These "all pinned" Info logs were removed. Now, we only have Warn logs for "unpinned at XYZ snippet" and Debug logs for parsing errors.

checks/raw/shell_download_validate.go
gabibguti: Ah yes! It is a bug.

checks/raw/shell_download_validate.go
gabibguti: My understanding is that all current cases of `downloadThenRun` identified by Scorecard are considered unpinned.

Here's a related issue where Joyce comments that yum, wget and curl generate warnings but cannot be pinned.
https://github.com/ossf/scorecard/issues/2578

gabibguti: I believe it's correct to hardcode as false.

e2e/pinned_dependencies_test.go
gabibguti: Adding some explanation here, I remove the "all pinned" Info logs, since now the proportion or pinned vs total is what matters, not only if all dependencies were pinned or not.

gabibguti: https://github.com/ossf/scorecard/pull/3436#issuecomment-1705687150

gabibguti:
Ok, ready to take a look again!

checker/check_result.go
```diff
@@ -50,6 +51,10 @@ const (
 	DetailDebug
 )
 
+// ErrSuccessTotal indicates a runtime error because number of success cases should
+// be smaller than the total cases to create a proportional score.
+var ErrSuccessTotal = errors.New("unexpected number of success is higher than total")
```
spencerschrock, line 56: could probably be unexported, thoughts?

checker/check_result.go
```diff
@@ -97,6 +110,42 @@ func CreateProportionalScore(success, total int) int {
...
+	for _, score := range scores {
+		if score.Success > score.Total {
+			return InconclusiveResultScore, fmt.Errorf("%w: %d, %d", ErrSuccessTotal, score.Success, score.Total)
+		}
+		if score.Total != 0 {
+			noScoreGroups = false
+		} else {
+			// Group with 0 total, does not count for score
+			continue
+		}
```
spencerschrock, line 129: this is super nit and optional, but in general [keep the happy path left](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88).

```go
if score.Total == 0 {
    continue // Group with 0 total, does not count for score
}
noScoreGroups = false
```

spencerschrock (approved):
Looks good, just some small optional things.

spencerschrock:
I plan on running a quick A/B of scoring changes to make sure everything looks good, and following up with my earlier request for feedback from the other maintainers.

checker/check_result.go
gabibguti: How do I make it unexported? Should I declare it inside CreateProportionalScoreWeighted function?

spencerschrock: rename it to `errSuccessTotal`

gabibguti: Learned. I didn't know about that.

spencerschrock:
> I plan on running a quick A/B of scoring changes to make sure everything looks good

In general, I think this is working as intended. Across a random selection of 10 popular repos, we see 6 of them drop to a score of 0 as none of their dependencies are pinned:
```
-1: 0 -> 0
 0: 0 -> 6
 1: 1 -> 0
 2: 1 -> 1
 3: 0 -> 0
 4: 1 -> 1
 5: 1 -> 0
 6: 0 -> 0
 7: 2 -> 0
 8: 2 -> 0
 9: 0 -> 0
10: 2 -> 2
```

For the repos that still score non-zero, should there be any `Info` logging for ecosystems we do find? For example
```console
go run main.go --repo github.com/apache/commons-codec --checks Pinned-Dependencies --format json --show-details | jq
{
  "date": "2023-09-18T13:41:25-07:00",
  "repo": {
    "name": "github.com/apache/commons-codec",
    "commit": "585497f09b026f6602daf986723a554e051bdfe6"
  },
  "scorecard": {
    "version": "",
    "commit": "unknown"
  },
  "score": 10,
  "checks": [
    {
      "details": null,
      "score": 10,
      "reason": "all dependencies are pinned",
      "name": "Pinned-Dependencies",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies",
        "short": "Determines if the project has declared and pinned the dependencies of its build process."
      }
    }
  ],
  "metadata": null
}
```

> and following up with my earlier request for feedback from the other maintainers.

There was no objection from other maintainers.

gabibguti:
> For the repos that still score non-zero, should there be any Info logging for ecosystems we do find? For example

Yes. I will add an Info log for "2 out of 10 go install dependencies pinned" for ecosystems we do find. So we have the information if all dependencies of an ecossystem are pinned, and how much work was that. I will address this tomorrow.

checks/evaluation/pinned_dependencies.go
```diff
@@ -208,11 +209,20 @@ func addWorkflowPinnedResult(rr *checker.Dependency, w *worklowPinningResult, is
 	}
 }
 
-func createScoreForGitHubActionsWorkflow(wp *worklowPinningResult) []checker.ProportionalScoreWeighted {
+func logPinnedResult(dl checker.DetailLogger, p pinnedResult, name string) {
+	dl.Info(&checker.LogMessage{
+		Text: fmt.Sprintf("%3d out of %3d %s dependencies pinned", p.pinned, p.total, name),
```
spencerschrock, line 214: opinion/question: Is the `%3d` verb needed compared to just `%d`?

spencerschrock (approved):
I think this looks good thanks.
```console
go run main.go --repo ossf/scorecard  --checks Pinned-Dependencies --format json --show-details | jq
{
  "date": "2023-09-24T20:13:47-07:00",
  "repo": {
    "name": "github.com/ossf/scorecard",
    "commit": "5a5a6561d65067678990e69dc3d16a711c98c048"
  },
  "scorecard": {
    "version": "",
    "commit": "unknown"
  },
  "score": 9,
  "checks": [
    {
      "details": [
        "Warn: third-party GitHubAction not pinned by hash: .github/workflows/goreleaser.yaml:89: update your workflow using https://app.stepsecurity.io/secureworkflow/ossf/scorecard/goreleaser.yaml/main?enable=pin",
        "Warn: third-party GitHubAction not pinned by hash: .github/workflows/slsa-goreleaser.yml:47: update your workflow using https://app.stepsecurity.io/secureworkflow/ossf/scorecard/slsa-goreleaser.yml/main?enable=pin",
        "Info: 51 out of 51 GitHub-owned GitHubAction dependencies pinned",
        "Info: 46 out of 48 third-party GitHubAction dependencies pinned",
        "Info: 25 out of 25 containerImage dependencies pinned",
        "Info: 1 out of 1 goCommand dependencies pinned"
      ],
```

spencerschrock:
By the way, I did the A/B comparison, and I think these info statements help. Repos in general had a score drop as expected.

More repos had a `-1` score than I anticipated but mainly from our lack of detecting certain ecosystems (e.g. yarn, which would be handled in other PRs). I'll get to merging this today, but trying to get some higher priority fixes in first.

checks/evaluation/pinned_dependencies.go
spencerschrock: This is fine for now, not sure how much it helps with readability but we can always remove in the future if needed.

gabibguti:
Good to know, thanks for the feedback!

PR #3440, opened 2023-09-01
:seedling: Remove old go.mod replaces

spencerschrock:

dependency `replace` cleanup

We have `replace` directives from long ago. I think part of this is due to old dependabot resolution behavior:
- [blog post](https://github.blog/changelog/2023-03-07-dependency-graph-removes-go-sum-support/)
- [related discussion](https://www.github.com/dependabot/dependabot-core/issues/4740):

All of these `replace` directives are either for older versions than are otherwise specified, or aren't even in the project's go.mod anymore.

While some of these were introduced to force changes in go.sum, I don't think they were ever true dependencies. If you look at the deps.dev screenshot from one of the PRs (https://github.com/ossf/scorecard/pull/2675) today, it no longer has the vuln listed for the old version of scorecard from that time. https://deps.dev/go/github.com%2Fossf%2Fscorecard%2Fv4/v4.10.2

With all replace directives gone, Scorecard can be installed via `go install` again.

##### Top level module
Not in `go.mod` or `go.sum`:
- github.com/gorilla/handlers
- github.com/ulikunitz/xz

No longer used (live in `go.sum` only):
- github.com/buger/jsonparser
- github.com/dgrijalva/jwt-go
- github.com/miekg/dns
- github.com/satori/go.uuid
- github.com/emicklei/go-restful/v2

Replace was ineffective:
- github.com/docker/distribution (specified v2.8.2, but so does our go.mod)
- github.com/opencontainers/image-spec (specified v1.0.2, but go.mod declares v1.1.0-rc3, so this was a downgrade)

##### Tools module
Ineffective:
- github.com/sigstore/cosign (we use v2 now, v1 not used)

Replace was harmful:
- github.com/theupdateframework/go-tuf (specified v0.3.0, which is a vulnerable version. gomod declared v0.5.2, a patched version)

NONE

`osv-scanner` (v1.3.6) doesn't detect any vulns for these changes. Just the aws one, which is uncalled:
```
osv-scanner --experimental-call-analysis -r .
```
| OSV URL                      | CVSS | ECOSYSTEM | PACKAGE                   | VERSION  | SOURCE       |
|------------------------------|------|-----------|---------------------------|----------|--------------|
| Uncalled vulnerabilities     |      |           |                           |          |              |
| https://osv.dev/GO-2022-0646 |      | Go        | github.com/aws/aws-sdk-go | 1.44.314 | go.mod       |
| https://osv.dev/GO-2022-0646 |      | Go        | github.com/aws/aws-sdk-go | 1.44.314 | tools/go.mod |

the `release-note`

```release-note

```


spencerschrock:
Leaving as a draft while I confirm some assumptions with the deps.dev team.

spencerschrock:
I believe this is an example of [module graph pruning](https://go.dev/ref/mod#graph-pruning). Since none of these vulnerable modules are in our `go.mod`, we have nothing to worry about. (`osv-scanner` and `govulncheck` also agree).

> Since a go 1.17 go.mod file includes a [require directive](https://go.dev/ref/mod#go-mod-file-require) for every dependency needed to build any package or test in that module, the pruned module graph includes all of the dependencies needed to go build or go test the packages in any dependency explicitly required by the [main module](https://go.dev/ref/mod#glos-main-module). A module that is not needed to build any package or test in a given module cannot affect the run-time behavior of its packages, so the dependencies that are pruned out of the module graph would only cause interference between otherwise-unrelated modules.

Since our go.mod currently specifies 1.19, only dependencies in our go.mod can affect our build. These modules do still show up in `go list -m all`, but that's expected:
> Modules whose requirements have been pruned out still appear in the module graph and are still reported by go list -m all: their [selected versions](https://go.dev/ref/mod#glos-selected-version) are known and well-defined, and packages can be loaded from those modules (for example, as transitive dependencies of tests loaded from other modules).

raghavkaul (approved):
How did these dependencies affect `go install` behavior? Other than that, lgtm.

raghavkaul:
> > How did these dependencies affect `go install` behavior? Other than that, lgtm.
> 
> `replace` directives aren't supported by `go install`: https://www.github.com/golang/go/issues/44840#issuecomment-792270485 https://www.github.com/golang/go/issues/44840#issuecomment-1651863470

Gotcha. Would it make sense to have a `go install` integration test?

spencerschrock:
> Gotcha. Would it make sense to have a `go install` integration test?

As far as I know, it's only a matter of "are there replace directives?".  So if anything, we should just avoid using them (or a non-blocking linter to do it for us?)

PR #3458, opened 2023-09-06
:sparkles: scdiff: add basic stats command to count scores by buckets

spencerschrock:

feature

The `stats command:
```
Summarize stats for a golden file

Usage:
  scdiff stats [flags] FILE

Flags:
  -c, --check string   Analyze breakdown of a single check
  -h, --help           help for stats
```

Which currently produces a table:
```console
 Aggregate Score Count
              -1     0
               0     0
               1     0
               2     0
               3     0
               4     2
               5     1
               6     2
               7     4
               8     1
               9     0
              10     0
```

Related to #2462, 5/n

Any thoughts on output format? I tried doing a simple tabwriter approach, but happy to go simpler (csv?) or more complicated if warranted.

the `release-note`

```release-note

```

### DIFF
--- a/cmd/internal/scdiff/app/stats.go
+++ b/cmd/internal/scdiff/app/stats.go
@@ -59,35 +59,45 @@ var (
 	errInvalidScore    = errors.New("invalid score")
 )
 
-func calcStats(input io.Reader, output io.Writer) error {
+// countScores quantizes the scores into 12 buckets, from [-1, 10]
+// If a check is provided, that check score is used, otherwise the aggregate score is used.
+func countScores(input io.Reader, check string) ([12]int, error) {
 	var counts [12]int // [-1, 10] inclusive
 	var score int
 	scanner := bufio.NewScanner(input)
-	scanner.Buffer(nil, 1024*1024)
+	scanner.Buffer(nil, 1024*1024) // TODO, how big is big enough?
 	for scanner.Scan() {
 		result, aggregateScore, err := pkg.ExperimentalFromJSON2(strings.NewReader(scanner.Text()))
 		if err != nil {
-			return fmt.Errorf("parsing result: %w", err)
+			return [12]int{}, fmt.Errorf("parsing result: %w", err)
 		}
-		if statsCheck == "" {
+		if check == "" {
 			score = int(aggregateScore)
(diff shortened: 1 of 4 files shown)


cmd/internal/scdiff/app/stats.go
```diff
@@ -0,0 +1,129 @@
...
+	statsCheck string
+	statsCmd   = &cobra.Command{
+		Use:   "stats [flags] FILE",
+		Short: "Summarize stats for a golden file",
+		Long:  `Summarize stats for a golden file`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 {
+				return errMissingInputFiles
+			}
+			f1, err := os.Open(args[0])
```
raghavkaul, line 53: Can you make this a required `StringVar` with cobra?

cmd/internal/scdiff/app/stats.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+			score = result.Checks[i].Score
+		}
+		if score < -1 || score > 10 {
+			return [12]int{}, errInvalidScore
+		}
+		bucket := score + 1 // score of -1 is index 0, score of 0 is index 1, etc.
+		counts[bucket]++
+	}
+	if err := scanner.Err(); err != nil {
```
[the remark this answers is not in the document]
spencerschrock: No.

https://pkg.go.dev/bufio#Scanner.Scan
> After Scan returns false, the Err method will return any error that occurred during scanning, except that if it was io.EOF, Err will return nil

cmd/internal/scdiff/app/stats.go
spencerschrock: Hmm, I was taking a `diff` approach, where the file is expected as an argument but isn't a flag.
It seems like there are better ways of doing what I was trying to do:
https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#positional-and-custom-arguments

spencerschrock: applied this change using the cobra arg validation, if you could please take another look

raghavkaul (approved):
LGTM!

Issue #3439, opened 2023-09-01
BUG: High Signed-Releases score given to project that only signed releases long time ago


diogoteles08:
**Describe the bug**
I've bumped into a case of a repository that have published some signed releases years ago, but all their last 10 releases are not signed. In this case, Scorecard is still granting a 8/10 score on Signed-Releases

**Reproduction steps**
1. Run the following command to run Scorecard on project github.com/AcademySoftwareFoundation/openexr
```
scorecard --repo=http://github.com/AcademySoftwareFoundation/openexr --checks=Signed-Releases --show-details --format=json | jq .
```
2. Note that the project scores 8/10 on Signed releases because of the signed artifacts of he releases v2.4.2, v2.5.2, v2.5.3 , but the current version of the project is v3.2.0, and there were several different releases between them.

**Expected behavior**
The score 8/10 is given to any project that signs their releases but don't emit a provenance. However, this check should consider mostly the most recents releases, not only old ones.

spencerschrock:
Possible duplicate of #2169?

raghavkaul:
The last release with any signing artifacts was v2.5.3, since then the releases are source code-only (only tar/zip files, which GitHub creates automatically), so we don't require them to be signed, since we implicitly trust GitHub/GitLab to zip correctly. Opened a PR to clarify this in our docs.

diogoteles08:
Oh that makes sense, thanks for letting us know and creating the PR!

PR #3469, opened 2023-09-11
📖 Update docs for Signed-Releases check

raghavkaul:
Update docs to clarify that the Signed-Releases check ignores source code-only releases.

Addresses #3439.


diogoteles08:
Thanks for the PR! It indeed makes things clearer.

I'd just recommend a rephrase of the initial text, because in that version the verb "looks" can provoke some readers (I felt it hehe) to think it refers to "releases". Also I was not familiar with the "bundle", so I'd also change to something simpler.

My suggestion: `
This check looks for the most recent release associated to an artifact. It ignores the source code-only releases that are created automatically by GitHub.
`

Initial text: `
This check ignores source code-only releases that are bundled by GitHub and looks for the most recent release with a associated artifact.
`

docs/checks/internal/checks.yaml
```diff
@@ -630,6 +630,8 @@ checks:
       If a signature is found in the assets for each release, a score of 8 is given.
       If a [SLSA provenance file](https://slsa.dev/spec/v0.1/index) is found in the assets for each release (*.intoto.jsonl), the maximum score of 10 is given.
 
+      This check looks for the most recent releases associated with an artifact. It ignores the source code-only releases that are created automatically by GitHub.
```
spencerschrock, line 633: there may be some confusion about what the most recent means. I think it's 30 for github by defualt?

spencerschrock (approved):

docs/checks/internal/checks.yaml
diogoteles08: Good point, it's better to specify the number of days

spencerschrock: I don't think it's related to the number of days currently.

PR #3409, opened 2023-08-21
✨ checks/evaluation logs findings

laurentsimon:
closes https://github.com/ossf/scorecard/issues/3406

```release-note
Let checks/evaluation logs findings
```

PR #3473, opened 2023-09-12
:sparkles: Add additional fuzzing probes

DavidKorczynski:

Feature

There is no checking of C, CPP, Rust, Java or Python fuzzing.

There is checking of C, CPP, Rust, Swift, Java or Python fuzzing.

Ref: https://github.com/ossf/scorecard/issues/897

This adds additional changes to the fuzzing logic in order to support the new probes:

- Adds the ability to have multiple file extensions checked. This is needed for CPP fuzzing and will likely also be impacting others down the line.
- Removes a check for test files `if isTestdataFile(filepath) {` in `OnMatchingFileContentDo`. This is blocking existing fuzzing checks in that often fuzzers will be located under `*test*` directories, but the check avoids checking these folders. This is likely causing a number of false negatives. Furthermore, the documentation of this function says it's matched against all functions but this is not the case as some are excluded. I removed it now as the doc is then right + fuzzing will be more accurately detected.

the `release-note`

```release-note
Adds fuzzing probes for C, CPP, Python, Rust and Java.
```


checks/fileparser/listing.go
```diff
@@ -76,10 +76,6 @@ func OnMatchingFileContentDo(repoClient clients.RepoClient, matchPathTo PathMatc
 	onFileContent DoWhileTrueOnFileContent, args ...interface{},
 ) error {
 	predicate := func(filepath string) (bool, error) {
-		// Filter out test files.
-		if isTestdataFile(filepath) {
-			return false, nil
-		}
```
spencerschrock, line 82: This is going to have changes beyond this check. E.g. we don't penalize test binaries in Binary-Artifacts.

If it makes sense to look at test files, I think it's better to have another function which does that. We can extract the common logic so we don't have two implementations.

DavidKorczynski: I wrapped it in a conditional, wdyt?

inferno-chromium:
This would be nice to have. @oliverchang @jonathanmetzman - can you please review the diff as well and suggest any improvements/gaps to address ?

DavidKorczynski:
> This would be nice to have. @oliverchang @jonathanmetzman - can you please review the diff as well and suggest any improvements/gaps to address ?

It would be great to have your thoughts on this as well https://github.com/ossf/scorecard/issues/3475

checks/evaluation/fuzzing_test.go
```diff
@@ -47,6 +47,26 @@ func TestFuzzing(t *testing.T) {
...
+					},
+					{
+						Probe:   "fuzzedWithRustCargofuzz",
+						Outcome: finding.OutcomeNegative,
+					},
+					{
+						Probe:   "fuzzedWithJavaJazzerFuzzer",
+						Outcome: finding.OutcomeNegative,
+					},
 					{
```
jonathanmetzman, line 70: How about swift and javascript too?

checks/raw/binary_artifact.go
```diff
@@ -54,7 +54,7 @@ func BinaryArtifacts(c clients.RepoClient) (checker.BinaryArtifactData, error) {
 	err := fileparser.OnMatchingFileContentDo(c, fileparser.PathMatcher{
```
jonathanmetzman, line 54: I probably would use a second function that calls this one rather than change all the callers.

checks/evaluation/fuzzing.go
```diff
@@ -35,6 +40,11 @@ func Fuzzing(name string,
 	expectedProbes := []string{
 		fuzzedWithClusterFuzzLite.Probe,
 		fuzzedWithGoNative.Probe,
+		fuzzedWithPythonAtheris.Probe,
```
jonathanmetzman, line 43: Might be good to add a check OR TODO for github.com/google/fuzztest

checks/raw/binary_artifact.go
spencerschrock: Agreed, that's what I had intended with https://github.com/ossf/scorecard/pull/3473#discussion_r1323343891

checks/evaluation/fuzzing_test.go
spencerschrock: javascript has some code already to look for `fast-check`, which was added as part of #3073

checks/raw/fuzzing.go
```diff
@@ -80,7 +85,7 @@ var languageFuzzSpecs = map[clients.LanguageName]languageFuzzConfig{
 	//
 	// This is not an exhaustive list.
 	clients.Haskell: {
-		filePattern: "*.hs",
+		filePatterns: []string{"*.hs"},
```
spencerschrock, line 88: can you add `*.lhs` as well? This was a shortcoming back when #2843 was merged

checks/raw/fuzzing.go
```diff
@@ -105,14 +110,55 @@ var languageFuzzSpecs = map[clients.LanguageName]languageFuzzConfig{
 				"and test that specific properties are satisfied."),
 	},
 	clients.TypeScript: {
-		filePattern: "*.ts",
+		filePatterns: []string{"*.ts"},
```
spencerschrock, line 113: does it make sense to merge JS and TS now that multiple patterns are supported? or keep them separate?

checks/raw/fuzzing.go
```diff
@@ -105,14 +110,55 @@ var languageFuzzSpecs = map[clients.LanguageName]languageFuzzConfig{
...
 		funcPattern: `(from\s+['"]fast-check['"]|require\(\s*['"]fast-check['"]\s*\))`,
 		Name:        fuzzerPropertyBasedTypeScript,
 		Desc: asPointer(
 			"Property-based testing in TypeScript generates test instances randomly or exhaustively " +
 				"and test that specific properties are satisfied."),
 	},
+	clients.Python: {
+		filePatterns: []string{"*.py"},
+		// Look for direct imports of fast-check.
+		funcPattern: `import atheris`,
```
spencerschrock, line 124: It's been a while since I've done python. How unique are these package names? Is this a good signature, in terms of false positives/negatives?

also all of these comments will need to be changed:
> // Look for direct imports of fast-check.

checks/raw/fuzzing.go
```diff
@@ -171,6 +217,7 @@ func Fuzzing(c *checker.CheckRequest) (checker.FuzzingData, error) {
 	prominentLangs := getProminentLanguages(langs)
 	for _, lang := range prominentLangs {
 		usingFuzzFunc, files, e := checkFuzzFunc(c, lang)
+		fmt.Print(files)
```
spencerschrock, line 220: leftover print statements should be removed

checks/raw/fuzzing.go
```diff
@@ -254,22 +301,30 @@ func checkFuzzFunc(c *checker.CheckRequest, lang clients.LanguageName) (bool, []
...
+		matcher := fileparser.PathMatcher{
+			Pattern:       filePattern,
+			CaseSensitive: false,
+		}
+		data.pattern = funcPattern
+		err := fileparser.OnMatchingFileContentDo(c.RepoClient, matcher, getFuzzFunc, false, &data)
+		if err != nil {
+			return false, nil, fmt.Errorf("error when OnMatchingFileContentDo: %w", err)
+		}
+		if len(data.files) != 0 {
```
spencerschrock, line 317: if statement here is redundant? I think `data.files...` will handle scenarios where data.files is empty or nil

checks/raw/fuzzing.go
```diff
@@ -254,22 +301,30 @@ func checkFuzzFunc(c *checker.CheckRequest, lang clients.LanguageName) (bool, []
...
-	matcher := fileparser.PathMatcher{
-		Pattern:       filePattern,
-		CaseSensitive: false,
-	}
-	data.pattern = funcPattern
-	err := fileparser.OnMatchingFileContentDo(c.RepoClient, matcher, getFuzzFunc, &data)
-	if err != nil {
-		return false, nil, fmt.Errorf("error when OnMatchingFileContentDo: %w", err)
+	filePatterns, funcPattern := pattern.filePatterns, pattern.funcPattern
+	dataFiles := []checker.File{}
```
spencerschrock, line 305: nit: can we use `var` style for zero vals inits? Or is there some intended initialization I'm missing
```
var dataFiles []checker.File
```

checks/raw/fuzzing.go
```diff
@@ -322,7 +377,7 @@ func getProminentLanguages(langs []clients.Language) []clients.LanguageName {
 	// Languages that have lines of code above average will be considered prominent.
 	ret := []clients.LanguageName{}
 	for lName, loC := range langMap {
-		if loC >= avgLoC {
+		if loC >= avgLoC/4.0 {
```
spencerschrock, line 380: where is this magic 4.0 coming from?

checks/raw/fuzzing_test.go
```diff
@@ -274,7 +274,10 @@ func Test_fuzzFileAndFuncMatchPattern(t *testing.T) {
 			if !ok && !tt.wantErr {
 				t.Errorf("retrieve supported language error")
 			}
-			fileMatchPattern := langSpecs.filePattern
+			fileMatchPattern := ""
```
spencerschrock, line 277: same nit: 
```go
var fileMatchPattern string
```

checks/fuzzing_test.go
```diff
@@ -76,7 +76,7 @@ func TestFuzzing(t *testing.T) {
 			},
 			wantErr: false,
 			expected: scut.TestReturn{
-				NumberOfWarn:  6,
+				NumberOfWarn:  11,
```
spencerschrock, line 79: might be a good time to go back and add the filtering of negative findings to the fuzzing check, which got added in #3409
e.g for another check the code was changed to something like:
```go
for i := range findings {
		f := &findings[i]
		if f.Outcome == finding.OutcomePositive {
			// Log all findings except the negative ones.
			checker.LogFindings(nonNegative(findings), dl)
			return checker.CreateMaxScoreResult(name, "update tool detected")
		}
}
```

basically we are assuming a project only needs one fuzzing tool (this does fall apart a little for large projects)

checks/raw/fuzzing.go
spencerschrock: I think C and C++ have a similar question. Some fuzzers will support both, i assume some fuzzers might be specific to one vs the other?

spencerschrock (commented):
left some initial thoughts. I think the diff will be smaller once all the modifications to existing callers of matching file contents are reversed.

checks/raw/fuzzing.go
DavidKorczynski: I think it's a good signature. Going through https://github.com/search?q=import+atheris&type=code I only find fuzzing-relevant code using this.

checks/raw/binary_artifact.go
DavidKorczynski: The problem is the function does not do what its documentation and name suggest. The documentation says: ```// OnMatchingFileContentDo matches all files ... and on every successful match, runs onFileContent fn on the file's contents.```. However, the function does not match all files as there is filtering going on, which at the moment isn't clear until reading the actual loop that does the filtering -- and it was confusing to debug as someone new to the code.

Adding the argument and changing the callers is an effort to correct this.

I saw the options as:
1) Change the name of the function + change the documentation `OnMatchingFileContentDo` --> `OnMatchingNonTestFileContentDo` keep the argument list.
2) Add arguments to the function to make it more obvious that filtering is happening and make it possible to avoid the filtering
3) Add a new function that allows for actually matching all files and keep the semantics of `OnMatchingFileContentDo` as is (or perhaps minor refactoring to avoid duplicate logic), but renaming it to indicate the matching happens on a subset of files.

I went for 2, but all of the above options require changes to all the callers.

checks/raw/fuzzing.go
DavidKorczynski: yep, fixed.

checks/raw/fuzzing.go
DavidKorczynski: Minor empirical testing. I ran it through a testing repo where I was confused why some fuzzers present wasn't showing up. The cause was the project had 2 languages and due to the language with fewer lines of code being the one with the fuzzer it wasn't showing up. Then, I didn't see a strong argument for why we need the average and would like to lower it significantly. Examples where the average is relevant is e.g. https://github.com/Instagram/LibCST where the native code in Rust has ~14% of the coverage, and Python has the rest. What was the reason for having the average mark?

DavidKorczynski:
> once all the modifications to existing callers of matching file contents are reversed.

Please consider the logic in https://github.com/ossf/scorecard/pull/3473#discussion_r1329938618 and let me know what you prefer

checks/evaluation/fuzzing_test.go
DavidKorczynski: I suggest doing Javascript in a follow-up PR because the logic of the fuzzing check isn't set up well for having multiple different fuzzers in a given language. I think it would be nice to avoid adding more complexity to this PR.

DavidKorczynski: Added Swift support. Added jazzer.js as an issue for now and will address in a follow-up PR https://github.com/ossf/scorecard/issues/3491

checks/fuzzing_test.go
DavidKorczynski: Am not sure I understand -- This was already added?

https://github.com/ossf/scorecard/blob/5c93fe63b4d011a5e7e299fd136ec4280479472f/checks/evaluation/fuzzing.go#L54-L56

The reason the numbers are increased in this PR is because in the relevant tests there are no success cases and that case all warnings are kept. Is there perhaps a bug here in that the second use of `LogFindings` does not sort it:

https://github.com/ossf/scorecard/blob/5c93fe63b4d011a5e7e299fd136ec4280479472f/checks/evaluation/fuzzing.go#L55-L60

spencerschrock: I may have just read the diff wrong. If it already calls it, then this can be resolved.

checks/raw/binary_artifact.go
spencerschrock: I would say there's an option 4.

Add a new function, leave the existing function as is (perhaps clarifying the comment) and adding a new function called `OnMatchingFileContentWithTestDataDo` (or similar). Which from a backwards compatibility standpoint is the best.

checks/raw/fuzzing.go
spencerschrock: Written before my time, but my guess is we were concerned about if the main language was fuzzed.

If the Rust code is fuzzed, but the Python isn't does that project deserve a 10?

spencerschrock (approved):
In general, I think this is fine. Is your plan to do the file content matching with testdata in a follow up PR?

checks/raw/fuzzing.go
spencerschrock: If keeping the 4.0, can you make it a constant with a constant expressing this?

DavidKorczynski:
> Is your plan to do the file content matching with testdata in a follow up PR?

yes

checks/raw/fuzzing.go
DavidKorczynski: > Written before my time, but my guess is we were concerned about if the main language was fuzzed.
> If the Rust code is fuzzed, but the Python isn't does that project deserve a 10?

I think in the current spirit of the fuzzing checks then the answer is yes it deserves a 10. However, I think medium-term it makes sense to consider a more refined fuzzing assessment, including some of the points considered in https://github.com/ossf/scorecard/issues/3475

checks/raw/fuzzing_test.go
```diff
@@ -274,7 +274,10 @@ func Test_fuzzFileAndFuncMatchPattern(t *testing.T) {
 			if !ok && !tt.wantErr {
 				t.Errorf("retrieve supported language error")
 			}
-			fileMatchPattern := langSpecs.filePattern
+			var fileMatchPattern string
+			if len(langSpecs.filePatterns) > 0 {
+				fileMatchPattern = langSpecs.filePatterns[0]
+			}
```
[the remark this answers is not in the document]
DavidKorczynski: Done!

checks/raw/binary_artifact.go
DavidKorczynski: I'll address this in a follow-up PR. Left out changing the logic of the file matching for now.

checks/raw/fuzzing.go
spencerschrock: Sorry to belabor this, but naming the constant `fourPointZero` doesn't help with readability compared to just using 4.0, which is what my initial comment was about.

I was thinking something more similar to:
```go
	// Calculate the average lines of code in the current repo.
	// This var can stay as an int, no need for a precise float value.
	avgLoC := totalLoC / numLangs
	// Languages that have lines of code above 1/8th will be considered prominent.
	prominentThreshold = avgLoC / 4.0
	ret := []clients.LanguageName{}
	for lName, loC := range langMap {
		if loC >= prominentThreshold {
```

DavidKorczynski: Set it to what you pasted

PR #3477, opened 2023-09-12
🐛 GitLab: Fix URI() used for OSS-Fuzz detection

### LINKED ISSUE #3476: BUG: Fuzzing check doesn't detect OSS-Fuzz for GitLab repos
**Describe the bug**
GitLab repos that are registered in OSS-Fuzz don't get Fuzzing credit in Scorecard.

Ex. libtiff in deps.dev: https://deps.dev/project/gitlab/libtiff%2Flibtiff

Has a project.yaml in oss-fuzz here: https://github.com/google/oss-fuzz/blob/master/projects/libtiff/project.yaml
But this repo gets 0/10 in the Fuzzing check.

raghavkaul:
Update URI() for GitLab repo client. Previously, it used the numeric project ID, which isn't the actual URI.

Fixes #3476.


e2e/fuzzing_test.go
```diff
@@ -60,6 +61,38 @@ var _ = Describe("E2E TEST:"+checks.CheckFuzzing, func() {
 			Expect(repoClient.Close()).Should(BeNil())
 			Expect(ossFuzzRepoClient.Close()).Should(BeNil())
 		})
+		It("Should return use of OSS-Fuzz - GitLab", func() {
```
spencerschrock, line 64: Does it matter if this runs with the GitHub integration or the GitLab ones, or both? 
Since this uses the ossfuzz client, I don't think so, but I know some of the gitlab repo init stuff reaches out to the network.

Feel free to resolve, I think the answer is no, but wanted to post.

spencerschrock (approved):

e2e/fuzzing_test.go
raghavkaul: This test shouldn't make a request to either GitHub or GitLab (OSS Fuzz results are fetched from the GCS bucket). But it does connect to the network, and is a GitLab-specific test, so I opted to leave it as an E2E test with the ` - GitLab` marker.

PR #3483, opened 2023-09-14
🌱 workflows/stale: Update workflow to increase operations-per-run to process more issues

afmarcum:
Increases operations-per-run from default of 30 to 100 to test whether older issues will be marked as inactive

Issues are only marked as inactive as they reach the 60 day trigger, but are not being marked if they have been at 60+ days of inactivity
Older issues will be marked as inactive as well

NONE

No

```NONE

```

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -43,5 +43,5 @@ jobs:
         days-before-pr-stale: '10'
         days-before-pr-close: '20'
         days-before-issue-stale: '60'
-        days-before-issue-close: '67'
-        operations-per-run: '400'
+        days-before-issue-close: '7'
+        operations-per-run: '100'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -44,3 +44,4 @@ jobs:
         days-before-pr-close: '20'
         days-before-issue-stale: '60'
         days-before-issue-close: '67'
+        operations-per-run: '400'


spencerschrock (approved):
> GitHub has a [rate limit](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting) and if reached will block these API calls for one hour (or API calls from other actions using the same user (a.k.a.: the github-token from the [repo-token](https://github.com/actions/stale/tree/main#repo-token) option)).
This option helps you to stay within the GitHub rate limits, as you can use this option to limit the number of operations for a single run.

I'm wondering if 400 is too aggressive. We can always try it and find out. And if it's a problem we can always run the action more than once per day to help get through more. (it's currently configured to once at 1:30 am UTC)

> If the action ends because of [operations-per-run](https://github.com/actions/stale/tree/main#operations-per-run) then the next run will start from the first unprocessed issue skipping the issues processed during the previous run(s). The state is reset when all the issues are processed. This should be considered for scheduling workflow runs.
> 
> The saved state lifetime is the same as the [actions cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) configured for the repo.

.github/workflows/stale.yml
```diff
@@ -44,3 +44,4 @@ jobs:
         days-before-pr-close: '20'
         days-before-issue-stale: '60'
         days-before-issue-close: '67'
```
spencerschrock, line 46: I'm pretty sure this value is wrong though:

> The number of days to wait to close an issue or a pull request after it being marked stale

We should set this as 7 right? Not 67?

afmarcum: Yes. I thought it was the total days, not days since the stale label was added.

Issue #3388, opened 2023-08-14
Code coverage hit after updating to go1.21


justaugustus:
We took a code coverage hit after updating to go1.21 in https://github.com/ossf/scorecard/pull/3387.
Not sure if this is a blip, but worth investigating.

## [Codecov](https://app.codecov.io/gh/ossf/scorecard/pull/3387?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ossf) Report
> Merging [#3387](https://app.codecov.io/gh/ossf/scorecard/pull/3387?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ossf) (9b7c9ae) into [main](https://app.codecov.io/gh/ossf/scorecard/commit/198aa963ebf52922dcd8ed9096e43b0b0ee6d3ab?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ossf) (198aa96) will **decrease** coverage by `6.22%`.
> The diff coverage is `n/a`.

<details><summary>Additional details and impacted files</summary>

```diff
@@            Coverage Diff             @@
##             main    #3387      +/-   ##
==========================================
- Coverage   74.30%   68.08%   -6.22%     
==========================================
  Files         178      175       -3     
  Lines       12876    12856      -20     
==========================================
- Hits         9567     8753     -814     
- Misses       2766     3593     +827     
+ Partials      543      510      -33     
```

</details>

_Originally posted by @codecov[bot] in https://github.com/ossf/scorecard/issues/3387#issuecomment-1677822921_

spencerschrock:
This is due to e2e tests which run only on push to main, due to their use of a PAT. So all PRs currently show this 6% decrease, but have the higher coverage when merged.
cc @naveensrinivasan

naveensrinivasan:
> This is due to e2e tests which run only on push to main, due to their use of a PAT. So all PRs currently show this 6% decrease, but have the higher coverage when merged. cc @naveensrinivasan

Yes, It is the case. I am closing it. Please reopen if you have concerns.

PR #3492, opened 2023-09-19
:seedling: Reduce confusion around codecov check status.

spencerschrock:

- With our current coverage upload setup, it will always show a drop of 6-7% (see #3388). This is also confusing to contributors (https://github.com/ossf/scorecard/pull/3426#issuecomment-1724484499)
- The patch config is broken, so it essentially always passes: `50.00% of diff hit (within 90.00% threshold of 74.34%)`
  - The goal here was probably a target of 90%, not a threshold

The project status is converted to [informational](https://docs.codecov.com/docs/commit-status#informational), so it will always pass. 
> Use Codecov in informational mode. Default is false. If true is specified the resulting status will pass no matter what the coverage is or what other settings are specified. Informational mode is great to use if you want to expose codecov information to other developers in your pull request without necessarily gating PRs on that information.

I still find the patch status helpful (especially the annotations while reviewing), so I'm leaving it as a check which can pass or fail. I've just fixed the threshold, it is now looking for patch coverage to be within 10% of the current code coverage level. Note: this is still a guideline and non-blocking since the check isn't marked as required.

I've validated the file using their recommended approach.

```console
curl -X POST --data-binary @.codecov.yml https://codecov.io/validate
Valid!
```

the `release-note`

```release-note

```

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -23,16 +23,15 @@ coverage:
   status:
     project:
       default:
-        enabled: true
-        # allowed to drop coverage and still result in a "success" commit status
-        threshold: null
+        informational: true
         if_not_found: success
         if_no_uploads: success
         if_ci_failed: error
     patch:
       default:
-        enabled: true
-        threshold: 90%
+        # patch coverage should be within 10% of existing coverage
+        target: auto
+        threshold: 10%
         if_not_found: success
         if_no_uploads: success
         if_ci_failed: error


spencerschrock:
```
codecov/project Successful in 1s — 66.93% (-7.40%) compared to 5c93fe6
```

If this is still confusing, another alternative is to disable the `codecov/project` status entirely, but keep the `codecov/patch` one?  The project code coverage could still be viewed on their site.

laurentsimon (approved):

PR #3501, opened 2023-09-20
:seedling: Switch test import to remove gotest.tools dependency.

spencerschrock:

Fixing a likely typo in test code

We import `gotest.tools/assert/cmp` to compare two string slices. This library is only used in one place.

The comparison now uses `cmp.Equal` from `github.com/google/go-cmp/cmp`, allowing us to eliminate the dependency.

NONE

I moved `github.com/rhysd/actionlint` down to be with the other imports.

the `release-note`

```release-note

```


raghavkaul (approved):
Thanks! This will hopefully get rid of that lint.

PR #3503, opened 2023-09-20
🌱 Monitoring: use opentelemetry for cron metrics

raghavkaul:
Use opentelemetry instead of opencensus


clients/githubrepo/stats/stats.go
```diff
@@ -15,29 +15,50 @@
...
+var once sync.Once
+
+func InitMetrics() error {
+	var err error
+	once.Do(func() {
+		c, err := Meter.Int64Histogram("RetryAfter",
+			metric.WithDescription("Measures the retry delay when dealing with secondary rate limits"),
+			metric.WithUnit("s"))
+		if err != nil {
+			return
```
spencerschrock, line 40: I think there's some unintentional shadowing here since you use `c, err :=` above

clients/githubrepo/stats/stats.go
```diff
@@ -15,29 +15,50 @@
...
-	}
+// // GithubTokens tracks the usage/remaining stats per token per resource-type.
+//
+//	GithubTokens = view.View{
+//		Name:        "GithubTokens",
+//		Description: "Token usage/remaining stats for Github API tokens",
+//		Measure:     RemainingTokens,
+//		TagKeys:     []tag.Key{TokenIndex, ResourceType},
+//		Aggregation: view.LastValue(),
+//	}
```
spencerschrock, line 63: this comment should probably be removed?

checker/check_runner.go
```diff
@@ -96,21 +95,23 @@ func (r *Runner) Run(ctx context.Context, c Check) CheckResult {
...
+	if err != nil {
+		panic(err)
+	}
+
 	if err != nil {
 		l.Warn(&LogMessage{Text: fmt.Sprintf("tag.New: %v", err)})
 	}
 
-	ctx, err = tag.New(ctx, tag.Upsert(stats.RepoHost, r.CheckRequest.Repo.Host()))
+	ctx = context.WithValue(ctx, stats.RepoHost, r.CheckRequest.Repo.Host())
```
spencerschrock, line 107: does it make a difference when you pass the value as an attribute, vs in the context? You do both in this PR.

 Skimming one migration page, I thought otel didnt support context based attributes?

cron/internal/worker/main.go
```diff
@@ -284,23 +279,31 @@ func processRequest(ctx context.Context,
...
-		return nil, fmt.Errorf("error during view.Register: %w", err)
+	// Set the global meter provider.
+	otel.SetMeterProvider(meterProvider)
+
+	// Handle shutdown properly so nothing leaks.
+	defer func() {
+		if err := meterProvider.Shutdown(context.Background()); err != nil {
+			fmt.Printf("couldn't shutdown meterProvider: %v\n", err)
+		}
+	}()
```
spencerschrock, line 301: wouldn't this close the meter provider as soon as the function returns? 

I think the Go idiom you're looking forward is returning a cancel/close/cleanup function, and then the caller can call/defer that when appropriate.

cron/monitoring/exporter.go
```diff
@@ -20,9 +20,11 @@ import (
 	"fmt"
 	"time"
 
-	"contrib.go.opencensus.io/exporter/stackdriver"
-	"contrib.go.opencensus.io/exporter/stackdriver/monitoredresource/gcp"
-	"go.opencensus.io/stats/view"
+	mexporter "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric"
```
spencerschrock, line 23: would `gcpexporter` provide more context?

cron/monitoring/exporter.go
```diff
@@ -38,51 +40,58 @@ const (
...
+func newStackDriverExporter() (metric.Exporter, error) {
+	exp, err := mexporter.New()
 	if err != nil {
-		return nil, fmt.Errorf("error getting stackdriver prefix: %w", err)
+		return nil, fmt.Errorf("couldn't get cloudmetrics exporter: %w", err)
 	}
-	exporter, err := stackdriver.NewExporter(stackdriver.Options{
-		ProjectID:         projectID,
-		MetricPrefix:      prefix,
-		MonitoredResource: gcp.Autodetect(),
```
spencerschrock, line 78: Is this information being lost in the new implementation?

cron/monitoring/exporter.go
```diff
@@ -38,51 +40,58 @@ const (
...
 // GetExporter defines a factory for returning opencensus Exporter.
 // Ensure config.ReadConfig() is called at some point before this function.
-func GetExporter() (Exporter, error) {
+func GetExporter() (metric.Exporter, error) {
 	exporter, err := config.GetMetricExporter()
 	if err != nil {
 		return nil, fmt.Errorf("error during GetMetricExporter: %w", err)
 	}
+
+	fmt.Printf("Using %s exporter for metrics\n", exporter)
```
spencerschrock, line 51: leftover print statement?

checker/check_runner.go
```diff
@@ -96,21 +95,23 @@ func (r *Runner) Run(ctx context.Context, c Check) CheckResult {
 				fmt.Sprintf("requiredType: %s not supported by check %s", fmt.Sprint(unsupported), r.CheckName)))
 	}
 
-	l := NewLogger()
-	ctx, err := tag.New(ctx, tag.Upsert(stats.CheckName, r.CheckName))
+	err := stats.InitMetrics()
+	if err != nil {
+		panic(err)
```
spencerschrock, line 100: is this something that should be setup in an `init`?

spencerschrock (commented):
Left some partial thoughts since this is still a draft. I'm unfamiliar with the migration, so trying to get some details there.

PR #3506, opened 2023-09-21
:sparkles: broaden job matcher for semantic release

secustor:

This PR broaden the job matcher for semantic release. 
That way other execution methods of semantic-release trigger no violation of the TokenPermission check. 

Only if semantic-release is executed using npx is detected as valid use of contents write permissions. 

also runs via pnpm and yarn are detected

NONE

Tested against https://github.com/renovatebot/renovate/blob/main/.github/workflows/build.yml#L577

the `release-note`

```release-note
Detecting yarn and pnpm as valid package managers to run `semantic-release`
```


secustor:
The linter seems to have simply timed out

spencerschrock:
> The linter seems to have simply timed out

New contributors require a maintainer to approve the CI/CD run, which I've pushed. So it's running now

checks/raw/permissions.go
```diff
@@ -473,7 +473,7 @@ func isReleasingWorkflow(workflow *actionlint.Workflow, fp string, pdata *permis
 			// Commonly JavaScript packages, but supports multiple ecosystems
 			Steps: []*fileparser.JobMatcherStep{
 				{
-					Run: "npx.*semantic-release",
+					Run: "(npx|pnpm|yarn).*semantic-release",
```
spencerschrock, line 476: Can you add a test?

I think this would be the easiest one to modify
https://github.com/ossf/scorecard/blob/6aa3bcc7f50e7c0efc293321284db15b1689b1a7/checks/permissions_test.go#L294-L303

spencerschrock (approved):
Thanks

PR #3508, opened 2023-09-22
:seedling: convert binary artifact check to probe

AdamKorcz:

Feature

This PR converts the binary artifacts check to a probe.

It does change the processing of requests in `checks/raw` to align it with the style of other probes.

NONE

NONE

```release-note

```


checks/evaluation/binary_artifacts.go
```diff
@@ -18,36 +18,47 @@ import (
...
 
 	return checker.CreateResultWithScore(name, "binaries present in source code", score)
 }
+
+func validateFindings(findings []finding.Finding, expectedProbes []string) error {
+	if !finding.UniqueProbesEqual(findings, expectedProbes) {
+		return sce.WithMessage(sce.ErrScorecardInternal, "invalid probe results")
+	}
+
+	if len(findings) == 0 {
```
laurentsimon, line 60: I think this is implied by `UniqueProbesEqual` so is not needed

checks/evaluation/binary_artifacts_test.go
```diff
@@ -17,273 +17,164 @@ package evaluation
...
 	tests := []struct {
-		name    string
-		args    args
-		want    checker.CheckResult
-		wantErr bool
+		name     string
+		findings     []finding.Finding
+		result     scut.TestReturn
+		expected []struct {
+			lineNumber uint
```
laurentsimon, line 34: I dont see lineNumber used in the tests

probes/hasBinaryArtifacts/def.yml
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+id: hasBinaryArtifacts
+short: Checks if the project has binary files in its source tree.
+motivation: >
+  Including generated executables in the source repository increases user risk. Many programming language systems can generate executables from source code (e.g., C/C++ generated machine code, Java .class files, Python .pyc files, and minified JavaScript). Users will often directly use executables if they are included in the source repository, leading to many dangerous behaviors.
```
laurentsimon, line 18: maybe we could say binary files are not readable so users can't see what they do, which is dangerous

probes/hasBinaryArtifacts/def.yml
```diff
@@ -0,0 +1,27 @@
...
+implementation: >
+  The implementation looks for the presence of binary files.
```
laurentsimon, line 20: shall we add a list of type we look for?

probes/hasBinaryArtifacts/impl.go
```diff
@@ -0,0 +1,79 @@
...
+func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
+	if raw == nil {
+		return nil, "", fmt.Errorf("%w: raw", uerror.ErrNil)
+	}
+
+	r := raw.BinaryArtifactResults
+
+	// Apply the policy evaluation.
+	if r.Files == nil || len(r.Files) == 0 {
+		return positiveOutcome()
```
laurentsimon, line 41: if this function is needed only here, do we need a separate function? Other probes create the positive result inline.

laurentsimon:
The check also looks for the installation of maven (or gradle) Action to verify the integrity of wrapper binaries. I think we need more than one probe to handle these cases. Suggestion: `hasMavenWrapperBinary`, `hasMavenWrapperActionInstalled`, `hasOtherBinaries`.

probes/hasBinaryArtifacts/impl.go
AdamKorcz: It reads better for me to wrap it. Have removed.

AdamKorcz:
> The check also looks for the installation of maven (or gradle) Action to verify the integrity of wrapper binaries. I think we need more than one probe to handle these cases. Suggestion: `hasMavenWrapperBinary`, `hasMavenWrapperActionInstalled`, `hasOtherBinaries`.

I only see checks for Gradle Wrappers, however, the filtering happens before the files are added to the `BinaryArtifactData`. As such, the current Scorecard check + evaluation does not filter out wrapper binaries. Is that something we should change and convert to probes?

laurentsimon:
> > The check also looks for the installation of maven (or gradle) Action to verify the integrity of wrapper binaries. I think we need more than one probe to handle these cases. Suggestion: `hasMavenWrapperBinary`, `hasMavenWrapperActionInstalled`, `hasOtherBinaries`.
> 
> I only see checks for Gradle Wrappers, however, the filtering happens before the files are added to the `BinaryArtifactData`. As such, the current Scorecard check + evaluation does not filter out wrapper binaries. Is that something we should change and convert to probes?

yes I think that would be useful. I could image end-users wanted to know if there's a gradle wrapper present in the repo or not. I think it's ok to do this in this. @spencerschrock wdut?

spencerschrock:
> > I only see checks for Gradle Wrappers, however, the filtering happens before the files are added to the `BinaryArtifactData`. As such, the current Scorecard check + evaluation does not filter out wrapper binaries. Is that something we should change and convert to probes?
> 
> yes I think that would be useful. I could image end-users wanted to know if there's a gradle wrapper present in the repo or not. I think it's ok to do this in this. @spencerschrock wdut?

I think it's fine to do the filtering on the evaluation side, but would need to double check against the Allstar code. 

> > > The check also looks for the installation of maven (or gradle) Action to verify the integrity of wrapper binaries. I think we need more than one probe to handle these cases. Suggestion: `hasMavenWrapperBinary`, `hasMavenWrapperActionInstalled`, `hasOtherBinaries`.

With regard to probes, I haven't read the PR, but my proposed breakdown was probably going to have multiple probes (although only one would likely be used by the evaluation code):
- free of binary artifacts
- free of binary artifacts (except for the "approved" maven ones)

spencerschrock:
I think this is another example where it makes sense to clarify what a finding is? If it's some data point which optionally has a file location, then would each binary artifact have its own finding (note: this is a little different than the repeated contributor/commit findings). If it's gradle verified it could be a gradle validated finding, otherwise a normal binary finding.

And the policy could later walk both, and choose to either count or ignore the gradle verified ones

checks/binary_artifact.go
```diff
@@ -38,17 +40,22 @@ func init() {
...
-		c.RawResults.BinaryArtifactResults = rawData
+	// Set the raw results.
+	pRawResults := getRawResults(c)
+	pRawResults.BinaryArtifactResults = rawData
+
+	// Evaluate the probes.
+	findings, err := zrunner.Run(pRawResults, probes.BinaryArtifacts)
+	if err != nil {
+		e := sce.WithMessage(sce.ErrScorecardInternal, err.Error())
+		return checker.CreateRuntimeErrorResult(CheckVulnerabilities, e)
```
spencerschrock, line 57: Drive by comment to fix this typo of the wrong check name.

spencerschrock:
> With regard to probes, I haven't read the PR, but my proposed breakdown was probably going to have multiple probes (although only one would likely be used by the evaluation code):
> 
> * free of binary artifacts
> * free of binary artifacts (except for the "approved" maven ones)

I believe this is the outcome of our discussion today. @laurentsimon your comment about different logic for each finding, I don't think that's true so long as the policy picks one. 

e.g. if you have 3 binaries: foo, bar, and gradle-wrapper.jar (verified)

The first probe would complain about all 3.
the second probe would only complain about foo and bar.

Scorecard evaluation code would use the second probe. The first probe would be dangling/uncalled, until someone writes a policy to use it.

laurentsimon:
> > With regard to probes, I haven't read the PR, but my proposed breakdown was probably going to have multiple probes (although only one would likely be used by the evaluation code):
> > 
> > * free of binary artifacts
> > * free of binary artifacts (except for the "approved" maven ones)
> 
> I believe this is the outcome of our discussion today. @laurentsimon your comment about different logic for each finding, I don't think that's true so long as the policy picks one.

I think I was wrong too. So long as there's one finding per binary, it should work. So the main benefit of having different probes would be to accommodate different remediations for each "type" of verification. Today I don't know if we need this flexibility. But if we want to not be blocked in the future, that could be beneficial to do it now. I don't see a strong argument for trying to pack multiple verification types into a single "unverifiedProbe". What is the main motivtion? 

> 
> e.g. if you have 3 binaries: foo, bar, and gradle-wrapper.jar (verified)
> 
> The first probe would complain about all 3. the second probe would only complain about foo and bar.

I thought we agreed the were mutually exclusive? (can re-visit if it's better not to)

> Scorecard evaluation code would use the second probe. The first probe would be dangling/uncalled, until someone writes a policy to use it.

spencerschrock:
> I thought we agreed the were mutually exclusive? (can re-visit if it's better not to)

I thought the binary type was mutually exclusive, but the probes may not be. Can discuss offline if needed.

AdamKorcz:
I have refactored the probe into two probes. Consumers are meant to only consume one of these. The use case would be:

1. `freeOfAnyBinaryArtifacts`: This will return a negative outcome even if the only binary artifact is a gradle wrapper. The finding will include the location and a gradle wrapper file type, that consumers can use to filter on their end.
2. `freeOfUntrustedBinaryArtifacts`: Will not return a negative outcome if the binary artifact is a gradle wrapper. This is a convenient probe for users that want to adopt Scorecards recommendations for trusted binary artifacts.

laurentsimon:
> I have refactored the probe into two probes. Consumers are meant to only consume one of these. The use case would be:
> 
> 1. `freeOfAnyBinaryArtifacts`: This will return a negative outcome for any type of binary artifact including gradle wrappers. The finding will include the location and a gradle wrapper file type, that consumers can use to filter on their end.
> 2. `freeOfUntrustedBinaryArtifacts`: Will not return a negative outcome if the binary artifact is a gradle wrapper. This is a convenient probe for users that want to adopt Scorecards recommendations for trusted binary artifacts.

LGTM. Maybe Untrusted -> Unverified

probes/freeOfUntrustedBinaryArtifacts/def.yml
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+id: freeOfUntrustedBinaryArtifacts
+short: Checks if the project has binary files in its source tree. The probe skips trusted binary files which currently are gradle-wrappers.
+motivation: >
+  Binary files are not readable so users can't see what they do. Many programming language systems can generate executables from source code (e.g., C/C++ generated machine code, Java .class files, Python .pyc files, and minified JavaScript). Users will often directly use executables if they are included in the source repository, leading to many dangerous behaviors.
+implementation: >
+  The implementation looks for the presence of binary files. This is a more permissive probe than "freeOfAnyBinaryArtifacts" which does not trust any binary files.
```
laurentsimon, line 20: nit: `The implementation looks for the presence of binary files that are not "verified". A verified binary is one that ...`

probes/freeOfAnyBinaryArtifacts/def.yml
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+id: freeOfAnyBinaryArtifacts
+short: Checks if the project has any binary files in its source tree.
+motivation: >
+  Binary files are not readable so users can't see what they do. Many programming language systems can generate executables from source code (e.g., C/C++ generated machine code, Java .class files, Python .pyc files, and minified JavaScript). Users will often directly use executables if they are included in the source repository, leading to many dangerous behaviors.
+implementation: >
+  The implementation looks for the presence of binary files. This is a more restrictive probe than "freeOfUntrustedBinaryArtifacts" which excludes trusted binary files.
+outcome:
+  - If the probe finds binary files, it returns a number of negative outcomes equal to the number of binary files found. Each outcome includes a location of the file.
```
laurentsimon, line 22: trust -> verified

probes/freeOfUntrustedBinaryArtifacts/def.yml
laurentsimon, line 20: trust -> verify

laurentsimon (approved):

probes/freeOfUntrustedBinaryArtifacts/def.yml
AdamKorcz: Done in https://github.com/ossf/scorecard/pull/3508/commits/c7b8915f14f4db04664e220b9cb3f3ec2affa4e3

probes/freeOfAnyBinaryArtifacts/def.yml
AdamKorcz: Done in https://github.com/ossf/scorecard/pull/3508/commits/37bfaa6153e09bdae323220dd2825a78f021d82e

checks/raw/binary_artifact.go
```diff
@@ -89,9 +90,10 @@ func excludeValidatedGradleWrappers(c clients.RepoClient, files []checker.File)
 	// Remove Gradle wrapper JARs from files.
 	filterFiles := []checker.File{}
 	for _, f := range files {
-		if filepath.Base(f.Path) != "gradle-wrapper.jar" {
-			filterFiles = append(filterFiles, f)
+		if filepath.Base(f.Path) == "gradle-wrapper.jar" {
+			f.Type = finding.FileTypeGradleWrapper
 		}
+		filterFiles = append(filterFiles, f)
```
spencerschrock, line 96: We're not filtering anything anymore, just labeling. can we just loop in place and modify the type?

```go
for i := range files {
		if filepath.Base(files[i].Path) == "gradle-wrapper.jar" {
			files[i].Type = finding.FileTypeGradleWrapper
		}
```

finding/finding.go
```diff
@@ -40,6 +40,8 @@ const (
 	FileTypeText
 	// FileTypeURL for URLs.
 	FileTypeURL
+	// FileTypeGradleWrapper for gradle-wrapper.jar files.
+	FileTypeGradleWrapper
```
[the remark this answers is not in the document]
spencerschrock: At the very least, this should be `FileTypeVerifiedGradleWrapper`, as it doesnt necessarily apply to all gradle wrappers

probes/freeOfAnyBinaryArtifacts/impl.go
```diff
@@ -0,0 +1,66 @@
...
+func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
+	if raw == nil {
+		return nil, "", fmt.Errorf("%w: raw", uerror.ErrNil)
+	}
+
+	r := raw.BinaryArtifactResults
+	var findings []finding.Finding
+
+	// Apply the policy evaluation.
+	if r.Files == nil || len(r.Files) == 0 {
```
spencerschrock, line 41: nit: `len` works on nil slices 
```
if len(r.Files) == 0 {
```

AdamKorcz: https://github.com/ossf/scorecard/pull/3508/commits/7a63302e1646be71a4d58f82cd3c575f6af0ad42

checks/raw/binary_artifact.go
AdamKorcz: https://github.com/ossf/scorecard/pull/3508/commits/7cb09fcc6543fdebc8ea57e337aa79c0dd326cc6

probes/entries.go
```diff
@@ -121,6 +122,9 @@ var (
 	CIIBestPractices = []ProbeImpl{
 		hasOpenSSFBadge.Run,
 	}
+	BinaryArtifacts = []ProbeImpl{
+		freeOfUnverifiedBinaryArtifacts.Run,
```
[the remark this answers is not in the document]
AdamKorcz: No. We only run one of the two probes in evaluation and make the other optional for consumers to integrate into their own policies.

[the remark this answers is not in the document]
laurentsimon: ah, got it!

Issue #1327, opened 2021-11-22
Feature: more robust way of logging errors in checks


laurentsimon:
Today, we return an inconclusive results (accompanied by an error) if a sub-check catches an error. For example, the Pinned-Dependencies check is currently fairly (too!) large, so a single error makes the entire check return `-1`.

In https://github.com/ossf/scorecard/pull/1324#issuecomment-975709844, we implemented a heuristic to avoid running the sub-check, using the filename. I feel this approach is fragile. Dockerfile templates are common, and are not always called `template`, `tpl`, etc. See https://github.com/ossf/scorecard/issues/710 and https://raw.githubusercontent.com/kubernetes/kubernetes/master/build/server-image/Dockerfile.

It'd be useful to have a more robust solution. I'n not sure what a better solution is. 
Maybe we should create an `Error()` function to log errors in sub-checks without affecting the entire check result. 

Any other ideas?

laurentsimon:
@chrismcgehee @azeemsgoogle @naveensrinivasan

laurentsimon:
relevant for shell parsing https://github.com/ossf/scorecard/issues/1307

chrismcgehee:
I do think we should simply log when we can't parse a file instead of failing the entire check, essentially do what https://github.com/ossf/scorecard/pull/1312 does but for any parsing error. It would be nice if we could examine the files that can't be parsed in case we can improve our code to parse better, but inevitably we're going to encounter files that are formatted incorrectly. Maybe we can keep a list of files that we know are not parseable, and just ignore them.

In regards to skipping template Dockerfiles, I think using the filename is a decent start. The file https://raw.githubusercontent.com/kubernetes/kubernetes/master/build/server-image/Dockerfile is not, strictly speaking, a template; it is a fully valid Dockerfile, just with build arguments. To handle a file like this would should enhance are logic to look for hashes as you suggest in https://github.com/ossf/scorecard/issues/710#issuecomment-975715013, but we shouldn't skip it because it's a template.

laurentsimon:
> I do think we should simply log when we can't parse a file instead of failing the entire check, essentially do what #1312 does but for any parsing error. It would be nice if we could examine the files that can't be parsed in case we can improve our code to parse better, but inevitably we're going to encounter files that are formatted incorrectly. Maybe we can keep a list of files that we know are not parseable, and just ignore them.

that'd be great, yes!

[...]

PR #3515, opened 2023-09-25
🐛 Pinned-Dependencies continues on error

### LINKED ISSUE #3316: BUG: Runtime error on Pinned-Dependencies check causes a -1 on its score
**Describe the bug**
When running Pinned-Dependencies check on [apache/arrow](github.com/apache/arrow) or [apache/beam](github.com/apache/beam), it's returning a -1 and reporting an Internal Error. **It's not the same error on both repository**, but both errors are breaking the execution and disabling users to view any actual result that could have been achieved. 

See above the errors in details:
```
diogoteles:~$ scorecard --repo=https://github.com/apache/beam --show-details --format=json --checks=Pinned-Dependencies | jq .
Error: check runtime error: Pinned-Dependencies: internal error: internal error: unable to determine OS for job: Build python wheels on ${{matrix.arch}} for ${{ matrix.os_python.os }}
2023/07/26 10:47:12 error during command execution: check runtime error: Pinned-Dependencies: internal error: internal error: unable to determine OS for job: Build python wheels on ${{matrix.arch}} for ${{ matrix.os_python.os }}
{
  "date": "2023-07-26T10:47:05-03:00",
  "repo": {
    "name": "github.com/apache/beam",
    "commit": "907c5110163b0efe52e9e12127fd013c7fc455d7"
  },
  "scorecard": {
    "version": "4.11.0",
    "commit": "4edb07802fdad892fa8d10f8fd47666b6ccc27c9"
  },
  "score": -1,
  "checks": [
    {
      "details": null,
      "score": -1,
      "reason": "internal error: internal error: unable to determine OS for job: Build python wheels on ${{matrix.arch}} for ${{ matrix.os_python.os }}",
      "name": "Pinned-Dependencies",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/4edb07802fdad892fa8d10f8fd47666b6ccc27c9/docs/checks.md#pinned-dependencies",
        "short": "Determines if the project has declared and pinned the dependencies of its build process."
      }
    }
  ],
  "metadata": null
}
```

```
diogoteles:~$ scorecard --repo=https://github.com/apache/arrow --show-details --format=json --checks=Pinned-Dependencies | jq .
Error: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: ci/docker/python-wheel-windows-test-vs2017.dockerfile:1:2: "if <cond>" must be followed by "then"
2023/07/26 10:51:40 error during command execution: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: ci/docker/python-wheel-windows-test-vs2017.dockerfile:1:2: "if <cond>" must be followed by "then"
{
  "date": "2023-07-26T10:51:35-03:00",
  "repo": {
    "name": "github.com/apache/arrow",
    "commit": "8503c869fa80090bc849bae0b4e68a84605ea82e"
  },
  "scorecard": {
    "version": "4.11.0",
    "commit": "4edb07802fdad892fa8d10f8fd47666b6ccc27c9"
  },
  "score": -1,
  "checks": [
    {
      "details": null,
      "score": -1,
      "reason": "internal error: error parsing shell code: ci/docker/python-wheel-windows-test-vs2017.dockerfile:1:2: \"if <cond>\" must be followed by \"then\"",
      "name": "Pinned-Dependencies",
      "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/4edb07802fdad892fa8d10f8fd47666b6ccc27c9/docs/checks.md#pinned-dependencies",
        "short": "Determines if the project has declared and pinned the dependencies of its build process."
      }
    }
  ],
  "metadata": null
}
```

**Reproduction steps**
1.  Run Scorecard (I've reproduced it on version 4.11.0 and 4.10.2) with Pinned-Dependencies check for apache/arrow or apache/beam

**Expected behavior**

At first, we should consider enhancing the error handling to avoid not showing any results to the user other than the internal error. I'd say that ideally the internal error should not break the whole execution of the check; the output should report the error(and probably keep reporting a -1), but should also contain any results that could be achieved before and after the internal error.

Additionally, we should work to avoid those specific internal errors, of course.

pnacht:

Whenever Pinned-Dependencies hits a runtime error (i.e. can't detect job OS, can't parse a Dockerfile, etc), the check crashes entirely and the project gets an inconclusive score for the check.

If Pinned-Dependencies hits a runtime error, the "element" that caused the error (i.e. workflow job, Dockerfile) is skipped and the check progresses as well as possible.

#### TO-DO

At this stage of the PR, the proposed change only applies to failures to detect a job's operating system (as happens with `apache/beam`). The idea is for this to cover other similar cases (i.e. `apache/arrow`, caused by an error parsing a Dockerfile).

Also, at this stage, the failure to detect a job's operating system isn't logged anywhere. The job is simply skipped "invisibly". This is useful information that should be displayed as a warning in the check's details. However, there's currently no place to store this information since `PinningDependenciesData` only stores data on problematic dependencies, not on problems encountered while investigating.

The most straightforward solution I see is to modify `PinningDependenciesData` to also contain new `SkippedWorkflowJobs` and `SkippedFiles` fields (each a struct containing the job/filename and skip reason). The data in these fields can later be added to the logs as warnings, i.e:

```
Warn: myWorkflow.yaml's job 'myJob' skipped: could not determine operating system
```

The job/filenames can be either parsed from the error message or a [new error type](https://go.dev/blog/error-handling-and-go#:~:text=you%20can%20use%20arbitrary%20data%20structures%20as%20error%20values) can be defined which also includes this metadata. A type assertion can  then be used to identify such cases and extract the data from the error. (I don't know where I'd store this new error type)

However, I'm no expert in either Go or the Scorecard codebase, so would like to get feedback on this solution before implementing it.

I have added the same test workflow to multiple test functions in `raw/pinned-dependencies-test.go`. The results show that a problematic job affects non-GHA-pinning scores, but GHA-pinning is unaffected.

#### Notes for the reviewer

I couldn't see where to neatly add anything regarding this change to the Pinned-Deps documentation (which is somewhat high-level, without much detail on how the score is calculated). I have therefore left the docs as-is, without changes. Let me know if you wish to add a comment on such "error-handling" to the check's docs.

Fixes #3316. 

```release-note
Pinned-Dependencies now continues after encountering runtime errors
```


spencerschrock:
> If Pinned-Dependencies hits a runtime error, the "element" that caused the error (i.e. workflow job, Dockerfile) is skipped and the check progresses as well as possible.

I think this is fine. There's the chance of missing dangerous workflows this way, but if we log it (either in the execution, or as a detail) then hopefully we still get bug reports. 

> The most straightforward solution I see is to modify PinningDependenciesData

I think the simplest is to just throw in a log that mentions it, but I can see the value in having it in the details. We currently handle some parse errors just as a silent debug statement:
https://github.com/ossf/scorecard/blob/703430609bd5481f84bf4b53ff0df772b62dc43c/checks/evaluation/pinned_dependencies.go#L80-L89

> A type assertion can then be used to identify such cases and extract the data from the error

These days, it's usually done with [errors.As](https://pkg.go.dev/errors#As)


Of course the other option is to fix the root cause of #3316, instead of skipping over it. But that may be a more complicated fix. Any thoughts @laurentsimon @raghavkaul ?

pnacht:
> I think the simplest is to just throw in a log that mentions it, but I can see the value in having it in the details. We currently handle some parse errors just as a silent debug statement:
> https://github.com/ossf/scorecard/blob/703430609bd5481f84bf4b53ff0df772b62dc43c/checks/evaluation/pinned_dependencies.go#L80-L89

I've just pushed some commits such that the skipped steps are displayed in the details.

Currently, `scorecard --repo apache/beam --checks Pinned-Dependencies [...] | jq '.checks[0]'` returns:

```json
{
  "details": null,
  "score": -1,
  "reason": "internal error: internal error: unable to determine OS for job: Build python wheels on ${{matrix.arch}} for ${{ matrix.os_python.os }}",
}
```

With the proposed change, the score is now correctly calculated (skipping the problematic steps), and everything that had to be skipped is logged in the details.

```json
{
  "details": [
    "Warn: Possibly incomplete results: error parsing job operating system: .github/workflows/build_wheels.yml's job 'Build python wheels on ${{matrix.arch}} for ${{ matrix.os_python.os }}' (step 5)",
    "Warn: Possibly incomplete results: error parsing job operating system: .github/workflows/build_wheels.yml's job 'Build python wheels on ${{matrix.arch}} for ${{ matrix.os_python.os }}' (step 7)",
    "Warn: GitHub-owned GitHubAction not pinned by hash: .github/workflows/assign_milestone.yml:34: update your workflow using https://app.stepsecurity.io/secureworkflow/apache/beam/assign_milestone.yml/master?enable=pin",
    "... 1000+ other warnings ..."
  ],
  "score": 0,
  "reason": "dependency not pinned by hash detected -- score normalized to 0",
}
```

It was also quite straightforward to do the same for the apache/arrow error, so I've implemented that as well.

`scorecard --repo apache/arrow --checks Pinned-Dependencies [...] | jq '.checks[0]'` before:

```
Error: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: ci/docker/python-wheel-windows-test-vs2017.dockerfile:1:2: "if <cond>" must be followed by "then"
2023/10/20 23:06:16 error during command execution: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: ci/docker/python-wheel-windows-test-vs2017.dockerfile:1:2: "if <cond>" must be followed by "then"
```
```json
{
  "details": null,
  "score": -1,
  "reason": "internal error: error parsing shell code: ci/docker/python-wheel-windows-test-vs2017.dockerfile:1:2: \"if <cond>\" must be followed by \"then\"",
}
```

Now:

```json
{
  "details": [
    "Warn: Possibly incomplete results: error parsing shell code: ci/docker/python-wheel-windows-test-vs2017.dockerfile:1:2: \"if \u003ccond\u003e\" must be followed by \"then\"",
    "Warn: Possibly incomplete results: error parsing shell code: ci/docker/python-wheel-windows-vs2017.dockerfile:1:2: \"if \u003ccond\u003e\" must be followed by \"then\"",
    "Warn: GitHub-owned GitHubAction not pinned by hash: .github/workflows/archery.yml:53: update your workflow using https://app.stepsecurity.io/secureworkflow/apache/arrow/archery.yml/main?enable=pin",
    "... 300+ warnings ..."
  ],
  "score": 0,
  "reason": "dependency not pinned by hash detected -- score normalized to 0",
}
```

I've seen a few other places where this maybe could also be applied, but I haven't dug into them to be sure yet.

> These days, it's usually done with [errors.As](https://pkg.go.dev/errors#As)

Thanks for the tip, I used it! (I'm still getting my golang sea legs...)

> Of course the other option is to fix the root cause of #3316, instead of skipping over it. But that may be a more complicated fix. Any thoughts @laurentsimon @raghavkaul ?

The issue here isn't necessarily in Scorecard, though:
- for apache/beam, the problem is a failure to identify a job's OS if the matrix is too complex (matrix variables that are dictionaries). That could probably be figured out within Scorecard.
- however, for apache/arrow, the error is thrown directly by `mvdan.cc/sh/v3/syntax`.

spencerschrock:
> With the proposed change, the score is now correctly calculated (skipping the problematic steps), and everything that had to be skipped is logged in the details.

I'd lean towards Info instead of warn? just so it doesn't end up as something in the security dashboard?


> for apache/beam, the problem is a failure to identify a job's OS if the matrix is too complex (matrix variables that are dictionaries). That could probably be figured out within Scorecard.

I have [a change](https://github.com/ossf/scorecard/compare/main...spencerschrock:scorecard:bug/matrix-os-pinned-dep) I was playing around with, but haven't had time to finish testing.
 

> * however, for apache/arrow, the error is thrown directly by `mvdan.cc/sh/v3/syntax`.

Because we don't have a shell parser for powershell. So partially scorecard's fault

checker/raw_result.go
```diff
@@ -111,6 +111,7 @@ const (
 // PinningDependenciesData represents pinned dependency data.
 type PinningDependenciesData struct {
 	Dependencies []Dependency
+	Incomplete   []error // jobs or files with errors may have incomplete results
```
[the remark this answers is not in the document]
pnacht: I left it as a generic since it doesn't always have to be an ElementError. For example, the error stored when parsing a Dockerfile is whatever kind of error is returned by `syntax.NewParser().Parse()`:

https://github.com/ossf/scorecard/pull/3515/files#diff-d861376c45f3cee9d7b3959c8bb1954fc1698ac3549a53ce369ce3a2e7c5962dR1062-R1067

ElementError is (currently) only used when the problem is parsing a workflow.

I didn't use `errors.New` or `sce.WithMessage` because I don't have full location information at the moment the problem arises, just the job name. I could have used those anyway, but I'd then have to parse the error string to extract the job name and recreate the error with the filename and step number.

I therefore thought it best to create an error type, especially since I think this approach (skipping problematic code and logging it in the details) can be applied in a few other cases and having an error type with location metadata might be useful.

laurentsimon:
> > If Pinned-Dependencies hits a runtime error, the "element" that caused the error (i.e. workflow job, Dockerfile) is skipped and the check progresses as well as possible.
> 
> I think this is fine. There's the chance of missing dangerous workflows this way, but if we log it (either in the execution, or as a detail) then hopefully we still get bug reports.
> 
> > The most straightforward solution I see is to modify PinningDependenciesData
> 
> I think the simplest is to just throw in a log that mentions it, but I can see the value in having it in the details. We currently handle some parse errors just as a silent debug statement:
> 
> https://github.com/ossf/scorecard/blob/703430609bd5481f84bf4b53ff0df772b62dc43c/checks/evaluation/pinned_dependencies.go#L80-L89
> 
> > A type assertion can then be used to identify such cases and extract the data from the error
> 
> These days, it's usually done with [errors.As](https://pkg.go.dev/errors#As)
> 
> Of course the other option is to fix the root cause of #3316, instead of skipping over it. But that may be a more complicated fix. Any thoughts @laurentsimon @raghavkaul ?

I'm fine with logging only for now, until we have better parsing code

checker/raw_result.go
spencerschrock, line 114: Two questions, since these structs are intended to be flexible to avoid breakages:
1. Is `Incomplete` better than `Skipped` or `ProcessingErrors` or something like that?
2. Do you intend this to be final? Since the type is defined as a slice of errors, we can't necessarily expand it later. (until the next major version)

checks/evaluation/pinned_dependencies_test.go
```diff
@@ -267,6 +269,29 @@ func Test_PinningDependencies(t *testing.T) {
...
+				},
+			},
+			incomplete: []error{
+				sce.ErrorJobOSParsing,
+			},
+			expected: scut.TestReturn{
+				Error:         nil,
+				Score:         8,
+				NumberOfWarn:  2,
+				NumberOfInfo:  8, // 7 for all but npm, +1 for skipped job
```
spencerschrock, line 291: this test will probably change once you rebase

checks/raw/pinned_dependencies_test.go
```diff
@@ -65,6 +65,11 @@ func TestGithubWorkflowPinning(t *testing.T) {
 			name:     "Matrix as expression",
 			filename: "./testdata/.github/workflows/github-workflow-matrix-expression.yaml",
 		},
+		{
+			name:     "Can't detect OS, but still detects unpinned Actions",
+			filename: "./testdata/.github/workflows/github-workflow-unknown-os.yaml",
+			warns:    2, // 1 in job with unknown OS, 1 in job with known OS
```
spencerschrock, line 71: do these warn numbers (or the comments) needed updated since the statement is an info?

checks/raw/pinned_dependencies_test.go
```diff
@@ -213,6 +218,11 @@ func TestNonGithubWorkflowPinning(t *testing.T) {
 			filename: "./testdata/.github/workflows/workflow-mix-pinned-and-non-pinned-non-github.yaml",
 			warns:    1,
 		},
+		{
+			name:     "Can't detect OS, but still detects unpinned Actions",
+			filename: "./testdata/.github/workflows/github-workflow-unknown-os.yaml",
+			warns:    2, // 1 in job with unknown OS, 1 in job with known OS
```
spencerschrock, line 224: do these warn numbers (or the comments) needed updated since the statement is an info?

errors/public.go
```diff
@@ -30,6 +30,8 @@ var (
 	ErrorInvalidURL = errors.New("invalid repo flag")
 	// ErrorShellParsing indicates there was an error when parsing shell code.
 	ErrorShellParsing = errors.New("error parsing shell code")
+	// ErrorJobOSParsing indicates there was an error when detecting a job's operating system.
+	ErrorJobOSParsing = errors.New("error parsing job operating system")
```
spencerschrock, line 34: once you rebase, this name will get flagged by the linter.

errors/public.go
```diff
@@ -60,3 +62,14 @@ func GetName(err error) string {
 		return "ErrUnknown"
 	}
 }
+
+// ElementError adds generic "element" information (file path, workflow job, step ID,
+// etc) to an error.
+type ElementError struct {
+	Err     error
```
spencerschrock, line 69: if your custom error type wraps another error, you need to provide a way to unwrap it so that `errors.Is` and `errors.As` work with your type.

For example, here's a [similar wrapper in the webapp](https://github.com/ossf/scorecard-webapp/blob/623b0d90aab250ee566518dbd26238f1f1b3dfac/app/server/verify_workflow.go#L65-L75)

errors/public.go
```diff
@@ -60,3 +62,14 @@ func GetName(err error) string {
 		return "ErrUnknown"
 	}
 }
+
+// ElementError adds generic "element" information (file path, workflow job, step ID,
+// etc) to an error.
+type ElementError struct {
```
spencerschrock, line 68: Since this is one of the more public parts of the code. Can you clarify what an Element is?  The examples are good, but what is an element? Some discrete portion of analysis data for the various scorecard checks?

spencerschrock:
Related discussion here: https://github.com/ossf/scorecard/issues/1327

pnacht (commented):
I've just pushed a few changes. 

`ElementError` is now in checker/raw_result, where it should've been from the start. It's `.Element` field is no longer a string, but a `Location`, making it easier to register precisely where the error occurred.

Due to this change, we also have Improved logging of the errors, now showing the specific lines and following the styling in the other details (since we're using the same plumbing).

Lastly, `PinningDependenciesData.Incomplete []error` has been replaced with `.ProcessingErrors []ElementError`. The name is better and using a struct makes more sense in this case.

I'd thought of using the `error` interface in case other error types come up which aren't a good fit for `ElementError`, but using the interface only makes sense if we actually work with the errors as `errors` – using `.Error()` to display them. However, since our logging relies on passing `LogMessage` structs to the logger, the string returned by `.Error()` isn't a good fit.

Unless we want to create our own error interface (or, more broadly, a something-to-be-logged interface) that expects a LogMessage instead of a string?

```go
type ScorecardError /* or */ LogMessenger interface {
  Message() checker.LogMessage
}
```

The errors are still displayed as `Info`, not `Error` as suggested in #1327. Don't know if that change (adding a new logger function) fits in this PR or if it should come in a follow-up.

checks/raw/shell_download_validate_test.go
```diff
@@ -102,8 +102,12 @@ func TestValidateShellFile(t *testing.T) {
 
 	var r checker.PinningDependenciesData
 	err = validateShellFile(filename, 0, 0, content, map[string]bool{}, &r)
-	if err == nil {
-		t.Errorf("failed to detect shell parsing error: %v", err)
+	if err != nil {
+		t.Errorf("error validating shell file")
```
spencerschrock, line 106: let's keep the `: %v", err)` bit

errors/public.go
```diff
@@ -33,6 +33,8 @@ var (
 	ErrorInvalidURL = errors.New("invalid repo flag") //nolint:errname
 	// ErrorShellParsing indicates there was an error when parsing shell code.
 	ErrorShellParsing = errors.New("error parsing shell code") //nolint:errname
+	// ErrorJobOSParsing indicates there was an error when detecting a job's operating system.
+	ErrorJobOSParsing = errors.New("error parsing job operating system") //nolint:errname
```
spencerschrock, line 37: Don't ignore the linter please, the `errname` exceptions are for the old values

spencerschrock (approved):

spencerschrock:
Please fix the unit tests too

errors/public.go
pnacht: Ah, I was just following the pattern! Done!

checks/raw/shell_download_validate_test.go
pnacht: Done.

checker/raw_result.go
```diff
@@ -402,3 +404,23 @@ func (f *File) Location() *finding.Location {
...
+// ElementError allows us to identify the "element" that led to the given error.
+// The "element" is the specific "code under analysis" that caused the error. It should
+// describe what caused the error as precisely as possible.
+//
+// For example, if a shell parsing error occurs while parsing a Dockerfile `RUN` block
+// or a GitHub workflow's `run:` step, the "element" should point to the Dockerfile
+// lines or workflow job step that caused the failure, not just the file path.
+type ElementError struct {
+	Err     error
+	Element *finding.Location
```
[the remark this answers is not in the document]
pnacht: Done.

checks/fileparser/github_workflow.go
```diff
@@ -208,8 +208,15 @@ func GetOSesForJob(job *actionlint.Job) ([]string, error) {
 	}
 
 	if len(jobOSes) == 0 {
-		return jobOSes, sce.WithMessage(sce.ErrScorecardInternal,
-			fmt.Sprintf("unable to determine OS for job: %v", GetJobName(job)))
+		// This error is caught by the caller, which is responsible for adding more
+		// precise location information
```
spencerschrock, line 212: optional: If the caller needs to add more information anyway, can we just return `sce.ErrJobOSParsing` here and let the caller create the `ElementError`?

It seems a little brittle to have 2 functions composing the error when the jobName is available to the caller too.

PR #3585, opened 2023-10-19
:seedling: use forbidigo linter to prevent print statements

spencerschrock:

linter change

print statements aren't flagged

print statements are flagged with our reasoning. This should help with reviews

NONE

the `release-note`

```release-note
NONE
```


attestor/command/cli.go
```diff
@@ -114,7 +114,7 @@ func init() {
 
 func Execute() {
 	if err := RootCmd.Execute(); err != nil {
-		fmt.Println(err)
+		fmt.Fprintln(os.Stdout, err) // can this just be stderr?
```
raghavkaul, line 117: Yes, I think this should be `os.Stderr`

raghavkaul (approved):

attestor/command/cli.go
spencerschrock: Is this something that any scorecard attestor users might be relying on?

PR #3619, opened 2023-10-27
:sparkles: Add WithValues function to findings

laurentsimon:

feature

Add WithValues function to findings to simplify the creation of probes findings.

the `release-note`

```release-note
Add WithValues function to findings
```

### DIFF
--- a/finding/finding.go
+++ b/finding/finding.go
@@ -221,6 +221,13 @@ func (f *Finding) WithLocation(loc *Location) *Finding {
 	return f
 }
 
+// WithValues adds values to an existing finding.
+// No copy is made.
+func (f *Finding) WithValues(values map[string]int) *Finding {
+	f.Values = values
+	return f
+}
+
 // WithPatch adds a patch to an existing finding.
 // No copy is made.
 func (f *Finding) WithPatch(patch *string) *Finding {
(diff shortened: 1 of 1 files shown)


finding/finding.go
```diff
@@ -221,6 +221,13 @@ func (f *Finding) WithLocation(loc *Location) *Finding {
 	return f
 }
 
+// WithValues adds values to an existing finding.
+// No copy is made.
+func (f *Finding) WithValues(values map[string]int) *Finding {
+	f.Values = values
```
spencerschrock, line 227: The doc says "WithValues adds values to an existing finding", but the code replaces the values of a finding.

What do you expect if the code does this?
```

foo.WithValues(map[string]int{"foo": 1, "bar": 2})
foo.WithValues(map[string]int{"baz": 3, "qux": 4})
```

Should it have all 4 values or just the 2?

laurentsimon: +1, I copied the comment from another function.

spencerschrock (approved):

finding/finding.go
spencerschrock: Had a chat with Laurent, we're fine with it either way in the future if the incremental approach is needed later.

Abeanna (approved):

PR #3634, opened 2023-10-31
🌱 Update stale workflow to exempt Structured Results milestone

afmarcum:

Exempts Structured Results issues from being labeled stale

Exemptions are based on labels only and it is not immediately clear which labels are exempt.
Contributors can apply Structured Results milestone rather than having to remember labels for exemption.

NONE

the `release-note`

```None

```

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -35,12 +35,9 @@ jobs:
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'This issue is stale because it has been open for 60 days with no activity.'
-        stale-pr-message: 'Stale pull request message'
-        stale-issue-label: 'no-issue-activity'
+        stale-pr-message: 'This pull request is stale because it has been open for 10 days with no activity'
         exempt-issue-labels: 'priority,bug,good first issue'
         exempt-issue-milestones: 'Structured results'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -38,6 +38,8 @@ jobs:
         stale-pr-message: 'Stale pull request message'
         stale-issue-label: 'no-issue-activity'
         exempt-issue-labels: 'priority,bug,good first issue'
+        exempt-issue-milestones: 'Structured results'
+        stale-pr-label: 'no-pr-activity'
```
spencerschrock, line 42: the `stale-pr-label` line is duplicated. 

do we need separate labels for stale issues and stale PRs? should we just have one? The default value for both (if the parameter isn't provided) is just `Stale`

spencerschrock (approved):
Exempting the milestone is good, just please address the duplicated line.

.github/workflows/stale.yml
afmarcum: I removed the duplicated `stale-pr-label`. 

If I want to change to the default label for both stale issues and stale PRs, do I remove the entire line or just the defined name?

I also updated the stale PR message to match the stale issue message. Should I change the `days-before-pr-close` value to -1 as well so that PRs are not automatically closed either?

Issue #3662, opened 2023-11-10
Remove OneFuzz from Fuzzing Checks Remediation


roger2hk:
[OneFuzz project](https://github.com/microsoft/onefuzz) was archived on September 30th, 2023. The fuzzing remediation/alert still recommends OneFuzz.

https://github.com/ossf/scorecard/blob/934f17049c46b6c55b77ce95bad7cdd776bacbee/probes/fuzzedWithOneFuzz/def.yml#L28-L29

https://github.com/ossf/scorecard/blob/934f17049c46b6c55b77ce95bad7cdd776bacbee/docs/checks/internal/checks.yaml#L405

PR #3666, opened 2023-11-11
:warning: Remove OneFuzz from fuzzing checks

DavidKorczynski:
This is removed because OneFuzz has been archived https://github.com/microsoft/onefuzz

Note that OneFuzz is not *deprecated* but development of it in the open source has stopped. In that sense I think it's a position statement from Scorecard: the project can still be used for fuzzing but is no longer maintained, should it be recommended as a fuzzing set up or considered in the scoring process?

I set this as a breaking feature since it's removing a probe, but am not sure if this should be considered breaking as such.

OneFuzz is recommended as a continuous fuzzing solution.

OneFuzz is not considered when recommending fuzzing.

Fixes: https://github.com/ossf/scorecard/issues/3662

the `release-note`

```release-note

```


spencerschrock (approved):
> Note that OneFuzz is not _deprecated_ but development of it in the open source has stopped. In that sense I think it's a position statement from Scorecard: the project can still be used for fuzzing but is no longer maintained, should it be recommended as a fuzzing set up or considered in the scoring process?

So previously we did something similar for Sonatype Lift (#3605), which was a managed service. When it was shutdown, the tool no longer worked. I'm curious if being self-hosted changes how we handle this in the future. For anyone currently using OneFuzz, they would continue to benefit from it since it's self-hosted.

From some quick searching, there don't seem to be any public repos which use this on GitHub so I don't think we have an issue here. Eventually we'll run into the issue of a deprecated probe that we still want to score, but don't want to recommend.

PR #3667, opened 2023-11-12
:seedling: refactor pinned dependencies

AdamKorcz:
The goal of this PR is to refactor the Pinned Dependencies evaluation to process findings instead of dependencies. The reason for doing this is to structure the code to resemble how it would look once it gets split into probes without modifying the score. As such, this is an intermediary code change towards migrating the Pinned Dependencies check into probes. Some of the code changes in this PR will be moved into probes in the near future.

My motivation for this PR is to reduce the review overhead. This PR is already substantial, and adding probes would add to that. As such, with this PR I want to refactor the code so it works as-is and make incremental steps towards migrating the check to probes.

The core change of this PR is an added step of converting the dependencies to findings and calculate the score + log from the findings instead of directly from the dependencies. Other code changes are utilities and tests around that core change.

NONE

NONE

```release-note

```

### DIFF
--- a/checks/evaluation/pinned_dependencies.go
+++ b/checks/evaluation/pinned_dependencies.go
@@ -21,6 +21,8 @@ import (
 	"github.com/ossf/scorecard/v4/checks/fileparser"
 	sce "github.com/ossf/scorecard/v4/errors"
 	"github.com/ossf/scorecard/v4/finding"
+	"github.com/ossf/scorecard/v4/finding/probe"
+	"github.com/ossf/scorecard/v4/rule"
 )
 
 type pinnedResult struct {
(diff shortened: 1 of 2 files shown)


checks/evaluation/pinned_dependencies.go
```diff
@@ -49,6 +51,141 @@ const (
...
+			f := &finding.Finding{
+				Probe:    "",
+				Outcome:  finding.OutcomeNegative,
+				Message:  generateTextUnpinned(&rr),
+				Location: loc,
+			}
+			if rr.Remediation != nil {
+				f.Remediation = ruleRemToProbeRem(rr.Remediation)
+			}
+			f = f.WithValues(map[string]int{
```
laurentsimon, line 163: I think we might want to have a distinct probes for each ecosystem, because remediation steps will be different. Not a blocker for this PR since it's an intermediary PR. Unless you think it'd help to do it now?

AdamKorcz: Good to know, thanks! I think it is best to introduce the probes later to avoid a large PR.

AdamKorcz: Yes, it would be temporary, but perhaps we should only have either probe.Remediation or rule.Remediation in Scorecard, since they are identical?

spencerschrock: as far as I know, probes and rules are the same thing. It just got renamed. Laurent would know for sure

laurentsimon: I confirm. rule is the old name we used.

spencerschrock: > but perhaps we should only have either probe.Remediation or rule.Remediation in Scorecard, since they are identical?

If this is just a code smell thing, and not blocking anything, I'm still in favor of this being done later, tracked in an issue to be taken care of closer to the major version bump

AdamKorcz: Sounds good. I have created https://github.com/ossf/scorecard/issues/3677 to track this. Closing this issue.

checks/evaluation/pinned_dependencies_test.go
```diff
@@ -303,40 +311,47 @@ func Test_PinningDependencies(t *testing.T) {
 					Location: &checker.File{
 						Snippet: "actions/checkout@v2",
 					},
-					Type:   checker.DependencyUseTypeGHAction,
-					Pinned: asBoolPointer(false),
+					Type:        checker.DependencyUseTypeGHAction,
+					Pinned:      asBoolPointer(false),
+					Remediation: testRemediation,
```
spencerschrock, line 316: was there a reason to add all these Remediation fields? Any extra test parts adds confusion on if its needed or not for the test

spencerschrock (approved):
the scdiff run showed no diffs, and the tests are working, so good enough for me for a temporary refactor

checks/evaluation/pinned_dependencies_test.go
spencerschrock: need to update to get this merged anyway, so will remove + rebase/merge commit

AdamKorcz: Done in https://github.com/ossf/scorecard/pull/3667/commits/4d7886c59d6efe089dc69886ba4fe05ae2d6af65

PR #3672, opened 2023-11-13
:sparkles: Add commit depth support for GitLab

### LINKED ISSUE #2828: BUG: Commit-Depth Not Used on Querying Gitlab
**Describe the bug**
The new gitlab functionality is not using the parameter `--commit-depth` in querying the gitlab repo for commits. In addition the defaulted value of 30 was not being used either, for the same reason.

**Reproduction steps**
1. Provide a `--commit-depth` on command line `go run main.go --repo=gitlab.com/hoppr/hoppr --commit-depth 200`
2. Counting commits there are only 20 (gitlab's default) (print out and/or debugger)

**Expected behavior**
As a user providing the `--commit-depth`, I would expect that correct number of commits being evaluated/scored. I also expect that I don't provide a value, I'd expect the documented default of 30 to be used.

ashearin:

Bug fix for #2828, reopening of #2829 

The new gitlab functionality is not using the parameter --commit-depth in querying the gitlab repo for commits. In addition the defaulted value of 30 was not being used either, for the same reason.

Fixes #2828

the `release-note`

```release-note
Added `--commit-depth` support for GitLab repos
```


clients/gitlabrepo/commits.go
```diff
@@ -27,31 +27,52 @@ import (
...
 	handler.once = new(sync.Once)
+	handler.commitDepth = commitDepth
 }
 
 func (handler *commitsHandler) setup() error {
 	handler.once.Do(func() {
-		commits, _, err := handler.glClient.Commits.ListCommits(
-			handler.repourl.projectID,
-			&gitlab.ListCommitsOptions{
-				RefName: &handler.repourl.commitSHA,
```
spencerschrock, line 48: Would removing this `RefName` list option break the `--commit` option?

clients/gitlabrepo/commits.go
```diff
@@ -27,31 +27,52 @@ import (
...
+						Page:    i,
+						PerPage: handler.commitDepth,
+					},
+				})
+			if err != nil {
+				handler.errSetup = fmt.Errorf("request for commits failed with %w", err)
+				return
+			}
+
+			if len(c) == 0 {
```
spencerschrock, line 63: does this correspond to the scenario where a user requested more commits than exist in the repo?

clients/gitlabrepo/commits.go
```diff
@@ -27,31 +27,52 @@ import (
...
-			return
+		var commits []*gitlab.Commit
+		i := 1
+
+		for {
+			c, _, err := handler.glClient.Commits.ListCommits(handler.repourl.projectID,
+				&gitlab.ListCommitsOptions{
+					ListOptions: gitlab.ListOptions{
+						Page:    i,
+						PerPage: handler.commitDepth,
```
spencerschrock, line 55: Are there any limits on the GitLab side of things for how many commits we can ask for? On the GitHub side we have to request max of 100 at a time for example.

ashearin: So the Gitlab default for PerPage is 20, max of 100. However, if you enter a commit depth of 200 gitlab will still return a perpage of 100 and we just iterate until we hit the requested depth

spencerschrock: Got it, do you know if it's the `Page` value depends on `PerPage` at all? For example let's say we request a commit depth of 200. This is too large for the first iteration (gitlab only returns commits 1-100 as you said). On the second iteration we'd be sending something like:
```
Page: 2
PerPage: 200
```


Again, this will return 100 commits, but does it return commits 101 - 200? or commits 201-300?

ashearin: It would return commits 101-200, we're just getting the second page (of 100) of repository commits. I'm trying to find a good example in the docs, but nothing really illustrates this. Theres [this](https://github.com/xanzy/go-gitlab/blob/793bc3cdf558e2ec74e5873e1de9fcd535266aa0/examples/pagination.go#L25) from the go-gitlab docs which is essentially what we're doing.

clients/gitlabrepo/commits.go
ashearin: Yes exactly, if the requested page of commits is empty our list of commits already has the entire commit history

clients/gitlabrepo/commits.go
spencerschrock: Since this is an example of what we're doing, can we also handle pagination the same way, using the semantics from the example? I think the readability is better

```go
		// Exit the loop when we've seen all pages.
		if resp.NextPage == 0 {
			break
		}

		// Update the page number to get the next page.
		opt.Page = resp.NextPage
```

clients/gitlabrepo/commits.go
ashearin: Most likely, this was inadvertent. Will get it back in with the other changes

clients/gitlabrepo/commits.go
ashearin: Yeah, I think thats a good idea

spencerschrock (approved):
Looks good overall. Thanks for the small + tested PR.

I left one comment about an easy test I'd like to see before merge.


There's also a harder test for pagination, but I'm fine merging without it. Our existing `suffixStubTripper` isn't good enough, since some of the values `go-gitlab` cares about are in the header, not the response. 
```
curl -vvv --header "PRIVATE-TOKEN: $GITLAB_AUTH_TOKEN" "https://gitlab.com/api/v4/projects/10858056/repository/commits?page=1&per_page=101"
...
< x-next-page: 2
< x-page: 1
< x-per-page: 100
< x-prev-page: 
...
```
But testing it manually seemed to work for something like:

```console
go run main.go --repo https://gitlab.com/baserow/baserow --checks Maintained --commit-depth 101 --format json --show-details | jq
```

Issue #3549, opened 2023-10-09
[Feedback] Unclear value proposition for hash-pinning dependencies and keeping them up-to-date with dependency update tools


pnacht:
I've recently been trying to get projects to hash-pin their dependencies and keep them up-to-date with dependabot or renovatebot. I've had mixed results, with many projects questioning the value of such a change.

See here a few of the discussions I've had on this topic. They are all solid discussions with valuable maintainer feedback. I recommend reading them in order since they refer to each other.
- https://github.com/pypa/setuptools/issues/4025
- https://github.com/python/cpython/issues/109110
- https://github.com/WebAssembly/wabt/pull/2291
- https://github.com/keras-team/keras-core/pull/877

Issue #2761, opened 2023-03-16
Feature: determine if publishing is credential-based or OIDC-based


di:
**Is your feature request related to a problem? Please describe.**
GitHub Actions supports OIDC identities for workflow runs: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect

Some ecosystems now support OIDC-based publishing with these identities, e.g. for PyPI: https://github.com/pypi/warehouse/issues/12465 (currently in closed beta), which can be identified based on the usage of the official action without supplying credentials: https://github.com/pypa/gh-action-pypi-publish#publishing-with-openid-connect

This should be considered more secure than using long-lived credentials like API tokens or username/passwords for publishing. It also has the added benefit of linking the source repository to the published artifact in a verifiable way.

**Describe the solution you'd like**
For ecosystems that support OIDC-based publishing, Scorecard should penalize repositories that use long-lived credentials for publishing, and reward projects that adopt OIDC-based publishing, either via a new check or an existing check.

di:
As an example, here's an instance of the change that would be necessary to move from less-secure long-lived credentials to more secure OIDC-based publishing: https://github.com/sigstore/sigstore-python/pull/566/

pnacht:
Would this be a new metric for the Packaging check or a new check in itself? It seems a good fit for Packaging, but then that raises the issue of how to score projects in different ecosystems (those that support OIDC and those that don't).

As a separate check this becomes easy to handle: if a project belongs to an OIDC-friendly ecosystem, it gets scored. Otherwise, the check is ignored.

The same thing could be done within Packaging, but it might seem odd to the maintainer that a Python project gets an 8/10 for publishing (without OIDC) while an R project gets a 10/10 for the exact same thing.

laurentsimon:
+1 from me. Great to encourage users to use OIDC instead of long-lived tokens.

varunsh-coder:
JFYI - I have created an [issue](https://github.com/step-security/secure-repo/issues/2032) in https://github.com/step-security/secure-repo to automate transformation of GitHub Actions workflows to use OIDC.

di:
More examples of this already available other ecosystems: https://dart.dev/tools/pub/automated-publishing

laurentsimon:
I'm +1 to have this under Packaging check. Folks who have a 10 today will drop to a lower score (say, 8) and it will encourage them to update.

woodruffw:
[...]

Issue #2483, opened 2022-11-23
[FEEDBACK] Unclear value on dependency-pinning and packing checks on C/C++ projects


diogoteles08:
**Is your feature request related to a problem? Please describe.**
When inspected by scorecards, projects written in C get penalized for not having dependency-pinning tools or packings, but C does not have any widely adopted dependency manager, so there is not much they can do.

**Describe the solution you'd like**
Study the possibility of a smaller penalty for dependency-pinning tools or packing for C projects.

**Describe alternatives you've considered**
--

**Additional context**
This concern was raised while discussing Scorecards adoption with [cURL](https://github.com/curl/curl) project. You can check the whole discussion on [this history](https://curl.se/mail/lib-2022-11/0026.html) of the mailing list, but the part related to this issue says:

> We don't have depdencies like that so there is no such tool. You build curl to use the dependencies that are installed in the machine where curl is built/run. We cannot scan for those on GitHub.

pnacht:
Would it perhaps be possible to use the GH API to fetch the project's [main language](https://docs.github.com/en/rest/repos/repos#list-repository-languages), and then disable the check for any project where the main language isn't [supported by dependabot](https://github.com/dependabot/dependabot-core#whats-in-this-repo)?

Some filtering might be necessary in case there's a weird project with so many scripts that the primary language is "shell", but the actual "content" of the repo is in Python, for example.

Also might be relevant to detect other types of objects that dependabot can handle, such as git submodules and Docker/Terraform files. So a project written in C but that uses submodules should still be penalized.

varunsh-coder:
In addition, might be good to check if project uses GitHub Actions, in which case updating Actions is still a good idea. e.g. I checked and cURL repo does have GitHub Actions.

pnacht:
Very true! I think the summary here is: the check should only run in repos where it's relevant, and it might be quite valuable for the warning to indicate how dependabot (or renovate) would be useful.

Currently the output for the Dependency-Update-Tool check is:

```json
{
  "name":"Dependency-Update-Tool",
  "score":0,
  "reason":"no update tool detected",
  "details":[
    "Warn: Config file not detected in source location for dependabot, renovatebot, Sonatype Lift, or\n\t\t\tPyUp (Python). We recommend setting this configuration in code so it can be easily verified by others."
  ],
  "documentation":{
    "short":"Determines if the project uses a dependency update tool.",
[...]

Issue #10, opened 2020-10-14
Does the project use protected tags?


kimsterv:
https://github.community/t/feature-request-protected-tags/1742/21

gabibguti:
Updating GitHub community discussion link:
https://github.com/orgs/community/discussions/10906
It's still on beta, but seems like it's not being watched/maintained.

gabibguti:
@afmarcum Just to clarify, you mean using the [GitHub rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) for restricting who can create, update and delete tags inside Scorecard repository does not align with Scorecard's focus right now, right?

Does it make sense to integrate the GitHub rulesets logic for Scorecard Branch-Protection check though?

gabibguti:
> @afmarcum Just to clarify, you mean using the [GitHub rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) for restricting who can create, update and delete tags inside Scorecard repository does not align with Scorecard's focus right now, right?
> 
> Does it make sense to integrate the GitHub rulesets logic for Scorecard Branch-Protection check though?

Seems like the GitHub rulesets are already being considered for Branch-Protection, https://github.com/ossf/scorecard/pull/3354. 🎉

afmarcum:
@gabibguti I'm working through older issues to see which still have interest. PRs are welcome for any work, but the maintainers will not be focused on this work in the near future.

AdamKorcz:
I am working on this. @spencerschrock would you mind assigning me?

spencerschrock:
@AdamKorcz a few questions before beginning.

1. Data collection, are you going to adapt the existing graphQL call and the repoclient `GetBranch` call? Maybe we rename it to GetRef? or something that encompasses both (i think tags and branches can both be named the same name though)? I suspect we'll have breaking changes eitherway?
2. Does GitLab have a similar API to query?
3. What kind of scoring would you impose?
4. What severity would the check be marked as?

AdamKorcz:
@spencerschrock 

In addition to my answers below, we face the following substantial decisions:

[...]

PR #3690, opened 2023-11-21
🌱 Use backlog and "help wanted" labels on issues/PRs to keep stale-bot away

pnacht:

Tells the stale bot that issues/PRs labelled "never stale" shouldn't be considered stale.

The stale bot will mark any issue or PR without activity as stale after some time. The exceptions are the issues and PRs with a few specific labels: 

- issues: `priority`, `bug` and `good first issue`
- PRs: `awaiting-approval` and `work-in-progress`

However, there are other types of issues (and maybe PRs) that should remain open indefinitely: these may be more "philosophical" questions regarding Scorecard (i.e. #3549 and #2483) or "wishlist" feature requests that require changes in other places (i.e. #10). 

Issues and PRs can now be labelled as `never stale`. These issues will never be marked as stale or closed by the stale-bot. (A maintainer will still need to create these labels).

As discussed in https://github.com/ossf/scorecard/issues/2761#issuecomment-1817894945.

the `release-note`

```release-note
NONE
```


spencerschrock:
Agree that having the author/interested party needing to constantly say "not stale" is bad. I think adding a generic `not stale` label is probably ok, but can see it being over-used and we would end up with a bunch of stale issues that are marked `not stale`.

So this seems like a triaging process, and the issue should really reflect the current state. In the case of your linked issue (#2761), it has support from one of the maintainers. Maybe we should have a `good idea` or `tentatively approved` or something label, which gets exempted instead.

raghavkaul:
My 2c: The existing categories describe issues well, but we could be better about tagging proactively.

spencerschrock:
> My 2c: The existing categories describe issues well, but we could be better about tagging proactively.

What label would you add to the issue in this case, and then to stale exemption list?

raghavkaul:
Maybe I'd add `backlog` or `icebox` to indicate that an issue is low priority but still in the consideration as a feature with widespread acceptance that it should be in Scorecard.

pnacht:
Just pushed a change adopting `icebox` and `help wanted`, for anything the team would like to see but doesn't have time to implement itself.

Also removed the label from PRs because I can't really think of a use-case there that doesn't fit into the existing labels.

spencerschrock:
Is icebox a common term? I personally think it's a bit cryptic compared to backlog

raghavkaul:
Backlog is more commonly used, sounds fine with me.

pnacht:
Ah, I used "icebox" due to @raghavkaul's suggestion. The term isn't as widely used as backlog, but it [does exist](https://www.reddit.com/r/agile/comments/41204f/what_is_an_icebox/).

Also, "backlog" is a fairly overloaded term, and in some definitions it's "this is what we ARE going to work on" (in agile, the tasks that will be performed during a sprint are the "backlog" or "[sprint backlog](https://asana.com/resources/sprint-backlog)"... there's also "work backlog" which is the lower priority stuff).

Another term I've often seen used is "[backburner](https://scrumdictionary.com/term/back-burner/)".

pnacht:
> I think another aspect of confusion is who "we" is here. And my understanding is the issue is something we'd be fine with but want the community to pickup. Which sounds like "help wanted" (although we need to change the description as it's currently "Extra attention is needed")
> 
> So is "backlog" for maintainers to do, and "help wanted" for anyone/community to do?

That could be an option, yes. Basically:
- "backlog/icebox/backburner" = "stuff maintainers want to do but don't have time for right now, maybe we'll get to it in the future"
- "help wanted" = "stuff that'd be nice-to-have but is pretty low priority, maintainers likely will never do this themselves, but it'd be great if you could do it!"

spencerschrock (approved):

PR #3693, opened 2023-11-26
:seedling: refactor permissions

AdamKorcz:

Currently, the Permissions checks applies the score directly from the raw results.

This PR adds an extra step to the the Permissions check. Instead of Permissions applying the score directly from the raw results, the check now creates a slice of findings from the raw results, and then passes the slice of findings onto the scoring.

This is the workflow that Permissions will use, once it gets migrated to probes. The reason I don't migrate to probes at once is to reduce complexity and size of PRs. In addition, this PR highlights the limitations of the migration: The enums are used because findings have a `map[string]int` values map.

This check might require a few iterations of refactoring before introducing the probes.

the `release-note`

```release-note

```

### DIFF
--- a/checks/evaluation/permissions/permissions.go
+++ b/checks/evaluation/permissions/permissions.go
@@ -38,6 +38,58 @@ var (
 	topNoWriteID   = "gitHubWorkflowPermissionsTopNoWrite"
 )
 
+type PermissionLevel int
+
+const (
+	// PermissionLevelNone is a permission set to `none`.
+	PermissionLevelNone PermissionLevel = iota
+	// PermissionLevelRead is a permission set to `read`.
+	PermissionLevelRead
+	// PermissionLevelUnknown is for other kinds of alerts, mostly to support debug messages.
+	// TODO: remove it once we have implemented severity (#1874).
+	PermissionLevelUnknown
+	// PermissionLevelUndeclared is an undeclared permission.
+	PermissionLevelUndeclared
+	// PermissionLevelWrite is a permission set to `write` for a permission we consider potentially dangerous.
+	PermissionLevelWrite
+)
+
+// PermissionLocation represents a declaration type.
+type LocationType int
+
+const (
+	// PermissionLocationNil is in case the permission is nil.
+	PermissionLocationNil LocationType = iota
+	// PermissionLocationNotDeclared is for undeclared permission.
+	PermissionLocationNotDeclared
+	// PermissionLocationTop is top-level workflow permission.
+	PermissionLocationTop
+	// PermissionLocationJob is job-level workflow permission.
+	PermissionLocationJob
+)
+
+// PermissionLocation represents a declaration type.
+type PermissionType int
+
+const (
+	// PermissionTypeNone represents none permission type.
+	PermissionTypeNone PermissionType = iota
+	// PermissionTypeNone is the "all" github permission type.
+	PermissionTypeAll
+	// PermissionTypeNone is the "statuses" github permission type.
+	PermissionTypeStatuses
+	// PermissionTypeNone is the "checks" github permission type.
+	PermissionTypeChecks
+	// PermissionTypeNone is the "security-events" github permission type.
+	PermissionTypeSecurityEvents
+	// PermissionTypeNone is the "deployments" github permission type.
+	PermissionTypeDeployments
+	// PermissionTypeNone is the "packages" github permission type.
+	PermissionTypePackages
+	// PermissionTypeNone is the "actions" github permission type.
+	PermissionTypeActions
+)
+
 // TokenPermissions applies the score policy for the Token-Permissions check.
 func TokenPermissions(name string, c *checker.CheckRequest, r *checker.TokenPermissionsData) checker.CheckResult {
 	if r == nil {
@@ -49,7 +101,18 @@ func TokenPermissions(name string, c *checker.CheckRequest, r *checker.TokenPerm
 		return checker.CreateInconclusiveResult(name, "no tokens found")
 	}
 
-	score, err := applyScorePolicy(r, c)
+	if len(r.TokenPermissions) == 0 {
+		text := "no workflows found in the repository"
+		return checker.CreateInconclusiveResult(name, text)
+	}
+
+	// This is a temporary step that should be replaced by probes in ./probes
+	findings, err := rawToFindings(r)
+	if err != nil {
+		return checker.CreateInconclusiveResult(name, "could not convert raw data to findings")
+	}
+
+	score, err := applyScorePolicy(findings, c)
 	if err != nil {
 		return checker.CreateRuntimeErrorResult(name, err)
 	}
@@ -63,29 +126,10 @@ func TokenPermissions(name string, c *checker.CheckRequest, r *checker.TokenPerm
 		"GitHub workflow tokens follow principle of least privilege")
 }
 
-// avoid memory aliasing by returning a new copy.
-func newUint(u uint) *uint {
-	return &u
-}
-
-// avoid memory aliasing by returning a new copy.
-func newStr(s string) *string {
-	return &s
-}
-
-func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckRequest) (int, error) {
-	// See list https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/.
-	// Note: there are legitimate reasons to use some of the permissions like checks, deployments, etc.
-	// in CI/CD systems https://docs.travis-ci.com/user/github-oauth-scopes/.
-
-	hm := make(map[string]permissions)
-	dl := c.Dlogger
-	//nolint:errcheck
-	remediationMetadata, _ := remediation.New(c)
-	negativeProbeResults := map[string]bool{
-		stepsNoWriteID: false,
-		topNoWriteID:   false,
-	}
+// rawToFindings is a temporary step for converting the raw results
+// to findings. This should be replaced by probes in ./probes.
+func rawToFindings(results *checker.TokenPermissionsData) ([]finding.Finding, error) {
+	var findings []finding.Finding
 
 	for _, r := range results.TokenPermissions {
 		var loc *finding.Location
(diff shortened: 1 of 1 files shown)


checks/evaluation/permissions/permissions.go
```diff
@@ -49,7 +101,18 @@ func TokenPermissions(name string, c *checker.CheckRequest, r *checker.TokenPerm
 		return checker.CreateInconclusiveResult(name, "no github tokens found")
 	}
 
-	score, err := applyScorePolicy(r, c)
+	if len(r.TokenPermissions) == 0 {
+		text := "no workflows found in the repository"
+		return checker.CreateInconclusiveResult(name, text)
+	}
```
spencerschrock, line 107: any idea if/how this differs from the early exit above?

checks/evaluation/permissions/permissions.go
```diff
@@ -99,77 +143,215 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
 				loc.Snippet = newStr(r.File.Snippet)
 			}
 		}
-
 		text, err := createText(r)
 		if err != nil {
-			return checker.MinResultScore, err
+			return findings, err
```
spencerschrock, line 148: should this be `return nil, err`? Typically we return the zero value on error

checks/evaluation/permissions/permissions.go
```diff
@@ -99,77 +143,215 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
...
 		}
 
-		msg, err := createLogMsg(r.LocationType)
+		f, err := createFinding(r.LocationType, text, loc)
 		if err != nil {
-			return checker.InconclusiveResultScore, err
+			return findings, err
```
spencerschrock, line 153: same statement about `return nil, err`

checks/evaluation/permissions/permissions.go
```diff
@@ -99,77 +143,215 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
...
+		case checker.PermissionLevelRead:
+			f = f.WithOutcome(finding.OutcomePositive)
+			f = f.WithValues(map[string]int{
+				"PermissionLevel": int(PermissionLevelRead),
+			})
 
+		case checker.PermissionLevelUnknown:
+			f = f.WithValues(map[string]int{
+				"PermissionLevel": int(PermissionLevelUnknown),
+			})
```
spencerschrock, line 171: should this have a `finding.OutcomeError` or similar? by default `createFinding` creates a finding with `finding.OutcomeNegative`

checks/evaluation/permissions/permissions.go
```diff
@@ -99,77 +143,215 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
...
+			//nolint:gocritic
 			if r.LocationType == nil {
-				return checker.InconclusiveResultScore,
-					sce.WithMessage(sce.ErrScorecardInternal, "locationType is nil")
+				locationType = PermissionLocationNil
+			} else if *r.LocationType == checker.PermissionLocationTop {
+				locationType = PermissionLocationTop
+			} else {
+				locationType = PermissionLocationNotDeclared
+			}
```
spencerschrock, line 181: im slightly unfamiliar with this code, is `checker.PermissionLocationJob` possible for `checker.PermissionLevelUndeclared`?

checks/evaluation/permissions/permissions.go
```diff
@@ -99,77 +143,215 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
...
+			default:
+				locationType = PermissionLocationNotDeclared
 			}
+			permissionType := permTypeToEnum(r.Name)
+			f = f.WithValues(map[string]int{
+				"PermissionLevel": int(PermissionLevelWrite),
+				"LocationType":    int(locationType),
+				"PermissionType":  int(permissionType),
+			})
+			f = f.WithOutcome(finding.OutcomeNegative)
```
spencerschrock, line 204: hmm, I don't think all write permissions are considered bad under our current scoring policy, just top level ones. but this seems to match what the probe definition for job level currently seems to say.

spencerschrock (commented):
few comments:
1. do these types/consts/enums need to be exported here?
2. do you see these enum/string -> int -> enum/string conversions staying after the next PR. We discussed today how this is a limitation of the current finding type. Does any of this get better once the probes are acting on the raw data? I don't think so?

Thinking about what findings these probes will report, the data we would want to convey is:
permission X, declared at level Y, is Z. (e.g. permission contents, declared at top level is write)

these are more or less all strings, but the value map doesn't support it, hence the ints.

all of that data is already in the `Message` string of the finding, but it becomes difficult to act on that data when it's buried in the message string.

AdamKorcz:
> We discussed today how this is a limitation of the current finding type. Does any of this get better once the probes are acting on the raw data? I don't think so?

Not without changes to Scorecard, no.

AdamKorcz:
> do you see these enum/string -> int -> enum/string conversions staying after the next PR

Long-term this is not the solution. I will have a look at what I have bandwidth for to optimize it.

checks/evaluation/permissions/permissions.go
```diff
@@ -38,6 +38,58 @@ var (
...
+)
+
+// PermissionLocation represents a declaration type.
+type LocationType int
+
+const (
+	// PermissionLocationNil is in case the permission is nil.
+	PermissionLocationNil LocationType = iota
+	// PermissionLocationNotDeclared is for undeclared permission.
+	PermissionLocationNotDeclared
```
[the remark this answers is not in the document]
AdamKorcz: I have added both to accommodate to this part:

https://github.com/ossf/scorecard/blob/2ef20f17fb2e64147c83440cd2c769653454015a/checks/evaluation/permissions/permissions.go#L120-L131

Here, `r.LocationType == nil` should return inconclusive score, `PermissionLocationTop` should log at warning level and other cases should debug. In my changes, I therefore need a `PermissionLocationType` that is not `PermissionLocationNil` so I can debug those instead of returning inconclusive score. This happens at these lines:

https://github.com/ossf/scorecard/blob/f38d5f3521caf14c1a26a8c12dfb93ae3efec668/checks/evaluation/permissions/permissions.go#L317-L329

checks/evaluation/permissions/permissions.go
```diff
@@ -38,6 +38,58 @@ var (
...
+	// TODO: remove it once we have implemented severity (#1874).
+	PermissionLevelUnknown
+	// PermissionLevelUndeclared is an undeclared permission.
+	PermissionLevelUndeclared
+	// PermissionLevelWrite is a permission set to `write` for a permission we consider potentially dangerous.
+	PermissionLevelWrite
+)
+
+// PermissionLocation represents a declaration type.
+type LocationType int
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/ossf/scorecard/pull/3693/commits/c89fc07ec2eab2f0b41156aee1a7b005804cb285

checks/evaluation/permissions/permissions.go
AdamKorcz: This was unnecessary. Sorry for that. Removed in https://github.com/ossf/scorecard/pull/3693/commits/d49d788ac83c09e32c59bd43d499c6599ca050eb

checks/evaluation/permissions/permissions.go
AdamKorcz: https://github.com/ossf/scorecard/pull/3693/commits/ef8518e2e5340b83c617cda46cd649e004d5b10f

checks/evaluation/permissions/permissions.go
AdamKorcz: https://github.com/ossf/scorecard/pull/3693/commits/ef8518e2e5340b83c617cda46cd649e004d5b10f

checks/evaluation/permissions/permissions.go
AdamKorcz: According to the current evaluation, yes:

https://github.com/ossf/scorecard/blob/2ef20f17fb2e64147c83440cd2c769653454015a/checks/evaluation/permissions/permissions.go#L120-L131

checks/evaluation/permissions/permissions.go
```diff
@@ -38,6 +38,58 @@ var (
 	topNoWriteID   = "gitHubWorkflowPermissionsTopNoWrite"
 )
 
+type PermissionLevel int
+
```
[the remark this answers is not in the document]
AdamKorcz: Fair point. I would prefer to defer that to a PR that adds the probes to not solve too many problems at once. There will still be a few more things to pick up, and I would prefer to do that later.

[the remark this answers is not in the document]
spencerschrock: Did you see my other comment about leaving them unexported if they're not needed?
> few comments:
> 
> do these types/consts/enums need to be exported here?

checks/evaluation/permissions/permissions.go
```diff
@@ -49,7 +101,13 @@ func TokenPermissions(name string, c *checker.CheckRequest, r *checker.TokenPerm
 		return checker.CreateInconclusiveResult(name, "no github tokens found")
 	}
 
-	score, err := applyScorePolicy(r, c)
+	// This is a temporary step that should be replaced by probes in ./probes
+	findings, err := rawToFindings(r)
+	if err != nil {
+		return checker.CreateInconclusiveResult(name, "could not convert raw data to findings")
```
spencerschrock, line 107: `checker.CreateRuntimeErrorResult`

checks/evaluation/permissions/permissions.go
```diff
@@ -99,77 +138,215 @@ func applyScorePolicy(results *checker.TokenPermissionsData, c *checker.CheckReq
...
+
+func createFinding(loct *checker.PermissionLocation, text string, loc *finding.Location) (*finding.Finding, error) {
+	probe := stepsNoWriteID
+	if loct == nil || *loct == checker.PermissionLocationTop {
+		probe = topNoWriteID
+	}
+	content, err := probes.ReadFile(probe + ".yml")
+	if err != nil {
+		return nil, fmt.Errorf("%w", err)
+	}
```
spencerschrock, line 262: add some context to the wrap. "%w" is circumventing the linter

spencerschrock (approved):
Approved, but left a few comments about minor changes I'd like to see still. 

> these are more or less all strings, but the value map doesn't support it

One alternative for future cleanup is just making the Values map a string -> string map. Int values can be parsed from strings, but this might just be delaying the problem for the next time.

checks/evaluation/permissions/permissions.go
AdamKorcz: Not sure I understood correctly, but added a comment here: https://github.com/ossf/scorecard/pull/3693/commits/e8762564509bd47b452864780d1bb496c0d74262

checks/evaluation/permissions/permissions.go
AdamKorcz: https://github.com/ossf/scorecard/pull/3693/commits/a4eaced370dc3a208c345f067f737639ff91c15d

checks/evaluation/permissions/permissions.go
spencerschrock: I was talking about the "Wrapping errors with %w" section of this blogpost.
https://go.dev/blog/go1.13-errors#:~:text=Wrapping%20errors%20with%20%25w

The linter's purpose is to add additional information to an error, but `fmt.Errorf("%w", err)` isn't adding anything. Instead, I meant something like:
```go
if err != nil {
	return nil, fmt.Errorf("reading probe file: %w", err)
}
```

or include the probe:
```go
if err != nil {
	return nil, fmt.Errorf("reading %v.yml: %w", probe, err)
}
```

AdamKorcz: Done in https://github.com/ossf/scorecard/pull/3693/commits/ee44701c50df49b8775c9c8089fe47ba2b10c1ed

PR #3706, opened 2023-11-29
🐛 Parse Gitlab Status fields to align w/Github Status and Conclusion

### LINKED ISSUE #3396: GitLab job/pipeline don't have conclusion fields
GitHub worksflows have two fields, `Status` and `Conclusion`, which is what we built into our model.

GitLab, just has a `Status`. I think we can parse the status field to set `status` and `conclusion` to their GitHub equivalent, or we can adjust the model going forward to be more generic. We may not have an answer immediately.

This seems to affect the `CI-Tests` and `SAST` checks.

ashearin:

Introduces code to parse status returned from gitlab to align with github status and conclusion fields

When populating checkruns from Gitlab commits, we currently assign the gitlab status to the checkrun status and leave the conclusion blank (Since gitlab does not have an equivalent field, the gitlab status field is roughly a combination of the github status and conclusion fields)

Added a function to parse the gitlab status into the appropriate github equivalent status and conclusion. In most cases the Conclusion is still empty (since the pipeline is not completed), but the status will match the equivalent Github status. 

References: 
- GItlab Status [Enum](https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/graphql/types/ci/pipeline_status_enum.rb?ref_type=heads)
- Github CheckRun [Docs](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28)
  - looking at the response schema for status and conclusion fields

Fixes #3396

the `release-note`

```release-note

```


clients/gitlabrepo/checkruns.go
```diff
@@ -50,10 +50,42 @@ func checkRunsFrom(data []*gitlab.PipelineInfo) []clients.CheckRun {
...
 	return checkRuns
 }
+
+// Conclusion does not exist in the pipelines for gitlab,
+// so we parse the status to determine the conclusion if it exists.
+func parseGitlabStatus(info *gitlab.PipelineInfo) clients.CheckRun {
+	checkrun := clients.CheckRun{
+		URL: info.WebURL,
+	}
+	completed := "completed"
```
spencerschrock, line 64: nit: 
```go
const completed = "completed"
````

File: clients/gitlabrepo/checkruns.go [hunk 2]
```diff
@@ -50,10 +50,42 @@ func checkRunsFrom(data []*gitlab.PipelineInfo) []clients.CheckRun {
...
+		checkrun.Conclusion = "success"
+	case "canceled":
+		checkrun.Status = completed
+		checkrun.Conclusion = "cancelled"
+	case "skipped":
+		checkrun.Status = completed
+		checkrun.Conclusion = "skipped"
+	case "manual":
+		checkrun.Status = completed
+		checkrun.Conclusion = "action_required"
```
spencerschrock, line 85: Should the status be `queued` here as well?

spencerschrock: I'm trying to find an example of `action_required`, but didn't see anything in the few minutes I looked.

spencerschrock: Ah never mind, it's correct as written
```
      "status": "completed",
      "conclusion": "action_required",
```
https://api.github.com/repos/ossf/scorecard/actions/runs/7030417252

spencerschrock (approved):

PR #3729, opened 2023-12-12
🌱  differentiate between refs and sha gitab

### LINKED ISSUE #3395: Difference between GitLab refs and GitHub refs
As part of #3394 , I noticed the GitLab handler assumes `ref` refers to a SHA. But in the GitLab API [link](https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines), ref is assumed to be 
> The branch or tag to check for the latest pipeline. Defaults to the default branch when not specified.

Where as on GitHub [doc link](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#list-check-runs-for-a-git-reference) refs can be commits, or branches, or tags:
> The commit reference. Can be a commit SHA, branch name (heads/BRANCH_NAME), or tag name (tags/TAG_NAME).

We may need to fix the GitLab code to distinguish between branches, tags, and commits.

ashearin:
Adds a check to `listCheckRunsForRef` in the Gitlab Handler to differentiate between shas and branches/tags and makes appropriate changes to the gitlab api call based on that check. 

`listCheckRunsForRef` in the GItlab Handler takes in a `ref` string, but assumes that the ref is a sha when making the gitlab api call. Gitlab API sometimes handles refs as a branch or tag name, while sha refers to a commit hash. If in the future we passed in a branch or tag to `listCheckRunsForRef` the API would fail since we would handle it as a sha. 

Added a check to compare the passed in `ref` string against a regex matching a SHA value. If there is a match we handle the ref as a SHA, otherwise we handle it as a ref (branch/tag)

- [x] Tests for the changes have been updated

Fixes #3395 

the `release-note`

```release-note

```

### DIFF
--- a/clients/gitlabrepo/checkruns.go
+++ b/clients/gitlabrepo/checkruns.go
@@ -16,6 +16,7 @@ package gitlabrepo
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/xanzy/go-gitlab"
 
@@ -32,11 +33,20 @@ func (handler *checkrunsHandler) init(repourl *repoURL) {
 }
 
 func (handler *checkrunsHandler) listCheckRunsForRef(ref string) ([]clients.CheckRun, error) {
+	commit := regexp.MustCompile("^[a-f0-9]{40}$")
+	options := gitlab.ListProjectPipelinesOptions{
+		ListOptions: gitlab.ListOptions{},
+	}
+
+	// In Gitlab: ref refers to a branch or tag name, and sha refers to commit sha
+	if commit.MatchString(ref) {
+		options.SHA = &ref
+	} else {
+		options.Ref = &ref
+	}
+
 	pipelines, _, err := handler.glClient.Pipelines.ListProjectPipelines(
-		handler.repourl.projectID, &gitlab.ListProjectPipelinesOptions{
-			SHA:         &ref,
-			ListOptions: gitlab.ListOptions{},
-		})
+		handler.repourl.projectID, &options)
 	if err != nil {
 		return nil, fmt.Errorf("request for pipelines returned error: %w", err)
 	}
--- a/clients/gitlabrepo/checkruns.go
+++ b/clients/gitlabrepo/checkruns.go
@@ -23,6 +23,8 @@ import (
 	"github.com/ossf/scorecard/v4/clients"
 )
 
+var gitCommitHashRegex = regexp.MustCompile(`^[a-fA-F0-9]{40}$`)
+
 type checkrunsHandler struct {
 	glClient *gitlab.Client
 	repourl  *repoURL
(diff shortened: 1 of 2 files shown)


clients/gitlabrepo/checkruns.go
```diff
@@ -32,11 +33,20 @@ func (handler *checkrunsHandler) init(repourl *repoURL) {
 }
 
 func (handler *checkrunsHandler) listCheckRunsForRef(ref string) ([]clients.CheckRun, error) {
+	commit := regexp.MustCompile("^[a-f0-9]{40}$")
+	options := gitlab.ListProjectPipelinesOptions{
+		ListOptions: gitlab.ListOptions{},
+	}
```
spencerschrock, line 39: optional/style: lets favor zero value if we don't need to set any fields
```go
var options gitlab.ListProjectPipelinesOptions
```

clients/gitlabrepo/checkruns.go
```diff
@@ -32,11 +33,20 @@ func (handler *checkrunsHandler) init(repourl *repoURL) {
 }
 
 func (handler *checkrunsHandler) listCheckRunsForRef(ref string) ([]clients.CheckRun, error) {
+	commit := regexp.MustCompile("^[a-f0-9]{40}$")
```
spencerschrock, line 36: let's only compile the regex once as this gets called multiple times, similar to like we do here:
https://github.com/ossf/scorecard/blob/663e1a9bad0f62b635cce71af8a1b3a0dabe7ba1/checks/raw/shell_download_validate.go#L60

On a similar note: let's add `A-Z` as a possibility too.

clients/gitlabrepo/checkruns.go
```diff
@@ -32,11 +33,20 @@ func (handler *checkrunsHandler) init(repourl *repoURL) {
...
+	} else {
+		options.Ref = &ref
+	}
+
 	pipelines, _, err := handler.glClient.Pipelines.ListProjectPipelines(
-		handler.repourl.projectID, &gitlab.ListProjectPipelinesOptions{
-			SHA:         &ref,
-			ListOptions: gitlab.ListOptions{},
-		})
+		handler.repourl.projectID, &options)
```
spencerschrock, line 49: optional/style: now that we're not initing a struct, let's just keep this all on one line instead of splitting

spencerschrock (approved):
Thanks, left a few small comments

clients/gitlabrepo/checkruns.go
```diff
@@ -32,11 +35,16 @@ func (handler *checkrunsHandler) init(repourl *repoURL) {
...
-			SHA:         &ref,
-			ListOptions: gitlab.ListOptions{},
-		})
+	var options gitlab.ListProjectPipelinesOptions
+
+	// In Gitlab: ref refers to a branch or tag name, and sha refers to commit sha
+	if gitCommitHashRegex.MatchString(ref) {
+		options.SHA = &ref
+	} else {
+		options.Ref = &ref
```
raghavkaul, line 44: Does options.Ref also work if the user passes in something like a 7-character SHA?

ashearin: So initially I set it up to only count full length shas, since thats what is pulled in the `ListCommits()` function, even if a user passes in the shortsha, tag or branch with the `--commit` flag. I hadn't considered to try if the shortsha works in the ref param. 

Testing with postman against the [ListProjectPipelines](https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines) endpoint:

- Sha Param: the full length Sha returns the expected commits, the short sha (8 characters in gitlab, I also tried 7) returns an empty list. 
- Ref Param: Branch name returns the expected commits, tags and shas (40 or 8 characters) returns an empty list. Given the documentation I expected tags to also work. 

I'll have to update the comment on 40 to reflect this.

spencerschrock (approved):

clients/gitlabrepo/checkruns.go
spencerschrock: That's weird about the tags. Can always fix it later if we find out we're doing something wrong.

e.g. if we needed to use `tags/foo` instead of `foo` (i tried a few variations and couldn't get it to work)

PR #3743, opened 2023-12-19
🌱 Add SAST Pysa probe

DavidKorczynski:

Adds a probe checking the presence of Pysa SAST Github action (https://github.com/facebook/pysa-action)

Scorecard will not recognize Pysa scanning as a SAST action

Scorecard will recognize Pysa scanning as a SAST action.

Yes, in that it will chnge scoring of projects that use Pysa and users may now see Pysa-related details in the output.

```release-note

```

### DIFF
--- a/checks/raw/sast.go
+++ b/checks/raw/sast.go
@@ -72,6 +72,13 @@ func SAST(c *checker.CheckRequest) (checker.SASTData, error) {
 		return data, err
 	}
 	data.Workflows = append(data.Workflows, snykWorkflows...)
+
+	pysaWorkflows, err := getPysaWorkflows(c)
+	if err != nil {
+		return data, err
+	}
+
+	data.Workflows = append(data.Workflows, pysaWorkflows...)
 	return data, nil
 }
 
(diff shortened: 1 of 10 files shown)


DavidKorczynski:
Putting this in draft as I'll tackle https://github.com/ossf/scorecard/issues/3745 first

checks/raw/sast.go
```diff
@@ -67,19 +66,19 @@ func SAST(c *checker.CheckRequest) (checker.SASTData, error) {
...
 
-	pysaWorkflows, err := getPysaWorkflows(c)
+	pysaWorkflows, err := getSastUsesWorkflows(c, "^facebook/pysa-action$", checker.PysaWorkflow)
 	if err != nil {
 		return data, err
 	}
 	data.Workflows = append(data.Workflows, pysaWorkflows...)
 
-	qodanaWorkflows, err := getQodanaWorkflows(c)
+	qodanaWorkflows, err := getSastUsesWorkflows(c, "^JetBrains/qodana-action$", checker.QodanaWorkflow)
```
spencerschrock, line 81: Do we need to worry about case sensitivity here (or generally for any of them)?

checks/raw/sast.go
```diff
@@ -142,13 +141,17 @@ func sastToolInCheckRuns(c *checker.CheckRequest) ([]checker.SASTCommit, error)
 	return sastCommits, nil
 }
 
-func codeQLInCheckDefinitions(c *checker.CheckRequest) ([]checker.SASTWorkflow, error) {
+func getSastUsesWorkflows(
+	c *checker.CheckRequest,
+	useRegex string,
```
spencerschrock, line 146: should this be `usesRegex`? Is this labeled this way because it's applied to the `uses` field? Might be worth a little doc comment.

checks/raw/sast.go
```diff
@@ -159,100 +162,38 @@ func codeQLInCheckDefinitions(c *checker.CheckRequest) ([]checker.SASTWorkflow,
...
 ) (bool, error) {
 	if !fileparser.IsWorkflowFile(path) {
 		return true, nil
 	}
 
-	if len(args) != 1 {
+	if len(args) != 2 {
 		return false, fmt.Errorf(
-			"searchGitHubActionWorkflowCodeQL requires exactly 1 arguments: %w", errInvalid)
+			"searchGitHubActionWorkflowUseRegex requires exactly 1 arguments: %w", errInvalid)
```
spencerschrock, line 183: error message says 1 arg, needs to be updated

checks/raw/sast.go
```diff
@@ -269,141 +210,11 @@ var searchGitHubActionWorkflowSnyk fileparser.DoWhileTrueOnFileContent = func(pa
...
-	if len(errs) > 0 && workflow == nil {
-		return false, fileparser.FormatActionlintError(errs)
-	}
-
-	for _, job := range workflow.Jobs {
-		for _, step := range job.Steps {
-			e, ok := step.Exec.(*actionlint.ExecAction)
-			if !ok || e == nil || e.Uses == nil {
+			m, err := regexp.MatchString(useRegex, action)
+			if err != nil {
```
spencerschrock, line 214: Any preference between this and trying to force the pattern to compile? I worry errors from bad regexes here may slip through since you just silently `continue` below. `MustCompile` causes a panic, but this would be a programmer error we'd want to catch.

```go
re := regexp.MustCompile(useRegex)
if re.MatchString(action) {
    *paths = append(*paths, path)
}
```

DavidKorczynski: Good point -- I set it to the pattern you suggested.

checks/raw/sast.go
DavidKorczynski: Fixed.

checks/raw/sast.go
DavidKorczynski: > Is this labeled this way because it's applied to the uses field?

Yes -- I renamed it from `useRegex` to `usesRegex` and added a doc comment.

checks/raw/sast.go
DavidKorczynski: I'm not 100% sure, but as far as I can tell the answer is *yes* considering `uses` can refer to Docker container images https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses which have case sensitive tags.

spencerschrock: docker images also start with `docker://`, but we can start with case sensitivity until we see evidence otherwise.
https://docs.github.com/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-docker-hub-action

DavidKorczynski: I experimented a bit https://github.com/DavidKorczynski/intellij-ton/commit/b97abd243e910eba1bb562ad98e38f2d8da7284f and https://github.com/DavidKorczynski/intellij-ton/commit/2b1cc7cbe9c6abcdc6a9656a9cc32a5c4d9fcf45 -- and in this case the action is case insensitive. So, I'm kinda leaning towards there being no case sensitivity, but will just check up a bit more.

spencerschrock (approved):
Thanks.

PR #3753, opened 2023-12-28
:bug: Fix signed release error for empty gitlab repo

naveensrinivasan:
- Fixed the issue where an empty gitlab repo is causing this error. `Error: check runtime error: Signed-Releases: internal error: could not get release name 2023/12/27 18:07:19 error during command execution: check runtime error: Signed-Releases: internal error: could not get release name exit status 1`

the `release-note`

```release-note

```

### DIFF
--- a/checks/evaluation/signed_releases.go
+++ b/checks/evaluation/signed_releases.go
@@ -15,6 +15,7 @@
 package evaluation
 
 import (
+	"errors"
 	"fmt"
 	"math"
 
@@ -24,6 +25,8 @@ import (
 	"github.com/ossf/scorecard/v4/probes/releasesAreSigned"
 	"github.com/ossf/scorecard/v4/probes/releasesHaveProvenance"
 )
+
+var errNoReleaseFound = errors.New("no release found")
 
 // SignedReleases applies the score policy for the Signed-Releases check.
 func SignedReleases(name string,
@@ -48,19 +51,13 @@ func SignedReleases(name string,
 
 		// Debug release name
 		if f.Outcome == finding.OutcomeNotApplicable {
-			dl.Warn(&checker.LogMessage{
-				Text: "no releases found",
-			})
 			// Generic summary.
 			return checker.CreateInconclusiveResult(name, "no releases found")
 		}
 		releaseName := getReleaseName(f)
 		if releaseName == "" {
-			dl.Warn(&checker.LogMessage{
-				Text: "no releases found",
-			})
 			// Generic summary.
-			return checker.CreateInconclusiveResult(name, "no releases found")
+			return checker.CreateRuntimeErrorResult(name, errNoReleaseFound)
 		}
 
 		if !contains(loggedReleases, releaseName) {
@@ -83,7 +80,7 @@ func SignedReleases(name string,
 
 		releaseName := getReleaseName(f)
 		if releaseName == "" {
-			return checker.CreateInconclusiveResult(name, "no releases found")
+			return checker.CreateRuntimeErrorResult(name, errNoReleaseFound)
 		}
 		if !contains(uniqueReleaseTags, releaseName) {
 			uniqueReleaseTags = append(uniqueReleaseTags, releaseName)
--- a/checks/evaluation/signed_releases.go
+++ b/checks/evaluation/signed_releases.go
@@ -47,10 +47,13 @@ func SignedReleases(name string,
 		f := &findings[i]
 
 		// Debug release name
-		releaseName, err := getReleaseName(f)
-		if err != nil {
-			e := sce.WithMessage(sce.ErrScorecardInternal, "could not get release name")
-			return checker.CreateRuntimeErrorResult(name, e)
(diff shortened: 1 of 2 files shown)


checks/evaluation/signed_releases.go
```diff
@@ -47,10 +47,13 @@ func SignedReleases(name string,
 		f := &findings[i]
 
 		// Debug release name
```
spencerschrock, line 49: We just need to move this block higher up:
```go
		// Check if outcome is NotApplicable
		if f.Outcome == finding.OutcomeNotApplicable {
			dl.Warn(&checker.LogMessage{
				Text: "no GitHub releases found",
			})
			// Generic summary.
			return checker.CreateInconclusiveResult(name, "no releases found")
		}
```

spencerschrock (commented):
As part of our structured result works, we try to use the finding outcomes when determining things like "no releases".

In this case, the convention is a single finding with `OutcomeNotApplicable`.
https://github.com/ossf/scorecard/blob/5d8767ebb3a3f135a78f41cc6061ba0c28cad005/probes/releasesAreSigned/def.yml#L15-L24

checks/evaluation/signed_releases.go
spencerschrock: note: this `Text` says `no GitHub releases found`, but really should just be `no releases found`

naveensrinivasan:
@spencerschrock Let me know if this is good.

checks/evaluation/signed_releases.go
```diff
@@ -47,11 +47,22 @@ func SignedReleases(name string,
...
-		releaseName, err := getReleaseName(f)
-		if err != nil {
-			e := sce.WithMessage(sce.ErrScorecardInternal, "could not get release name")
-			return checker.CreateRuntimeErrorResult(name, e)
+		if f.Outcome == finding.OutcomeNotApplicable {
+			dl.Warn(&checker.LogMessage{
+				Text: "no releases found",
+			})
```
spencerschrock, line 53: optional: this is redundant with the reason string, and could probably be removed

checks/evaluation/signed_releases.go
```diff
@@ -47,11 +47,22 @@ func SignedReleases(name string,
...
+			// Generic summary.
+			return checker.CreateInconclusiveResult(name, "no releases found")
+		}
+		releaseName := getReleaseName(f)
+		if releaseName == "" {
+			dl.Warn(&checker.LogMessage{
+				Text: "no releases found",
+			})
+			// Generic summary.
```
spencerschrock, line 62: optional: same comment about redundant

checks/evaluation/signed_releases.go
```diff
@@ -47,11 +47,22 @@ func SignedReleases(name string,
...
+			return checker.CreateInconclusiveResult(name, "no releases found")
```
spencerschrock, line 63: I actually think this was better as a `CreateRuntimeErrorResult` before, as not having a release name would be an error, and isn't indicative of not having releases

checks/evaluation/signed_releases.go
```diff
@@ -77,11 +81,10 @@ func SignedReleases(name string,
 	for i := range findings {
 		f := &findings[i]
 
-		releaseName, err := getReleaseName(f)
-		if err != nil {
-			return checker.CreateRuntimeErrorResult(name, err)
+		releaseName := getReleaseName(f)
+		if releaseName == "" {
+			return checker.CreateInconclusiveResult(name, "no releases found")
```
spencerschrock, line 86: same comment about `CreateRuntimeErrorResult`

spencerschrock (approved):
I think moving the `if f.Outcome == finding.OutcomeNotApplicable` block up is all that was needed. 

I'm fine with the `getReleaseName` function signature changes, but still think not having a release name should be considered an error as it would indicate a programming mistake on our side, not anything wrong with the repo.

spencerschrock:
/scdiff generate Signed-Releases

(just a tool to help visualize changes across 15 test repos)

checks/evaluation/signed_releases.go
naveensrinivasan: Thanks, I agree.

checks/evaluation/signed_releases.go
naveensrinivasan: 👍

checks/evaluation/signed_releases.go
naveensrinivasan: 👍

checks/evaluation/signed_releases.go
naveensrinivasan: 👍

naveensrinivasan:
> I think moving the `if f.Outcome == finding.OutcomeNotApplicable` block up is all that was needed.
> 
> I'm fine with the `getReleaseName` function signature changes, but still think not having a release name should be considered an error as it would indicate a programming mistake on our side, not anything wrong with the repo.

Thanks

PR #3759, opened 2023-12-29
:seedling: Add branch protection evaluation

AdamKorcz:

- Changes Branch Protection evaluation to process probe findings. Some of the probes were submitted here: https://github.com/ossf/scorecard/pull/3691
- Adds another probe
- Rewrites a bunch of tests

NONE

NONE
the `release-note`

```release-note

```


checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
+				continue
+			}
+			branchName = k
+		}
 		// Protected field only indates that the branch matches
 		// one `Branch protection rules`. All settings may be disabled,
 		// so it does not provide any guarantees.
-		protected := !(b.Protected != nil && !*b.Protected)
-		if !protected {
+		protected := (f.Values["branchProtected"] == 1)
```
AdamKorcz, line 113: The `protected` value comes from the probe findings. `1` means `true`, `0` means `false`.

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
+			score, max = basicNonAdminProtection(f, dl)
+			branchScores[branchName].scores.basic += score
+			branchScores[branchName].maxes.basic += max
+
+		case dismissesStaleReviews.Probe, branchProtectionAppliesToAdmins.Probe:
+			score, max = adminThoroughReviewProtection(f, dl)
+			branchScores[branchName].scores.adminThoroughReview += score
+			branchScores[branchName].maxes.adminThoroughReview += max
+
+		case requiresApproversForPullRequests.Probe:
```
AdamKorcz, line 142: This probe processes the `branch.BranchProtectionRule.RequiredPullRequestReviews.RequiredApprovingReviewCount` value. This value is used in two places in the evaluation. As such, we have a single probe that is used two places. The differentiating factor is the `"numberOfRequiredReviewers"` value.

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
+				Text: fmt.Sprintf("branch protection not enabled for branch '%s'", branchName),
 			})
+			warnedBranches = append(warnedBranches, branchName)
+		}
+		if branchName == "" {
+			e := sce.WithMessage(sce.ErrScorecardInternal, "probe is missing branch name")
+			return checker.CreateRuntimeErrorResult(name, e)
+		}
+
+		if _, ok := branchScores[branchName]; !ok {
```
AdamKorcz, line 125: `branchScores` is a map that has the score for each branch. Map key = branch name. Map value = the score.

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
-
-		scores = append(scores, score)
 	}
 
-	if len(scores) == 0 {
+	if len(branchScores) == 0 {
 		return checker.CreateInconclusiveResult(name, "unable to detect any development/release branches")
 	}
 
+	var scores []levelScore
```
AdamKorcz, line 179: Convert the map to a slice to avoid modifying `computeFinalScore`.

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
+
+		case requiresApproversForPullRequests.Probe:
+			// Scorecard evaluation scores twice with this probe:
+			// Once if the count is above 0
+			// Once if the count is above 2
+			score, max = nonAdminThoroughReviewProtection(f, dl)
+			branchScores[branchName].scores.thoroughReview += score
+			branchScores[branchName].maxes.thoroughReview += max
+
+			reviewerWeight := 2
```
AdamKorcz, line 150: These lines were an internal api on their own called `nonAdminReviewProtection` which I have removed.

checks/raw/branch_protection.go
```diff
@@ -51,7 +51,8 @@ func (set branchSet) contains(branch string) bool {
 }
 
 // BranchProtection retrieves the raw data for the Branch-Protection check.
-func BranchProtection(c clients.RepoClient) (checker.BranchProtectionsData, error) {
+func BranchProtection(cr *checker.CheckRequest) (checker.BranchProtectionsData, error) {
+	c := cr.RepoClient
```
AdamKorcz, line 55: Change the parameter from `clients.RepoClient` to `*checker.CheckRequest` to align it with similar APIs for other checks.

checks/raw/branch_protection_test.go
```diff
@@ -273,7 +273,11 @@ func TestBranchProtection(t *testing.T) {
 					return tt.releases, tt.releasesErr
 				})
 			mockRepoClient.EXPECT().ListFiles(gomock.Any()).AnyTimes().Return(tt.repoFiles, nil)
-			rawData, err := BranchProtection(mockRepoClient)
+
+			c := &checker.CheckRequest{
+				RepoClient: mockRepoClient,
```
AdamKorcz, line 278: Same as https://github.com/ossf/scorecard/pull/3759/files#r1438390648

probes/blocksDeleteOnBranches/impl.go
```diff
@@ -37,9 +37,26 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
...
+			return nil, Probe, fmt.Errorf("create finding: %w", err)
+		}
+		findings = append(findings, *f)
+		return findings, Probe, nil
+	}
+
 	for i := range r.Branches {
 		branch := &r.Branches[i]
 
+		protected := !(branch.Protected != nil && !*branch.Protected)
```
AdamKorcz, line 52: Previously, this was done in the evaluation here:https://github.com/ossf/scorecard/blob/69bb742f12a56d223577b4984cf88b43fa56cd7d/checks/evaluation/branch_protection.go#L76-L81. Now every probe does this.

probes/requiresCodeOwnersReview/impl_test.go
```diff
@@ -40,7 +40,7 @@ func Test_Run(t *testing.T) {
 		err      error
 	}{
 		{
-			name: "1 branch requires code owner reviews with viles = 1 positive outcome",
+			name: "1 branch requires code owner reviews with files = 1 positive outcome",
```
AdamKorcz, line 43: This probe was not _entirely_ correct so it needed minor changes.

probes/requiresPRsToChangeCode/def.yml
```diff
@@ -0,0 +1,32 @@
+# Copyright 2023 OpenSSF Scorecard Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
```
AdamKorcz, line 4: This is a new probe that was missing from the previous PR with the Branch Protection probes.

checks/evaluation/branch_protection_test.go
```diff
@@ -17,564 +17,1396 @@ package evaluation
...
-		name            string
-		branch          *clients.BranchRef
-		codeownersFiles []string
-		expected        scut.TestReturn
+		name     string
+		findings []finding.Finding
+		result   scut.TestReturn
 	}{
 		{
-			name: "GitHub default settings",
```
AdamKorcz, line 53: To a large extent, `&falseVal` has been changed to `OutcomeNegative` and `&trueVal` has been changed to `OutcomePositive`. Missing values or `nil` can be either `OutcomeNegative` or `OutcomeNotAvailable` depending on the probe.

`RequireLinearHistory` existed in the test but is not used in the evaluation.

AdamKorcz: I have not changed results from the tests to make a point out of compatibility. 

I have not added new tests, nor have I removed any.

probes/requiresCodeOwnersReview/impl.go
```diff
@@ -50,9 +77,6 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
 		case !*reqOwnerReviews:
 			text = fmt.Sprintf("codeowners review is not required on branch '%s'", *branch.Name)
 			outcome = finding.OutcomeNegative
-		case len(r.CodeownersFiles) == 0:
-			text = "codeowners review is required - but no codeowners file found in repo"
-			outcome = finding.OutcomeNegative
```
spencerschrock, line 55: out of curiosity, what was there a reason for the change?

At first glance the current implementation seems reasonable?

spencerschrock (commented):
only a partial review (ran out of time before lunch), but wanted to get a few of these questions to you for discussion. Particularly the "protected" branch value/probe discussion

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
 			dl.Warn(&checker.LogMessage{
-				Text: fmt.Sprintf("branch protection not enabled for branch '%s'", *b.Name),
+				Text: fmt.Sprintf("branch protection not enabled for branch '%s'", branchName),
 			})
+			warnedBranches = append(warnedBranches, branchName)
+		}
+		if branchName == "" {
+			e := sce.WithMessage(sce.ErrScorecardInternal, "probe is missing branch name")
+			return checker.CreateRuntimeErrorResult(name, e)
+		}
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/ossf/scorecard/pull/3759/commits/ddb074fba1a2ad6af3313ed96a92d992142f94c9

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
+			return checker.CreateInconclusiveResult(name, "unable to detect any development/release branches")
+		}
 
+		var branchName string
+		for k := range f.Values {
+			if k == "branchProtected" || k == "numberOfRequiredReviewers" || k == "codeownersFiles" {
+				continue
+			}
+			branchName = k
+		}
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/ossf/scorecard/pull/3759/commits/5558e2f3463eacfcea9a9cc19878211cdc4388d3

AdamKorcz:
https://github.com/ossf/scorecard/pull/3759#discussion_r1439768200 fixed in https://github.com/ossf/scorecard/pull/3759/commits/ba40d034438c6ac5bd64801c531076720d69ae45

checks/evaluation/branch_protection.go
AdamKorcz: Added in https://github.com/ossf/scorecard/pull/3759/commits/7d8ce6fd83f2445ed20b08a589eaac21d07d4487

probes/requiresCodeOwnersReview/impl.go
AdamKorcz: In short: To determine when to log at info level here: 

https://github.com/ossf/scorecard/blob/fdf3fb23652cb69e19204421c0cede904bd16061/checks/evaluation/branch_protection.go#L412

The relevant evaluation is here: 

https://github.com/ossf/scorecard/blob/fdf3fb23652cb69e19204421c0cede904bd16061/checks/evaluation/branch_protection.go#L409-L421

`*branch.BranchProtectionRule.RequiredPullRequestReviews.RequireCodeOwnerReviews` can be true while `codeOwnerFiles` can be 0. 

If the probe returns `OutcomeNegative` if `len(codeownersFiles)` is 0, then the evaluation won't know whether to log the `info()` call.

probes/blocksDeleteOnBranches/impl.go
AdamKorcz: Added in https://github.com/ossf/scorecard/pull/3759/commits/7d8ce6fd83f2445ed20b08a589eaac21d07d4487

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +70,117 @@ const (
...
+			branchScores[branchName] = &levelScore{}
+		}
+
+		var score, max int
+
+		switch f.Probe {
+		case blocksDeleteOnBranches.Probe, blocksForcePushOnBranches.Probe:
+			score, max = basicNonAdminProtection(f, dl)
+			branchScores[branchName].scores.basic += score
+			branchScores[branchName].maxes.basic += max
```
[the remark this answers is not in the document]
AdamKorcz: I just reused the existing names. `basicNonAdminProtection` is the name used here:

https://github.com/ossf/scorecard/blob/fdf3fb23652cb69e19204421c0cede904bd16061/checks/evaluation/branch_protection.go#L227-L234

AdamKorcz:
The failing integration test looks unrelated, but reliably reproduces:

```
• [FAILED] [0.487 seconds]
E2E TEST:SearchCommits E2E TEST:SearchCommits [It] Should return commits by dependabot
/home/runner/work/scorecard/scorecard/e2e/searchCommits_test.go:29

  [FAILED] Expected
      <int>: 0
  to be >
      <int>: 0
  In [It] at: /home/runner/work/scorecard/scorecard/e2e/searchCommits_test.go:37 @ 01/09/24 17:24:45.055
```

spencerschrock:
> The failing integration test looks unrelated, but reliably reproduces:

I think this had to do with [today's GitHub status incident](https://www.githubstatus.com/incidents/pxg3dz4yg7lp).

AdamKorcz:
> > The failing integration test looks unrelated, but reliably reproduces:
> 
> I think this had to do with [today's GitHub status incident](https://www.githubstatus.com/incidents/pxg3dz4yg7lp).

Thanks! Ran again, and now it works.

probes/requiresPRsToChangeCode/impl.go
```diff
@@ -0,0 +1,102 @@
...
+		} else {
+			protectedValue = 0
+		}
+
+		nilMsg := fmt.Sprintf("could not determine whether branch '%s' requires PRs to change code", *branch.Name)
+		trueMsg := fmt.Sprintf("PRs are required in order to make changes on branch '%s'", *branch.Name)
+		falseMsg := fmt.Sprintf("PRs are not required to make changes on branch '%s'; ", *branch.Name) +
+			"or we don't have data to detect it." +
+			"If you think it might be the latter, make sure to run Scorecard with a PAT or use Repo " +
+			"Rules (that are always public) instead of Branch Protection settings"
```
spencerschrock, line 68: @diogoteles08 how attached are you to the `false` reason string here mentioning we may be unable to detect it? Otherwise we could probably simplify by using the `GetTextOutcomeFromBool` helper we use elsewhere. Which would just say "PRs are not required" (or something similar)

checks/evaluation/branch_protection.go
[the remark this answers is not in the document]
spencerschrock: Yes, but the context of the function has changed. Previously it looked at things tied to the basic tier. Now if you look at the body of `basicNonAdminProtection`, nothing there has to do with the basic tier. 

My comment was mainly around whether there's a better name for the helper function now.

checks/evaluation/branch_protection.go
```diff
@@ -60,41 +71,120 @@ const (
...
+		protected, err := isBranchProtected(findings, branchName)
+		if err != nil {
+			e := sce.WithMessage(sce.ErrScorecardInternal, err.Error())
+			return checker.CreateRuntimeErrorResult(name, e)
+		}
+
+		if !protected && !contains(warnedBranches, branchName) {
 			dl.Warn(&checker.LogMessage{
-				Text: fmt.Sprintf("branch protection not enabled for branch '%s'", *b.Name),
+				Text: fmt.Sprintf("branch protection not enabled for branch '%s'", branchName),
```
spencerschrock, line 124: question: now that protection status is its own probe, should this logic just be rolled into to the `switch f.Probe` below?  

1. as written it's quadratic on number of branch names (for each finding, search through all findings for protected probe)
2. We probably don't need to maintain `warnedBranches` anymore.

spencerschrock (commented):
Couple of things:
1. The protected calculation and "Value" still exists in most probes
2. Some of the e2e tests are broken. It's a long story, but they're not run as part of the PR CI due to token requirements. So for anything Branch Protection related you should run `make e2e-pat`.
```
Summarizing 2 Failures:
  [FAIL] E2E TEST PAT:Branch-Protection E2E TEST:Validating branch protection [It] Should fail to return branch protection on other repositories patch
  ~/scorecard.git/e2e/branch_protection_test.go:115
  [FAIL] E2E TEST PAT:Branch-Protection E2E TEST:Validating branch protection [It] Should fail to return branch protection on other repositories
  ~/scorecard.git/e2e/branch_protection_test.go:87

Ran 61 of 73 Specs in 12.012 seconds
FAIL! -- 59 Passed | 2 Failed | 0 Pending | 12 Skipped
```

If you want to just run the branch protection e2e tests for a faster feedback cycle:
```console
TOKEN_TYPE=PAT ginkgo -v --focus-file branch_protection_test.go e2e
```

probes/requiresPRsToChangeCode/impl.go
diogoteles08: I personally care about this clarification, because I AFAIK this is the only case where we give a negative response when the project might even be doing everything right (for other cases we make it clear we don't have data and we don't penalize them). The documentation doesn't make it clear either, so that can bring frustration to maintainers as they might not understand what they're doing wrong.

Maybe a alternative would be to review [Branch protection documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection) to make that explicit. I.e., at the line we mention the rule of Require PRs, we add a note saying something like "when using GitHub traditional Branch Protections without admin permissions, we can't differentiate the case of zero reviewer from not requiring PRs. That's due to limitations on GitHub API and causes a false negative when requiring PRs without requiring reviewers."

probes/requiresCodeOwnersReview/impl.go
```diff
@@ -37,8 +37,35 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
...
 	for i := range r.Branches {
 		branch := &r.Branches[i]
+
+		protected := !(branch.Protected != nil && !*branch.Protected)
+		var protectedValue int
+		if protected {
+			protectedValue = 1
+		} else {
+			protectedValue = 0
+		}
```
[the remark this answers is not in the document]
AdamKorcz: removed in https://github.com/ossf/scorecard/pull/3759/commits/46449531ce9136fb80c9c488e6a5039e6e5f59df

probes/branchesAreProtected/def.yml
```diff
@@ -0,0 +1,27 @@
...
+  Branches that are not protected may allow excessive actions that could compromise the projects security. 
+implementation: >
+  Checks the protection rules of default and release branches.
+outcome:
+  - The probe returns one OutcomePositive for each branch that is protected, and one OutcomeNegative for branches that are not protected. Scorecard only considers default and releases branches.
+remediation:
+  effort: Low
+  text:
+    - For Gitlab-hosted project, follow [the documentation on how to protect branches](https://docs.gitlab.com/ee/user/project/protected_branches.html)
+    - For GitHub-hosted projects, follow [the documentation on protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches)
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/ossf/scorecard/pull/3759/commits/cff2845bc1a712b1febac0555ebaf5a15fc6076a

checks/evaluation/branch_protection.go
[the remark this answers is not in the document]
AdamKorcz: Changed the name in https://github.com/ossf/scorecard/pull/3759/commits/2a8adff14dd8b006749cde686ab56141271b3ef4

AdamKorcz:
> The protected calculation and "Value" still exists in most probes

Removed in https://github.com/ossf/scorecard/pull/3759/commits/4ca56de2b7c445fa8501e7c3dc7104bae55eef12

probes/requiresCodeOwnersReview/impl.go
spencerschrock: Mentioned this to Laurent offline, and he agreed with `OutcomeNegative`. We lose the info statement, but it simplifies remediation later.

AdamKorcz: https://github.com/ossf/scorecard/pull/3759/commits/db37bc1aee27aca1b0aa8ab59e82f73e1970b174

probes/requiresCodeOwnersReview/impl_test.go
```diff
@@ -62,7 +62,7 @@ func Test_Run(t *testing.T) {
 			},
 		},
 		{
-			name: "1 branch requires code owner reviews without files = 1 negative outcome",
+			name: "1 branch requires code owner reviews without files = 1 positive outcome",
```
spencerschrock, line 65: this should be reverted back to negative due to the codeowner probe change

checks/branch_protection_test.go
```diff
@@ -174,7 +174,7 @@ func TestReleaseAndDevBranchProtected(t *testing.T) {
 				Error:         nil,
 				Score:         4,
 				NumberOfWarn:  9,
-				NumberOfInfo:  12,
+				NumberOfInfo:  11,
```
spencerschrock, line 177: this change is expected, just posting for my own record, the decrease is from codeowners being enabled, but no files. Previously this situation would produce 1 warn + 1 log. now it's just 1 warn.

checks/evaluation/branch_protection_test.go
```diff
@@ -18,563 +18,1691 @@ import (
...
-				Protected: &trueVal,
-				BranchProtectionRule: clients.BranchProtectionRule{
-					AllowDeletions:          &falseVal,
-					AllowForcePushes:        &falseVal,
-					RequireLinearHistory:    &falseVal,
-					EnforceAdmins:           &falseVal,
-					RequireLastPushApproval: &falseVal,
-					RequiredPullRequestReviews: clients.PullRequestReviewRule{
-						Required: &falseVal,
+			name: "Required status check enabled",
```
spencerschrock, line 34: the name of this test is "Required status check enabled", but the test result is an error / inconclusive score due to the missing branch names. Something seems like a mismatch here.

checks/evaluation/branch_protection_test.go
```diff
@@ -18,563 +18,1691 @@ import (
...
+					Values: map[string]int{
+						"main": 1,
+					},
+				},
+				{
+					Probe:   "blocksForcePushOnBranches",
+					Outcome: finding.OutcomePositive,
+					Values: map[string]int{
+						"main": 1,
+					},
```
spencerschrock, line 135: optional and/or followup ( I can pick this up if you want):

can we just create a pre-defined var (for the probes that don't have other values, like the number of reviewers) so we can cut down on the lines and increase readability?

```
mainBranchValues := map[string]int{
	"main": 1,
},
```

And then the tests can just reference: 
```
Probe:   "blocksForcePushOnBranches",
Outcome: finding.OutcomePositive,
Values: mainBranchValues
```

Or even just a helper func to make a finding based on name + outcome?
```go
func testFinding(probe string, outcome finding.Outcome) finding.Outcome {
	return finding.Finding{
		Probe: probe,
		Outcome: outcome,
		Values: map[string]int{
			"main": 1,
		},
	}
}
```

And then we cut down on the toil/lines?
```go
findings: []finding.Finding{
	testFinding("blocksDeleteOnBranches", finding.OutcomePositive),
	testFinding("blocksForcePushOnBranches", finding.OutcomePositive),
	// and so on
```

checks/evaluation/branch_protection_test.go
```diff
@@ -18,563 +18,1691 @@ import (
...
+					Values: map[string]int{
+						"main": 1,
 					},
 				},
 			},
+			result: scut.TestReturn{
+				Score:        4,
+				NumberOfWarn: 1,
+				NumberOfInfo: 9,
+			},
```
spencerschrock, line 560: 1 less warn is due to codeowners file change

checks/evaluation/branch_protection_test.go
```diff
@@ -18,563 +18,1691 @@ import (
...
+					Values: map[string]int{
+						"main": 1,
 					},
 				},
 			},
+			result: scut.TestReturn{
+				Score:        8,
+				NumberOfWarn: 1,
+				NumberOfInfo: 9,
+			},
```
spencerschrock, line 1266: Note (can close): similar decrease of 1 warn due to codeowner file missing

checks/evaluation/branch_protection_test.go
```diff
@@ -18,563 +18,1691 @@ import (
...
-						RequiresStatusChecks: &falseVal,
-						UpToDateBeforeMerge:  &trueVal,
-						Contexts:             []string{"foo"},
-					},
-					RequiredPullRequestReviews: clients.PullRequestReviewRule{
-						Required:                     &trueVal,
-						DismissStaleReviews:          &trueVal,
-						RequireCodeOwnerReviews:      &trueVal,
-						RequiredApprovingReviewCount: &oneVal,
+			findings: []finding.Finding{
```
spencerschrock, line 1358: this one doesn't really make sense anymore at this level. Given the codeowner change, we can only express positive/negative. Delete this one maybe?

checks/evaluation/branch_protection_test.go
```diff
@@ -18,563 +18,1691 @@ import (
...
+					Values: map[string]int{
+						"main2": 1,
+					},
+				},
+			},
+			result: scut.TestReturn{
+				Score:        8,
+				NumberOfWarn: 3,
+				NumberOfInfo: 7,
+			},
```
spencerschrock, line 1608: if 1 branch is protected and 1 isn't, I would expect the score to be < 5?

checks/evaluation/branch_protection_test.go
```diff
@@ -18,563 +18,1691 @@ import (
...
+					Outcome: finding.OutcomePositive,
+					Values: map[string]int{
+						"main": 1,
+					},
+				},
+			},
+			result: scut.TestReturn{
+				Score:        5,
+				NumberOfWarn: 1,
+			},
```
spencerschrock, line 1695: similarly, if the branch is not protected, this would be a 0?

spencerschrock (approved):
Overall looks good. Left a few comments, minor cleanup, notes, etc.

 Please make sure to fix:

1.  the e2e tests (see other comment above)
2. The two unit tests that have to do with unprotected branches with weird scores

checks/evaluation/branch_protection_test.go
AdamKorcz: The `branchesAreProtected` probe covers these lines:

https://github.com/ossf/scorecard/blob/b03bd230e2e96cfc4ab3b4af1952d80a35865bc2/checks/evaluation/branch_protection.go#L73-L81

... which affect logging and not the score.

checks/evaluation/branch_protection_test.go
AdamKorcz: Same as https://github.com/ossf/scorecard/pull/3759/files/00e20ba47fd1d1825b9204d898fc10c4c960cf56#r1478152843

checks/evaluation/branch_protection_test.go
AdamKorcz: https://github.com/ossf/scorecard/pull/3759/commits/ef82d1225b946ef6a6f3f6f759043b399b371e10

checks/evaluation/branch_protection_test.go
AdamKorcz: Thanks! Renamed in https://github.com/ossf/scorecard/pull/3759/commits/ca2eae69d009ceef6c3e1f063d508e8059bb4bec

checks/evaluation/branch_protection_test.go
AdamKorcz: > ( I can pick this up if you want)

Thanks, that would be great.

checks/evaluation/branch_protection.go
AdamKorcz: I added a loop specifically for this instead of making it part of the main loop.

checks/evaluation/branch_protection_test.go
spencerschrock: Yes, but I think the findings here are an impossible combination of findings. If a branch isn't protected, the rest of the raw data would be empty.

spencerschrock: If you run it on an empty repo for example, you get the expected result: 
```console
$ go run main.go --repo https://github.com/spencerschrock/actions-test --checks Branch-Protection --show-details --format json --verbosity debug | jq
...
  "score": 0.0,
  "checks": [
    {
      "details": [
        "Warn: branch protection not enabled for branch 'main'"
      ],
      "score": 0,
      "reason": "branch protection not enabled on development/release branches",
      "name": "Branch-Protection",
...
```

Note: the probes are still there for all the branches. The probes are mainly `OutcomeNotAvailable` (two are `OutcomeNegative`), but because the branch isn't protected, it's not logged.

AdamKorcz:
@spencerschrock PTAL again.

spencerschrock:
> @spencerschrock PTAL again.

I'm seeing weird reversions of commits from earlier discussions. For example:
* 75c18eba30ac3a2f69711afee0834760e17e9afc was reverted by 00acf66ea6e9416d453b0990cfe1d5f018d076f0
* 5c9838896ee179142eeef36ed02b34caa317f359 was reverted by e5b16662ada3a647b3e7aab7a2a9f6aa9524a891

And one e2e test is still failing
```shell
TOKEN_TYPE=PAT ginkgo --focus-file branch_protection_test.go e2e

[FAILED] branch protection accessible patch: (-expected +actual)  utests.TestReturn{
        Error:         nil,
        Score:         1,
  -     NumberOfWarn:  3,
  +     NumberOfWarn:  6,
        NumberOfInfo:  4,
  -     NumberOfDebug: 4,
  +     NumberOfDebug: 8,
    }
```

checks/evaluation/branch_protection_test.go
spencerschrock: #3904 3904

checks/evaluation/branch_protection_test.go
spencerschrock: a4896eb9988ae94ea4146ed902344462b0c59350

raghavkaul (approved):
[93fbbc1](https://github.com/ossf/scorecard/pull/3759/commits/93fbbc1a0c693643016fcd4cb619525374e49387)-[f706b69](https://github.com/ossf/scorecard/pull/3759/commits/f706b693f7fca41b5d4c3f086b04480e6fa87654) LGTM.

PR #3770, opened 2024-01-04
:bug: Fix OSV URI in probe remediation text

### LINKED ISSUE #3609: Fix URI in OSVVulnerability probe
See https://github.com/ossf/scorecard/pull/3487#discussion_r1372117029

spencerschrock:

bug fix

The OSV IDs are joined together when generating a link, leading to a broken link:
For example, if a vuln is known by both `foo` and `bar`, the link was `https://osv.dev/bar,foo` which is broken, but `https://osv.dev/bar` or `https://osv.dev/foo` would have worked.

The first Vuln ID is used for the link, from the example above that leads to `https://osv.dev/bar`
Also added a space after the link, so the period wouldn't be interpreted as part of the link in the remediation text.

Fixes #3609

the `release-note`

```release-note
NONE
```

### DIFF
--- a/probes/hasOSVVulnerabilities/impl.go
+++ b/probes/hasOSVVulnerabilities/impl.go
@@ -62,6 +62,9 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
 	IDs := grouper.Group(aliasVulnerabilities)
 
 	for _, vuln := range IDs {
+		if len(vuln.IDs) == 0 {
+			return nil, Probe, errNoVulnID
+		}
 		f, err := finding.NewWith(fs, Probe,
 			"Project contains OSV vulnerabilities", nil,
 			finding.OutcomeNegative)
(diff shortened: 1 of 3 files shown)


probes/hasOSVVulnerabilities/impl.go
```diff
@@ -65,10 +68,12 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
 		if err != nil {
 			return nil, Probe, fmt.Errorf("create finding: %w", err)
 		}
-		f = f.WithMessage(fmt.Sprintf("Project is vulnerable to: %s",
-			strings.Join(vuln.IDs, " / ")))
+		f = f.WithMessage("Project is vulnerable to: " + strings.Join(vuln.IDs, " / "))
+		if len(vuln.IDs) == 0 {
+			return nil, Probe, errNoVulnID
+		}
```
pnacht, line 74:
```suggestion
		if len(vuln.IDs) == 0 {
			return nil, Probe, errNoVulnID
		}
		f = f.WithMessage("Project is vulnerable to: " + strings.Join(vuln.IDs, " / "))
```

Since `f` won't be used if we enter the `if` block.

spencerschrock: Done in 60c806e3deeada7b970701dffc6cc8828b6bb955, except I moved it slightly higher so we don't create `f` at all.

naveensrinivasan (approved):

Issue #3335, opened 2023-07-31
BUG: Pinned-Dependencies fails to handle Dockerfiles with here-docs


pnacht:
**Describe the bug**
Dockerfiles with here-docs crash the Pinned-Deps check

**Reproduction steps**
Run `scorecard --checks Pinned-Dependencies --repo moby/moby` (also works with `--repo distribution/distribution`)

```
Error: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: Dockerfile:1:1: unclosed here-document 'EOT'
2023/07/31 16:52:39 error during command execution: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: Dockerfile:1:1: unclosed here-document 'EOT'
```

`moby/moby/Dockerfile` and `distribution/distribution/dockerfiles/vendor.Dockerfile` have `RUN` commands such as

```shell
RUN --mount=target=/context \
    --mount=target=.,type=tmpfs  \
    --mount=target=/go/pkg/mod,type=cache <<EOT
  set -e
  rsync -a /context/. .
  go mod tidy
  go mod vendor
  mkdir /out
  cp -r go.mod go.sum vendor /out
EOT

```

**Expected behavior**
The Dockerfile should be correctly parsed.

jkreileder:
https://github.com/ossf/scorecard/pull/3774 fixes the problem for moby/moby:

```console
$ go run main.go --checks Pinned-Dependencies --repo moby/moby
Starting [Pinned-Dependencies]
Finished [Pinned-Dependencies]

RESULTS
-------
Aggregate score: 0.0 / 10

Check scores:
|--------|---------------------|--------------------------------|--------------------------------------------------------------------------------|
| SCORE  |        NAME         |             REASON             |                           DOCUMENTATION/REMEDIATION                            |
|--------|---------------------|--------------------------------|--------------------------------------------------------------------------------|
| 0 / 10 | Pinned-Dependencies | dependency not pinned by hash  | https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies |
|        |                     | detected -- score normalized   |                                                                                |
|        |                     | to 0                           |                                                                                |
|--------|---------------------|--------------------------------|--------------------------------------------------------------------------------|
```

And it also fixes the problem for me:

```console
$ scorecard --checks Pinned-Dependencies --repo=jkreileder/cf-ips-to-hcloud-fw
[...]
[...]

PR #3774, opened 2024-01-06
:bug: Refactor Dockerfile validation code to handle here-documents

jkreileder:

bug fix

The Pinned-Dependencies check fails on Dockerfiles that use [here-documents](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#here-documents).

```Dockerfile
RUN <<EOT1 <<EOT2
	pip install --no-deps -e git+https://github.com/username/repo.git@v1.0#egg=package
EOT1
	curl bla | bash
EOT2
```
=>
```
Error: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: Dockerfile:1:1: unclosed here-document 'EOF'
2024/01/06 17:55:47 error during command execution: check runtime error: Pinned-Dependencies: internal error: error parsing shell code: Dockerfile:1:1: unclosed here-document 'EOF'
```

Dockerfiles containing here-documents don't cause failures and the pin-status of the dependencies is correctly extracted.

Fixes https://github.com/ossf/scorecard/issues/3335

```release-note
Shell commands in Dockerfile here-documents are now parsed correctly by the Pinned-Dependencies check
```


checks/raw/pinned_dependencies.go
```diff
@@ -172,21 +171,33 @@ var validateDockerfileInsecureDownloads fileparser.DoWhileTrueOnFileContent = fu
...
+		} else {
+			var valueList []string
+			for n := child.Next; n != nil; n = n.Next {
+				valueList = append(valueList, n.Value)
+			}
 
-		if len(valueList) == 0 {
-			return false, sce.WithMessage(sce.ErrScorecardInternal, errInternalInvalidDockerFile.Error())
-		}
+			if len(valueList) == 0 {
```
jkreileder, line 191: This code isn't new but a check would be easy for this one: `RUN []`

jkreileder:
The other two missed branches (1 new, 1 old) are actually hard to trigger - at least I couldn't come up with something that triggers a parser error.

In the old code, the here-docs triggered that case (and raised a sh level error).  But that doesn't occur anymore with the fix.

checks/raw/pinned_dependencies_test.go
```diff
@@ -843,6 +875,149 @@ func TestDockerfileInsecureDownloadsLineNumber(t *testing.T) {
...
+			startLine uint
+			endLine   uint
+			pinned    bool
+			t         checker.DependencyUseType
+		}
+	}{
+		{
+			name:             "dockerfile heredoc downloads",
+			filename:         "./testdata/Dockerfile-download-heredoc",
+			processingErrors: 1,
```
spencerschrock, line 896: Is the python script necessary? I was a little confused at what the processing error was.
```
RUN <<EOT
#!/usr/bin/env python
print("hello world")
EOT
```

spencerschrock: If this is something only the heredoc would support, I can see why it would be here, but at the very least can we add a comment that the processing error is for the python script?

spencerschrock (approved):
Thanks for the contribution and the tests. Left a small comment about one of them

checks/raw/pinned_dependencies_test.go
jkreileder: The main point of the python test case is that she-bangs are allowed in here-docs (to specify another interpreter than a shell) and shouldn't cause a fatal error.

The mvdan/sh error is indeed slightly strange (the `foo` comes from [here](https://github.com/mvdan/sh/blob/v3.7.0/syntax/parser.go#L1809)).

I'll add a comment to the test case.

PR #3791, opened 2024-01-13
:seedling: Some more cleanup for field placement

naveensrinivasan:
- Cleanup for fieldplacement in struct

the `release-note`

```release-note

```

### DIFF
--- a/checker/check_result.go
+++ b/checker/check_result.go
@@ -56,17 +56,14 @@ const (
 var errSuccessTotal = errors.New("unexpected number of success is higher than total")
 
 // CheckResult captures result from a check run.
-//
-//nolint:govet
 type CheckResult struct {
-	Name    string
-	Version int
 	Error   error
-	Score   int
+	Name    string
 	Reason  string
 	Details []CheckDetail
-	// Structured results.
-	Rules []string // TODO(X): add support.
+	Rules   []string
+	Version int
+	Score   int
 }
 
 // CheckDetail contains information for each detail.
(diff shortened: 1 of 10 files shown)


checker/check_result.go
```diff
@@ -56,17 +56,14 @@ const (
 var errSuccessTotal = errors.New("unexpected number of success is higher than total")
 
 // CheckResult captures result from a check run.
-//
-//nolint:govet
 type CheckResult struct {
-	Name    string
-	Version int
```
spencerschrock, line 63: similar comment on human-readability. We should decide if we care about this

checker/raw_result.go
```diff
@@ -25,29 +25,27 @@ import (
...
-//nolint:govet
 type RawResults struct {
-	BinaryArtifactResults       BinaryArtifactData
+	Metadata                    MetadataData
 	BranchProtectionResults     BranchProtectionsData
-	CIIBestPracticesResults     CIIBestPracticesData
-	CITestResults               CITestData
-	CodeReviewResults           CodeReviewData
+	PinningDependenciesResults  PinningDependenciesData
+	SecurityPolicyResults       SecurityPolicyData
```
spencerschrock, line 32: So I know this was alphabetized recently to find the data easier.
https://github.com/ossf/scorecard/pull/3540

naveensrinivasan: https://github.com/ossf/scorecard/pull/3791#discussion_r1456296660

checker/check_result.go
spencerschrock: > So it helps to have performance IMO. Thoughts?

My first thought is, does it matter? I know we have the linter (and could disable it if the answer is no), but is this a performance bottleneck? Compared to the I/O and API operations we need to wait for anyway?

naveensrinivasan: The compiler utilizes the L1 cache based on the word size.

https://articles.wesionary.team/simple-data-alignment-technique-to-speed-up-your-struct-in-golang-954c43faa1fe
https://www.ardanlabs.com/blog/2013/07/understanding-type-in-go.html

It helps to make it better. 

> I understand API Operations.

PR #3799, opened 2024-01-16
:seedling: Fixed field alignment

naveensrinivasan:
- Fixed field alignment

the `release-note`

```release-note

```

### DIFF
--- a/finding/finding.go
+++ b/finding/finding.go
@@ -45,14 +45,12 @@ const (
 )
 
 // Location represents the location of a finding.
-//
-//nolint:govet
 type Location struct {
-	Type      FileType `json:"type"`
-	Path      string   `json:"path"`
 	LineStart *uint    `json:"lineStart,omitempty"`
 	LineEnd   *uint    `json:"lineEnd,omitempty"`
 	Snippet   *string  `json:"snippet,omitempty"`
+	Path      string   `json:"path"`
+	Type      FileType `json:"type"`
 }
 
 // Outcome is the result of a finding.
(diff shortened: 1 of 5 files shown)


finding/finding.go
```diff
@@ -96,23 +94,19 @@ const (
...
 type Finding struct {
-	Probe       string             `json:"probe"`
-	Outcome     Outcome            `json:"outcome"`
-	Message     string             `json:"message"`
 	Location    *Location          `json:"location,omitempty"`
 	Remediation *probe.Remediation `json:"remediation,omitempty"`
 	Values      map[string]int     `json:"values,omitempty"`
+	Probe       string             `json:"probe"`
+	Message     string             `json:"message"`
+	Outcome     Outcome            `json:"outcome"`
```
spencerschrock, line 103: So one thing about the non-test structs is these values affect the JSON marshaling order. So these `Finding` and `Location` structs may have been intentionally ordered the way they were. @laurentsimon ?

rule/rule.go
```diff
@@ -81,14 +80,13 @@ const (
 	RiskCritical
 )
 
-//nolint:govet
 type Rule struct {
+	Remediation *Remediation
```
spencerschrock, line 84: possibly same intentional alignment question

finding/finding.go
laurentsimon: I did not order them in particular order at the time. Do we need to?

spencerschrock: It's all preference and readability I think. e.g. if we prefer, a slightly more user-friendly order:
```
    {
      "probe": "toolDependabotInstalled",
      "outcome": 12,
      "message": "tool 'Dependabot' is used",
      "location": {
        "type": 1,
        "path": ".github/dependabot.yml",
        "lineStart": 1
      }
    },
```
or
```
    {
      "location": {
        "type": 1,
        "path": ".github/dependabot.yml",
        "lineStart": 1
      }
      "probe": "toolDependabotInstalled",
      "message": "tool 'Dependabot' is used",
      "outcome": 12,
    },
```

naveensrinivasan: Systems will use these after someone consumes the data.

spencerschrock: I'm fine continuing with this change if we want, since this is an unreleased feature.

spencerschrock (approved):
We can always revisit in the future if we decide to change the readability approach discussed here. https://github.com/ossf/scorecard/pull/3799#discussion_r1453688928

PR #3803, opened 2024-01-17
:bug: Handle osvscanner errors on projects with no dependencies

### LINKED ISSUE #3802: BUG: Running checks with --local option
**Describe the bug**
The "--local" option does not work.   I must explicitly name very few checks to get it to run, then I get a score of 10 because nothing is checked.

**Reproduction steps**
1. See the attached scorecard2.sh shell script and error log.

**Expected behavior**
Expecting "--local" to automatically exclude checks that are GitHub specific.    Expecting scorecard to run on an extracted source code tarball like those referenced in the Yocto Project which builds embedded Linux.  

**Additional context**
[scorecard2.sh.txt](https://github.com/ossf/scorecard/files/13966383/scorecard2.sh.txt)
[scorecard3.sh.txt](https://github.com/ossf/scorecard/files/13966445/scorecard3.sh.txt)
[hello-tgz-scorecard.txt](https://github.com/ossf/scorecard/files/13966447/hello-tgz-scorecard.txt)
[hello-github-scorecard.txt](https://github.com/ossf/scorecard/files/13966452/hello-github-scorecard.txt)

spencerschrock:

bug fix

`osvscanner.DoScan` throws an error when no packages are found. This bubbles up to the `Vulnerabilities` check as a runtime error.

We ignore `osvscanner.NoPackagesFoundErr`.

Fixes #3802

the `release-note`

```release-note
Projects without dependencies or packages no longer throw an error for the Vulnerabilities check.
```


clients/osv.go
```diff
@@ -56,7 +56,9 @@ func (v osvClient) ListUnfixedVulnerabilities(
 
 	response := VulnerabilitiesResponse{}
 
-	if err == nil { // No vulns found
+	// either no vulns found, or no packages detected by osvscanner, which likely means no vulns
+	// while there could still be vulns, not detecting any packages shouldn't be a runtime error.
+	if err == nil || errors.Is(err, osvscanner.NoPackagesFoundErr) {
```
spencerschrock, line 61: @another-rex is there anything big I'm missing with this assumption?

As written, this change would give empty repos a 10/10 for Vulnerabilities, which would be reasonable. The other option would be returning an inconclusive score (`-1`).

laurentsimon (approved):

clients/osv.go
spencerschrock: Chatted with Laurent, this is already something that could occur, so we are going to merge as written.

If osv-scanner detects 1 vuln-free package, and fails to detect other packages we wouldn't know.

another-rex: No, this assumption is correct!

Issue #3824, opened 2024-01-27
Avoiding probe explosion and boilerplate


spencerschrock:
## Background
When evaluating a repository, it’s common for Scorecard to detect different tools when scoring checks or probes.  The heuristic has to do with the usage of tools from certain categories, not the usage of an exact tool. This applies to many checks (and their underlying probes):
* Fuzzing
  * OSS-Fuzz / Clusterfuzz lite
  * Various property-based fuzzers for each of the supported languages: Go, Haskell, JS/TS, Python, C, C++, Swift, Rust, Java
* SAST
  * CodeQL PR stats (this one has some extra logic)
  * Tool config files
    * CodeQL, Pysa, Qodana, Snyk, Sonar
* Dependency-Update-Tool
  * Dependabot, PyUp, Renovate

## Proposal
We frame probes as answers to various questions about a repository, so this proposal boils down to figuring out what the question is:
1. Does the project use a tool from category X?
2. Does the project use tool Y from category X? (asked for each supported tool)

I’m going to argue we should favor the first question. Specifically, there should be a single probe per tool category, not per tool.

### Pros
* Avoids boilerplate
  * Each implementation includes about 250 lines of documentation, code, and tests. Of which very little differs
    * Probe YAML definition (~50 lines)
      * Slight differences in description and remediation
    * Probe implementation and tests (~200 lines)
      * As little as 1 changed difference, significant code duplication
* Less fragile tests
    * UniqueProbesEqual is called by most evaluation code to make sure we have all the probes we expect. So adding a new tool probe requires adding the new probe to all the other tests, or the test will throw errors.
* Better warnings
    * Users don’t need to get warned for not using every tool. Part of this is Scorecard’s implementation/consumption of probes, and could be alleviated by some of the policy work. But for now, users that don’t use 1 fuzzer get warned they don't use every fuzzer.

### Cons
* Remediation becomes less specific. But we can solve it with another layer of indirection.
E.g. check out our [link to some markdown] for instructions for specific tools for your language.
* Certain policies may become more difficult to write? Anyone that cares about a specific SAST/Fuzzing/DUT tool will need to check one more “tool” field.

### Example
Current
```json
{"probe": "fuzzedWithGoNative", "outcome": "Negative", ...}
{"probe": "fuzzedWithRustCargofuzz","outcome": "Negative",...}
...
{"probe": "fuzzedWithPropertyBasedTypescript","outcome": "Positive",...}
[...]

Issue #2928, opened 2023-04-28
Feature: re-visit the need for NotApplicableOutcome type


laurentsimon:

spencerschrock:
From a discussion: Should probes have a numeric value?

e.g number of contributors for the contributor check. or number of minimum reviewers in code review check.

AdamKorcz:
> From a discussion: Should probes have a numeric value?
> 
> e.g number of contributors for the contributor check. or number of minimum reviewers in code review check.

The checks that do proportional scoring will benefit from this, as the corresponding probes will need to return a numeric value. The checks that do proportional scoring are:

1. [CI Tests](https://github.com/ossf/scorecard/blob/c061367a8bbc2950cb7dda2a67b7a5624b8f9a5e/checks/evaluation/ci_tests.go#L76)
2. [Code review](https://github.com/ossf/scorecard/blob/c061367a8bbc2950cb7dda2a67b7a5624b8f9a5e/checks/evaluation/code_review.go#L74)
3. [Contributors](https://github.com/ossf/scorecard/blob/c061367a8bbc2950cb7dda2a67b7a5624b8f9a5e/checks/evaluation/contributors.go#L74)
4. [Maintained](https://github.com/ossf/scorecard/blob/c061367a8bbc2950cb7dda2a67b7a5624b8f9a5e/checks/evaluation/maintained.go#L72)
5. [Pinned Dependencies](https://github.com/ossf/scorecard/blob/c061367a8bbc2950cb7dda2a67b7a5624b8f9a5e/checks/evaluation/pinned_dependencies.go#L147)
6. [Webhooks](https://github.com/ossf/scorecard/blob/c061367a8bbc2950cb7dda2a67b7a5624b8f9a5e/checks/evaluation/webhooks.go#L58)

In most of the cases, the probes can handle this by themselves. For example. they can return the number of findings corresponding to a numeric value, however it does increase complexity in the probes, especially regarding understanding the probe outcomes; Some probes can return the number of findings, some probes can only return the number of findings in certain cases, other probes will not return the number of cases. A numeric field in the `Outcome` would in most cases replace a scenario where the number of finding outcomes is used as the numeric value. As such, which scenario is better: Using the number of findings as the numeric value or using a field in the `Outcome`? 

[...]

PR #3837, opened 2024-02-01
:seedling: Change finding Values to map[string]string

spencerschrock:

follow-up of #3558

The `finding.Finding.Values` maps strings to ints. This leads to problems with probes which want to store non-int values.

The `finding.Finding.Values` maps strings to strings. More things can be represented as a string, including ints (with an extra conversion step).

Related to #3824 (see "Why aren’t things like this already?") and #2928

the `release-note`

```release-note
NONE
```


spencerschrock:
Leaving as a draft while I figure out coordination with #3829

spencerschrock:
/scdiff generate CII-Best-Practices,Maintained,Token-Permissions,Pinned-Dependencies,SAST,Signed-Releases,Branch-Protection

checks/evaluation/maintained.go
```diff
@@ -59,17 +60,22 @@ func Maintained(name string,
...
 			switch f.Probe {
 			case issueActivityByProjectMember.Probe:
-				numberOfIssuesUpdatedWithinThreshold = f.Values[issueActivityByProjectMember.NoOfIssuesValue]
+				numberOfIssuesUpdatedWithinThreshold, err = strconv.Atoi(f.Values[issueActivityByProjectMember.NoOfIssuesKey])
+				if err != nil {
+					return checker.CreateRuntimeErrorResult(name, sce.WithMessage(sce.ErrScorecardInternal, err.Error()))
+				}
 			case hasRecentCommits.Probe:
-				commitsWithinThreshold = f.Values[hasRecentCommits.CommitsValue]
+				commitsWithinThreshold, err = strconv.Atoi(f.Values[hasRecentCommits.CommitsValue])
```
raghavkaul, line 75: For consistency (and since it's being used as a key to a map), should this be something like `hasRecentCommits.NoOfCommitsKey`?

probes/notCreatedRecently/impl.go
```diff
@@ -50,24 +52,20 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
 }
 
 func negativeOutcome() ([]finding.Finding, string, error) {
-	f, err := finding.NewWith(fs, Probe,
-		"Repository was created in last 90 days.", nil,
-		finding.OutcomeNegative)
+	text := fmt.Sprintf("Repository was created in last %d days.", lookBackDays)
+	f, err := finding.NewWith(fs, Probe, text, nil, finding.OutcomeNegative)
 	if err != nil {
```
raghavkaul, line 57: Do we need to create a finding `WithValue` even if it's a negative outcome here? That's how it's done with other probes in this PR, and it may be helpful to display it to the user.

raghavkaul (approved):
Looks much cleaner this way! 

For some of the keys that are used across multiple probes, like ` ReleaseNameKey` or `BranchNameKey`, it would be nice to define them once, but I'm not sure where.

spencerschrock:
> For some of the keys that are used across multiple probes, like  ReleaseNameKey or BranchNameKey, it would be nice to define them once, but I'm not sure where.

My goal was to have everything needed to interpret the probe, defined in the same package. Which leads to a little duplication.

checks/evaluation/maintained.go
spencerschrock: 3986739a733bc84c129704779b949b0f02192eb3

probes/notCreatedRecently/impl.go
spencerschrock: 12199913a8b57303c32072c64434164b9cc5bb2a

PR #3874, opened 2024-02-12
:seedling: Combine SAST probes into single probe

### LINKED ISSUE #3824: Avoiding probe explosion and boilerplate

spencerschrock:

There is one probe per SAST tool

There is only one probe, and a "tool" value which describes the tool.

First step for #3824 

the `release-note`

```release-note
The SAST probes have been combined into a single probe. The various tools are specified by the "tool" value.
```


spencerschrock:
This is just a draft for now, to show the approach, as it's still being discussed in #3824.

@laurentsimon you had left a comment to me privately about how this affects remediation.

> is there a way to keep the remediation tool-specific if we made the proposed changes? It's still useful if we want to be able to expand the remediation section. Also, we'd like to have a way to "auto-remediate" (piece of code that creates a diff, for example). Any thoughts how that'd work?

This is what I had in mind for the remediation aspect.

It includes a link to a doc page:
https://github.com/ossf/scorecard/blob/a8675eb9c44e8b9323a6f8ed1627b4310b72bcfd/docs/checks/sast/README.md

While this is pretty sparse, so was the remediation info beforehand. I can add a few more instructions under each entry to match what was there before:
```diff
- remediation:
-   effort: Medium
-   text:
-     - Follow the steps in https://github.com/facebook/pysa-action
```

```diff
- remediation:
-   effort: Medium
-   text:
-    - Follow the steps in https://github.com/snyk/actions
```

laurentsimon (approved):

PR #3876, opened 2024-02-16
:seedling: Add probe registration mechanism

spencerschrock:

refactor

A list of probes and their mappings (name to runner func, which raw data it needs, ect) are maintained away from the probe code. This makes is easy for people to forget to update 

The probes define it in their init block. Unfortunately, due to the fact that `init` is only executed when something imports it, we need to add the probes in the `entries.go` file anyway.

NONE

the `release-note`

```release-note
NONE
```


spencerschrock:
Leaving as a draft to revisit tomorrow one last time. Not sure if this is noticably better than what we have now.
I think it's slightly more structured and defined closer to the probe which is nice, but setting the raw data is still a little messy

probes/codeReviewOneReviewers/impl.go
```diff
@@ -22,23 +22,35 @@ import (
...
+func init() {
+	err := probes.Register(probes.Probe{
+		Name:            Probe,
+		Implementation:  Run,
+		RequiredRawData: []probes.CheckName{probes.CodeReview},
+	})
+	if err != nil {
+		panic(err)
+	}
+}
```
pnacht, line 38: This block is effectively identical across all probes. Is it possible to refactor this into:

```go
// probes/util.go or internal/probes.go
func ProbeInit(probe probes.Probe, impl func(raw *checker.RawResults), requiredRawData []probes.CheckName) {
    err := probes.Register(probes.Probe{
		Name:            probe,
		Implementation:  impl,
		RequiredRawData: requiredRawData,
	})
	if err != nil {
		panic(err)
	}
}

// $PROBE/impl.go
func init() {
    probes.ProbeInit(Probe, Run, []probes.CheckName{probes.CodeReview})
}
```

spencerschrock: Good suggestion. Done in 2ff7302b691d8964b5b0b6088abbdb707d8823cd.

laurentsimon (approved):
LGTM, thanks!

PR #3877, opened 2024-02-16
:seedling: Combine fuzzing probes

spencerschrock:

refactor

Each fuzzing tool has its own probe

There's only one fuzzing probe, and a "tool" value specifies which tool is used.

Related to #3824 

the `release-note`

```release-note
The Fuzzing probes have been combined into a single probe. The various tools are specified by the "tool" value.
```


spencerschrock:
Still have a couple TODOs, especially around remediation and supported tools so still a draft.

I know there has been discussion about continuous fuzzing in the past being different than the fuzzing libraries. While this PR combines everything into one probe, I don't think this limits the introduction of a continuous fuzzing probe in the future. 

For example: 
`fuzzed` is just a simple "is this fuzzed"
and maybe we add `fuzzedContinuously` later, and OSS-Fuzz would be reported by both.

laurentsimon (approved):

PR #3892, opened 2024-02-20
✨ Dangerous-Workflow: Check for workflow_run trigger

### LINKED ISSUE #3861: Feature: Check for usage of on: workflow_run trigger
GitHub Actions supports `on: workflow_run` as a [workflow trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run). This trigger (and potentially also `on: workflow_call`) allows chaining of workflows in a way that can lead to unintentional execution of a GitHub workflow through a contributor's open pull request.

Consider two workflows:
### workflow-a.yml:
```
on:
  push:
    branches: [ main ]
...
```

### workflow-b.yml:
```
on:
  workflow_run:
    workflows: [workflow-a]
    types: [completed]
    branches: [main]
jobs:
  build:
    steps:
    - name: Upload artifact
    - uses: actions/upload-artifact@v4.3.1
    ...
```

Now consider a pull request by a first-time contributor (or, if the maintainers require workflow approval for first-time contributor, consider a contributor that has submitted a small patch already):
### workflow-a.yml (modified in PR):
```
   on:
     push:
       branches: [ main ]
+++  pull_request:
+++    branches: [ main ]
```

**Mitigating factors**
If `workflow-a.yml` is not a "privileged action" (e.g. pushing or pulling code, modifying artifacts, modifying tags), then the risk from this workflow is low. This is because a workflow triggered with `on: workflow_run` uses the workflow definition on the default branch. We should discuss how to prevent false-positives here.

The Dangerous-Workflows check should flag for uses of the `on: workflow_run` trigger that could allow a pull request to execute privileged actions.

raghavkaul:
Check for privileged usage of `on: workflow_run` trigger in GitHub Workflows.

* **Privileged**: A workflow that uses this trigger contains an action that uses `secrets.GITHUB_TOKEN`

Fixes #3861.

The largest function, `checkUsesGithubToken` checks if `secrets.GITHUB_TOKEN` is passed to another workflow. There are a lot of `nil` checks. Another approach is to look through the GitHub Workflow AST, but I wasn't sure if this was possible with actionlint.


laurentsimon:
Couple of comments:
1. Do we care that a GITHUB_TOKEN is referenced? Actions called in a workflow can reference the token themselves. So I would say it does not matter if the token if referenced in the workflow or not
2. iiuc, the solution to pull_request_target in https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ would be detected as a vuln with the changes, but should not

I think what we're missing is capturing the context https://github.com/ossf/scorecard/issues/3861., which is that one of the workflow referenced in `workflows: [Run Tests, Another]` must be running _not_ on a PR. There is also an assumption that the workflow_run workflow uses on-sanitized inputs from the first workflow: This may be one of the `github.*` variables (already covered by the check), a checkout + run (already covered by the check), an artifact download (not covered by the check yet). So maybe we should start by detecting artifact download + first workflow not running on PR?

something unclear to me: does workflow_run accept a `branch` option? I don't see it documented in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

pnacht:
> something unclear to me: does workflow_run accept a `branch` option? I don't see it documented in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#limiting-your-workflow-to-run-based-on-branches

raghavkaul:
Ah, I assumed `GITHUB_TOKEN` had to be passed explicitly to be available. Yes, it's possible for there to be false positives. Is there another indicator of "sensitive workflow" here? And also, if the triggered workflow is restricted to `branch: [<default>]` as in Pedro's link, then it's not vulnerable to this attack.

>> one of the `github.*` variables (already covered by the check)
We only check for `secrets.GITHUB_TOKEN` being passed, why would passing `github.*` be unsafe?

>> a checkout + run (already covered by the check)
We cover this, but don't explicitly differentiate from composite actions that exist in the same repo. I can modify the code to cover this case though.

>> an artifact download
Isn't artifact upload (publishing) what we care about?

spencerschrock:
> We only check for `secrets.GITHUB_TOKEN` being passed, why would passing `github.*` be unsafe?

`github.token` is another way of accessing `secrets.GITHUB_TOKEN`

spencerschrock:
> Isn't artifact upload (publishing) what we care about?

We care about downloading an artifact produced by the other workflow. 

> And also, if the triggered workflow is restricted to branch: [<default>] as in Pedro's link, then it's not vulnerable to this attack.

See https://github.com/ossf/scorecard/issues/3861#issuecomment-1946278932

laurentsimon:
> We cover this, but don't explicitly differentiate from composite actions that exist in the same repo. I can modify the code to cover this case though.

My suggestion is to keep this PR simple and do this in another PR. There has not been many complains (AFAIK) about false positives introduced by the lack of permission checking. Maybe it's uncommon in practice?

>> Isn't artifact upload (publishing) what we care about?

> We care about downloading an artifact produced by the other workflow.

+1

>> And also, if the triggered workflow is restricted to branch: [] as in Pedro's link, then it's not vulnerable to this attack.

> See https://github.com/ossf/scorecard/issues/3861#issuecomment-1946278932

I thought it _does_ work `As for my previous comment, I did some digging and this does indeed seem to work, even though release.yml sets branches: [main]. This is very weird to me and honestly feels like a bug...`

raghavkaul:
Not stale, waiting for update on #3861.

laurentsimon:
@raghavkaul still interested in getting this PR cross the finish line?

PR #3897, opened 2024-02-22
:seedling: use ValidateTestReturn for Code-Review tests

spencerschrock:

test cleanup, similar to #3872. 
This is the first step before hooking up the Code-Review probes to the evaluation code

we only checked the result score, which produces test failures without clear actionable feedback.

* We use our test helper: ValidateTestReturn, which provides much more actionable feedback when tests fail
* `ctrl.Finish()` is removed, as it's not needed after Go 1.14
* simplified err checking logic
* removed some unneeded lints

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/checks/code_review_test.go
+++ b/checks/code_review_test.go
@@ -24,18 +24,16 @@ import (
 	"github.com/ossf/scorecard/v4/checker"
 	"github.com/ossf/scorecard/v4/clients"
 	mockrepo "github.com/ossf/scorecard/v4/clients/mockclients"
+	sce "github.com/ossf/scorecard/v4/errors"
 	scut "github.com/ossf/scorecard/v4/utests"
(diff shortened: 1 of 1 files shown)


checks/code_review_test.go
```diff
@@ -24,52 +24,31 @@ import (
...
-			expected: checker.CheckResult{
+			expected: scut.TestReturn{
 				Score: -1,
 			},
 		},
 		{
-			name:      "no PR's with error as well as commits",
-			err:       errNew,
-			commiterr: errNew,
-			expected: checker.CheckResult{
```
spencerschrock, line 71: these were combined as they tested the exact same things, but under 3 names

laurentsimon (approved):

Issue #3831, opened 2024-01-30
BUG: Binary-Artifact and Pinned-Dependencies kill Scorecard in a repo with large files


pnacht:
**Describe the bug**
Running Scorecard with Binary-Artifact and/or Pinned-Dependencies on a repo with large files crashes entirely.

**Reproduction steps**
I stumbled on this while trying to run Scorecard on a local clone of a HuggingFace model repository.

```shell
# Requires LFS support; ~6GB download!
$ git clone https://huggingface.co/microsoft/phi-2

# same outcome with each check individually, too
$ scorecard --local ~/phi-2 --checks Binary-Artifacts,Pinned-Dependencies
Starting [Binary-Artifacts]
Starting [Pinned-Dependencies]
signal: killed
```

Deleting the very large files (including the .git folder), the checks pass. (There may be other checks that would also fail, I only tested those that run with `--local`)

**Expected behavior**
The checks should work even with large files.

As described below, Binary-Artifacts doesn't need to load the entire file, and it's unlikely an actual script will ever be big enough to be a problem.

**Additional context**
I believe I understand why these checks are failing: both have at least one function ([BinaryArtifacts](https://github.com/ossf/scorecard/blob/v4.13.1/checks/raw/binary_artifact.go#L52) and [collectShellScriptInsecureDownloads](https://github.com/ossf/scorecard/blob/v4.13.1/checks/raw/pinned_dependencies.go#L73)) that runs `fileparser.OnMatchingFileContentDo` with `Pattern: "*"` (i.e. all files).

As the function name implies, this function sequentially opens and loads all matching files. I assume one of the files was simply too large.

This should be fixable, though:

- `BinaryArtifacts` uses `fileparser.OnMatchingFileContentDo` to call `checkBinaryFileContent`. That loads the file and then uses https://github.com/h2non/filetype to determine the file's type. This can be replaced by:
  - Simply trusting a file's extension, and only checking its contents if it has no extension (since it could be anything)
  - Replacing `fileparser.OnMatchingFileContentDo` with a function that only loads the first [262 bytes h2non/filetype needs](https://github.com/h2non/filetype#:~:text=Only%20first%20262%20bytes%20representing%20the%20max%20file%20header%20is%20required%2C%20so%20you%20can%20just%20pass%20a%20slice)
- Pinned-Deps' `collectShellScriptInsecureDownloads` could be set to only run on files with common script extensions (i.e. `.sh`, `.bash`, `.ps`, and no extension).

[...]

PR #3912, opened 2024-03-04
:warning: Switch RepoClient file access to io.ReadCloser

spencerschrock:

breaking change

Checks which deal with file content must work with the whole file contents. Some usages don't require any data, some require only the first line, others require 1 KiB. This is inefficient, and leads to crashes. (See https://github.com/ossf/scorecard/issues/3831#issuecomment-1969978434)

- Clients will provide an `io.ReadCloser` when accessing a file, so the callers can use as much or as little data as they need.
  - Existing callers of `GetFileContent` will use `io.ReadAll` to get the whole file contents, and then `Close` the Reader.
  - Individual uses can be migrated in follow-up PRs.
- Cleaned up the tests to return `io.ReadClosers`.
  - For tests that had strings, this uses a strings/bytes Reader, and an io.NopCloser
  - For tests that previously opened and read files, the test now just opens the file.

Related to #3831, however follow-up PRs are needed to allow Binary-Artifacts and Pinned-Dependencies to take advantage of the `io.Reader`

https://github.com/ossf/scorecard/issues/3831#issuecomment-1969978434 talks about a non-breaking way of doing this, but with V5 around the corner, breaking changes are fine. 

This is simpler than maintaining two versions of the functions (`[]byte` and `io.Reader` versions)

the `release-note`

```release-note
File access through RepoClient now returns an io.ReadCloser, instead of the full file contents. This is a breaking change
```

### DIFF
--- a/checks/fileparser/listing.go
+++ b/checks/fileparser/listing.go
@@ -17,6 +17,7 @@ package fileparser
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"path"
 	"strings"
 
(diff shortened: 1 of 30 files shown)


spencerschrock:
/scdiff generate Fuzzing,Token-Permissions,Pinned-Dependencies,SAST,Security-Policy,Binary-Artifact,Dangerous-Workflow,Packaging

laurentsimon (approved):

PR #3923, opened 2024-03-05
:bug: Limit Binary Artifact file reads to first 1024 bytes

spencerschrock:

bug fix

The whole file is read, even though we only use 1024 bytes when determining if a file is text or binary

Only up to the first 1024 bytes are read.

Part 1 of 2 to address #3831

the `release-note`

```release-note
NONE
```

### DIFF
--- a/checks/fileparser/listing.go
+++ b/checks/fileparser/listing.go
@@ -64,6 +64,21 @@ type PathMatcher struct {
 	CaseSensitive bool
 }
 
+// DoWhileTrueOnFileReader takes a filepath, its reader and
+// optional variadic args. It returns a boolean indicating whether
+// iterating over next files should continue.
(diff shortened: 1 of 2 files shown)


pnacht:
This is already a big improvement, but why 1024 bytes? From what I understood (described in #3831), h2non/filetype (the package Scorecard uses to determine the filetype) only needs the [first 262 bytes](https://github.com/h2non/filetype#:~:text=Only%20first%20262%20bytes%20representing%20the%20max%20file%20header%20is%20required).

spencerschrock:
> This is already a big improvement, but why 1024 bytes

Getting the file type is just one part, we also check some of the file for false positives. You can read a bit more here: https://github.com/ossf/scorecard/issues/2412.

PR #3932, opened 2024-03-11
:seedling: Cleanup branch protection tests

### LINKED ISSUE #3904: cleanup branch protection tests
From https://github.com/ossf/scorecard/pull/3759#discussion_r1457842572:

can we just create a pre-defined var (for the probes that don't have other values, like the number of reviewers) so we can cut down on the lines and increase readability?

```go
mainBranchValues := map[string]int{
	"main": 1,
},
```

And then the tests can just reference: 
```
Probe:   "blocksForcePushOnBranches",
Outcome: finding.OutcomePositive,
Values: mainBranchValues
```

Or even just a helper func to make a finding based on name + outcome?
```go
func testFinding(probe string, outcome finding.Outcome) finding.Outcome {
	return finding.Finding{
		Probe: probe,
		Outcome: outcome,
		Values: map[string]int{
			"main": 1,
		},
	}
}
```

And then we cut down on the toil/lines?
```go
findings: []finding.Finding{
	testFinding("blocksDeleteOnBranches", finding.OutcomePositive),
	testFinding("blocksForcePushOnBranches", finding.OutcomePositive),
	// and so on
```

spencerschrock:

test refactor

Findings are defined as structs, which can be very verbose:
https://github.com/ossf/scorecard/blob/e1f54831df5522129c61e9faf7b1e5cf6cb50c80/checks/evaluation/branch_protection_test.go#L34-L49

helper functions are used to make it easier to specify the findings:

Fixes #3904

The replacements were done using regex find and replace in vscode, so no copy/paste errors to worry about. Only the `findings` fields were changed, and the tests still pass. 

Find
```regex
\{
.+Probe:.+(".+"),
.+Outcome:.+(finding.Outcome.+),
.+Values:.+map\[string\]string\{
.+"branchName":.+(".+"),
.+\},
.+\},
```
Replace
```
branchFinding($1, $3, $2),
```
so assuming you're happy with the new look, should be an easy review:

the `release-note`

```release-note
NONE
```


naveensrinivasan (approved):
I couldn't verify every change.

PR #3948, opened 2024-03-15
🌱 Add script to set up probe boilerplate

raghavkaul:
Adds script to set create a new directory for a probe and set up *.go files with boilerplate for a probe.


probes/internal/scripts/setup.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/ossf/scorecard/v5/finding"
+)
+
+//go:embed *.yml
+var fs embed.FS
+
+const (
+	Probe = "%s"
+)
+
```
spencerschrock, line 67: There's registration code that can't be reflected here. Not that I'm sure it's something a template could help with, maybe have a placeholder comment describing the need to fill it out?

probes/internal/scripts/setup.go
```diff
@@ -0,0 +1,136 @@
...
+short: A short description of this probe
+motivation: >
+  What is the motivation for this probe?
+implementation: >
+  How does this probe work under-the-hood?
+outcome:
+  -
+remediation:
+  effort: # High, Medium, Low
+  text:
```
spencerschrock, line 123: this is missing `onOutcome:`

probes/internal/scripts/setup.go
```diff
@@ -0,0 +1,136 @@
...
+  - 
+ecosystem:
+  languages:
+  - all
+  clients:
+  - github
+  - gitlab
```
spencerschrock, line 130: should we be pre-filling these in? It may depend on the probe

probes/internal/scripts/setup.go
```diff
@@ -0,0 +1,142 @@
...
+
+const (
+	Probe = "%s"
+)
+
+// If your probe is associated with a Scorecard check, map it like so:
+// and create the entry in internal/probes/probes.go
+// func init() {
+// 	probes.MustRegister(Probe, Run, []probes.CheckName{probes.<ProbeName>})
+// }
```
spencerschrock, line 72: There's also `MustRegisterIndependent` for probes not associated with a check. Registration is needed either way so it works with the `--probe` CLI argument.

Although I think the entry you're referring to would be in `probes/entries.go`, not `internal/probes/probes.go`

spencerschrock (approved):
Please fix the DCO on your most recent commit and address the small registration feedback.

PR #3954, opened 2024-03-19
🌱 Add ProjectPackageClient interface and deps.dev default client

raghavkaul:
This change adds a new client that allows Scorecard to retrieve package versions for a repository. A `ProjectPackageVersion.Version` is a package version published by a repository.

Instead of looking up a repository directly in a package manager, this interface allows Scorecard to handle multiple project package versions, i.e. packages with different names than their source repos, and monorepos that publish multiple packages.

We implement a deps.dev client for this interface using [GetProjectPackageVersions](https://docs.deps.dev/api/v3alpha/#getprojectpackageversions):

> GetProjectPackageVersions returns known mappings between the requested project and package versions. At most 1500 package versions are returned. Mappings which were derived from a [SLSA provenance attestation](https://slsa.dev/spec/v1.0/provenance) are served first.


spencerschrock (commented):
Can we throw everything under an `internal` folder while experimenting, so we can change things later without breaking changes?

maybe `internal/clients` or `internal/clients/depsdev` ? We do this already with nuget client for example, which is marked as internal in `cmd/internal/nuget`

internal/packageclient/depsdev_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("E2E TEST: depsdevclient.GetProjectPackageVersions", func() {
+	var client ProjectPackageClient
+
+	Context("E2E TEST: Confirm ProjectPackageClient works", func() {
+		It("Should receive a non-empty response from deps.dev for existing projects", func() {
```
spencerschrock, line 28: Usually these ginkgo tests have a suite which hooks them up to the e2e tests.

raghavkaul: Moved back to `e2e/` which should force them to run during e2e test make targets.

spencerschrock (approved):

PR #3972, opened 2024-03-25
:seedling: Bump CodeQL Action version to 3.24.10 and remove whitespace

adamdmharvey:

I don't see Dependabot updating the `github/codeql` action in this particular action workflow, though it's updating a) other actions in this workflow and b) `github/codeql` in other actions.

Just a while speculation, but this is the only one with the whitespace between the actual name key and the uses? Just in case, firing this to see if the next Dependabot scan will start patching the codeql action in this file also. (and if so, it may infer a minor text checking bug in the upstream Dependabot algorithm?)

Should allow Dependabot to bump the codeql action in this file.

While here, I also bumped the actions manually to the latest v3.x series (which also includes Node 20, which should help eliminate any actions warnings about Node deprecations).

Note the only other call is this in a different workflow:

https://github.com/ossf/scorecard/blob/e780e089f512f12cd1fc3d090a2424f186ac1a78/.github/workflows/scorecard-analysis.yml#L53-L54

An attempt to help dig more into https://github.com/ossf/scorecard-action/pull/1354.

the `release-note`

```release-note

```

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -73,7 +73,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@cdcdbb579706841c47f7063dda365e292e5cad7a # v1
+      uses: github/codeql-action/init@83a02f7883b12e0e4e1a146174f5e2292a01e601 # v2.16.4
       with:
         languages: ${{ matrix.language }}
         queries: +security-extended
@@ -85,7 +85,7 @@ jobs:
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@cdcdbb579706841c47f7063dda365e292e5cad7a # v1
+      uses: github/codeql-action/autobuild@83a02f7883b12e0e4e1a146174f5e2292a01e601 # v2.16.4
 
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 https://git.io/JvXDl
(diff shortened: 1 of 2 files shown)


adamdmharvey:
Indeed, but notice that PR is updating the codeql action ONLY in the Scorecard Analysis workflow? It's not updating it in the CodeQL Workflow? (if I'm reading it correctly)

adamdmharvey:
What do you think about trying the v3.x series? (only difference being Node 20, which eliminates a GitHub Runner warning about them deprecating Node 16)

spencerschrock:
> What do you think about trying the v3.x series? (only difference being Node 20, which eliminates a GitHub Runner warning about them deprecating Node 16)

Interestingly I see that commit listed as the ref for the "CodeQL Bundle v2.16.5" release
https://github.com/github/codeql-action/releases

Where as if you look at https://github.com/github/codeql-action/tags:

[v3.24.8](https://github.com/github/codeql-action/releases/tag/v3.24.8) uses a different hash: `05963f47d870e2cb19a537396c1f668a348c7d8f`

I'm good with trying `@05963f47d870e2cb19a537396c1f668a348c7d8f # v3.24.8`, just please make sure to update the version in `scorecard-analysis.yml` too

spencerschrock:
here's an example of dependabot picking up the updates properly. so I recommend mirroring what we do here:
https://github.com/ossf/scorecard-action/pull/1355/files

spencerschrock:
@adamdmharvey are you still interested in patching this (based on `v3.24.10` at this point, you can see https://github.com/ossf/scorecard-action/pull/1360)? If not, I would be fine putting it as a good-first-issue for someone else to pickup.

adamdmharvey:
Yep!  I'll take a stab.

spencerschrock (approved):
Thanks!

PR #3979, opened 2024-03-27
:seedling: migrate code review check to probes

spencerschrock:

structured results!

eval code uses raw data

eval code uses the results of the `codeApproved` probe. which I had to make a few minor changes to:
* allowing out of platform review platforms (gerrit, phabricator etc)
* allowing empty reviewer/login names

NONE

I have questions! 

1. `codeApproved` seemed like the better fit, so I just adapted that one. There could arguably be a split for "approvedInPlatform" or "approvedOutOfPlatform" for things like GitHub / GitLab native reviews vs Gerrit / Phabricator etc.
2. The probes take a proportional approach based on `approvedChangesets` and `totalChangesets` keys. We lose some better control over logging, so I'm thinking of switching to the "one finding per PR" approach.

the `release-note`

```release-note
NONE
```


laurentsimon (approved):

spencerschrock:
@laurentsimon any thoughts on the "Special notes for your reviewer"?

laurentsimon:
> 1. codeApproved seemed like the better fit, so I just adapted that one. There could arguably be a split for "approvedInPlatform" or "approvedOutOfPlatform" for things like GitHub / GitLab native reviews vs Gerrit / Phabricator etc.

Iuuc, the advantages of separation would be:
1. Less likely that users ignore the type of review system by mistake when doing verification / applying a policy. The separation would make it more explicit
2. If there are different types of data we need to return based on the platform type, separation would make that cleaner. I'm not sure that will ever be the case? Wdut?

I think I lean towards separation due to (1).

> 2. The probes take a proportional approach based on approvedChangesets and totalChangesets keys. We lose some better control over logging, so I'm thinking of switching to the "one finding per PR" approach.

Do you mean "one finding per changeset"? I think having both probes (proportional and per changeset) is probably what we want. The proportional one is likely going to be what most people will use. Per-changeset probes will be used for manual troubleshooting and for policies where the non-reviewed bot contributions may be deemed acceptable... or something like that. Wdut?

spencerschrock:
> The proportional one is likely going to be what most people will use. Per-changeset probes will be used for manual troubleshooting and for policies where the non-reviewed bot contributions may be deemed acceptable... or something like that. Wdut?

From a scorecard perspective, a per-changeset approach would allow debug messages. I wasn't envisioning having two versions of the probe, but rather trying to decide between the two.

If we split the `codeApproved` we would have to do per-changeset or we wouldn't be able to make a decision about accepting in/out-of-platform for each PR.

raghavkaul (approved):
LGTM. As to the two `errNoReviewer` cases in `probes/codeApproved`, we get `changeset.Author` and `reviewer.Author` from GraphQL, so if the probe doesn't have the raw data, there was some problem fetching it.

spencerschrock:
re: splitting probes. In an offline sync last week, the three of us agreed splitting probably makes sense.

For time reasons before the prerelease/OSS NA, I'll save it as a follow-up PR, so at least the probe is converted

PR #4006, opened 2024-04-05
:warning: Switch `Outcome` type to string

spencerschrock:

structured result cleanup which is technically a breaking change

Outcome is an int based type

Outcome is a string based type, which leads to better output format

Related to https://github.com/ossf/scorecard/issues/2928#issuecomment-2040433089

the `release-note`

```release-note
NONE
```

### DIFF
--- a/finding/finding.go
+++ b/finding/finding.go
@@ -54,43 +54,25 @@ type Location struct {
 }
 
 // Outcome is the result of a finding.
-type Outcome int
+type Outcome string
 
 // TODO(#2928): re-visit the finding definitions.
 const (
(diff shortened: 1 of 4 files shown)


laurentsimon (approved):
You'll do the change from positive/negative -> true/false in another PR I suppose. LGTM

spencerschrock:
> You'll do the change from positive/negative -> true/false in another PR I suppose. LGTM

Yes. And that will also be 2 changes so the important changes don't get lost in a big diff. 

1. A mechanical Positive -> True and Negative -> False change. Plus the necessary log changes.
2. Flipping the negative probes. e.g `notArchived` -> `archived`

PR #4017, opened 2024-04-08
:warning: Replace Positive and Negative outcomes with True and False

### LINKED ISSUE #3866: Feature: Rename OutcomePositive/Negative to something more aligned with the security implication of a probe finding
`OutcomePositive/Negative` are big words for "yes" and "no". There have been various places/discussions requiring added logic to handle the fact that "yes" for one probe means it is safe, but for another probe it means it's dangerous (most recently, https://github.com/ossf/scorecard/pull/3816#discussion_r1483005286).

Replace `OutcomePositive/Negative` (i.e. "yes"/"no") with "safe"/"unsafe": `OutcomeSafe/Unsafe`, `OutcomeGood/Bad`, etc.

I believe this will be more immediately legible: "ah, I don't even know what this probe is looking for, but apparently the project is safe from it!"

It may also simplify evaluation logic, since we don't need to special-case probes where `Positive` is a bad thing: `OutcomeGood` will always increase a project's score, `OutcomeBad` will always decrease it.

Probes could be renamed so that `OutcomePositive` always means it's safe. But this can lead to "negative" probe names (`doesNotDoX`), which doesn't follow the general coding principle that "variables" should be in the affirmative.

spencerschrock:

structured result breaking change

Outcomes use Positive and Negative which is a confusing mix of both safe/unsafe and yes/no

Outcomes use True / False which is a less ambiguous equivalent of yes/no

Fixes #3866 (see https://github.com/ossf/scorecard/issues/3866#issuecomment-2043506236)

the `release-note`

```release-note
NONE
```


spencerschrock:
/scdiff generate Dangerous-Workflow,Fuzzing,SAST,Security-Policy,Packaging,Signed-Releases,Branch-Protection,Token-Permissions,CI-Tests,Code-Review,Contributors,Binary-Artifacts,Maintained,Pinned-Dependencies,CII-Best-Practices,Dependency-Update-Tool,License,Vulnerabilities

spencerschrock:
The current change is a find/replace for Positive -> True and Negative -> False. All of the behavior changes will be in a follow-up PR.

The unit and e2e tests pass, and the scdiff was a match. 


I will highlight a few important places:

The new definitions / documentation for `OutcomeTrue` and `OutcomeFalse`:
https://github.com/ossf/scorecard/blob/17d84c4d7ccb43e7f489d0d35787237f4d55a6e9/finding/finding.go#L60-L70

We currently only surface remediation data when the probe is negative. So this will likely need to go away in a followup.
https://github.com/ossf/scorecard/blob/17d84c4d7ccb43e7f489d0d35787237f4d55a6e9/finding/finding.go#L220-L232

While some of the probes are phrased so `OutcomeNegative` corresponded to a no, `hasDangerousWorkflowUntrustedCheckout` wasn't and will need to be fixed
https://github.com/ossf/scorecard/blob/17d84c4d7ccb43e7f489d0d35787237f4d55a6e9/probes/hasDangerousWorkflowUntrustedCheckout/impl.go#L57-L62

laurentsimon (approved):

PR #4021, opened 2024-04-10
✨ Add check ID

### LINKED ISSUE #2577: Feature: Add a check ID to each check
The current API only has a string available for a consumer to match, e.g.

```json
{
      "name": "Dangerous-Workflow",
      "reason": "dangerous workflow patterns detected",
      "details": [
        "Warn: untrusted code checkout '${{ github.event.pull_request.head.sha }}': .github/workflows/integration.yml:45"
      ],
      "documentation": {
        "short": "Determines if the project's GitHub Action workflows avoid dangerous patterns.",
        "url": "https://github.com/ossf/scorecard/blob/c40859202d739b31fd060ac5b30d17326cd74275/docs/checks.md#dangerous-workflow"
      }
    },
```

A more easily leveraged approach is the implement a check number. This way people can match against an integer (generally less error prone then string matching). They can then also easily do things such as implement a list of check numbers they wish to ignore.

I recommend something like:

```json
{
      "id": 1
      "name": "Binary-Artifacts",
      "score": 10,
      "reason": "no binaries found in the repo",
      "details": null,
      "documentation": {
        "short": "Determines if the project has generated executable (binary) artifacts in the source repository.",
        "url": "https://github.com/ossf/scorecard/blob/c40859202d739b31fd060ac5b30d17326cd74275/docs/checks.md#binary-artifacts"
      }
    },
```

ashearin:

Enhancement to our structured results. Adds a static Check ID for all checks, along with updates to check validation process to make sure the ID is unique. Only impacts Json results. 

I also ran fieldalignment on the files where I made changes to struct members. 

Json output only has name for identifying the check
```json
{
    "details": null,
    "score": 10,
    "reason": "no binaries found in the repo",
    "name": "Binary-Artifacts",
    "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/main/docs/checks.md#binary-artifacts",
        "short": "Determines if the project has generated executable (binary) artifacts in the source repository."
    }
},
```

Adding a uint as a unique ID allows users to filter/search for checks without string comparison. 
```json
{
    "id": 3,
    "details": null,
    "score": 10,
    "reason": "no binaries found in the repo",
    "name": "Binary-Artifacts",
    "documentation": {
        "url": "https://github.com/ossf/scorecard/blob/main/docs/checks.md#binary-artifacts",
        "short": "Determines if the project has generated executable (binary) artifacts in the source repository."
    }
},
```

Fixes #2577 

```release-note
Add Check ID 
```


pkg/json.go
```diff
@@ -30,6 +30,7 @@ import (
 type jsonCheckResult struct {
 	Name       string
 	Details    []string
+	ID         uint
 	Confidence int
 	Pass       bool
 }
```
spencerschrock, line 36: we shouldn't touch the old json format at this point in my opinion

cron/internal/format/json.go
```diff
@@ -29,6 +29,7 @@ import (
 type jsonCheckResult struct {
 	Name       string
 	Details    []string
+	ID         uint
```
spencerschrock, line 32: shouldn't touch the old format

cron/internal/format/json.go
```diff
@@ -46,13 +47,13 @@ type jsonCheckDocumentationV2 struct {
...
-//nolint:govet
 type jsonCheckResultV2 struct {
-	Details []string                 `json:"details"`
-	Score   int                      `json:"score"`
+	Doc     jsonCheckDocumentationV2 `json:"documentation"`
 	Reason  string                   `json:"reason"`
 	Name    string                   `json:"name"`
-	Doc     jsonCheckDocumentationV2 `json:"documentation"`
+	Details []string                 `json:"details"`
+	ID      uint                     `json:"id"`
```
spencerschrock, line 55: this requires coordinating a schema update for the bigquery table. which is more involved than I have time for this week with OSS NA prep.

pkg/json.go
```diff
@@ -77,14 +77,14 @@ func (s jsonFloatScore) MarshalJSON() ([]byte, error) {
 
 // JSONScorecardResultV2 exports results as JSON for new detail format.
 //
-//nolint:govet
+
 type JSONScorecardResultV2 struct {
```
spencerschrock, line 81: note: the blank line breaks the godoc. Please change to either:
```go
// JSONScorecardResultV2 exports results as JSON for new detail format.
//
type JSONScorecardResultV2 struct {
```

but preferably:
```go
// JSONScorecardResultV2 exports results as JSON for new detail format.
type JSONScorecardResultV2 struct {
```

spencerschrock (changes requested):
Declaring the ints in the checks.yaml and validating new duplicates seems reasonable to me. Although I'm curious what would happen if a check were deleted, how would we mark its old number as reserved.

I think this change represents an interesting challenge for the cron, as it's something we could technically go and backfill data for. Can we wait on this until after OSS NA? 

Since probes/structured results haven't been released yet, I had thought this could be an easy change to probes in the meantime.

> I also ran [betteralign](https://github.com/dkorunic/betteralign) on the files where I made changes to struct members. This resulted in changes to structs I did not touch, but I left the changes since they are beneficial changes. Can revert if thats preferred.

Note the linter we use currently is `fieldalignment` (mentioned in the betteralign readme)
```
go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment
fieldalignment -fix
```

pkg/json.go
ashearin: Removed

ashearin:
> I think this change represents an interesting challenge for the cron, as it's something we could technically go and backfill data for. Can we wait on this until after OSS NA?

No rush from me, will do a bit more cleanup and we can look at it after next week

ashearin:
> Note the linter we use currently is `fieldalignment`

Went back and reset the structs and ran `fieldalignment` on them.

spencerschrock:
@ashearin we are planning to merge this at the same time that we flip the switch on maintainer annotations, since both require a schema change to the BigQuery data. So going to do both at the same time.

spencerschrock:
/scdiff generate License,Vulnerabilities,Code-Review,Dangerous-Workflow,SAST,Pinned-Dependencies,Branch-Protection,Signed-Releases,Maintained,Token-Permissions,CII-Best-Practices,Packaging,Binary-Artifacts,Dependency-Update-Tool,Security-Policy,Fuzzing

checker/check_result.go
```diff
@@ -61,15 +61,15 @@ var errSuccessTotal = errors.New("unexpected number of success is higher than to
 //
 //nolint:govet
 type CheckResult struct {
-	Name    string
```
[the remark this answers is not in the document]
spencerschrock: Can you add this patch before merge?
```diff
diff --git a/cmd/internal/scdiff/app/compare/compare.go b/cmd/internal/scdiff/app/compare/compare.go
index 51dfe62f..7155a97c 100644
--- a/cmd/internal/scdiff/app/compare/compare.go
+++ b/cmd/internal/scdiff/app/compare/compare.go
@@ -49,6 +49,9 @@ func compareChecks(r1, r2 *scorecard.Result) bool {
                if r1.Checks[i].Name != r2.Checks[i].Name {
                        return false
                }
+               if r1.Checks[i].ID != r2.Checks[i].ID {
+                       return false
+               }
                if r1.Checks[i].Score != r2.Checks[i].Score {
                        return false
                }
```

Issue #2657, opened 2023-02-13
Fix documentation for structured results


laurentsimon:
This is a meta issue to go over all rule documentations and fix / improve it.

laurentsimon:
test

lolodroid9:
test

lolodroid9:
test

spencerschrock:
The Maintained probes still need some polish 
* https://github.com/ossf/scorecard/pull/3507#discussion_r1396510335
* https://github.com/ossf/scorecard/pull/3507#discussion_r1396510459
* https://github.com/ossf/scorecard/pull/3507#discussion_r1396511302

spencerschrock:
Probe language/ecosystem could use another review. #3797 
For example, free of binary artifact should work on localdir too
https://github.com/ossf/scorecard/blob/6fc7d4c061d2276fdf2427e40cd736e98598f476/probes/freeOfAnyBinaryArtifacts/def.yml#L32-L34

PR #4023, opened 2024-04-11
:book: Review and update some probe documentation

spencerschrock:

documentation

Related to #2657 but we'll need another review after the prerelease feedback.

the `release-note`

```release-note
NONE
```

### DIFF
--- a/probes/hasBinaryArtifacts/def.yml
+++ b/probes/hasBinaryArtifacts/def.yml
@@ -15,12 +15,12 @@
 id: hasBinaryArtifacts
 short: Checks if the project has any binary files in its source tree.
 motivation: >
-  Binary files are not readable so users can't see what they do. Many programming language systems can generate executables from source code (e.g., C/C++ generated machine code, Java .class files, Python .pyc files, and minified JavaScript). Users will often directly use executables if they are included in the source repository, leading to many dangerous behaviors.
+  Binary files are not human readable so users and reviewers can't easily see what they do.
 implementation: >
   The implementation looks for the presence of binary files. This is a more restrictive probe than "hasUnverifiedBinaryArtifacts" which excludes verified binary files.
 outcome:
-  - If the probe finds binary files, it returns a number of true outcomes equal to the number of binary files found. Each outcome includes a location of the file.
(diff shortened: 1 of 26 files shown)


probes/hasBinaryArtifacts/def.yml
```diff
@@ -15,12 +15,12 @@
...
 short: Checks if the project has any binary files in its source tree.
 motivation: >
-  Binary files are not readable so users can't see what they do. Many programming language systems can generate executables from source code (e.g., C/C++ generated machine code, Java .class files, Python .pyc files, and minified JavaScript). Users will often directly use executables if they are included in the source repository, leading to many dangerous behaviors.
+  Binary files are not human readable so users and reviewers can't easily see what they do.
 implementation: >
   The implementation looks for the presence of binary files. This is a more restrictive probe than "hasUnverifiedBinaryArtifacts" which excludes verified binary files.
 outcome:
-  - If the probe finds binary files, it returns a number of true outcomes equal to the number of binary files found. Each outcome includes a location of the file.
-  - If the probe finds no binary files, it returns a single false outcome.
+  - If the probe finds binary files, it returns one OutcomeTrue for each binary file found.
```
laurentsimon, line 22: should `OutcomeTrue` be simply `True`? Same for `OutcomeFalse` should be `False`? The `onOutcome` field is set to `True`, so I think we should re-use the same value?

laurentsimon (approved):
Added one comment. Approving to avoid blocking you for the merge

probes/hasBinaryArtifacts/def.yml
spencerschrock: so the constant is called `finding.OutcomeTrue`, but the value is `True` because otherwise the output would be:

```json
"outcome": "OutcomeTrue"
```

We can address this later if you want, but going to merge for now

laurentsimon: sure. What the constant is called in our code is an implementation detail imo. What users see is the yml and the json probe result, which both say `True`. Hence why I'm proposing using `True` in the yml.

PR #4052, opened 2024-04-24
:sparkles: allow probes to collect their own data from repo clients

### LINKED ISSUE #3235: Discussion: Can probes exist without belonging to a check
The probes (experimental) currently need to belong to a check to be run. What is users need probes that don't fit into a check? Do we create a new check? Do we put all new probes into their own check? In this case, shall we "hide" these checks from the `--checks` option?

I think this will eventually be a case-by-case discussion. But it would be good to have a story. 

I'm currently thinking that if there's not obvious check for a new probe, we could have a fallback "special" check that contains all the probes and not exposed the check within the `--checks` option. This would also allow us to migrate probes to part of a public check once we think that probe(s) should be "promoted" to a public check (based on probe popularity, etc).

spencerschrock:

feature

probes must have the data they rely on in some check's raw result

* introduce "independent" probes which fetch data directly from the repo client
* Allow these probes to run when specified via `--probe`
  * since this is only for probes which don't belong to checks, the logic around running probes from the various `check/*.go` functions doesn't need modified

Fixes #3235

the `release-note`

```release-note
NONE
```

### DIFF
--- a/internal/probes/probes.go
+++ b/internal/probes/probes.go
@@ -48,12 +48,15 @@ const (
 )
 
 type Probe struct {
-	Name            string
-	Implementation  ProbeImpl
-	RequiredRawData []CheckName
+	Name                      string
+	Implementation            ProbeImpl
+	IndependentImplementation IndependentProbeImpl
+	RequiredRawData           []CheckName
 }
 
 type ProbeImpl func(*checker.RawResults) ([]finding.Finding, string, error)
+
+type IndependentProbeImpl func(*checker.CheckRequest) ([]finding.Finding, string, error)
 
 // registered is the mapping of all registered probes.
 var registered = map[string]Probe{}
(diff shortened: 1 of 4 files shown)


spencerschrock:
You can see an example of a toy probe which does its own data collection [here](https://github.com/ossf/scorecard/commit/cc974423bbc203d018e671823cbdf727efb8e8d2).

```go
func init() {
	probes.MustRegisterIndependent(Probe, Run)
}

//go:embed *.yml
var fs embed.FS

const Probe = "foo"

func Run(cr *checker.CheckRequest) ([]finding.Finding, string, error) {
	if cr == nil {
		return nil, "", fmt.Errorf("%w: check request", uerror.ErrNil)
	}

	name, err := cr.RepoClient.GetDefaultBranchName()
	if err != nil {
		return nil, Probe, fmt.Errorf("fetching default branch name: %w", err)
	}

	// ...
}
```

laurentsimon (approved):

PR #4060, opened 2024-04-29
:bug: fix signed-releases lookback limit precedence

### LINKED ISSUE #4059: BUG Signed-Releases: internal error: too many releases, please report this
**Describe the bug**

Running v5.0.0-rc1

Signed-Releases check reports `internal error: too many releases, please report this`

**Reproduction steps**
1. Downloaded scorecard_5.0.0-rc1_linux_amd64.tar.gz
2. Untarred the binary and copied it to ~/.local/bin/scorecard so it's on my path 
3. Ran `scorecard --repo github.com/cpswan/release_automation`

**Expected behavior**

I get a score for my signed releases.

**Additional context**

The repo I was testing against presently has 31 releases. I was previously signing with sigstore, but I just added a workflow to add SLSA provenance instead, so the releases now have a mixture of `.sigstore` and `multiple.intoto.jsonl`. My first guess is that this might cause the problem.

Testing against a repo that only has (some) `.sigstore` releases seems to work fine (e.g. `scorecard --repo github.com/atsign-foundation/noports`) NB that repo has more releases than the one that's failing, so it's not simply that 31 is too many.

Testing against this repo, which has just `multiple.intoto.jsonl` is also fine

spencerschrock:

bug fix

if the 6th release had no assets, the lookback limit exit condition was being skipped. This led to scenarios where too many releases were being considered by the Signed-Releases check.

The lookback exit condition is now the first thing performed.

Fixes #4059

the `release-note`

```release-note
Fixed a Signed-Releases bug where more releases were being analyzed than intended.
```

### DIFF
--- a/probes/releasesAreSigned/impl.go
+++ b/probes/releasesAreSigned/impl.go
@@ -53,12 +53,12 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
 
 	totalReleases := 0
 	for releaseIndex, release := range releases {
-		if len(release.Assets) == 0 {
-			continue
-		}
-
 		if releaseIndex == releaseLookBack {
 			break
+		}
+
+		if len(release.Assets) == 0 {
+			continue
 		}
 
 		totalReleases++
--- a/probes/releasesAreSigned/impl.go
+++ b/probes/releasesAreSigned/impl.go
@@ -53,7 +53,7 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
 
 	totalReleases := 0
 	for releaseIndex, release := range releases {
-		if releaseIndex == releaseLookBack {
+		if releaseIndex >= releaseLookBack {
 			break
 		}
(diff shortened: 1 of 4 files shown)


probes/releasesAreSigned/impl.go
```diff
@@ -53,12 +53,12 @@ func Run(raw *checker.RawResults) ([]finding.Finding, string, error) {
 
 	totalReleases := 0
 	for releaseIndex, release := range releases {
```
raghavkaul, line 55: Maybe neater to use a slice here?
```go
for releaseIndex, release := range releases[:releaseLookBack] {
```

raghavkaul (approved):

probes/releasesAreSigned/impl.go
spencerschrock: that could unfortunately panic
https://go.dev/play/p/8zDplRlC5Yb

PR #4097, opened 2024-05-09
:bug: Support renamed gradle verification action and callers which pin to hash

### LINKED ISSUE #2357: Binary Artifact detection gradle-wrapper.jar Incorrect logic
This is hilarious bug, took me a few tries to get to the exact issue.

1. gradle-wrapper.jar is detected as a binary-artifact
2. suggested recommendation is to do a wrapper validation https://github.com/ossf/scorecard/issues/1815
3. even with wrapper validation script there was error

The issue **gets closed** only when following conditions are met. 
1. `gradle/wrapper-validation-action@v<version_number> is specified`
2. The configuration is in an independent file.

The issue **is not closed** in following conditions.
2. Config is in part of other files
3. `gradle/wrapper-validation-action@<hash> is specified` : Hash pinning which is the recommended action by scorecard and gets raised if my action doesnt have hashes is raised and if we fix that binary artifact issue reappears

I am no golang expert but if i am looking at the right file then there is a very specific `@v` condition check in this code.

https://github.com/ossf/scorecard/blob/main/checks/raw/binary_artifact.go#L36

### LINKED ISSUE #2477: Gradle wrapper action check for binary artifact exception does not recognize actions pinned by hash
**Describe the bug**
Cannot include a gradle wrapper and use the gradle/wrapper-validation-action to exempt the repo from violations if pinning the action by hash (instead of tag)

**Reproduction steps**
1. Add gradle wrapper to repo
2. Add gradle/wrapper-validation-action to ensure validity of the wrapper (but use a hash instead of tag)
3. Allstar hates it

Example: https://github.com/sigstore/protobuf-specs/issues/23#issuecomment-1322837470

**Expected behavior**
Allstar has no complaints.

**Additional context**
The action hash should be converted to a version (or compared against a version)

spencerschrock:
bug fix

`gradle/wrapper-validation-action` must be present and pinned to a hash

* detects `gradle/actions/wrapper-validation`
  * From gradle/wrapper-validation-action's readme: "As of v3 this action has been superceded by gradle/actions/wrapper-validation"
* Also support actions pinned to a hash.

Fixes #2477
Fixes #2357

Related to https://github.com/ossf/scorecard-action/issues/782#issuecomment-2101915990, but requires a release (which was going to be cut today anyway)

the `release-note`

```release-note
Binary-Artifacts check now supports the new gradle wrapper validation action, and can be pinned to a hash.
```

### DIFF
--- a/checks/raw/binary_artifact.go
+++ b/checks/raw/binary_artifact.go
@@ -19,11 +19,9 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"unicode/utf8"
 
-	semver "github.com/Masterminds/semver/v3"
 	"github.com/h2non/filetype"
 	"github.com/h2non/filetype/types"
 	"github.com/rhysd/actionlint"
@@ -35,22 +33,8 @@ import (
 	"github.com/ossf/scorecard/v5/finding"
 )
 
-var (
-	gradleWrapperValidationActionRegex             = regexp.MustCompile(`^gradle\/wrapper-validation-action@v?(.+)$`)
-	gradleWrapperValidationActionVersionConstraint = mustParseConstraint(`>= 1.0.0`)
-)
-
 // how many bytes are considered when determining if a file is text or binary.
 const binaryTestLen = 1024
-
-// mustParseConstraint attempts parse of semver constraint, panics if fail.
-func mustParseConstraint(c string) *semver.Constraints {
-	if c, err := semver.NewConstraint(c); err != nil {
-		panic(fmt.Errorf("failed to parse constraint: %w", err))
-	} else {
-		return c
-	}
-}
 
 // BinaryArtifacts retrieves the raw data for the Binary-Artifacts check.
 func BinaryArtifacts(req *checker.CheckRequest) (checker.BinaryArtifactData, error) {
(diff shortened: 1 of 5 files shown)


raghavkaul (approved):

loosebazooka:
nice

faern:
This fix does not seem to work for us.

We merged the CI workflow for `gradle/actions/wrapper-validation` back on August 29: https://github.com/mullvad/mullvadvpn-app/blob/2b0bd13088a1417ac6510746051043bf81dfc8c7/.github/workflows/android-validate-gradle-wrapper.yml. But the scorecard still warns us about `Warn: binary detected: android/gradle/wrapper/gradle-wrapper.jar:1`, which can be seen here: https://securityscorecards.dev/viewer/?uri=github.com/mullvad/mullvadvpn-app

Are we using the `gradle/actions/wrapper-validation` workflow wrong, or is this check faulty in some way?

faern:
Thanks for the quick response! Maybe our trigger rules are not optimal. I realize someone could potentially commit a malicious `gradle-wrapper.jar` without our CI noticing if they commit it to another path, or push it without a PR (branch protection rules would not allow that, but anyway).

But I also don't want to run it on every single commit (since `gradle-wrapper.jar` virtually never changes, it's a waste of CI minutes).

I'm thinking to change the trigger rule to something like:
```
on:
  push:
    paths:
      - .github/workflows/android-validate-gradle-wrapper.yml
      - '**/gradle-wrapper.jar'
```

This seems like it would be secure. I think `scorecard` should allow this at least. But maybe it should indeed deny our current setup, since it has flaws?

spencerschrock:
> This seems like it would be secure. I think `scorecard` should be made to allow this at least. But maybe it should indeed deny our current setup, since it has flaws. I can create a separate issue for this. But I have not yet verified if `scorecard` would warn on my new `on:`-setup.

Your new trigger sounds good, and following the intention of the Binary-Artifact check, even if Scorecard has trouble giving credit for this due to some implementation limitations. 

We do our analysis on the tarball produced from `git archive`, so we don't have the gradle-wrapper revision easily accessible to check against.

PR #4104, opened 2024-05-13
⚠️ Add ProjectPackageVersions to raw data collection

raghavkaul:
Adds ProjectPackageVersion data to `SignedReleasesData`, which will let a probe consume the data. Also add mock for `ProjectPackageClient`.

⚠️ **breaking change**: implementers of the `Repo` interface must now implement `Path()`.


checker/raw_result.go
```diff
@@ -286,7 +287,8 @@ type BinaryArtifactData struct {
 // SignedReleasesData contains the raw results
 // for the Signed-Releases check.
 type SignedReleasesData struct {
-	Releases []clients.Release
+	Releases               []clients.Release
+	ProjectPackageVersions packageclient.ProjectPackageVersions
```
spencerschrock, line 291: most of the other checks define the subset of data they care about within this `checker` package. Right now `ProjectPackageVersions` re-uses the exact struct from the client, which seems very tied to the deps.dev API response.

Also, stylistically can we just call this `Packages` or something?

clients/repo.go
```diff
@@ -16,6 +16,7 @@ package clients
 
 // Repo interface uniquely identifies a repo.
 type Repo interface {
+	Path() string
```
spencerschrock, line 19: does this need to be its own method or can the caller just call `URI()` and strip the `Host()` prefix?

raghavkaul: I think that's possible, but I'm wary of unexpectedly-shaped enterprise URLs or substring operations being finicky.

spencerschrock: SGTM.

Now that there are multiple, closely related parts, can you add some brief comments in the interface method so the godoc explains the differences?

```go
type Repo interface {
        // URI returns everything ... 
	URI() string
 	// Path returns ...
	Path() string
        // Host returns ...
	Host() string
	String() string
	IsValid() error
	Metadata() []string
	AppendMetadata(metadata ...string)
}
```

spencerschrock (commented):
Can you remind me what the intended purpose is? There's a lot of data being added to the raw results. But hard to reason about the necessity of it when there's no accompanying probe.

If we consider the ossf/scorecard response for example:
https://api.deps.dev/v3/projects/github.com%2Fossf%2Fscorecard:packageversions

raghavkaul:
The probe will be something like `releasesHaveVerifiedProvenance`, which returns a proportional finding depending on how many releases have provenance. I think we'll need `System` (ecosystem), `Name`, and `Version` to create a unique `VersionKey` (as deps.dev does) to return proportional results. Then we'll need `IsVerified` for the actual probe. I left `Commit` in there if we'll eventually want to deduplicate GitHub releases attestations from other system release attestations.

spencerschrock:
Getting segfaults and e2e test failures, which would need to be fixed

raghavkaul:
> how errors in the deps dev response would be handled right now (if it would mess up other data collection)

They're propagated up to the check level, so they'll cause the check to fail.

cron/internal/worker/main.go
```diff
@@ -124,7 +124,7 @@ func newScorecardWorker() (*ScorecardWorker, error) {
 	sw.logger = log.NewCronLogger(log.InfoLevel)
 	sw.githubClient = githubrepo.CreateGithubRepoClient(sw.ctx, sw.logger)
 	// TODO(raghavkaul): Read GitLab auth token from environment
-	if sw.gitlabClient, err = gitlabrepo.CreateGitlabClient(sw.ctx, "https://gitlab.com"); err != nil {
+	if sw.gitlabClient, err = gitlabrepo.CreateGitlabClient(sw.ctx, "gitlab.com"); err != nil {
```
spencerschrock, line 127: unrelated to this (it wouldn't let me comment on the unchanged line).

But in this file, I think you need to create the project client in `newScorecardWorker`

clients/gitlabrepo/client.go
```diff
@@ -283,7 +283,8 @@ func CreateGitlabClient(ctx context.Context, host string) (clients.RepoClient, e
 }
 
 func CreateGitlabClientWithToken(ctx context.Context, token, host string) (clients.RepoClient, error) {
-	client, err := gitlab.NewClient(token, gitlab.WithBaseURL(host))
+	url := "https://" + host
```
spencerschrock, line 286: Can you remind me why this change was needed again? I know you mentioned it, but I've forgotten and having a recorded answer is good. 

Is this just to strip the scheme from the host before sending it to deps dev? If yes, should that be an implementation detail of `func (d depsDevClient) GetProjectPackageVersions` instead?

clients/repo.go
```diff
@@ -16,9 +16,15 @@ package clients
...
+	// Path returns the specifier of the repository within its forge
+	Path() string
+	// URI returns the fully qualified address of the repository
 	URI() string
+	// Host returns the web domain of the repository
 	Host() string
+	// String returns a string representation of the repository URI
 	String() string
+	// IsValid returns whether the repository provided is a real URI
 	IsValid() error
```
spencerschrock, line 28: a real URI? Or one valid by a forge's standard or our support?

spencerschrock (approved):
overall looks good. I think the cron client (as-is) will panic on a nil pointer deref, so let's fix that.

clients/gitlabrepo/client.go
raghavkaul: Yes, it's because the deps.dev client needs the domain + path but not scheme. But also, this matches how the GitHub client works more closely.

PR #4106, opened 2024-05-13
:warning: Simplify RunScorecard with functional optionals

### LINKED ISSUE #3717: Redesigning the top-level Scorecard entrypoint
Note: still tweaking this, but happy to get any feedback.
# Problem
Our current top-level entrypoint, [`RunScorecard`](https://github.com/ossf/scorecard/blob/5dc03b72e109d67afe4c91a0fbe52a3c1678a97f/pkg/scorecard.go#L92-L102),  has a long list of parameters.
 ```go
func RunScorecard(ctx context.Context,
	repo clients.Repo,
	commitSHA string,
	commitDepth int,
	checksToRun checker.CheckNameToFnMap,
	repoClient clients.RepoClient,
	ossFuzzRepoClient clients.RepoClient,
) (ScorecardResult, error)
```
There’s two problems with this:
1. Adding new options would require breaking changes to the function signature. While Scorecard has historically not cared about breaking changes, this is the single most important one, and we’ve broken it [countless times](https://github.com/ossf/scorecard/blame/3cbafa90122b18b92beaa68d6421d43b2f7cecec/pkg/scorecard.go#L93-L102) when [adding features](https://github.com/ossf/scorecard/pull/3688#discussion_r1399559443).
2. Callers need to provide all parameters, even if they’re fine with the default values. Looking at our own usages, the vast majority of our calls use the default values.

# Solution
We can use [functional options](https://sagikazarmark.hu/blog/functional-options-on-steroids/) to solve this problem. Functional options are a paradigm to pass an arbitrary number of options to a function to modify its behavior without changing its signature. At its core is the option type, which is unsurprisingly an alias for a function signature.

```go
type Option func(*somePrivateState) error
```

The first step is splitting our current arguments into required and optional values. I’m arguing that the only required values are the context and the repo to analyze (for rationale, see Appendix). Every other existing argument will use default values, unless modified by an option:
```go
func Run(ctx context.Context, repo clients.Repo, options ...Option) error {
    c := config{
        // some sensible defaults
    }
    for _, option := range options {
        if err := option(&c); err != nil {
            return err
        }
  }
  ... 
```

With this starting point, options are easily added. Let’s use commit SHA and depth as an example, we just add a new field to our unexported state, and an exported option to manipulate it:

```go
type config struct {
    commit      string
    commitDepth int
}
func WithCommitDepth(depth int) Option {
    return func(c *config) error {
        c.commitDepth = depth
        return nil
    }
}
func WithCommitSHA(sha string) Option {
	return func(c *config) error {
		c.commit = sha
		return nil
	}
}
```

Existing callers don’t need to change their signature, but those that want to customize the behavior can:
```
Run(ctx, repo)
Run(ctx, repo, WithCommitDepth(42), WithCommitSHA("foo..."))
```

# Appendix

A lot of the arguments either have default behavior built in already. Or the callers tend to provide the same arguments.

- `commitSHA string`,
  - `clients.HeadSHA` is the default, unless someone provides the `--commit` option
- `commitDepth int`
  - defaults to 30 if no option provided [link](https://github.com/ossf/scorecard/blob/5dc03b72e109d67afe4c91a0fbe52a3c1678a97f/clients/githubrepo/client.go#L78-L80)
- `checksToRun checker.CheckNameToFnMap`
  - The default is to run all checks, but if `--checks` is provided, only those checks are run. However with structured results we're also seeing `--probes` added (see #3688)

- all of the repo clients
  - repoClient clients.RepoClient,
  - ossFuzzRepoClient clients.RepoClient,
  - ciiClient clients.CIIBestPracticesClient,
  - vulnsClient clients.VulnerabilitiesClient,
  - internal
    - internal usages tend to do one of two things:
      - use default clients. e.g. [`clients.DefaultCIIBestPracticesClient()`](https://github.com/ossf/scorecard/blob/5dc03b72e109d67afe4c91a0fbe52a3c1678a97f/clients/cii_client.go#L66-L67)
      - use [`checker.GetClients`](https://github.com/ossf/scorecard/blob/5dc03b72e109d67afe4c91a0fbe52a3c1678a97f/checker/client.go#L31-L32), which indirectly creates the default clients
  - external
    - use [`checker.GetClients`](https://github.com/Legit-Labs/legitify/blob/6ea81a72bef090fe722eac53ef038f4633b1b9ca/internal/scorecard/wrapper.go#L27)

spencerschrock:

breaking change / refactor

See #3717. Ultimately every change to the signature is another breaking change, and this breaking change hopefully stops future breaking changes.

A new top level entry point: `Run`. The hope is that the `pkg` package also gets renamed so it's just `scorecard.Run()` instead of `pkg.RunScorecard` or `pkg.Run`, but that's not the topic of this PR.

This entry point takes a context and a repo to be analyzed. Everything else is optional and can be passed as a list of `WithFoo()` arguments.

Fixes #3717 

the `release-note`

```release-note
The main function to call Scorecard as a library was changed to reduce future breaking changes.
```


spencerschrock:
This is still a WIP, as I'm thinking about `WithChecks`, if that should be a `[]string` or the existing `checker.CheckNameToFnMap`.

There's also some duplication that happens setting up the `CheckNameToFnMap`
https://github.com/ossf/scorecard/blob/6815161e15b9acaec08bc868869894dc906e3252/cron/internal/worker/main.go#L202-L207

spencerschrock:
In order to automatically create a repo client if one isn't provided, I need to be able to test the Repo type. 
So my latest changes make the various `clients.Repo` implementations public.
https://github.com/ossf/scorecard/blob/438f838d6cb4d59579ef3651d23603e51dc59ae9/pkg/scorecard.go#L382-L399

pkg/scorecard.go
```diff
@@ -277,3 +282,131 @@ func ExperimentalRunProbes(ctx context.Context,
...
+	case *gitlabrepo.Repo:
+		if c.client == nil {
+			c.client, err = gitlabrepo.CreateGitlabClient(ctx, repo.Host())
+			if err != nil {
+				return ScorecardResult{}, fmt.Errorf("creating gitlab client: %w", err)
+			}
+		}
+	}
+
+	if !strings.EqualFold(c.commit, clients.HeadSHA) {
```
raghavkaul, line 401: It appears `--repo github.com/foo/bar --commit HEAD` would have an empty slice of requiredRequestTypes in this example, is that intentional?

raghavkaul (approved):

pkg/scorecard.go
spencerschrock: `--commit HEAD` is a no-op since that's the default. This is what the existing code already did.



https://github.com/ossf/scorecard/blob/91532e12d129f4f377a785269e560fbba82b86d4/cmd/root.go#L113-L124

Issue #2441, opened 2022-11-08
BUG: Scorecard's npm search doesn't match exact


scovetta:
**Describe the bug**
The Scorecard npm package integration uses the npm registry full text search function, which seems to omit deprecated packages. This means when searching for `left-pad`, you get the package `repeat-string`, which causes a later failure:

```
Error: GetClients: getting local directory client: error during parse: invalid repo flag: . Exepted full repository url
```

https://registry.npmjs.org/-/v1/search?text=left-pad&size=1
![image](https://user-images.githubusercontent.com/732166/200683024-b1ac173b-6107-4267-adfb-70e2f3c4c6af.png)

I think the call could be changed to just `registry.npmjs.org/<package name>`, but you'd have to find the best repository URL (probably in `versions[dist-tags.latest].repository.url`). Or, if there's an advantage to doing it the current way, then try that first and then fall back to the direct URL.

**Reproduction steps**
1. Run Scorecard against left-pad:

```
docker run -e GITHUB_AUTH_TOKEN=XXXXXXXXXXXXXXXXXXXXXX gcr.io/openssf/scorecard:stable --npm=left-pad --show-details
```

2. See the error:

```
Error: GetClients: getting local directory client: error during parse: invalid repo flag: . Exepted full repository url
2022/11/08 22:47:42 error during command execution: GetClients: getting local directory client: error during parse: invalid repo flag: . Exepted full repository url
```

**Expected behavior**
For left-pad, it should attempt to load the GitHub repository URL.

**Additional context**
N/A

PR #4118, opened 2024-05-20
🐛 Use direct endpoint instead of search to find repository URL from npm database

### LINKED ISSUE #3166: BUG: --npm wrong input does not throw error
**Describe the bug**
Scorecard can receive as input the name of the package from `npm`, `pypi` and `rubygems` ecosystems as per [the documentation](https://github.com/ossf/scorecard/blob/4cd5446862ea4c470810fea81fc7f45a36d04dec/README.md?plain=1#L423-L427). Reading the documentation, it was not clear to me that I needed to provide the package name and providing a package URL does not throw an error but runs the evaluation with a weird behavior.

**Reproduction steps**
1. Run Scorecard v4.10.2 with `--npm=https://github.com/airbnb/lottie-web`
2. See output results for `repo: name: github.com/perrmadiafrrian/react-lottie-light`

**Expected behavior**
I expected Scorecard would warn me I made a mistake in the `--npm` flag input and not run the evaluation for an unexpected repository.

**Additional context**
None.

aklevans:

This PR is a bug fix. When finding the github repository URL for an npm package, the application uses the `GET·/{package}/latest`  endpoint instead of the `GET·/{package}`

The application finds the github URL of an npm package by using the search functionality.  If a user incorrectly enters a github URL instead of an npm package name, the user will not receive feedback about their mistake and the application will run the checks on the first package that the search finds. 
When a user enters a github URL instead of an npm package name, an error message is displayed saying that the npm package could not be found.

Fixes #3166.

NO

```
NONE

```


raghavkaul:
Thanks for the contribution! Could you add some unit tests that expect an error when a GitHub/NPM URL is passed to the `--npm` flag instead of a project name? Even better to test the projects from #2144 and #3166, which silently score the wrong projects because of the inexact name match.

aklevans:
I added unit tests for #3166 and #2441 with mock http responses

raghavkaul:
Still getting linter issues, try `make fix-linter`.

For `package_manager_test.go`, can we make the diff smaller by only returning the parts of the HTTP/JSON response in `args.result` that we'll actually end up parsing?

aklevans:
I don't think fix-linter is working correctly on my machine. It just adds line breaks between every single line. Any reason this may happen?

raghavkaul:
Hm, that's strange that `--fix` has that behavior, I think you'll need to fix those issues manually then. It looks like some lints were fixed but there are still some issues. Make sure that `make all` passes, because our CI requires that for merge.

spencerschrock:
> I don't think fix-linter is working correctly on my machine. It just adds line breaks between every single line. Any reason this may happen?

Can you share a little bit about your environment? You're running the linter with `make check-linter`? The double-space the whole file approach should be reverted.

aklevans:
I think the issue was stemming from the fact I was working with windows. I have switched to Linux and I believe it is working now.

cmd/package_managers.go
```diff
@@ -129,23 +125,24 @@ type rubyGemsSearchResults struct {
...
+	v := &npmResult{}
 	err = json.NewDecoder(resp.Body).Decode(v)
 	if err != nil {
 		return "", sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("failed to parse npm package json: %v", err))
 	}
-	if len(v.Objects) == 0 {
+	if resp.StatusCode == 404 || v.Repository.URL == "" {
 		return "", sce.WithMessage(sce.ErrScorecardInternal,
 			fmt.Sprintf("could not find source repo for npm package: %s", packageName))
 	}
```
spencerschrock, line 144: since this deals with the status code, should we check this before decoding?

cmd/package_managers.go
```diff
@@ -129,23 +125,24 @@ type rubyGemsSearchResults struct {
 
 // Gets the GitHub repository URL for the npm package.
 func fetchGitRepositoryFromNPM(packageName string, packageManager pmc.Client) (string, error) {
-	npmSearchURL := "https://registry.npmjs.org/-/v1/search?text=%s&size=1"
+	npmSearchURL := "https://registry.npmjs.org/%s/latest"
```
spencerschrock, line 128: nit: we're no longer using the search API, so the URL shouldn't be `npmSearchURL`

spencerschrock (approved):
Overall looks good, thanks!

cmd/package_managers.go
spencerschrock: although I see now the second part needs the decoded fields. so feel free to ignore this.

If you happen to be fixing the other comment, can you use `http.StatusNotFound` instead of 404?

cmd/package_managers.go
aklevans: fixed!

cmd/package_managers.go
aklevans: Sure

PR #4131, opened 2024-05-29
:bug: fix Docker remediations for unpinned GHA dependencies

raboof:

Bug fix

As both the check for unpinned dependencies in GitHub Actions and the check for unpinned Docker dependencies contribute to d.Dependencies, the loop that created remediations for Docker dependencies would also create try to create Docker remediations for the unpinned GitHub Actions dependencies.

This could get really slow, especially when scanning a repo with many GitHub Actions such as https://github.com/apache/beam.

Only create Docker remediations for Docker dependencies (and similar for GitHub workflow remediations)

NONE

the `release-note`

```release-note
Improve Pinned-Dependencies remediation creation
```

### DIFF
--- a/checks/raw/pinned_dependencies_test.go
+++ b/checks/raw/pinned_dependencies_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ossf/scorecard/v5/checker"
 	mockrepo "github.com/ossf/scorecard/v5/clients/mockclients"
 	"github.com/ossf/scorecard/v5/finding"
+	"github.com/ossf/scorecard/v5/remediation"
 	scut "github.com/ossf/scorecard/v5/utests"
 )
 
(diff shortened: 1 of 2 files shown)


checks/raw/pinned_dependencies_test.go
```diff
@@ -415,6 +416,75 @@ func TestDockerfilePinning(t *testing.T) {
...
+		PinnedAt: stringAsPointer("daadedc81d5f9d3c06d2c92f49202a3cc2b919ba"),
+		Location: &checker.File{
+			Path:      ".github/workflows/workflow-not-pinned.yaml",
+			Snippet:   "actions/checkout@daadedc81d5f9d3c06d2c92f49202a3cc2b919ba",
+			Offset:    31,
+			EndOffset: 31,
+			FileSize:  0,
+			Type:      1,
+		},
+		Pinned:      boolAsPointer(false),
```
spencerschrock, line 433: This looks pinned due to the `Snippet` and the `PinnedAt`, but the `Path` and `Pinned` fields seem to conflict with that.

spencerschrock (approved):
In the context of how the code works now, I think this change seems reasonable. 

How did the speed-up look on your end? Is there still need for improvement, we can try to get rid of the quadratic behavior currently exhibited as the code iterates over dependencies multiple times, which for large projects like the ASF may take a lot of time still.

These functions are all called from 
https://github.com/ossf/scorecard/blob/5447253ff192e7e37a7be3b911329c962519aae5/checks/raw/pinned_dependencies.go#L36-L63

So `collectGitHubActionsWorkflowPinning` identifies workflow deps W1, W2, and W3 via `validateGitHubActionWorkflow` and then cycles through them.

Then `collectDockerfilePinning` identifies docker images D1, D2, and D3, and via `validateDockerfilesPinning`, and then cycles through W1, W2, W3, **AND** D1, D2, and D3.

This continues with `collectDockerfileInsecureDownloads`, `collectShellScriptInsecureDownloads`, and `collectGitHubWorkflowScriptInsecureDownloads`. 

This could be fixed by changing where we append to the the final dep slice, and iterating over temporary slices, but if its not worth the effort we don't need to do it yet.

raboof:
> How did the speed-up look on your end?

`scorecard --repo=https://github.com/apache/beam --checks Pinned-Dependencies` went from 16 minutes to 24 seconds on my machine ;)

> Is there still need for improvement

Faster is always better, and I haven't done any further profiling, but I suspect the bottlenecks will be elsewhere now.

> This continues with `collectDockerfileInsecureDownloads`, `collectShellScriptInsecureDownloads`, and `collectGitHubWorkflowScriptInsecureDownloads`.

These don't iterate over the `Dependencies`, though, as far as I noticed

spencerschrock:
> `scorecard --repo=https://github.com/apache/beam --checks Pinned-Dependencies` went from 16 minutes to 24 seconds on my machine ;)

Fantastic. Thanks for catching this!

> Faster is always better, and I haven't done any further profiling, but I suspect the bottlenecks will be elsewhere now.
> 
> > This continues with `collectDockerfileInsecureDownloads`, `collectShellScriptInsecureDownloads`, and `collectGitHubWorkflowScriptInsecureDownloads`.
> 
> These don't iterate over the `Dependencies`, though, as far as I noticed

I did a quick and dirty test to avoid going over the GHA deps again in `collectDockerfilePinning` and got another 20% speedup: 24s to 20s on my machine for`github.com/apache/beam`, which is better but like you said probably no longer a bottleneck.

PR #4132, opened 2024-05-29
:sparkles: announce where results are written

raboof:

Minor usability improvement

When running with '-o foo' the output would end with:

```
RESULTS
-------
```

This was rather confusing. There's of course many ways to make this more clear, this commit adds a log line announcing where the output is written to:

```
RESULTS
-------
Writing to foo
```

NONE

the `release-note`

```release-note
NONE
```


raghavkaul (approved):
Thanks!

spencerschrock:
> I believe there is a lint that is supposed to prevent new print statements, not sure why it isn't complaining here...

The linter was added primarily from people writing to  `stdout`, so it's not currently configured for `fmt.Fprintf` which often redirects to a different `io.Writer`
https://github.com/ossf/scorecard/blob/3da6db56c91033e53933000fb7f26e8e617d4b77/.golangci.yml#L93-L96

Issue #4048, opened 2024-04-22
BUG: Patch Maintainers Annotations


gabibguti:
**Describe the bug**
Some things still need to be addressed in Maintainers Annotation feature (regarding PR https://github.com/ossf/scorecard/pull/3905):

- [X] In checker/check_result.go, rename IsExempted to Annotations (https://github.com/ossf/scorecard/pull/3905#discussion_r1548354284)
- [X] In config/annotations.go, change IsValidReason to private method as isValidReason (https://github.com/ossf/scorecard/pull/3905#discussion_r1548364147).
- [x] similarly, these errors can probably be private? https://github.com/ossf/scorecard/blob/19f8ae107daead9d25f3f7406f00a206bcd17eef/config/config.go#L28-L31
- [ ] Clarify the desired types for annotations. Define if Paths should be under a Reason or under an Annotation and same thing for Probes. We need to understand the better way for users to fill in the annotations (https://github.com/ossf/scorecard/pull/3905#discussion_r1530869510).
- [x] Define if one invalid check or reason during the config parsing should invalidate the whole config, or if it should be logged and skipped (https://github.com/ossf/scorecard/pull/3905#discussion_r1530857745 and https://github.com/ossf/scorecard/pull/3905#discussion_r1548411138).
- [X] Define if in the case a check is annotated it should not be removed from SARIF result `runs`. Or, if is working as expected and we should not populate `runs` if there are no relevant `results` for that check (https://github.com/ossf/scorecard/pull/3905#issuecomment-2030613931).
- [X] Do we need to pass in checks here? Also this **doesn't work if you only run a subset of checks but there's an annotation for a check which isn't being run**. https://github.com/ossf/scorecard/blob/19f8ae107daead9d25f3f7406f00a206bcd17eef/config/config.go#L73-L74
- [ ] Closing the file with t.Cleanup which is opened here: https://github.com/ossf/scorecard/blob/19f8ae107daead9d25f3f7406f00a206bcd17eef/config/config_test.go#L134-L138
- [X] Make all output format options pointers for consistency?  AsJSON2, AsString, etc: https://github.com/ossf/scorecard/blob/19f8ae107daead9d25f3f7406f00a206bcd17eef/pkg/json.go#L131-L134
- [x] Support more config file locations (e.g. `.scorecard.yml`, `.github/.scorecard.yml`, ...)
- [X] Update the config README, it uses outdated yaml tags / values. 
  - [X] Document config file locations

[...]

PR #4151, opened 2024-06-06
:warning: Make all ScorecardResult format options pointers

spencerschrock:

breaking change

Some of the `AsX` functions take pointers (`AsProbe`), some take structs (`AsString`, `AsJSON2`)

All three `AsX` functions which have option structs, take pointers.

One item from #4048 

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cmd/internal/scdiff/app/format/format.go
+++ b/cmd/internal/scdiff/app/format/format.go
@@ -56,7 +56,7 @@ func JSON(r *pkg.ScorecardResult, w io.Writer) error {
 		return err
 	}
 	Normalize(r)
-	o := pkg.AsJSON2ResultOption{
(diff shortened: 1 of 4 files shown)


spencerschrock:
> Just curious, what's the motivation to make all of the functions return pointers instead of all return values?

From my own experience, being able to pass `nil` is slightly shorter when I just want to use the defaults. But I took a look at what some of our other libraries do, and they both use pointers for options too, so I figured that was better.

For example the go-gitlab often uses pointers: https://pkg.go.dev/github.com/xanzy/go-gitlab#ProjectsService.GetProject
As does our go-github library: https://pkg.go.dev/github.com/google/go-github/github#ChecksService.ListCheckRunsForRef

raghavkaul (approved):

PR #4152, opened 2024-06-06
🌱 maintainer annotations: search for config

raghavkaul:
Look for maintainer annotation config files in multiple locations: `scorecard.yml`, `.scorecard.yml`, `.github/scorecard.yml`. Scorecard uses the first one it finds regardless of whether it's valid.

Only look for `scorecard.yml` and don't apply maintainer annotation if that file doesn't exist.

Addresses #4048

### DIFF
--- a/pkg/scorecard.go
+++ b/pkg/scorecard.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 	"sync"
@@ -168,21 +169,23 @@ func runScorecard(ctx context.Context,
 	go runEnabledChecks(ctx, repo, request, checksToRun, resultsCh)
 
 	if os.Getenv(options.EnvVarScorecardExperimental) == "1" {
-		// Get configuration
-		rc, err := repoClient.GetFileReader("scorecard.yml")
-		// If configuration file exists, continue. Otherwise, ignore
+		r, err := findConfigFile(repoClient)
+
+		logger := sclog.NewLogger(sclog.DefaultLevel)
+
 		if err == nil {
-			defer rc.Close()
+			defer r.Close()
 			checks := []string{}
 			for check := range checksToRun {
 				checks = append(checks, check)
 			}
-			c, err := config.Parse(rc, checks)
+			c, err := config.Parse(r, checks)
 			if err != nil {
-				logger := sclog.NewLogger(sclog.DefaultLevel)
 				logger.Error(err, "parsing configuration file")
 			}
 			ret.Config = c
+		} else {
+			logger.Info(fmt.Sprintf("no maintainer annotation: %v", err))
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/scorecard.go
```diff
@@ -168,21 +169,23 @@ func runScorecard(ctx context.Context,
...
 			}
-			c, err := config.Parse(rc, checks)
+			c, err := config.Parse(r, checks)
 			if err != nil {
-				logger := sclog.NewLogger(sclog.DefaultLevel)
 				logger.Error(err, "parsing configuration file")
 			}
 			ret.Config = c
+		} else {
+			logger.Info(fmt.Sprintf("no maintainer annotation: %v", err))
```
spencerschrock, line 188: I lean towards the opposite. Logging if a maintainer annotation file is found and used, not if it's absent.

For example, golangci-lint spits this out, and we could do the same:
```
INFO [config_reader] Used config file .golangci.yml
```

spencerschrock: Plus, I don't think `err` is set appropriately there and shouldn't be in the message.

pkg/scorecard_test.go
```diff
@@ -341,3 +346,65 @@ func TestExperimentalRunProbes(t *testing.T) {
...
+		{
+			desc:      ".github/scorecard.yml exists",
+			locs:      []string{".github/scorecard.yml"},
+			wantFound: true,
+		},
+		{
+			desc:      "multiple configs exist",
+			locs:      []string{"scorecard.yml", ".github/scorecard.yml"},
+			wantFound: true,
+		},
```
spencerschrock, line 378: do we want to test precedence at all? or nah?

pkg/scorecard_test.go
```diff
@@ -341,3 +346,65 @@ func TestExperimentalRunProbes(t *testing.T) {
...
+			mockRepoClient.EXPECT().GetFileReader(gomock.Any()).AnyTimes().DoAndReturn(func(filename string) (io.ReadCloser, error) {
+				if !slices.Contains(tt.locs, filename) {
+					return nil, fmt.Errorf("os.Open: %s", filename)
+				}
+				fullPath := filepath.Join("./testdata", filename)
+				f, err := os.Open(fullPath)
+				if err != nil {
+					return nil, fmt.Errorf("os.Open: %w", err)
+				}
+				return f, nil
```
spencerschrock, line 401: I don't think the slices.Contains block is needed. If the file doesn't exist, os.Open will fail anyway.

So something like  this should be sufficient (wrapping linter is disabled for test funcs)
```go
mockRepoClient.EXPECT().GetFileReader(gomock.Any()).AnyTimes().DoAndReturn(func(filename string) (io.ReadCloser, error) {
	fullPath := filepath.Join("./testdata", filename)
	return os.Open(fullPath)
})
```

spencerschrock (approved):
Seems reasonable. Let me know what you think about the logging flip. And please document the config precedence in `config/README.md`

pkg/scorecard_test.go
raghavkaul: Added a test for precedence

pkg/scorecard_test.go
raghavkaul: This is a bit confusing, but my intention was that the test files (which are there in testdata) should only "exist" if the test case def'n says they should, i.e. if the test case def'n includes them in `tt.locs`. The files in testdata are always there though, so os.Open on L397 would always work.

pkg/scorecard.go
raghavkaul: Fixed, only logging if annotation can't be parsed.

PR #4162, opened 2024-06-10
🌱 maintainer annotations: improve annotation file validation

raghavkaul:
Validate maintainer annotation check names against full list of check names instead of checks passed via `runScorecard`

Addresses #4048.


config/config.go
```diff
@@ -45,19 +45,63 @@ func parseFile(c *Config, content []byte) error {
...
+	Packaging            CheckName = "Packaging"
+	PinnedDependencies   CheckName = "Pinned-Dependencies"
+	SAST                 CheckName = "SAST"
+	SBOM                 CheckName = "SBOM"
+	SecurityPolicy       CheckName = "Security-Policy"
+	SignedReleases       CheckName = "Signed-Releases"
+	TokenPermissions     CheckName = "Token-Permissions"
+	Vulnerabilities      CheckName = "Vulnerabilities"
+	Webhooks             CheckName = "Webhooks"
+)
```
spencerschrock, line 72: Is there any way we can de-duplicate from 

https://github.com/ossf/scorecard/blob/20ec42c2b5b993e66293e6b34c04cf554e4990c8/internal/probes/probes.go#L25-L50

maybe by moving to `internal/checknames/checknames.go` or something? and referencing those here?

If redefined here, they shouldn't be public.

config/config.go
```diff
@@ -45,19 +45,63 @@ func parseFile(c *Config, content []byte) error {
...
+	SecurityPolicy,
+	SignedReleases,
+	TokenPermissions,
+	Vulnerabilities,
+	Webhooks,
+}
+
+func isValidCheck(check string) bool {
+	return slices.Contains(allValidChecks, check)
 }
```
spencerschrock, line 99: ideally this would be a map?

config/config_test.go
```diff
@@ -41,7 +40,7 @@ func Test_Parse_Checks(t *testing.T) {
 			want: config.Config{
 				Annotations: []config.Annotation{
 					{
-						Checks:  []string{"binary-artifacts"},
+						Checks:  []string{"Binary-Artifacts"},
```
spencerschrock, line 43: in my opinion this really should be case insensitive, so the old "binary-artifacts" should work.

config/config_test.go
```diff
@@ -125,20 +124,21 @@ func Test_Parse_Checks(t *testing.T) {
...
 			r, err := os.Open(tt.configPath)
 			if err != nil {
 				t.Fatalf("Could not open config test file: %s", tt.configPath)
 			}
-			result, err := config.Parse(r, allChecks)
+			defer r.Close()
+
+			result, err := config.Parse(r)
+			if (tt.wantErr != nil) != (err != nil) {
+				t.Errorf("expected %+v got %+v", tt.wantErr, err)
```
spencerschrock, line 137: this should likely be a `t.Fatalf` as it's an unexpected error?

config/config.go
raghavkaul: I made them private, but I think creating a new module will clutter up `utils` and lead to a lot more import statements.

If the issue with duplication is that a new check writer will have to update several check lists, we could have an E2E test that imports all these modules and checks they have the same lists? But that would require them to be public.

config/config.go
raghavkaul: How would a map help?

config/config.go
spencerschrock: > clutter up utils

I'm not sure which `utils` package you're talking about.

> lead to a lot more import statements.

it's being used in 2 spots, I dont think having 1 new import in 2 files is a problem. and seems simpler than the e2e test you're considering adding.

config/config.go
spencerschrock: Set membership is traditionally done in a map in constant time, instead of a linear search in a slice.

But at this scale (19 checks), it probably doesn't matter.

config/config_test.go
```diff
@@ -124,17 +123,13 @@ func Test_Parse_Checks(t *testing.T) {
 	}
 	for _, tt := range tests {
 		tt := tt // Re-initializing variable so it is not changed while executing the closure below
```
spencerschrock, line 125: optional: now that the import cycle is broken, lines 15-17 (wont let me pick them) can be changed just to `package config`?

But only if you feel like it, not important.

spencerschrock (approved):
Looks good thanks!

Issue #4171, opened 2024-06-13
BUG: osv-scanner panic: runtime error: index out of range [0] with length 0


Zxilly:
**Describe the bug**

osv-scanner panics on https://scorecard.dev/viewer/?uri=github.com/Zxilly/go-size-analyzer, but I can't reproduce it while run it locally with `osv-scanner -r .`

**Reproduction steps**

Failed to reproduce.

**Expected behavior**

Process correctly.

spencerschrock:
I also can't reproduce it locally with Scorecard
```shell
go run main.go --repo Zxilly/go-size-analyzer --checks Vulnerabilities
```
I've seen this panic before in our weekly cron infrastructure, but I've never been able to reproduce it locally.
The panic is in `osv-scanner`, but likely in an error handling pathway when it receives some error or empty response from osv.dev, as it's not easily triggered.

Zxilly:
So I created a PR to produce the stack info while we met the error. Please review it at https://github.com/ossf/scorecard/pull/4172

Zxilly:
@AdamKorcz  It is only helpful to identify the real bug, did not resolve it.
In fact I'm waiting for a new release to apply this to my real production environment.  I can't reproduce it locally, but it always happen in the github actions.

spencerschrock:
> In fact I'm waiting for a new release to apply this to my real production environment. I can't reproduce it locally, but it always happen in the github actions.

I had assumed this was a different panic, and was occurring in the cron. But looking again at your Scorecard Action, I see now what it is.

The latest scorecard action currently is v2.3.3 which uses v5.0.0-rc2 of scorecard. If you use that tag locally to analyze the repo you can reproduce it. Cherry picking #4172 shows the stack trace. 

osv-scanner is struggling to parse [your pnpm lockfile](https://github.com/Zxilly/go-size-analyzer/blob/08f323fb346d281dd143e70a4305a0c6602f3d46/ui/pnpm-lock.yaml). This was fixed in [osv-scanner v1.7.3](https://github.com/google/osv-scanner/blob/main/CHANGELOG.md#v173), which is why we're not seeing it at scorecard HEAD. This will be picked up at the next scorecard action release.

<details>
<summary> Click for trace </summary>

```
runtime/debug.Stack()
        ~/sdk/go1.22.4/src/runtime/debug/stack.go:24 +0x5e
github.com/ossf/scorecard/v5/clients.osvClient.ListUnfixedVulnerabilities.func1()
[...]

PR #4172, opened 2024-06-13
:seedling: add stack info to osv-scanner error

Zxilly:

feature

only print a error string for osv-scanner panic error

also print the stacktrace

Updates #4171 

```release-note
NONE
```

### DIFF
--- a/clients/osv.go
+++ b/clients/osv.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime/debug"
 
 	"github.com/google/osv-scanner/pkg/osvscanner"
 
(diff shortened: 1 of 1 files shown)


clients/osv.go
```diff
@@ -38,7 +39,10 @@ func (v osvClient) ListUnfixedVulnerabilities(
 ) (_ VulnerabilitiesResponse, err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			err = sce.CreateInternal(sce.ErrScorecardInternal, fmt.Sprintf("osv-scanner panic: %v", r))
+			err = sce.CreateInternal(
+				sce.ErrScorecardInternal,
+				fmt.Sprintf("osv-scanner panic: %v\n%s", r, string(debug.Stack())),
+			)
```
spencerschrock, line 45: This error ends up in the published results, API, etc, as you noted in #4171. I don't think this level of debugging info belongs there. 

We can still return the internal error saying osv-scanner had a panic, and put the stack info in stderr instead?

Zxilly:
One e2e test failed, reaching Github speed limit, not related with this patch.

clients/osv.go
spencerschrock: If the issue occurs in GitHub Actions, you could look for it in the GitHub action output. In our weekly scan infrastructure, stderr goes to our logging and metrics where I can search for it and grab the stack trace with the osv-scanner team.

spencerschrock (approved):

spencerschrock:
Thank you!

Zxilly:
I'm wondering when will we got a new release.

spencerschrock:
> I'm wondering when will we got a new release.

For our CLI and action, we are targeting end of the month for the full, non-release candidate v5. This code is already deployed in our infrastructure though.

PR #4207, opened 2024-07-03
:seedling: Use new entrypoint for cron worker

spencerschrock:

cron update

Uses the old `pkg.RunScorecard`

Uses the new `pkg.Run`
There are further simplifications that are possible, but leaving for later to keep this PR small.

Follow up of #3717 

the `release-note`

```release-note
NONE
```

### DIFF
--- a/cron/internal/worker/main.go
+++ b/cron/internal/worker/main.go
@@ -40,7 +40,6 @@ import (
 	"github.com/ossf/scorecard/v5/cron/worker"
 	docs "github.com/ossf/scorecard/v5/docs/checks"
 	sce "github.com/ossf/scorecard/v5/errors"
-	"github.com/ossf/scorecard/v5/internal/packageclient"
 	"github.com/ossf/scorecard/v5/log"
 	"github.com/ossf/scorecard/v5/pkg"
 	"github.com/ossf/scorecard/v5/policy"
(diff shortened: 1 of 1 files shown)


spencerschrock:
I used the [local cron environment](https://github.com/ossf/scorecard/tree/1c448ee6522578a20a89d7f924debd0624c7bd71/cron/internal/emulator#readme) to test with a random selection of 50 repos. The results were compared with `scdiff` and match.


We'll also be able to see how this performs in the release test environment after merging.

cron/internal/worker/main.go
```diff
@@ -176,7 +172,6 @@ func processRequest(ctx context.Context,
 	githubClient, gitlabClient clients.RepoClient, ossFuzzRepoClient clients.RepoClient,
 	ciiClient clients.CIIBestPracticesClient,
 	vulnsClient clients.VulnerabilitiesClient,
-	projectClient packageclient.ProjectPackageClient,
```
raghavkaul, line 179: Any reason this isn't being passed but the rest are?

spencerschrock: There isn't a WithProjectPackageClient option because there's only one client, so there's not currently a way of configuring another option. One will still get made in `pkg.Run`.

https://github.com/ossf/scorecard/blob/5f7cea363724ba03923eb07bdb4e6fa3b2927c15/pkg/scorecard.go#L388-L390

spencerschrock: Happy to make one for testing / mocking purposes if needed, but functionally isn't need here.

raghavkaul: Gotcha, I didn't realize it was the only Client that just had one possible implementation.

raghavkaul (approved):

justaugustus (approved):

Issue #3438, opened 2023-08-31
Missing checks in API response


UlisesGascon:
**Describe the bug**

While comparing the scoring with the API:
- Request1: [Nodejs (bd31d489615b424202553ffe1a5093a19d5e5e1f)](https://api.securityscorecards.dev/projects/github.com/nodejs/nodejs.org/?commit=bd31d489615b424202553ffe1a5093a19d5e5e1f)
- Request2: [Nodejs (1fbd908497ece0c5b08c7cc8f7494d7a12e7001f)](https://api.securityscorecards.dev/projects/github.com/nodejs/nodejs.org/?commit=1fbd908497ece0c5b08c7cc8f7494d7a12e7001f)

In the Request2 the payload does not include all the checks, for example `"CI-Tests"` is missing inside [the checks](https://api.securityscorecards.dev/#/results/getResult). 

**Reproduction steps**
1. You can check this in action at: https://kooltheba.github.io/openssf-scorecard-api-visualizer/#/projects/github.com/nodejs/nodejs.org/compare/1fbd908497ece0c5b08c7cc8f7494d7a12e7001f/bd31d489615b424202553ffe1a5093a19d5e5e1f (explore the network request in the chrome dev tools) or do the HTTP request directly in the browser and compare the payloads

**Expected behavior**

I will expect both payloads to include all the checks included in the scorecard, or at least an error code that explains this behaviour, so the API consumers can control the error.

**Additional context**

This bug is affecting:
- [openssf-scorecard-api-visualizer](https://github.com/KoolTheba/openssf-scorecard-api-visualizer) cc: @kooltheba
- [OpenSSF Scorecard Monitor](https://github.com/marketplace/actions/openssf-scorecard-monitor)

spencerschrock:
I believe this is a matter of where the result is sourced. The API pulls from the GitHub action published results if available, or the weekly cron if not.
https://github.com/ossf/scorecard-webapp/blob/e7be60069b18df32c7d3b003dd9daca6a622e6c5/app/server/get_results.go#L70-L88

```diff
diff checks1 checks2
4d3
< "CI-Tests"
6d4
< "Contributors"
8d5
< "Dependency-Update-Tool"
```

These are the checks disabled in the weekly cron:
https://github.com/ossf/scorecard/blob/b68ef18c9691a7be16dbbc60c32cd736dd372d1d/cron/config/config.yaml#L48

The output is basically tied to the checks requested, so
```
[...]

PR #4226, opened 2024-07-08
🌱 Update active `cisco` projects, remove `cisco-open` projects

lelia:

Updates the active list of [cisco](https://github.com/cisco) projects, and removes [cisco-open](https://github.com/cisco-open) projects from the weekly cronjob. 

We have recently installed the Scorecard Action on all `cisco-open` projects, and are using Scorecard Monitor to further evaluate each project, so we want to ensure only the results from the Action are being used, given the [known limitations](https://github.com/ossf/scorecard/issues/3438) with Scorecard results from the cronjob.

Currently, only some `cisco` projects are checked by Scorecard, and all `cisco-open` projects are included.

NONE

```release-note
NONE
```


spencerschrock:
> are using Scorecard Monitor to further evaluate each project, so we want to ensure only the results from the Action are being used

Currently the API prioritizes the action results over the cron results already. Were you seeing otherwise?

lelia:
> > are using Scorecard Monitor to further evaluate each project, so we want to ensure only the results from the Action are being used
> 
> Currently the API prioritizes the action results over the cron results already. Were you seeing otherwise?

No issues with that currently, but we have some older Scorecard Monitor reports with results from the cronjob that were causing some confusion when trying to analyze overall project health, so I wanted to create a clean separation.

I don't anticipate there being collisions with reports we run in the future, but there's less value in having `/cisco-open` projects analyzed by the cronjob, compared with `/cisco` projects, which we have limited access to influence directly.

spencerschrock (approved):

PR #4231, opened 2024-07-09
:sparkles: remove experimental gate on maintainer annotation parsing

spencerschrock:

feature enablement

Looking for maintainer annotations is gated behind the SCORECARD_EXPERIMENTAL flag

The feature is no longer gated

NONE

the `release-note`

```release-note
NONE
```

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -229,13 +229,6 @@ func (o *Options) Probes() []string {
 	return o.ProbesToRun
 }
 
-// isExperimentalEnabled returns true if experimental features were enabled via
-// environment variable.
-func (o *Options) isExperimentalEnabled() bool {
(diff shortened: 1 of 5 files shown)


options/options.go
```diff
@@ -229,13 +229,6 @@ func (o *Options) Probes() []string {
 	return o.ProbesToRun
 }
 
-// isExperimentalEnabled returns true if experimental features were enabled via
```
raghavkaul, line 232: nit: Why remove? We'll probably need this flag in the future.

raghavkaul (approved):

options/options.go
spencerschrock: `unused` linter

PR #4335, opened 2024-09-04
:ghost: git ignore vendor folder

6543:
minor .gitignore change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ dist/*
 repos.txt
 oldRelease.json
 newRelease.json
+
+# Ignore golang's vendored files
+vendor/
--- a/.gitignore
+++ b/.gitignore
@@ -61,4 +61,5 @@ oldRelease.json
 newRelease.json
 
 # Ignore golang's vendored files
-vendor/
+/vendor/
+/tools/vendor/


6543:
-> added better description

spencerschrock:
Can you explain the motivation for this change? We don't vendor dependencies in this project, are you trying to locally?

6543:
Some tooling does work better if you `go mod vendor` a golang prohect first ...

... but as you mention you dont vendor this files in git, so it should be ignored to not interfear

6543:
@spencerschrock ☝️ ?

.gitignore
```diff
@@ -59,3 +59,6 @@ dist/*
 repos.txt
 oldRelease.json
 newRelease.json
+
+# Ignore golang's vendored files
+vendor/
```
spencerschrock, line 64: Let's make this an absolute path, 
```
/vendor/
```

we have a folder with some testdata named "vendor"
```
$ find -name vendor
./checks/raw/testdata/vendor
```

spencerschrock: And probably our other go.mod folder too then.
```
/tools/vendor/
```

6543, line 64:
```suggestion
/vendor/
/tools/vendor/
```

spencerschrock (approved):

PR #4369, opened 2024-10-06
:sparkles: Support Nuget Central Package Management

### LINKED ISSUE #4252: BUG: .Net pinned dependency should support Central Package Management
**Describe the bug**
In the .NET ecosystem all packages that are uploaded to NuGet.org are digitally signed with a x.509 certificate, and NuGet checks the validity of these signatures (and their certificate lineage) as part of the NuGet Restore action.
 
The current implementation of .Net pinned dependency only checks for using lock file mechanism, but instead it should check if either a lock file mechanism is used as implemented and described [here](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#restore-behavior-with-lock-file), OR if the [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management) feature had been enabled together with pinning to specific versions of dependencies.

**Reproduction steps**
1. The pinned dependency check for .net repositories should return the dependencies as pinned in cases where CPM feature is enabled and the versions of dependencies are pinned to specific versions.
2. for example: go run main.go --repo Noggog/CSharpExt   --checks Pinned-Dependencies --format json --show-details | jq

**Expected behavior**
.net repositories can pin dependencies with an exact version by using Central Package Management or utilize lock files.
for the example above (Noggog/CSharpExt), it is expected that 2 out of 2 nugetCommand dependencies will be pinned.

balteravishay:

Support nuget [central package management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management) in detecting pinned dependencies

Today, scorecard only supported nuget pinning dependencies with lockfiles. However, nuget provides package-manager-level assurance that when enabling CPM AND declaring specific versions for all direct dependencies the same level of pinning is provided to the end-user.

Scorecard will check if Directory.*.props file is present and both: 
a. ManagePackageVersionsCentrally attribute is set to true 
b. all direct dependencies have a specific version declared

if both are true, all unpinned dependencies will be marked as pinned.

Fixes #4252

This was discussed in a community call with @spencerschrock and it continues the work that was started in PR #4351

Here is a repo that would be found as pinned by this new checks: [foesmm/WolvenKit](https://github.com/foesmm/WolvenKit) since it has a [github workflow](https://github.com/foesmm/WolvenKit/blob/main/.github/workflows/check-only.yml) with dotnet restore (without lockfile flag) but declares CPM and all dependencies are pinned to specific versions in the [properties file](https://github.com/foesmm/WolvenKit/blob/main/Directory.Packages.props)

-->

```release-note
NONE
```


spencerschrock:
> all direct dependencies have a specific version declared

I seem to remember listing specific versions was considered pinned due to some immutable property of the nuget package manager? (Similar to how the Go ecosystem has a checksum database?) Is there a good link for that? I found https://devblogs.microsoft.com/nuget/nuget-package-signing/, but it seems to be a few years old, and didn't know what the default policy of the tool was these days.

JonDouglas:
@jeffkl could you advise on this PR on supporting CPM in this project? 😄

checks/raw/testdata/Directory.Pinned.packages.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageVersion Include="SomePackage" Version="4.4.5" />
```
jeffkl, line 6: Package versions are not pinned just because a version is declared.  NuGet just uses this version specified when it comes across a `<PackageReference Include="SomePackage" />` and emits an error if you specified a version in a project.

To truly "pin" a transitive package version that you do not have a direct `<PackageReference />` for, you have to enable transitive pinning: https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management#transitive-pinning

jeffkl: Nevermind, I was wrong what "pinning" meant in this context.

spencerschrock: [We define "pinned"](https://github.com/ossf/scorecard/blob/28db9a99a4c2e43e1f90a5afed6753e6ef2c36d3/docs/checks.md#pinned-dependencies) in a reproducibility sense: "a dependency that is explicitly set to a specific hash instead of
allowing a mutable version or range of versions". Which Avishay added support for via `nuget --locked-mode`, or the `RestoreLockedMode` property .

The NuGet claim in the PR description is that:

> Today, scorecard only supported nuget pinning dependencies with lockfiles. However, nuget provides package-manager-level assurance that when enabling CPM AND declaring specific versions for all direct dependencies the same level of pinning is provided to the end-user.

Which to my limited understanding, is based on the [package immutability nuget.org provides](https://github.com/NuGet/Home/wiki/Package-Immutability). But I'm not sure that fully covers all scenarios mentioned in the [lockfile blog post](https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/#why-use-a-lock-file).

balteravishay: @spencerschrock, could you please elaborate which scenarios mentioned in the [lockfile blog post](https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/#why-use-a-lock-file) you are referring to as missing from this PR:

- nuget.config mismatch: is covered with CPM, since CPM suggests that packages are managed centrally and the user gets [warnings](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#warning-when-using-multiple-package-sources) when using multiple package sources
- Intermediate versions and Floating versions - This PR checks for specifying specific versions and will deem floating versions as "unpinned"
- Package deletion is granted by [nuget](https://learn.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages)
- Package content mismatch is granted by [Package Immutability/signing](https://github.com/NuGet/Home/wiki/Package-Immutability)

spencerschrock: I've been out on the office on personal time, will take a look in the next day or two.

spencerschrock: > Intermediate versions and Floating versions - This PR checks for specifying specific versions and will deem floating versions as "unpinned"

I believe my comment was for this example:
* Day 1: If you specified `<PackageReference Include="My.Sample.Lib" Version="4.0.0"/>` but the versions available on the NuGet repositories were 4.1.0, 4.2.0 and 4.3.0, NuGet resolves to 4.1.0 as it the nearest minimum version.
* Day 2: Version 4.0.0 gets published. NuGet now restores version 4.0.0 as it is an exact match.

JonDouglas: I’m a bit unclear on the terminology being used here, and I’d appreciate some clarification. In the example from `checks.md`, it says:

> If your project is producing an application, declare all your dependencies with **specific versions** in your package format file (e.g. package.json for npm, requirements.txt for python, **packages.config for nuget**)

My understanding has always been that specifying exact versions is what we consider a pinned dependency check, and this is how we refer to it. Pinning provides consistency, and we typically **improve reproducibility further** with lock files for added determinism.

The example given, where the package manager resolves to the nearest version when an exact match isn’t available, doesn’t seem that compelling since other package managers handle similar cases. In practice, people pin versions that actually exist, and our registry follows an immutability policy, ensuring versions aren’t removed and that these dependencies remain reliable.

It might be helpful to create documentation that explains the different levels/scenarios of "pinned" dependency checks, with clear examples for each scenario. For instance, detailing how an exact version match qualifies as "pinned," a lock file represents "locked," and so forth, to ensure clarity across all cases.

aortiz-msft: > * Day 2: Version 4.0.0 gets published. NuGet now restores version 4.0.0 as it is an exact match.

Typically, we see folks publishing higher versions of their packages as time goes by, not lower versions.

spencerschrock: > It might be helpful to create documentation that explains the different levels/scenarios of "pinned" dependency checks, with clear examples for each scenario. For instance, detailing how an exact version match qualifies as "pinned," a lock file represents "locked," and so forth, to ensure clarity across all cases.

Most of our documentation uses "pinned" to refer to lock files, so we can certainly try to clarify "pinned" vs "locked".

I did some digging into the history of the check, as well as checking with relevant contributors.   It started off as [`Frozen-Deps`, which looked for lockfiles.](https://github.com/ossf/scorecard/blob/68e801f1f80a3cc3f36ff12cca4114b07e839a47/checks/checks.md#frozen-deps), before being renamed `Pinned-Dependencies`. Most of the check is focused on lockfiles and immutable hashes (OCI or GitHub), but there is some pinning recognized for package managers with immutable versions (Go and Nuget).

> In practice, people pin versions that actually exist

> Typically, we see folks publishing higher versions of their packages as time goes by, not lower versions.

I recognize this is an edge case we probably don't need to worry about. I was just highlighting one difference of lockfiles and version pinning, as I clarified the semantics of the check.

JonDouglas: > > It might be helpful to create documentation that explains the different levels/scenarios of "pinned" dependency checks, with clear examples for each scenario. For instance, detailing how an exact version match qualifies as "pinned," a lock file represents "locked," and so forth, to ensure clarity across all cases.
> 
> Most of our documentation uses "pinned" to refer to lock files, so we can certainly try to clarify "pinned" vs "locked".
> 
> I did some digging into the history of the check, as well as checking with relevant contributors. It started off as [`Frozen-Deps`, which looked for lockfiles.](https://github.com/ossf/scorecard/blob/68e801f1f80a3cc3f36ff12cca4114b07e839a47/checks/checks.md#frozen-deps), before being renamed `Pinned-Dependencies`. Most of the check is focused on lockfiles and immutable hashes (OCI or GitHub), but there is some pinning recognized for package managers with immutable versions (Go and Nuget).
> 
> > In practice, people pin versions that actually exist
> 
> > Typically, we see folks publishing higher versions of their packages as time goes by, not lower versions.
> 
> I recognize this is an edge case we probably don't need to worry about. I was just highlighting one difference of lockfiles and version pinning, as I clarified the semantics of the check.

thanks! idk who owns the naming but might be worth calling this out in a future meeting.

**should package managers provide basic support for the definition of pinned dependencies I provided earlier and what's in the docs? (i.e. exact version)?** or is the expectation to contribute to the "locked" only definition?

lock files aren't that prevalent in NuGet/.NET. this means that the .NET/NuGet ecosystem would score very low:

image

it seems like this check might benefit from partial points for exact versions and full points for locked/frozen/etc deps?

or perhaps a new check for "pinned" / "locked" package dependencies would be needed to take all this into account to separate from the CI/CD & docker checks?

checks/raw/pinned_dependencies.go
```diff
@@ -67,14 +73,87 @@ func PinningDependencies(c *checker.CheckRequest) (checker.PinningDependenciesDa
...
+
+func collectPostProcessNugetCPMDependencies(unpinnedNugetDependencies []*checker.Dependency,
+	postProcessingData *NugetPostProcessData,
+) {
+	packageVersions := postProcessingData.CpmConfig.PackageVersions
+
+	numFixedVersions, unfixedVersions := countFixedVersions(packageVersions)
+	// if all dependencies are fixed to specific versions, pin all dependencies
+	if numFixedVersions == len(packageVersions) {
+		pinAllNugetDependencies(unpinnedNugetDependencies)
```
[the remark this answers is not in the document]
jeffkl: CPM is all or nothing for an individual project.  By enabling it in `Directory.Packages.props` its on for all projects but individual projects can still opt-out of it.  Once enabled, you'll receive an error if any PackageReference declares a version or if any PackageReference has no corresponding PackageVersion item.

checks/raw/testdata/Directory.Pinned.packages.props
jeffkl: Normally, the transitive graph is resolved for you based on what the packages you consume have declared. To "pin" a transitive package to a higher version, you just add a direct dependency on it via `<PackageReference Include="SomeTransitivePackage" />`.  With CPM and transitive pinning, NuGet just does this for you during graph resolution without you having to declare the direct dependency.

balteravishay:
hey @spencerschrock, thanks for your feedback! any other questions about CPM or can we start the review of this PR?

spencerschrock:
> hey @spencerschrock, thanks for your feedback! any other questions about CPM or can we start the review of this PR?

Yep my questions are answered (thanks everyone), I'll get started on the review when I get some time.

checks/raw/pinned_dependencies.go
```diff
@@ -186,6 +259,17 @@ func countUnlocked(csprojFiles []dotnetCsprojLockedData) (int, string) {
 	return len(unlockedPaths), strings.Join(unlockedPaths, ", ")
 }
 
+func countFixedVersions(packages []properties.NugetPackage) (int, string) {
+	var unfixedVersions []string
```
spencerschrock, line 263: nit:  this counts unfixed versions

checks/raw/pinned_dependencies.go
```diff
@@ -67,14 +73,87 @@ func PinningDependencies(c *checker.CheckRequest) (checker.PinningDependenciesDa
...
+}
+
+func collectPostProcessNugetCPMDependencies(unpinnedNugetDependencies []*checker.Dependency,
+	postProcessingData *NugetPostProcessData,
+) {
+	packageVersions := postProcessingData.CpmConfig.PackageVersions
+
+	numFixedVersions, unfixedVersions := countFixedVersions(packageVersions)
+	// if all dependencies are fixed to specific versions, pin all dependencies
+	if numFixedVersions == len(packageVersions) {
```
spencerschrock, line 111: the variable is called `numFixedVersions` but it should really be `numUnfixedVersions` ? this logic needs fixed

internal/dotnet/properties/properties_test.go
```diff
@@ -0,0 +1,60 @@
...
+		{"fixed beta version", "10.1.1-beta", true},
+		{"fixed beta patch", "10.1.1-beta.1", true},
+		{"fixed version label zzz", "1.0.1-zzz", true},
+		{"fixed version RC with label", "1.0.1-rc.10", true},
+		{"fixed version RC with label 2", "1.0.1-rc.2", true},
+		{"fixed version with label open", "1.0.1-open", true},
+		{"fixed version alpha", "1.0.1-alpha2", true},
+		{"fixed version RC with label aaa", "1.0.1-aaa", true},
+		{"fixed version range", "[1.0]", true},
+		{"version as variable", "$(ComponentDetectionPackageVersion)", true},
```
spencerschrock, line 38: can you clarify the version as a variable result? do we know it's fixed?

internal/dotnet/properties/properties_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func TestIsValidFixedVersion(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name    string
+		version string
+		isFixed bool
+	}{
+		{"fixed version", "10.1.1", true},
+		{"fixed beta version", "10.1.1-beta", true},
```
spencerschrock, line 29: according to [your link in the regex comment](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#version-ranges), this is equivalent to "Minimum version, inclusive"?

so this conflicts with some of your tests below?

balteravishay: fixed this by allowing only "Exact version match" using [] brackets.

internal/dotnet/properties/properties_test.go
balteravishay: done. this now is only deemed as fixed if it is surrounded by [] brackets.

checks/raw/pinned_dependencies.go
balteravishay: fixed, thank you, this is a great catch!

checks/raw/pinned_dependencies.go
balteravishay: fixed

checks/raw/pinned_dependencies.go
```diff
@@ -67,14 +73,87 @@ func PinningDependencies(c *checker.CheckRequest) (checker.PinningDependenciesDa
...
+		for i := range unpinnedNugetDependencies {
+			(unpinnedNugetDependencies)[i].Remediation.Text = (unpinnedNugetDependencies)[i].Remediation.Text +
+				": some of dependency versions are not fixes to specific versions: " + unfixedVersions
+		}
+	}
+}
+
+func retrieveNugetCentralPackageManagement(c *checker.CheckRequest, nugetPostProcessData *NugetPostProcessData) error {
+	if err := fileparser.OnMatchingFileContentDo(c.RepoClient, fileparser.PathMatcher{
+		Pattern:       "Directory.*.props",
```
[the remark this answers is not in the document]
balteravishay: it does, but the product group decided to allow this use other types of props file, since folks (and older versions of nuget) use/allow usage of other props file to declare dependencies such as Directory.Build.pros and others.

checks/raw/pinned_dependencies.go
```diff
@@ -67,14 +73,87 @@ func PinningDependencies(c *checker.CheckRequest) (checker.PinningDependenciesDa
...
+		if err := retrieveCsprojConfig(c, &nugetPostProcessData); err != nil {
+			return err
+		}
+		if nugetPostProcessData.CpmConfig.IsCPMEnabled {
+			collectPostProcessNugetCPMDependencies(unpinnedDependencies, &nugetPostProcessData)
+		} else {
+			collectPostProcessNugetCsprojDependencies(unpinnedDependencies, &nugetPostProcessData)
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
balteravishay: fixed, thanks for the best-practice!

spencerschrock (approved):
Thanks all!

(ignore any unit test failures when I update/merge, we had a slight flake introduced in a previous PR that should be fixed soon)

PR #4415, opened 2024-11-18
:bug: Add a workaround for organisations in GitLab

verm:
This fixes Contributors.

Currently organisations are an experimental features in GitLab this change makes it use the users email domain as a pseudo organisation.  The domains are anonymised as these are printed as part of the report and shouldn't be exposed.

It's uncertain if the uptake in organisation will be large considering GitLab users are used to using groups for this purpose.

Fixes a bug where Contributors were always set to 0 I did not see an open issue for this.

Contributors will always be 0.

NONE

What I've done here is used the email domains as a pseudo orginisation and anonymomised it using the slice index as `gitlab-N`.  This lets Contributors give somewhat useful data and in the future when Orginisation is no longer experimental we'll be able to remove this.  I found this to be the least invasive method of handling the situation.

NONE

```release-note
Fix Contributors to return somewhat useful data for GitLab.
```

### DIFF
--- a/clients/gitlabrepo/contributors.go
+++ b/clients/gitlabrepo/contributors.go
@@ -84,7 +84,34 @@ func (handler *contributorsHandler) retrieveUsers(queryName string) ([]*gitlab.U
 	return users, nil
 }
 
+// Orginisation is an experimental feature in GitLab. Takes a slice
+// pointer to maintain list of domains and email returns an anonymised
+// string with the slice index as a unique identifier.
+// Pointer is required as we are modifying the slice length.
+func getOrginisation(orgDomain *[]string, orgReal string) string {
+	// Some repositories date before email was widely used.
+	emailSplit := strings.Split(orgReal, "@")
+	var orgName string // Orginisation name
+	if len(emailSplit) > 1 {
+		orgName = emailSplit[1]
+	} else {
+		orgName = emailSplit[0] // Not an "email"
+	}
+
+	*orgDomain = append(*orgDomain, orgName)
+
+	// Search for domain and use index as unique marker
+	for i, domain := range *orgDomain {
+		if domain == orgName {
+			return fmt.Sprint("GitLab-", i)
+		}
+	}
+	return "GitLab"
+}
+
 func (handler *contributorsHandler) setup() error {
+	var orgDomain []string // Slice of email domains
+
 	handler.once.Do(func() {
 		if !strings.EqualFold(handler.repourl.commitSHA, clients.HeadSHA) {
 			handler.errSetup = fmt.Errorf("%w: ListContributors only supported for HEAD queries",
@@ -126,9 +153,17 @@ func (handler *contributorsHandler) setup() error {
 				user = users[0]
 			}
 
+			// In case someone is using the experimental feature.
+			var orgName string
+			if user.Organization == "" {
+				orgName = getOrginisation(&orgDomain, contrib.Email)
+			} else {
+				orgName = user.Organization
+			}
+
 			contributor := clients.User{
 				Login:            contrib.Email,
-				Companies:        []string{user.Organization},
+				Companies:        []string{orgName},
 				NumContributions: contrib.Commits,
 				ID:               int64(user.ID),
 				IsBot:            user.Bot,


clients/gitlabrepo/contributors.go
```diff
@@ -84,7 +84,34 @@ func (handler *contributorsHandler) retrieveUsers(queryName string) ([]*gitlab.U
...
+	if len(emailSplit) > 1 {
+		orgName = emailSplit[1]
+	} else {
+		orgName = emailSplit[0] // Not an "email"
+	}
+
+	*orgDomain = append(*orgDomain, orgName)
+
+	// Search for domain and use index as unique marker
+	for i, domain := range *orgDomain {
```
raghavkaul, line 104: Consider `slices.Index`.

clients/gitlabrepo/contributors.go
```diff
@@ -84,7 +84,34 @@ func (handler *contributorsHandler) retrieveUsers(queryName string) ([]*gitlab.U
 	return users, nil
 }
 
+// Orginisation is an experimental feature in GitLab. Takes a slice
+// pointer to maintain list of domains and email returns an anonymised
+// string with the slice index as a unique identifier.
+// Pointer is required as we are modifying the slice length.
+func getOrginisation(orgDomain *[]string, orgReal string) string {
```
raghavkaul, line 91: Should be in `clients/gitlabrepo/commits.go` since we're dealing with commit email addresses.

raghavkaul (commented):
Email domains aren't indicative of an organization that a contributor belongs to. Is there any other API feature we can use to get contributor organizations?

clients/gitlabrepo/contributors.go
verm: I did not put that there since this should only be a temporary measure and this is specific to GitLab if you want me to move it after happy to do so.

verm:
> Email domains aren't indicative of an organization that a contributor belongs to. Is there any other API feature we can use to get contributor organizations?

Orginisations in GitLab are an experimental feature with no interface other than the API to get to it.  There is no other unique identifier.  I agree that it's not ideal nore a good indicator the only other option is to remove it completely but this at least gives us *something* that may be useful.

I did it this way as it seemed the least invasive to support this until Orginisation is settled as a public feature.

raghavkaul:
We use the Users API to get the Organization, not the Organizations API. As I understand it, it's just the API that's experimental, and Organizations are part of GitLab to stay.

verm:
> We use the Users API to get the Organization, not the Organizations API. As I understand it, it's just the API that's experimental, and Organizations are part of GitLab to stay.

Yes they are here to stay the way GitLab wants you to set them up is using groups which don't exist in GitHub.  If you go to GitLab docs they just want you to setup groups -- this is not what the API being used is it's an entirely new concept for Orginizations similar to how GitHub has them.

The concept of 'orginization' does not exist in GitLab yet it is only exposed via an API and it's all experimental.

This is why in nearly every case using this API right now is going to return an empty string.  The web UI isn't even enabled for it:
  * https://gitlab.com/gitlab-org/gitlab/-/merge_requests/122866

To summerise: The only way you're going to get a user with an orgnization is if you enable experimental flags and use the API directly.  It won't show up anywhere on the web UI nore can you edit it.  You can also see here that all of the feature flags are disabled by default for orginization:

  * https://docs.gitlab.com/ee/user/feature_flags.html

verm:
> e.g if a user is in group `foo/bar/baz`, then they're a part of "organization" foo?

The problem with this is groups are used for organising permissions.  While some may use it in that way there are plenty that don't look at RTEMS for example:

  * https://gitlab.rtems.org/dashboard/groups

I think it will be years until Orgs in GitLab will be used in the same way as GitHub.  I'd argue at least email domains being no worse than using groups but my original point is valid.  Orgs are coming to GitLab and this is just a temp measure and the ones that causes the least churn since we are running these queries anyway.

spencerschrock:
Going to discuss this alternative to Contributors at the weekly sync, as the use of emails also came up in the AzureDevops client

verm:
> Going to discuss this alternative to Contributors at the weekly sync, as the use of emails also came up in the AzureDevops client

Apologies for the delay here I'll be on the meetings again the week of the 6th we can talk about it then happy to make whatever changes required here to get it working so I can move on to the next issues.

clients/gitlabrepo/contributors.go
ashearin: The example found here is good: https://pkg.go.dev/slices#Index

ashearin: better example: https://goplay.tools/snippet/0CXt57McB3Z

verm:
This has been closed is there anything else I need to do to get this change in?  I was not able to attend meetings but I will start attending again consistently soonish.  Thank you.

PR #4423, opened 2024-11-27
:sparkles: Allow incomplete local checks 

### LINKED ISSUE #3832: Feature: Allow "incomplete" checks for --local repos
Running scorecard with `--local` only allows it to run 6 checks. Of the 12 that aren't run, I propose that 6 be allowed to run, even if the results may be incomplete when compared with the upstream:
- Fuzzing
- License
- Maintained
- Packaging
- SAST
- Security-Policy

- Fuzzing only requires a non-local repo for OSS-Fuzz, all other fuzzers are detected using files in the repository itself. With `--local`, we simply don't look for OSS-Fuzz.
- License tries to use the platform's API, but it already has fallback methods to scan the repository itself. With `--local`, we simply use the fallbacks.
- Maintained can score a `--local` repo using just the commit activity: use `git log` to detect the frequency of commits and use the date of the root commit to determine whether the repo is 90+ days old.
- Packaging currently looks for release workflows and then checks to see if they've ever passed. With `--local`, it could skip that later step.
- SAST looks for workflows with CodeQL or SonarCloud and sees if they're run on PRs (or something like that?). With `--local`, it could simply reward for having a SAST workflow.
- Security-Policy looks for a policy in the repo or the org's .github repo. With `--local`, it could just look in the repo.

In all of these cases, Scorecard should log that the results are potentially incomplete/unrepresentative of the upstream repository.

lharrison13:

Certain checks are currently not run when using the `local` option

Now allows incomplete versions of the following checks to be run when using the `local` option

Fuzzing
License
Packaging
SAST
Security-Policy

Fixes #3832 

- For the packaging check I just ran both the github and gitlab checks since its not really clear which one it might be. I also exported the LocalDirClient so it can be used in the type assertion.
- I did not enable the maintained check as per https://github.com/ossf/scorecard/issues/3832#issuecomment-1918132684 and https://github.com/ossf/scorecard/issues/1709

the `release-note`

```release-note
Enabled Fuzzing, License, Packaging, SAST, and Security-Policy checks when using --local option
```
Screenshot 2024-11-26 at 11 59 38 PM


checks/raw/fuzzing.go
```diff
@@ -208,6 +209,10 @@ func Fuzzing(c *checker.CheckRequest) (checker.FuzzingData, error) {
 
 	langs, err := c.RepoClient.ListProgrammingLanguages()
 	if err != nil {
+		// ListProgrammingLanguages is unsupported by local client so just return
+		if errors.Is(err, clients.ErrUnsupportedFeature) {
+			return checker.FuzzingData{Fuzzers: detectedFuzzers}, nil
+		}
```
spencerschrock, line 215: we miss out on most of the file-based detection if we return early here. (only clusterfuzzlite is detected).

I wonder if it makes sense to scan for every language if the client doesn't have the info

checks/raw/github/packaging.go
```diff
@@ -73,7 +75,12 @@ func Packaging(c *checker.CheckRequest) (checker.PackagingData, error) {
 
 		runs, err := c.RepoClient.ListSuccessfulWorkflowRuns(filepath.Base(fp))
 		if err != nil {
-			return data, fmt.Errorf("Client.Actions.ListWorkflowRunsByFileName: %w", err)
+			// appending empty run for localdir client
```
spencerschrock, line 78: lets clarify the purpose of adding the empty run in the comment. maybe something like (but open to suggestions):

```go
// assume the workflow will run for localdir client
```

checks/raw/security_policy.go
```diff
@@ -74,7 +74,9 @@ func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error)
 	case errors.Is(err, sce.ErrRepoUnreachable), errors.Is(err, clients.ErrUnsupportedFeature):
 		break
 	default:
-		return checker.SecurityPolicyData{}, err
+		if !errors.Is(err, clients.ErrUnsupportedFeature) {
+			return checker.SecurityPolicyData{}, err
+		}
```
spencerschrock, line 79: does the case above `default` not already handle `clients.ErrUnsupportedFeature`?

clients/localdir/client.go
```diff
@@ -34,12 +34,12 @@ import (
...
 
 var (
-	_                clients.RepoClient = &localDirClient{}
+	_                clients.RepoClient = &LocalDirClient{}
 	errInputRepoType                    = errors.New("input repo should be of type repoLocal")
 )
 
 //nolint:govet
-type localDirClient struct {
+type LocalDirClient struct {
```
spencerschrock, line 42: nit: now that we're exporting this type, lets make the name match the others:
```go
type Client struct {
```

so that users will write:
```go
case *localdir.Client:
```

instead of having the repetition
```go
case *localdir.LocalDirClient:
```

policy/policy_test.go
```diff
@@ -333,7 +333,7 @@ func TestGetEnabled(t *testing.T) {
 			name:                  "request types limit enabled checks",
 			argsChecks:            []string{},
 			requiredRequestTypes:  []checker.RequestType{checker.FileBased, checker.CommitBased},
-			expectedEnabledChecks: 5, // All checks which are FileBased and CommitBased
+			expectedEnabledChecks: 7, // All checks which are FileBased and CommitBased
```
spencerschrock, line 336: hmm, i thought this PR added FileBased to 3 checks that weren't before? Any idea what's going on here?

spencerschrock (commented):
Thanks for sending a PR, overall looks pretty good! Left a few comments

policy/policy_test.go
lharrison13: I think that test is for FileBased **and** Commit Based. I added 3 checks that were FileBased but two originally had CommitBased.

checks/raw/security_policy.go
lharrison13: I honestly have no clue why i put that there

spencerschrock (approved):
Thank you!

I ran this through my local repo with `scdiff`, and the changes are working as expected, with the 5 new checks showing up.

Issue #4177, opened 2024-06-15
Feature: Support for Azure DevOps


JamieMagee:
**Is your feature request related to a problem? Please describe.**
Scorecards currently supports GitHub and GitLab as source code platforms, but Azure DevOps could also benefit from Scorecard

**Describe the solution you'd like**
Support for most of the checks that are supported on GitHub and GitLab to also be supported on Azure DevOps

**Describe alternatives you've considered**
Not supporting Azure DevOps? 😅

**Additional context**
I am interested in implementing this myself, but it'll likely be across multiple smaller PRs.

JamieMagee:
Not stale. I'm working on this.

JamieMagee:
Remaining unimplemented:
- `GetOrgRepoClient`
  - This is a GitHub-only feature
- `ListLicenses`
  - Azure DevOps doesn't have a license detection feature. Thankfully the `License` check falls back to file-based detection
- `ListReleases`
  - Azure DevOps doesn't have the concept of 'a release' in the same way that GitHub or GitLab do
  - In Azure DevOps a release is a pipeline
  - Scorecard expects a 'release' to have a git tag and release assets

I believe the Azure DevOps client initial implementation is complete.

JamieMagee:
Still need to setup integration tests in https://dev.azure.com/openssf-scorecard/

justaugustus:
From @JamieMagee in https://github.com/ossf/scorecard/pull/4994/changes#r3036131267:

> More of a question: what would graduating ADO from `scorecard.experimental` look like?

PR #4456, opened 2024-12-13
:sparkles: implement more of the Azure DevOps client

JamieMagee:

Includes:
  - `GetBranch`
  - `GetSuccessfulWorkflowRuns`
  - `ListCheckRunsForRef`
  - `ListStatuses`
  - `ListWebhooks`
  - `SearchCommits`

Also, includes comments about methods which can never be implemented:
  - `GetOrgRepoClient`
    - Org repository, AKA the `<org>/.github` repository, is a GitHub-specific feature
  - `ListLicenses`
    - Azure DevOps doesn't have a license detection feature. Thankfully, the License check falls back to file-based detection.

Still need to implement:
  - `ListReleases`
    - Needs a little more investigation to line up the Azure DevOps implementation with what Scorecard expects

Missing implementations for the methods listed above

Next step of #4177

the `release-note`

```release-note
implement more of the Azure DevOps client
```


clients/azuredevopsrepo/zip.go
```diff
@@ -110,23 +110,30 @@ func (z *zipHandler) getZipfile() error {
...
 		queryParams.Add("versionDescriptor.version", z.repourl.defaultBranch)
 	} else {
 		queryParams.Add("versionDescriptor.versionType", "commit")
 		queryParams.Add("versionDescriptor.version", z.repourl.commitSHA)
 	}
 
-	parsedURL, err := url.Parse(baseURL + "?" + queryParams.Encode())
+	parsedURL := fmt.Sprintf("%s?%s", baseURL, queryParams.Encode())
+
+	req, err := z.client.CreateRequestMessage(
```
JamieMagee, line 123: The previous implementation didn't add the `Authorization` header.

clients/azuredevopsrepo/commits.go
```diff
@@ -205,18 +215,102 @@ func (handler *commitsHandler) listPullRequests() (map[string]clients.PullReques
...
+	if err := c.setup(); err != nil {
 		return time.Time{}, fmt.Errorf("error during commitsHandler.setup: %w", err)
 	}
 
-	return handler.firstCommitCreatedAt, nil
+	return c.firstCommitCreatedAt, nil
+}
+
+func (c *commitsHandler) listStatuses(ref string) ([]clients.Status, error) {
+	matched, err := regexp.MatchString("^[0-9a-f]{5,40}$", ref)
```
spencerschrock, line 235: what about `A-F`? also most of our other regexes are declared in the form of package vars. 

In this other example, it expects a full 40, but doesn't necessarily have to be.
```
var gitCommitHashRegex = regexp.MustCompile(`^[a-fA-F0-9]{40}$`)
```

clients/azuredevopsrepo/commits_test.go
```diff
@@ -0,0 +1,139 @@
...
+						},
+					},
+				}, nil
+			},
+			want:    []clients.Status{},
+			wantErr: false,
+		},
+		{
+			name: "head - single status",
+			ref:  "main",
```
spencerschrock, line 87: technically head is "HEAD"? not sure if it matters here

spencerschrock:
How do Branch-Protection and SAST look now?

clients/azuredevopsrepo/builds.go
```diff
@@ -0,0 +1,108 @@
...
+	) (*build.GetDefinitionsResponseValue, error)
+	fnGetBuilds func(
+		ctx context.Context,
+		args build.GetBuildsArgs,
+	) (*build.GetBuildsResponseValue, error)
+)
+
+func (b *buildsHandler) init(ctx context.Context, repourl *Repo) {
+	b.ctx = ctx
+	b.once = new(sync.Once)
```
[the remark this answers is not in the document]
JamieMagee: No. I think I had gotten into a habit of copy-pasting the same struct for each new handler.

clients/azuredevopsrepo/commits.go
JamieMagee: I didn't realize regex were case-sensitive by default in Go.

As for package vars versus local vars, is there any reason why one over the other if this is the only use?

spencerschrock:
> SAST works perfectly

I'll note this is falling back to the github detection of the `.github/workflows`, and not through your added `vstfs:///build/build` string (which will be more likely for AZDO projects I assume?)

> The issue with ListReleases on Azure DevOps is it doesn't have a concept of a release in the same way that GitHub and GitLab do

I'm curious if we should just continue on the ListReleases `ErrUnsupportedCheck` so we get branch protection data for the default branch at least

clients/azuredevopsrepo/commits.go
spencerschrock: > As for package vars versus local vars, is there any reason why one over the other if this is the only use?

Mainly about when it gets compiled. If it's a package var, it's one time when the package is imported. If it's local to the function, it only gets compiled when used, but every time the function is called.

We can keep it local

clients/azuredevopsrepo/commits_test.go
spencerschrock: Sorry, I could've been more clear. I wasn't talking about `head` vs `HEAD`. But instead the ref is `main` instead of `HEAD`, and that gets passed into `c.listStatuses` and I thought the code does a comparison to `HEAD`, but I think it's in a different function

```go
if c.repourl.commitSHA == headCommit {
```

JamieMagee: Yeah, I think the solution here is actually to change the name of the test. The `head` it's referring to is `refs/heads/main`, not `HEAD`.

JamieMagee:
> I'll note this is falling back to the github detection of the `.github/workflows`, and not through your added `vstfs:///build/build` string (which will be more likely for AZDO projects I assume?)

That's an issue with the check itself than with the Azure DevOps client. There are definitely a couple of checks like that.

The SAST check specifically would need to be updated to look for all YAML files in the repository, because Azure Pipelines files can be in any arbitrary location in the repository, and look for the `AdvancedSecurity-Codeql-Analyze` task. See [Code scanning for GitHub Advanced Security for Azure DevOps](https://learn.microsoft.com/en-us/azure/devops/repos/security/github-advanced-security-code-scanning?view=azure-devops) for more information.

> I'm curious if we should just continue on the ListReleases `ErrUnsupportedCheck` so we get branch protection data for the default branch at least

Yes, that would be one approach. But I don't want to expand the scope of this PR too much more.

spencerschrock:
> The SAST check specifically would need to be updated to look for all YAML files in the repository, because Azure Pipelines files can be in any arbitrary location in the repository, and look for the `AdvancedSecurity-Codeql-Analyze` task

Whoops, I had meant CI-Tests, as that's what you added the `vstfs:///build/build` string for.

spencerschrock (approved):

PR #4499, opened 2025-01-22
✨ New Probe: Memory safety

balteravishay:

This PR provides the foundation for addressing issue #3736 by adding a new probe that checks if the code uses non memory safe practices for the repository languages.
The goal is to automate the detection of as many of the practices that the memory safety SIG provides under the [Best Practices - Memory-Safe By Default Languages](https://github.com/ossf/Memory-Safety/blob/main/docs/best-practice-memory-safe-by-default-languages.md) and the [Best Practices - Non Memory-Safe By Default Languages](https://github.com/ossf/Memory-Safety/blob/main/docs/best-practice-non-memory-safe-by-default-languages.md) guides.

Today scorecard does not detect memory safe practices in it's core features or in any of the probes.

Probe detects the following:

- for golang it detects if the code imports the unsafe package and points to the locations where it is used.
- for c# it detects if the projects allow for unsafe blocks which is a requirement for any project that would use any form of .Net [unsafe code, pointer types, and function pointers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code)

This code change addresses issue #3736 but does not close it (to be discussed)

This code change and the implementation of it were discussed in scorecard community calls with @spencerschrock 

the `release-note`

```release-note
Added independent probe that checks for ecosystem specific non-memory safety practices in the codebase and flags them.
```


balteravishay:
here are some repos to test the code:

go run main.go  --repo github.com/microsoft/midi  --probes memorysafe --format probe (.net unsafe code)
go run main.go  --repo github.com/microsoft/winget-cli  --probes memorysafe --format probe (.net safe code)
go run main.go  --repo github.com/pkujhd/goloader  --probes memorysafe --format probe (go unsafe code)
go run main.go  --repo github.com/ossf/scorecard  --probes memorysafe --format probe (go safe code)

spencerschrock:
Hoping to have some time to look at this tomorrow. Quick question about organization:

> Added independent probe that checks for ecosystem specific non-memory safety practices

Probes are usually one specific behavior. In this case, the probe is concerned with looking for `unsafe` usage across ecosystems. I'm not sure if your intention was to use this probe for the go race detector follow up too, but generally different behavior tests get broken up into different probes.

probes/memorysafe/impl.go
```diff
@@ -0,0 +1,229 @@
...
+		// panic if it is not correct type
+		panic(fmt.Sprintf("expected type findings, got %v", reflect.TypeOf(args[0])))
+	}
+	fset := token.NewFileSet()
+	f, err := parser.ParseFile(fset, "", content, parser.ImportsOnly)
+	if err != nil {
+		dl, ok := args[1].(checker.DetailLogger)
+		if !ok {
+			// panic if it is not correct type
+			panic(fmt.Sprintf("expected type checker.DetailLogger, got %v", reflect.TypeOf(args[1])))
```
spencerschrock, line 152: we should move the type checking out of the `err != nil` block to make the panics when passing wrong types more obvious.

probes/memorysafe/impl.go
```diff
@@ -0,0 +1,229 @@
...
+			Text: fmt.Sprintf("malformed golang file: %v", err),
+		})
+		return true, nil
+	}
+	for _, i := range f.Imports {
+		if i.Path.Value == `"unsafe"` {
+			found, err := finding.NewWith(fs, Probe,
+				"Golang code uses the unsafe package", &finding.Location{
+					Path: path,
+				}, finding.OutcomeFalse)
```
spencerschrock, line 165: we have much more info than path available if we use : 
https://pkg.go.dev/go/ast#ImportSpec.Pos and https://pkg.go.dev/go/ast#ImportSpec.End

Something like:
```
position := fset.Position(i.Pos())
```

and then we have access to line info too and can set `finding. LineStart`. https://pkg.go.dev/go/token#Position

probes/memorysafe/impl.go
```diff
@@ -0,0 +1,229 @@
...
+
+	if err := fileparser.OnMatchingFileContentDo(client.RepoClient, fileparser.PathMatcher{
+		Pattern:       "*.go",
+		CaseSensitive: false,
+	}, goCodeUsesUnsafePackage, &findings, client.Dlogger); err != nil {
+		return nil, err
+	}
+	if len(findings) == 0 {
+		found, err := finding.NewWith(fs, Probe,
+			"Golang code does not use the unsafe package", nil, finding.OutcomeTrue)
```
spencerschrock, line 131: we could also move this (and the c# version) to one combined top level check in `Run` before the return,depending on the semantics we want the probe to have.

probes/memorysafe/impl.go
```diff
@@ -0,0 +1,229 @@
...
+	if !ok {
+		// panic if it is not correct type
+		panic(fmt.Sprintf("expected type findings, got %v", reflect.TypeOf(args[0])))
+	}
+	unsafe, err := csproj.IsAllowUnsafeBlocksEnabled(content)
+	if err != nil {
+		dl, ok := args[1].(checker.DetailLogger)
+		if !ok {
+			// panic if it is not correct type
+			panic(fmt.Sprintf("expected type checker.DetailLogger, got %v", reflect.TypeOf(args[1])))
```
spencerschrock, line 209: same comment about moving type check

spencerschrock (commented):
didn't quite go through the tests yet, since the semantics of the probe will probably change through this discussion.

balteravishay:
> Hoping to have some time to look at this tomorrow. Quick question about organization:
> 
> > Added independent probe that checks for ecosystem specific non-memory safety practices
> 
> Probes are usually one specific behavior. In this case, the probe is concerned with looking for `unsafe` usage across ecosystems. I'm not sure if your intention was to use this probe for the go race detector follow up too, but generally different behavior tests get broken up into different probes.

the intention of this probe is to save the end user the need to know which best practices they are expected to have in the repo, and to give them a kind of a one stop shop for finding out if there's anything they're missing. Our goal was to try and implement the different practices that are suggested in the memory safety guides for [memory-safe-by-default](https://github.com/ossf/Memory-Safety/blob/main/docs/best-practice-memory-safe-by-default-languages.md) and [non-memory-safe-by-default](https://github.com/ossf/Memory-Safety/blob/main/docs/best-practice-non-memory-safe-by-default-languages.md) guides into this single probe.
The concern with breaking that up for different probes is that user will have to run a bunch of different probes, per ecosystem, per guideline, just to get the overall view.
hope this makes sense.

probes/memorysafe/def.yml
```diff
@@ -0,0 +1,41 @@
...
+  text:
+    - Visit the OpenSSF Memory Safety SIG guidance on how to make your project memory safe.
+    - Guidance for [Memory-Safe By Default Languages](https://github.com/ossf/Memory-Safety/blob/main/docs/best-practice-memory-safe-by-default-languages.md)
+    - Guidance for [Non Memory-Safe By Default Languages](https://github.com/ossf/Memory-Safety/blob/main/docs/best-practice-non-memory-safe-by-default-languages.md)
+ecosystem:
+  languages:
+    - go
+    - c#
+  clients:
+    - github
```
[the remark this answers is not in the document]
balteravishay: fixed.

probes/memorysafe/impl.go
```diff
@@ -0,0 +1,229 @@
...
+	prominentLangs := getRepositoryLanguageChecks(raw)
+	findings := []finding.Finding{}
+
+	for _, lang := range prominentLangs {
+		for _, langFunc := range lang.funcPointers {
+			if langFunc == nil {
+				raw.Dlogger.Warn(&checker.LogMessage{
+					Text: fmt.Sprintf("no function pointer found for language %s", lang.Desc),
+				})
+			}
```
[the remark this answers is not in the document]
balteravishay: agree, removed!

probes/memorysafe/impl.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return findings, Probe, nil
+}
+
+func getRepositoryLanguageChecks(raw *checker.CheckRequest) []languageMemoryCheckConfig {
+	langs, err := raw.RepoClient.ListProgrammingLanguages()
+	if err != nil {
+		raw.Dlogger.Warn(&checker.LogMessage{
+			Text: fmt.Sprintf("RepoClient retured error for ListProgrammingLanguages: %v", err),
+		})
```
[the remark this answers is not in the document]
balteravishay: (y)

probes/memorysafe/impl.go
balteravishay: (y)

probes/memorysafe/impl.go
balteravishay: added linestart. do you think any other information should be added?

spencerschrock:
> The concern with breaking that up for different probes is that user will have to run a bunch of different probes, per ecosystem, per guideline, just to get the overall view.

On the other hand, our goal for probes is that consumers are concerned with specific behavior and by having them do one thing it's easier to build policy.

spencerschrock:
> ok, so just for clarity, for this PR do you want to separate also to two different probes (one for c# and one for go)?

No, for example the `fuzzed` probe is the same regardless of ecosystem. For both C# and Go, the behavior being tested is the same: is the project enabling/using some feature to step around the normally memory safe (type) features.

balteravishay:
> > ok, so just for clarity, for this PR do you want to separate also to two different probes (one for c# and one for go)?
> 
> No, for example the `fuzzed` probe is the same regardless of ecosystem. For both C# and Go, the behavior being tested is the same: is the project enabling/using some feature to step around the normally memory safe (type) features.

@spencerschrock, this is now fixed to only test unsafe code blocks.

probes/memorysafe/def.yml
```diff
@@ -0,0 +1,41 @@
...
+id: memorysafe
+lifecycle: experimental
+short: Flags non memory safe practices in this project.
+motivation: >
+  Memory safety in software should be considered a continuum, rather than being binary.
+  This probe does not consider a specific ecosystem more or less memory safe than another, but rather tries to surface non memory safe code or practices in the project, in the context of the ecosystems it is using.
+implementation: >
+  The probe is ecosystem-specific and tries to flag non memory safe code blocks in the project by looking at the code and practices used in the project.
+  It may look for specific memory safety practices, such as the use tools or non memory-safe patterns and code.
+  The probe supports multiple checks for each ecosystem, and the outcome is based on the results of these checks.
```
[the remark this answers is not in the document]
balteravishay: fixed these.

spencerschrock:
> @spencerschrock, this is now fixed to only test unsafe code blocks.

Thanks will take a look today or tomorrow.

probes/unsafeblock/impl.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func getLanguageChecks(raw *checker.CheckRequest) ([]languageMemoryCheckConfig, error) {
+	langs, err := raw.RepoClient.ListProgrammingLanguages()
+	if err != nil {
+		return nil, fmt.Errorf("cannot get langs of repo: %w", err)
+	}
+	if len(langs) == 0 {
+		return []languageMemoryCheckConfig{}, nil
+	}
```
spencerschrock, line 92: nit: this seems redundant as not entering the loop would produce the same result.

probes/unsafeblock/impl.go
```diff
@@ -0,0 +1,209 @@
...
+		})
+		return true, nil
+	}
+	for _, i := range f.Imports {
+		if i.Path.Value == `"unsafe"` {
+			lineStart := uint(fset.Position(i.Pos()).Line)
+			found, err := finding.NewWith(fs, Probe,
+				"Golang code uses the unsafe package", &finding.Location{
+					Path: path, LineStart: &lineStart,
+				}, finding.OutcomeFalse)
```
spencerschrock, line 153: flagging outcome flip

probes/unsafeblock/impl_test.go
```diff
@@ -0,0 +1,514 @@
...
+				return tt.repoLanguages, nil
+			}).AnyTimes()
+			mockRepoClient.EXPECT().ListFiles(gomock.Any()).DoAndReturn(func(predicate func(string) (bool, error)) ([]string, error) {
+				return tt.filenames, nil
+			}).AnyTimes()
+			mockRepoClient.EXPECT().GetFileReader(gomock.Any()).DoAndReturn(func(file string) (io.ReadCloser, error) {
+				return os.Open(file)
+			}).AnyTimes()
+			raw.RepoClient = mockRepoClient
+			raw.Dlogger = checker.NewLogger()
```
spencerschrock, line 441: normally in tests we use the TestDetailLogger
https://github.com/ossf/scorecard/blob/3b42b6e7d8c66543d10c7f5ac5d3ecc0d1e56e5a/utests/utlib.go#L38-L41

probes/unsafeblock/impl_test.go
```diff
@@ -0,0 +1,514 @@
...
+func Test_Run_Error_ListProgrammingLanguages(t *testing.T) {
+	t.Parallel()
+	ctrl := gomock.NewController(t)
+	raw := &checker.CheckRequest{}
+	mockRepoClient := mockrepo.NewMockRepoClient(ctrl)
+	mockRepoClient.EXPECT().ListProgrammingLanguages().DoAndReturn(func() ([]clients.Language, error) {
+		return nil, fmt.Errorf("error")
+	}).AnyTimes()
+	raw.RepoClient = mockRepoClient
+	raw.Dlogger = checker.NewLogger()
```
spencerschrock, line 463: same comment about logger

probes/unsafeblock/impl_test.go
```diff
@@ -0,0 +1,514 @@
...
+			raw := &checker.CheckRequest{}
+			mockRepoClient := mockrepo.NewMockRepoClient(ctrl)
+			mockRepoClient.EXPECT().ListProgrammingLanguages().DoAndReturn(func() ([]clients.Language, error) {
+				return tt.repoLanguages, nil
+			}).AnyTimes()
+			mockRepoClient.EXPECT().ListFiles(gomock.Any()).DoAndReturn(func(predicate func(string) (bool, error)) ([]string, error) {
+				return nil, fmt.Errorf("error")
+			}).AnyTimes()
+			raw.RepoClient = mockRepoClient
+			raw.Dlogger = checker.NewLogger()
```
spencerschrock, line 503: logger

spencerschrock (approved):
a few minor things to fix

probes/unsafeblock/def.yml
```diff
@@ -0,0 +1,44 @@
...
+implementation: >
+  The probe is ecosystem-specific and will surface non memory safe practices in the project by identifying unsafe code blocks.
+  Unsafe code blocks are supported in rust, go, c#, and swift, but only go and c# are supported by this probe at this time:
+  - for go the probe will look for the use of the `unsafe` include directive.
+  - for c# the probe will look at the csproj and identify the use of the `AllowUnsafeBlocks` property.
+outcome:
+  - For supported ecosystem, the probe returns OutcomeFalse per unsafe block.
+  - If the project has no unsafe blocks, the probe returns OutcomeTrue.
+remediation:
+  onOutcome: False
```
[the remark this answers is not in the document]
balteravishay: reversed the True/False returns of the probe accordingly

balteravishay:
@spencerschrock fixed all comments here.

spencerschrock:
@balteravishay had to tweak this to fix an issue similar to the one in #4531 . Parse errors now return a finding.OutcomeError, similar to Pinned-Dependencies

balteravishay:
Thank you very much @spencerschrock. Patch looks good to me.

PR #4533, opened 2025-02-20
🐛 do not allow unsupported `ListReleases` to fail a run

JamieMagee:

This check allows clients which currently do not support `ListReleases` to complete a Scorecard run. Currently `azuredevopsrepo`, `localdir`, and `git` do not support `ListReleases`. While I am working on support for Azure DevOps, I suspect that `localdir` and `git` may never support `ListReleases`.

Currently, if a client does not support `ListReleases`, the Scorecard run fails with an error like:

```
Error: check runtime error: Branch-Protection: internal error: unsupported feature
```

This is the case for `Branch-Protection` and `Signed-Releases`.

NONE

This is similar to the fix done in #4423

the `release-note`

```release-note
Prevent unimplemented ListReleases from failing a run
```

### DIFF
--- a/checks/raw/branch_protection.go
+++ b/checks/raw/branch_protection.go
@@ -15,6 +15,7 @@
 package raw
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"regexp"
@@ -66,6 +67,9 @@ func BranchProtection(cr *checker.CheckRequest) (checker.BranchProtectionsData, 
 	// Get release branches.
 	releases, err := c.ListReleases()
 	if err != nil {
+		if errors.Is(err, clients.ErrUnsupportedFeature) {
+			return checker.BranchProtectionsData{}, nil
+		}
 		return checker.BranchProtectionsData{}, fmt.Errorf("%w", err)
 	}
 	for _, release := range releases {
--- a/checks/raw/sbom.go
+++ b/checks/raw/sbom.go
@@ -15,6 +15,7 @@
 package raw
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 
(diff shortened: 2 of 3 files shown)


checks/raw/branch_protection.go
```diff
@@ -66,6 +67,9 @@ func BranchProtection(cr *checker.CheckRequest) (checker.BranchProtectionsData,
 	// Get release branches.
 	releases, err := c.ListReleases()
 	if err != nil {
+		if errors.Is(err, clients.ErrUnsupportedFeature) {
+			return checker.BranchProtectionsData{}, nil
+		}
```
spencerschrock, line 72: we could still query branch protection of the default branch even if we can't determine the release branches. 
Would something like this be better?:
```go
// Get release branches.
releases, err := c.ListReleases()
if err != nil && !errors.Is(err, clients.ErrUnsupportedFeature) {
		return checker.BranchProtectionsData{}, fmt.Errorf("%w", err)
}
```

checks/raw/sbom.go
```diff
@@ -38,6 +39,9 @@ func SBOM(c *checker.CheckRequest) (checker.SBOMData, error) {
 
 	releases, lerr := c.RepoClient.ListReleases()
 	if lerr != nil {
+		if errors.Is(lerr, clients.ErrUnsupportedFeature) {
+			return results, nil
+		}
 		return results, fmt.Errorf("RepoClient.ListReleases: %w", lerr)
 	}
```
spencerschrock, line 46: similarly, this would prevent looking for SBOMs in source.

checks/raw/signed_releases.go
```diff
@@ -15,15 +15,20 @@
...
+	"github.com/ossf/scorecard/v5/clients"
 )
 
 // SignedReleases checks for presence of signed release check.
 func SignedReleases(c *checker.CheckRequest) (checker.SignedReleasesData, error) {
 	releases, err := c.RepoClient.ListReleases()
 	if err != nil {
+		if errors.Is(err, clients.ErrUnsupportedFeature) {
+			return checker.SignedReleasesData{}, nil
+		}
```
spencerschrock, line 31: I'm not sure we should change this one. the Signed-Releases check relies on this data alone.

Issue #3931, opened 2024-03-10
Feature: Probe whether repo has up-to-date CODEOWNERS


raghavkaul:
Scorecard should have a probe for whether users in a CODEOWNERS file are still members of the org that the repo belongs to. An up-to-date CODEOWNERS file makes it easier for a contributor to know who can help with a PR or a question about the project. This could use the "Get Organization membership for user" [API](https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#get-organization-membership-for-a-user) (which requires a PAT).

Might be a good fit for either the Contributors or Maintained checks.

spencerschrock:
> Might be a good fit for either the Contributors or Maintained checks.

It may be better to  have an "OSPO" focused category for these sort of admin required checks (like Webhooks).

> whether users in a CODEOWNERS file are still members of the org that the repo belongs to

Note: I think this makes an assumption about no external collaborators. I'm guessing the GitHub OSPO didn't have these sort of scenarios.

schaeferka:
I'm interested in this issue.

wbeckler:
Is there a way to tell whether a repo or org is enforcing org membership for maintainer activities? Maybe the check ignores org membership where it's not enforced.

spencerschrock:
Duplicate of #1554 

(trying to do some issue bookkeeping)

raghavkaul:
Users listed in the CODEOWNERS file should also be listed as contributors/maintainers, informing either the Contributors or Maintained check.

lharrison13:
I can try to pick this one up since its been here for a while.

> Is there a way to tell whether a repo or org is enforcing org membership for maintainer activities? Maybe the check ignores org membership where it's not enforced.

To @wbeckler's point, since there is no way to currently enforce this on github repos that I know of, maybe we can allow for a verified external approvers comment that will ignore this check for certain users in the file. Not sure if there is a better way to do this.

So the todos are:

- [ ] Replace the Webhooks check with an OSPO one - as per https://github.com/ossf/scorecard/issues/3931#issuecomment-1987475376
- [ ] Add the webhooksUseSecrets probe into the new OSPO check - as per https://github.com/ossf/scorecard/issues/3931#issuecomment-1987475376
- [ ] Create a new codeOwnersInOrganization probe for the OSPO check - as per the original issue
[...]

PR #4551, opened 2025-03-08
✨ Add contributorsFromCodeOwners probe and listCodeOwners github client method

lharrison13:

This PR adds the contributorsFromCodeOwners probe mentioned in here https://github.com/ossf/scorecard/issues/3931#issuecomment-2214220439 to supplement the contributor check.

It also adds a listCodeOwners method in the github repo client to support further checks and probes using code owners. As per #3931 

Currently the contributors check only checks if recent contributors are from companies and organizations.

Now the contributors check also test to see if code owners from the code owners file are included in the recent contributors list.

Contributes to  #3931

This is my first time creating a probe so let me know if I missed something.

I also tried to solve the issue of "verified external contributors" by allowing users to add an `# @verified` comment to their CODEOWNERS file along with the user names of the verified external contributors as per https://github.com/ossf/scorecard/issues/3931#issuecomment-2076272102.
This feature is not used in the probe I wrote but can be used if you want to write the new OSPO check described in the above issue.

Furthermore this does not actually implement the check described in the original issue as I just wanted to get your input on what I did so far. I can implement it in another PR later if this one ends up getting merged.

To test the new probe you can run the following:

`go run main.go --repo=https://github.com/carbon-design-system/carbon --checks=Contributors`
5/10

`go run main.go --repo=https://github.com/apache/superset --checks=Contributors`
10/10

Screenshot 2025-03-08 at 5 01 03 PM

This probe fails on a lot of repos because many orgs keep their teams private, so we cant check their members. Im not really sure if making org teams public is a security issue but it just seems a lot of orgs do this. If you do have access to see the team the probe should pass.

Lastly to parse the CODEOWNERS file i am using this [package](https://github.com/hmarr/codeowners) by @hmarr. I can manually parse it with regex but this was pretty easy to use and made things a lot simpler.

```release-note
adds new contributorsFromCodeOwners probe
```


spencerschrock:
Just wanted to acknowledge the PR. I am back from travel, and this is on my list of things to review this week.

clients/repo_client.go
```diff
@@ -50,6 +50,7 @@ type RepoClient interface {
 	ListLicenses() ([]License, error)
 	ListReleases() ([]Release, error)
 	ListContributors() ([]User, error)
+	ListCodeOwners() ([]User, error)
```
spencerschrock, line 53: Looking at the GitHub implementation of `ListCodeOwners()` I see a few different logical chunks:

1. Where to find the CODEOWNERS file and how to parse it. Which I understand is partly forge specific, but there's also overlap between forges. At a glance, the syntax is similar (wont say identical), and the locations are almost equivalent.
* GitHub: `CODEOWNERS`, `docs/CODEOWNERS`, `.github/CODEOWNERS`.
* GitLab:  `CODEOWNERS`, `docs/CODEOWNERS`, `.gitlab/CODEOWNERS`
2. Calls to map teams/usernames/emails in codeowners to users
3. A call to `handler.ghClient.Organizations.List` to get info about each discovered codeowner. A similar call already happens in `ListContributors`.

My immediate thought reviewing this is "Do we need to introduce a new interface method which would be a breaking change?"

I don't think there's any way to current do 2. or 3. with the current methods, unless we bundle the functionality under `ListContributors`.

checker/raw_result.go
```diff
@@ -218,7 +218,8 @@ type Changeset struct {
 
 // ContributorsData represents contributor information.
 type ContributorsData struct {
-	Users []clients.User
```
spencerschrock, line 221: another breaking change

docs/checks/internal/checks.yaml
```diff
@@ -360,19 +360,23 @@ checks:
...
 
       The check looks at the `Company` field on the GitHub user profile for authors of
-      recent commits. To receive the highest score, the project must have had
+      recent commits. It also compares the users specified in the codeowners file 
+      to the list of active contributors. To receive the highest score, the project must have had
       contributors from at least 3 different companies in the last 30 commits; each of
-      those contributors must have had at least 5 commits in the last 30 commits.
+      those contributors must have had at least 5 commits in the last 30 commits. 
+      Furthermore, the project should also have a CODEOWNERS file with at least 3 different users
+      and have commits in the last 30 commits.
```
spencerschrock, line 379: I'm not sure I'm interpreting the linked comment the same way:
https://github.com/ossf/scorecard/issues/3931#issuecomment-2214220439

Instead of scoring if recent commits are from contributors, I thought it was saying anyone listed as a codeowner should be counted as a contributor regardless of recent commits.

checks/evaluation/contributors.go
```diff
@@ -36,32 +39,54 @@ func Contributors(name string,
...
 		return checker.CreateMaxScoreResult(name, reason)
 	}
 
-	return checker.CreateProportionalScoreResult(name, reason, numberOfTrue, numberCompaniesForTopScore)
+	return checker.CreateProportionalScoreResult(
+		name,
+		reason,
+		numberOfTrueCompanies+numberOfTrueOwners,
+		numberCompaniesForTopScore+numberCodeOwnersForTopScore,
+	)
```
spencerschrock, line 81: i'm hesitant to make half the check something that usually relies on a privileged token

clients/repo_client.go
spencerschrock: There are also ways of doing this without adding `ListCodeowners` to this interface for right now, but keeping it in the GitHub client.

For example if we add the probe as an independent probe for now (meaning it doesn't use any underlying check data), and do a type check to a GitHub client:

```go
ghClient, ok := checkRequest.RepoClient.(*githubrepo.Client)
if !ok {
// something to skip or indicate not supported
}
ghClient.ListCodeowners()
```

docs/checks/internal/checks.yaml
spencerschrock: cc @raghavkaul for details maybe?

spencerschrock (commented):
Let's discuss this a little before iterating on the specifics (skipped most files in this review).

My individual comments tried to focus on the big picture, which to me is availability of the data, and focusing on this as something part of an OSPO type check.

docs/checks/internal/checks.yaml
spencerschrock: > Perhaps we could add a new field to the clients.user struct called RepoAssociation of type RepoAssociation and set it to RepoAssociationOwner so we can identify the contributor as an owner

`RepoAssociationOwner` has a slightly different meaning as it's currently used. But yes we could add some sort of metadata to the `clients.User` struct. Maybe a `Maintainer` or `Codeowner` bool?

But yes other than that, it sounds reasonable to me (pending Raghav's interpretation).

lharrison13: @raghavkaul thoughts?

spencerschrock: let's just go ahead and proceed.

lharrison13: Okay ill probably take a look again this weekend. Ill have to figure out where i left off lol.

lharrison13: @spencerschrock sorry for the delay I worked on it today and I will probably open a new pr tomorrow with the new changes. It should be a lot simpler to review than this one.

spencerschrock:
Continued in #4611

Issue #4627, opened 2025-05-15
How can I get the details when using scorecard serve?


Fix3dP0int:
I deployed scorecard as a service by `./scorecard serve`. I send a HTTP request to the port, but it returns the json whose `details` field is null. And I check the usage of `./scorecard serve` by `./scorecard serve --help`. It only support the flags `-h --help` but no any flags like `--show-details`.

Fix3dP0int:
Hi, I found the reson why lead to this result. I'm not sure yet whether this is a bug. In `serve.go`, the line [79](https://github.com/ossf/scorecard/blob/a553e1e6054807e043b607ca089e7b45ba8014db/cmd/serve.go#L79C16-L79C80), the usage of `repoResult.AsJson` is old, which doesn't compute the aggregate score. I replace it by `format.JSON`, which uses `AsJSON2`. Then I get the right result. Besides, I think that the serve should invoke `rootCmd` functions and inherit flags of root command by `o.AddFlags(cmd)`. If you agree with this, I'm glad to create a pull request to fix them.

spencerschrock:
Thanks for taking a look at this, yeah, `serve.go` is pretty old, I don't think it's ever worked in my ~3 years on the project. We haven't really had any interest/report until this issue. Do you have a use case for the functionality? 

In terms of things like `--show-details`, I would need to think if that would be a CLI flag, or something like a GET parameter for the request

Fix3dP0int:
Thank you for your reply. Due to GitHub token limitations, the Scorecard API service officially deployed is not real-time.
Our product relies on a large number of open-source components, and we use the Scorecard score as one of the criteria for technology selection. This helps us quickly identify open-source software that may pose security risks and prevent potential supply chain attacks.

Therefore, we need to evaluate multiple open-source projects **in bulk and in an automated way**. A stable and reliable HTTP service would significantly improve both efficiency and security. We plan to deploy a real-time internal Scorecard service for shared use across our teams, and to provide machine-to-machine interfaces for other services within our system.

I also fully agree with your idea to treat CLI flags as parameters — it might be a good time for us to consider designing a complete REST API.

spencerschrock:
Fixed by #4665

PR #4665, opened 2025-06-17
:sparkles: feat: Refactor scorecard serve cmd

Fix3dP0int:

feature

The current serve uses Go’s built-in http package, which lacks modern features. And It fails to correctly aggregate the total score, and parameters and details cannot be retrieved properly.

This PR refactors the serve component by migrating the original CLI-based parameter input to a RESTful API interface. Additionally, I replaced the native net/http logic with the [chi](https://github.com/go-chi/chi) router, which is lightweight yet expressive and well-suited for modular HTTP services in Go.

Related to #4627


spencerschrock:
Don't have time to look in-depth, but wanted to make one comment:

> The current serve uses Go’s built-in http package, which lacks modern features

What features specifically? Go 1.22 made good strides at least for the routing
https://go.dev/blog/routing-enhancements

gcanlin:
Sorry. I'm unfamiliar with new features of Go. It seems that `chi` isn't necessary. I will revert it. Thx.

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+	Checks      []string `json:"checks,omitempty"`
+	Commit      string   `json:"commit,omitempty"`
+	CommitDepth int      `json:"commit_depth,omitempty"`
+	ShowDetails bool     `json:"show_details,omitempty"`
+	Format      string   `json:"format,omitempty"`
+	LogLevel    string   `json:"log_level,omitempty"`
+	Probes      []string `json:"probes,omitempty"`
+	FileMode    string   `json:"file_mode,omitempty"`
+	PolicyFile  string   `json:"policy_file,omitempty"`
+}
```
spencerschrock: In the CLI, policy file is a local file. How do we expect to pass a policy file to a server, with a URI?

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+	}
+}
+
+func (s *server) handleScorecard(w http.ResponseWriter, r *http.Request) {
+	var req scorecardRequest
+	if r.Method == http.MethodPost {
+		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			http.Error(w, "invalid request body", http.StatusBadRequest)
+			return
+		}
```
spencerschrock, line 74: what is the goal of having POST? Passing parameters as JSON instead of as query parameters?

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+	}
+	s.opts.CommitDepth = req.CommitDepth
+	s.opts.ShowDetails = req.ShowDetails
+	s.opts.Format = req.Format
+	if s.opts.Format == "" {
+		s.opts.Format = options.FormatDefault
+	}
+	s.opts.LogLevel = req.LogLevel
+	if s.opts.LogLevel == "" {
+		s.opts.LogLevel = "info"
```
spencerschrock: `log.InfoLevel` is a constant with this value.

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+		req.Probes = strings.Split(r.URL.Query().Get("probes"), ",")
+		req.FileMode = r.URL.Query().Get("file_mode")
+		req.PolicyFile = r.URL.Query().Get("policy_file")
+	}
+
+	// Set options
+	s.opts.Repo = req.Repo
+	s.opts.Local = req.Local
+	s.opts.NPM = req.NPM
+	s.opts.PyPI = req.PyPI
```
spencerschrock: Right now we have one `options.Options`, which is passed in `serveCmd`, and then this one copy is modified for each request. This seems like a race condition.

we should create a new struct for each request. either through `options.New`, or manually if we want to avoid the env var parsing.

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+	if s.opts.LogLevel == "" {
+		s.opts.LogLevel = "info"
+	}
+	s.opts.FileMode = req.FileMode
+	if s.opts.FileMode == "" {
+		s.opts.FileMode = options.FileModeArchive
+	} else if s.opts.FileMode != options.FileModeArchive && s.opts.FileMode != options.FileModeGit {
+		http.Error(w, fmt.Sprintf("unsupported file mode: %s", s.opts.FileMode), http.StatusBadRequest)
+		return
+	}
```
spencerschrock: this validation is already covered by the `s.opts.Validate` call below

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+
+	ctx := r.Context()
+
+	var repo clients.Repo
+	if s.opts.Local != "" {
+		repo, err = localdir.MakeLocalDirRepo(s.opts.Local)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("making local dir: %v", err), http.StatusInternalServerError)
+			return
+		}
```
spencerschrock: do we expect people to pass the `serve` command repos which are local to the server?

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+	var requiredRequestTypes []checker.RequestType
+	if s.opts.Local != "" {
+		requiredRequestTypes = append(requiredRequestTypes, checker.FileBased)
+	}
+	if !strings.EqualFold(s.opts.Commit, clients.HeadSHA) {
+		requiredRequestTypes = append(requiredRequestTypes, checker.CommitBased)
+	}
+
+	enabledChecks, err := policy.GetEnabled(pol, s.opts.Checks(), requiredRequestTypes)
+	stdlog.Printf("DEBUG: enabledChecks = %#v", enabledChecks)
```
spencerschrock: should this be removed? was this for your testing?

cmd/serve.go
```diff
@@ -15,107 +15,316 @@
...
+	sort.Slice(repoResult.Checks, func(i, j int) bool {
+		return repoResult.Checks[i].Name < repoResult.Checks[j].Name
+	})
+
+	// Return results
+	w.Header().Set("Content-Type", "application/json")
+	if err := repoResult.AsJSON2(w, checkDocs, &scorecard.AsJSON2ResultOption{
+		LogLevel:    log.ParseLevel(s.opts.LogLevel),
+		Details:     s.opts.ShowDetails,
+		Annotations: false,
```
spencerschrock: we would probably want `ShowAnnotations` as a query parameter.

spencerschrock (commented):
At a high level this looks like what we want, an http wrapper around the CLI. You said MCP will be built on top of this API, so is it matching what you need for that? 

There are a few things that need changed, which I've left individual comments on. The linter also has some thoughts with this file.

cmd/serve.go
gcanlin: Sorry, I'm late. I rarely use this policy file. I think that it should be a cmd parameter rather than a server parameter, so now I've temporarily removed it.

cmd/serve.go
gcanlin: Both POST and GET APIs are provided. For clients that need to **pass multiple parameters**, using the request body is more convenient, whereas for simple cases with just one or two parameters, query parameters work very well.

cmd/serve.go
gcanlin: Thanks. I have fixed it.

cmd/serve.go
gcanlin: Yeah. `options.New` would be a better choice. Have fixed it.

cmd/serve.go
gcanlin: OK. It has been removed.

cmd/serve.go
gcanlin: This error also stems from my naive approach of simply copying the logic from `root.go`. When the server starts up, the repo should be a parameter provided by the client's request. Now `local` would be not supported.

cmd/serve.go
gcanlin: Yeah. Sorry for my carelessness. Done removing.

cmd/serve.go
gcanlin: Thanks for the suggestion! I've added `ShowAnnotations` as a query parameter now.

spencerschrock:
Back from vacation, going through my backlog from before and reopenning PRs that went stale waiting for me.

Issue #4617, opened 2025-05-06
BUG packaging not recognized


miparnisari:
**Describe the bug**
In https://securityscorecards.dev/viewer/?uri=github.com/authzed/spicedb 

it says

```
Reason
packaging workflow not detected
Details
Warn: no GitHub/GitLab publishing workflow detected.
```

however, we _do_ publish to GitHub, see the artifacts in https://github.com/orgs/authzed/packages?repo_name=spicedb

and the workflow step is https://github.com/authzed/spicedb/blob/9622de7c2f550b285f9c8d089f136412b9851476/.github/workflows/release.yaml#L41-L53

**Reproduction steps**
n/a

**Expected behavior**
The `packaging` check should pass for this repo.

AdamKorcz:
Should be fixed by https://github.com/ossf/scorecard/pull/4673

spencerschrock:
> however, we _do_ publish to GitHub, see the artifacts in https://github.com/orgs/authzed/packages?repo_name=spicedb
> 
> and the workflow step is

Currently we search for `actions/setup-go`, followed by `goreleaser/goreleaser-action`, where as your workflow doesn't call `actions/setup-go`. 

Instead it uses the Go distribution present in the runner already (which probably uses Go toolchains to download 1.24.0 as specified in your go.mod). I've left a comment in the linked PR about this, and whether or not we should look for setup-go

AdamKorcz:
To add a bit of context, @miparnisari s workflow calls `authzed/actions/setup-go`

PR #132, opened 2021-01-15
Add packaging check.

inferno-chromium:

PR #4673, opened 2025-06-26
🐛 remove setup-go requirement for Packaging with goreleaser

AdamKorcz:

Currently, the Go `JobMatcher` requires both the `setup-go` and `goreleaser` actions to be used to consider the project to use and automated releaser. Some users have adopted other actions that install Go and still use the `goreleaser` action in which case Scorecard will not see the `goreleaser` action.

This PR removes the requirement that users _must_ use the `setup-go` action to Scorecard recognizing `goreleaser`.

Fixes https://github.com/ossf/scorecard/issues/4617

the `release-note`

```release-note
Fix bug in parsing workflow files.
```

### DIFF
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -383,8 +383,6 @@ func AnyJobsMatch(workflow *actionlint.Workflow, jobMatchers []JobMatcher, fp st
 // matches returns true if the job matches the job matcher.
 func (m *JobMatcher) matches(job *actionlint.Job) bool {
 	for _, stepToMatch := range m.Steps {
-		hasMatch := false
-
 		// First look for re-usable workflow calls.
 		if job.WorkflowCall != nil &&
 			job.WorkflowCall.Uses != nil &&
@@ -395,15 +393,11 @@ func (m *JobMatcher) matches(job *actionlint.Job) bool {
 		// Second looks for steps in the job.
 		for _, step := range job.Steps {
 			if stepsMatch(stepToMatch, step) {
-				hasMatch = true
-				break
+				return true
 			}
 		}
-		if !hasMatch {
-			return false
-		}
 	}
-	return true
+	return false
 }
 
 // stepsMatch returns true if the fields on 'stepToMatch' match what's in 'step'.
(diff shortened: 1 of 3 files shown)


spencerschrock:
> Currently, `matches` will return `false` if the first `Step` of the `JobMatcher` fails. This is a problem if there are multiple steps in the `JobMatcher` where the second one should return true.

This is by design, the matcher is looking for every step in the pattern to be in the job. "Each step in this field has a matching step in the job"

https://github.com/ossf/scorecard/blob/42e0ac2ef4ad9cdc0bd71f35a69f45257b8d3276/checks/fileparser/github_workflow.go#L328-L333

As written, this PR would give credit to any repo which uses `actions/setup-go` regardless of if it uses goreleaser. e.g.
https://github.com/spencerschrock/actions-test/blob/f4f716a8b0df876e28868d879246863b2c4bcfa8/.github/workflows/disk.yml#L12

```
  "score": 10.0,
  "checks": [
    {
      "details": [
        "Info: Project packages its releases by way of GitHub Actions.: .github/workflows/disk.yml:8"
      ],
      "score": 10,
      "reason": "packaging workflow detected",
      "name": "Packaging",
```


I think the question we want to ask is: do we need actions/setup-go to give credit for a goreleaser release workflow.
https://github.com/ossf/scorecard/blob/42e0ac2ef4ad9cdc0bd71f35a69f45257b8d3276/checks/fileparser/github_workflow.go#L557-L565

AdamKorcz:
Thanks @spencerschrock Essentially this means that we (in this case) only detect a goreleaser use if it is preceded by a step using `actions/setup-go`. Do you know what the logic behind this is? As in, why do we _only_ want to detect goreleaser if it is preceded by `setup-go`?

spencerschrock:
> only detect a goreleaser use if it is preceded by a step using `actions/setup-go`

as long as all steps are present, I don't think order matters to our current implementation.

> Do you know what the logic behind this is?

This predates my time on the project. Detection for Go was first added in #800, but similar logic (expecting language setup) was added in #132 for npm and Python.

My guess is it was to cut down on false positives, as the entire workflow file was scanned with a regex, instead of the workflow parsing and step analysis that happens now.

AdamKorcz:
okay, it may be better to remove `actions/setup-go` from the `JobMatcher` `Uses`. It seems unnecessary that both actions need to exist for the goreleaser action to be valid.

spencerschrock:
> okay, it may be better to remove `actions/setup-go` from the `JobMatcher` `Uses`. It seems unnecessary that both actions need to exist for the goreleaser action to be valid.

Seems reasonable to me.

AdamKorcz:
Updated the PR and the PR description.

Issue #4531, opened 2025-02-19
BUG Probe for SAST throws nil pointer exception for some Repos


aunovis-heidrich:
**Describe the bug**
If I run the sastToolConfigured probe on the https://github.com/toml-rs/toml repository it throws a memory exception.

**Reproduction steps**
```
scorecard.exe --repo=https://github.com/toml-rs/toml --probes=sastToolConfigured
```
On my setup (git-bash, Windows 11) this produces a stack trace:
```
Starting probe [sastToolConfigured]
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x18 pc=0x19f9b4f]

goroutine 1 [running]:
github.com/ossf/scorecard/v5/checks/raw.sastToolInCheckRuns(0xc0004a86e0)
        github.com/ossf/scorecard/v5@v5.1.1/checks/raw/sast.go:128 +0x2ef
github.com/ossf/scorecard/v5/checks/raw.SAST(0xc0004a86e0)
        github.com/ossf/scorecard/v5@v5.1.1/checks/raw/sast.go:53 +0x69
github.com/ossf/scorecard/v5/pkg/scorecard.assignRawData({0x21d249e, 0x12?}, 0xc0004a86e0, 0xc000554708)
        github.com/ossf/scorecard/v5@v5.1.1/pkg/scorecard/scorecard_result.go:350 +0xf6
github.com/ossf/scorecard/v5/pkg/scorecard.populateRawResults(0xc0004a86e0, {0xc000502230, 0x1, 0x440b9e?}, 0xc000554708)
        github.com/ossf/scorecard/v5@v5.1.1/pkg/scorecard/scorecard_result.go:407 +0x2ad
github.com/ossf/scorecard/v5/pkg/scorecard.runEnabledProbes(0xc0004a86e0, {0xc000502230, 0x1, 0x9?}, 0xc000554708)
        github.com/ossf/scorecard/v5@v5.1.1/pkg/scorecard/scorecard.go:222 +0x45
github.com/ossf/scorecard/v5/pkg/scorecard.runScorecard({_, _}, {_, _}, {_, _}, _, _, {0xc000502230, 0x1, ...}, ...)
        github.com/ossf/scorecard/v5@v5.1.1/pkg/scorecard/scorecard.go:171 +0xb1b
github.com/ossf/scorecard/v5/pkg/scorecard.Run({_, _}, {_, _}, {_, _, _})
[...]

PR #4675, opened 2025-06-26
🐛 add nil check to loggers in checks/raw

AdamKorcz:

bug fix

Currently Scorecard logs even though the logger technically can be nil.

Scorecard now checks whether the logger is nil before logging.

I have added a test case to the existing test.

Fixes https://github.com/ossf/scorecard/issues/4531

the `release-note`

```release-note
add nil check for loggers in checks raw
```

### DIFF
--- a/checks/raw/sast_test.go
+++ b/checks/raw/sast_test.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"go.uber.org/mock/gomock"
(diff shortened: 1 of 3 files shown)


checks/raw/sast_test.go
```diff
@@ -28,14 +29,17 @@ import (
 	"github.com/ossf/scorecard/v5/finding"
 )
 
+var mergedOneHourAgo = time.Now().Add(time.Hour - 1)
+
```
spencerschrock: this is one hour in the future btw (well, 1 hour less one second). not that it really matters, just that the timestamp is non-zero

AdamKorcz: https://github.com/ossf/scorecard/pull/4675/commits/d35f1ee35e5d65569a54949f3ac3a69a801b81f9

PR #4676, opened 2025-06-26
:sparkles: recognize escaped emails in security policy documents

ralphbean:

I think this counts as a feature (a micro feature).

Some security policy documents may include an email address partially
obfuscated in the form security[at]example.org or, especially in
markdown, escaped like security\\[at\\]example.org.

This change updates the regex to recognize those two forms, with the
tests expanded to cover those cases as well.

Only emails of the form security@example.org are recognized.

Emails of the form security[at]example.org and security\[at\]example.org are recognized.

NONE

This depends first on https://github.com/ossf/scorecard/pull/4674 being merged.

```release-note
Start recognizing escaped emails in security policy documents
```

### DIFF
--- a/checks/raw/security_policy.go
+++ b/checks/raw/security_policy.go
@@ -199,7 +199,7 @@ func collectPolicyHits(policyContent []byte) []checker.SecurityPolicyInformation
 	// pattern for URLs
 	reURL := regexp.MustCompile(`(http|https)://[a-zA-Z0-9./?=_%:-]*`)
 	// pattern for emails
-	reEML := regexp.MustCompile(`\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,6}\b`)
+	reEML := regexp.MustCompile(`\b[A-Za-z0-9._%+-]+(@|\\?\[at\\?\])[A-Za-z0-9.-]+\.[A-Za-z]{2,6}\b`)
 	// pattern for 1 to 4 digit numbers
 	// or
 	// strings 'disclos' as in "disclosure" or 'vuln' as in "vulnerability"
--- a/checks/raw/security_policy_test.go
+++ b/checks/raw/security_policy_test.go
@@ -204,13 +204,41 @@ func Test_collectPolicyHits(t *testing.T) {
 		},
 		// Email regex
 		{
-			name:  "Email positive",
+			name:  "Email positive with @",
 			input: "Contact us at security@example.org.",
 			expected: []checker.SecurityPolicyInformation{
 				{
 					InformationType: checker.SecurityPolicyInformationTypeEmail,
 					InformationValue: checker.SecurityPolicyValueType{
 						Match:      "security@example.org",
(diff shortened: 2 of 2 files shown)


checks/raw/security_policy_test.go
```diff
@@ -173,3 +173,191 @@ func TestSecurityPolicy(t *testing.T) {
...
+						Match:      "security[at]example.org",
+						LineNumber: 1,
+						Offset:     14,
+					},
+				},
+			},
+		},
+		{
+			name:  "Email positive with [at] escaped",
+			input: "Contact us at security\\[at\\]example.org.",
```
spencerschrock, line 236: In the README you would be matching this?

`Contact us at security\[at\]example.org.`

Just trying to confirm the level of escaping since you're also escaping the Go here. (p.s. go supports raw string literals.
https://go.dev/ref/spec#String_literals
> Raw string literals are character sequences between back quotes, as in `foo`. Within the quotes, any character may appear except back quote. The value of a raw string literal is the string composed of the uninterpreted (implicitly UTF-8-encoded) characters between the quotes; in particular, backslashes have no special meaning and the string may contain newlines

so this would be equivalent (but not a required change):
```go
input: `Contact us at security\[at\]example.org.`,
```

checks/raw/security_policy.go
```diff
@@ -199,7 +199,7 @@ func collectPolicyHits(policyContent []byte) []checker.SecurityPolicyInformation
 	// pattern for URLs
 	reURL := regexp.MustCompile(`(http|https)://[a-zA-Z0-9./?=_%:-]*`)
 	// pattern for emails
-	reEML := regexp.MustCompile(`\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,6}\b`)
+	reEML := regexp.MustCompile(`\b[A-Za-z0-9._%+-]+(@|\\?\[at\\?\])[A-Za-z0-9.-]+\.[A-Za-z]{2,6}\b`)
```
spencerschrock, line 202: do people always do lowercase `[at]`?

spencerschrock (commented):
generally looks good, one or two small questions

checks/raw/security_policy.go
ralphbean: I haven't seen that yet, but I suppose its possible.

checks/raw/security_policy_test.go
ralphbean: > In the README you would be matching this?
```
Contact us at security\[at\]example.org.
```
Yes, that's my intent. I'm assuming people will be trying to avoid markdown tools misinterpreting that `[at]` as a link. Github-flavored markdown won't, but - I already saw this in a SECURITY.md that I tried scorecard on in the wild.

checks/raw/security_policy.go
spencerschrock: we can always add it later if other forms are seen

spencerschrock (approved):
Thanks

PR #4677, opened 2025-06-27
🐛 do not allow unsupported `ListReleases` to fail a run

JamieMagee:

This check allows clients which currently do not support `ListReleases` to complete a Scorecard run. Currently `azuredevopsrepo`, `localdir`, and `git` do not support `ListReleases`.

This is a follow-up from #4533, with PR comments addressed, after I let that PR go stale.

Currently, running `scorecard` against a client that does not support `ListReleases` results in a non-zero exit code, and an error printed to the console:

```
Error: check runtime error: Branch-Protection: internal error: unsupported feature
2025/06/26 20:08:25 error during command execution: check runtime error: Branch-Protection: internal error: unsupported feature
```

Partial results for the checks that partially rely on `ListReleases` (`Branch-Protection`, `SBOM`) and no results for `Signed-Releases`.

NONE

the `release-note`

```release-note
Prevent unimplemented ListReleases from failing a run
```

### DIFF
--- a/checks/raw/signed_releases.go
+++ b/checks/raw/signed_releases.go
@@ -15,15 +15,17 @@
 package raw
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/ossf/scorecard/v5/checker"
+	"github.com/ossf/scorecard/v5/clients"
 )
 
 // SignedReleases checks for presence of signed release check.
 func SignedReleases(c *checker.CheckRequest) (checker.SignedReleasesData, error) {
 	releases, err := c.RepoClient.ListReleases()
-	if err != nil {
+	if err != nil && !errors.Is(err, clients.ErrUnsupportedFeature) {
 		return checker.SignedReleasesData{}, fmt.Errorf("%w", err)
 	}
 
--- a/checks/raw/branch_protection.go
+++ b/checks/raw/branch_protection.go
@@ -15,6 +15,7 @@
 package raw
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"regexp"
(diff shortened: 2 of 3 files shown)


checks/raw/signed_releases.go
```diff
@@ -15,15 +15,17 @@
...
 
 	"github.com/ossf/scorecard/v5/checker"
+	"github.com/ossf/scorecard/v5/clients"
 )
 
 // SignedReleases checks for presence of signed release check.
 func SignedReleases(c *checker.CheckRequest) (checker.SignedReleasesData, error) {
 	releases, err := c.RepoClient.ListReleases()
-	if err != nil {
+	if err != nil && !errors.Is(err, clients.ErrUnsupportedFeature) {
```
spencerschrock: release data is critical to the SignedReleases check, thoughts on keeping it?

spencerschrock (commented):
Branch-Protection and SBOM changes LGTM. Question about Signed-Releases

checks/raw/signed_releases.go
JamieMagee: Does removing a check from the YAML policy definition prevent it from running?

https://github.com/JamieMagee/scorecard-azure-pipelines-task/blob/63f20d3575ab167a9bc9bdda94db07a98d25fac5/assets/policy.yml#L50-L52

If so, I could revert this change here.

spencerschrock: I'm not as familiar with the policy file, but I believe so, since it iterates over checks in the file
https://github.com/ossf/scorecard/blob/cb65b9b772049cb943b66c215902a5b86d216f05/policy/policy.go#L164-L170

JamieMagee: Okay, I'll revert this change then.

JamieMagee: I just double checked, and even with `mode: disabled` the `Signed-Releases` check fails with:

```
Error: check runtime error: Signed-Releases: internal error: unsupported feature
2025/06/27 10:25:30 error during command execution: check runtime error: Signed-Releases: internal error: unsupported feature
```

I need to completely omit it from `policy.yml` to get Scorecard to pass. I think it is valid to exclude it, given Azure DevOps doesn't support the Scorecard concept of a 'release'.

spencerschrock (approved):
Thanks!

Issue #4589, opened 2025-04-09
[npmCommand] npm install github:package.repo#sha is detected as non pinned dependency


kuisathaverat:
**Describe the bug**

The following step in a GitHub workflow is detected as a `Warn: npmCommand not pinned by hash` when it really is pinned.

```yaml
       
      - name: Install Dependencies
        run: |
          npm install github:nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12
```

**Reproduction steps**
1. Create a workflow like the following
```yaml
name: test
on:
  workflow_dispatch:
permissions:
  contents: read
jobs:
  process:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
      - name: Install Dependencies
        run: |
          npm install github:nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12
```
3. Run the command `scorecard --repo=github.com/owner/repo  --show-details`
4. Check that the new workflow reports the Warning `npmCommand not pinned by hash`

**Expected behavior**

The command is detected as a pinned dependency 

https://docs.npmjs.com/about-packages-and-modules#npm-package-git-url-formats

**Additional context**

PR #4680, opened 2025-06-27
🐛 check for npm package git URLs

AdamKorcz:

Currently Scorecard only check for `npm ci` to determine whether packages are pinned, however, users can also do that with other URL formats like [these](https://docs.npmjs.com/about-packages-and-modules#npm-package-git-url-formats). 

Scorecard supports [git URLs](https://docs.npmjs.com/about-packages-and-modules#npm-package-git-url-formats) for `npm install`

Fixes https://github.com/ossf/scorecard/issues/4589

the `release-note`

```release-note
Support git URLs for calls to npm install.
```

### DIFF
--- a/checks/raw/shell_download_validate.go
+++ b/checks/raw/shell_download_validate.go
@@ -460,6 +460,25 @@ func isNpmUnpinnedDownload(cmd []string) bool {
 		if strings.EqualFold(cmd[i], "ci") {
 			return false
 		}
+
+		// Check if the package is of any of the
+		// following formats: https://docs.npmjs.com/about-packages-and-modules#npm-package-git-url-formats
+		if strings.HasPrefix(cmd[i], "github:") ||
+			strings.HasPrefix(cmd[i], "git:") ||
+			strings.HasPrefix(cmd[i], "http://") ||
+			strings.HasPrefix(cmd[i], "https://") {
+			// We expect the URL to only have 1 #, but
+			// we can check whether it has more and
+			// is invalid.
+			s := strings.SplitN(cmd[i], "#", 5)
+			if len(s) != 2 {
+				return false
+			}
+			// git commit hashes have 40 chars
+			if len(s[1]) == 40 {
+				return false
+			}
+		}
 	}
 	return true
 }
--- a/checks/raw/shell_download_validate_test.go
+++ b/checks/raw/shell_download_validate_test.go
@@ -456,6 +456,55 @@ func Test_isNpmUnpinnedDownload(t *testing.T) {
 			},
 			want: false,
 		},
+		{
+			name: "npm install with github: prefix and valid hash",
+			args: args{
+				cmd: []string{"npm", "install", "github:nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12"},
+			},
+			want: false,
+		},
+		{
+			name: "npm install with git: prefix and valid hash",
+			args: args{
+				cmd: []string{"npm", "install", "github:nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12"},
+			},
+			want: false,
+		},
+		{
+			name: "npm install with http: prefix and valid hash",
+			args: args{
+				cmd: []string{"npm", "install", "github:nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12"},
+			},
+			want: false,
+		},
+		{
+			name: "npm install with https: prefix and valid hash",
+			args: args{
+				cmd: []string{"npm", "install", "github:nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12"},
+			},
+			want: false,
+		},
+		{
+			name: "npm install invalid github url (has too many hash characters)",
+			args: args{
+				cmd: []string{"npm", "install", "githu#b:n#odeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12"},
+			},
+			want: true,
+		},
+		{
+			name: "npm install with wrong prefix (githu instead of github)",
+			args: args{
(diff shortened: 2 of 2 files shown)


checks/raw/shell_download_validate.go
```diff
@@ -460,6 +460,25 @@ func isNpmUnpinnedDownload(cmd []string) bool {
...
+		// Check if the package is of any of the
+		// following formats: https://docs.npmjs.com/about-packages-and-modules#npm-package-git-url-formats
+		if strings.HasPrefix(cmd[i], "github:") ||
+			strings.HasPrefix(cmd[i], "git:") ||
+			strings.HasPrefix(cmd[i], "http://") ||
+			strings.HasPrefix(cmd[i], "https://") {
+			// We expect the URL to only have 1 #, but
+			// we can check whether it has more and
+			// is invalid.
+			s := strings.SplitN(cmd[i], "#", 5)
```
spencerschrock, line 473: do we know if the array of commands might have comments? My guess says no, but just brainstorming

checks/raw/shell_download_validate.go
```diff
@@ -460,6 +460,25 @@ func isNpmUnpinnedDownload(cmd []string) bool {
...
+			if len(s) != 2 {
+				return false
+			}
+			// git commit hashes have 40 chars
+			if len(s[1]) == 40 {
+				return false
+			}
```
spencerschrock: Use `gitCommitHashRegex` defined higher in the file?
```
var gitCommitHashRegex = regexp.MustCompile(`^[a-fA-F0-9]{40}$`)
```

spencerschrock: you could have a tag or branch ref that's 40 char for example

checks/raw/shell_download_validate_test.go
```diff
@@ -456,6 +456,55 @@ func Test_isNpmUnpinnedDownload(t *testing.T) {
...
+			},
+			want: false,
+		},
+		{
+			name: "npm install with https: prefix and valid hash",
+			args: args{
+				cmd: []string{"npm", "install", "github:nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12"},
+			},
+			want: false,
+		},
```
spencerschrock: these URLs are all `github:`, i assume this is copy paste error?

checks/raw/shell_download_validate.go
AdamKorcz: I guess technically it can, but to catch such cases, we likely need to use a shell parser and ignore comments? Or implement some logic ourselves that removes everything after ` #` (hash after a space) before splitting.

checks/raw/shell_download_validate_test.go
AdamKorcz: yep, it was a miss. Fixed in https://github.com/ossf/scorecard/pull/4680/commits/bfd450edfd486b3cad600ab45a3aaeb70fe0e64b

checks/raw/shell_download_validate.go
AdamKorcz: https://github.com/ossf/scorecard/pull/4680/commits/f5f758d559aa2b81736eedbcd4a9293e5f5c49ca

checks/raw/shell_download_validate.go
spencerschrock: I believe we use a shell parser to get here, so it will probably get removed

checks/raw/shell_download_validate.go
```diff
@@ -460,6 +460,25 @@ func isNpmUnpinnedDownload(cmd []string) bool {
 		if strings.EqualFold(cmd[i], "ci") {
 			return false
 		}
+
+		// Check if the package is of any of the
+		// following formats: https://docs.npmjs.com/about-packages-and-modules#npm-package-git-url-formats
+		if strings.HasPrefix(cmd[i], "github:") ||
+			strings.HasPrefix(cmd[i], "git:") ||
+			strings.HasPrefix(cmd[i], "http://") ||
+			strings.HasPrefix(cmd[i], "https://") {
```
spencerschrock: Can we tweak these prefixes? From the link these are the expected formats:
```
git://github.com/user/project.git#commit-ish
git+ssh://user@hostname:project.git#commit-ish
git+http://user@hostname/project/blah.git#commit-ish
git+https://user@hostname/project/blah.git#commit-ish
```

So I would expect something like this:
```go
if strings.HasPrefix(cmd[i], "github:") ||
			strings.HasPrefix(cmd[i], "git://") ||
			strings.HasPrefix(cmd[i], "git+http://") ||
			strings.HasPrefix(cmd[i], "git+https://") {
```

checks/raw/shell_download_validate_test.go
```diff
@@ -456,6 +456,55 @@ func Test_isNpmUnpinnedDownload(t *testing.T) {
...
+		{
+			name: "npm install with https: prefix and valid hash",
+			args: args{
+				cmd: []string{"npm", "install", "http://nodeca/js-yaml#2cef47bebf60da141b78b085f3dea3b5733dcc12"},
+			},
+			want: false,
+		},
```
spencerschrock: Can we make these match the example format from the documentation (but with a commit hash for the commitish)

AdamKorcz: https://github.com/ossf/scorecard/pull/4680/commits/cdb99618d1630820b874d259522ee7ab15695182

checks/raw/shell_download_validate.go
AdamKorcz: https://github.com/ossf/scorecard/pull/4680/commits/cdb99618d1630820b874d259522ee7ab15695182

spencerschrock (commented):
/scdiff generate Pinned-Dependencies

spencerschrock (approved):

Issue #2906, opened 2023-04-25
BUG: Dockerfile named build stages with incomplete remediation report


gabibguti:
**Describe the bug**

Pinned-Dependencies check does not correctly report the remediation for snippets such as `FROM alpine as build` in Dockerfiles.

Using [named build stages](https://docs.docker.com/build/building/multi-stage/#name-your-build-stages) as in `FROM alpine as build` results in a incomplete remediation report:

```Warn: containerImage not pinned by hash: Dockerfile:{line}```

While using `FROM alpine` results in a complete remediation report:

```Warn: containerImage not pinned by hash: Dockerfile:{line}: pin your Docker image by updating alpine to alpine@sha256:124c7d2707904eea7431fffe91522a01e5a861a624ee31d03372cc1d138a3126```

**Reproduction steps**
1. Create a Dockerfile with 2 lines, one `FROM alpine as build` and another `FROM alpine`.
2. Run Scorecard@4.10.5 detailed report
3. Check Pinned-Dependencies check details
4. Verify in report the `FROM alpine as build` line does NOT report the remediation suggesting the sha256
5. Verify in report the `FROM alpine` line does report the remediation suggesting the sha256

**Expected behavior**
I expect `FROM alpine as build` example and other named build stages to report the remediation suggesting the sha256.

**Additional context**
None.

naveensrinivasan:
https://github.com/ossf/scorecard/issues/2988#issue-1704497599

pnacht:
I'm not sure if #2988 is the same issue as this one. That one effectively requires parsing the dockerfile as an actual dockerfile, evaluating variables (potentially environment variables?)... all of which may be very complex.

I'm not fluent on the Scorecard codebase, but this issue likely comes from some slight difference in how Scorecard interprets `FROM xyz` and `FROM xyz AS abc` which is causing Scorecard to stumble when generating remediation for the latter case.

https://github.com/ossf/scorecard/tree/main/checks/raw/pinned_dependencies.go#L254-L305

PR #4683, opened 2025-06-27
🐛 do not include asName when creating digest of FROM name as asName

AdamKorcz:

bug fix

When calculating the digest in remediation, Scorecard would do that from `python:3.7:build` when the line was `FROM python:3.7 as build`. This would lead to issues such as https://github.com/ossf/scorecard/issues/2906.

We are creating the digest on the following lines: 

https://github.com/ossf/scorecard/blob/f96f4f49e4846c4846fd7cb8a2846eec83be8774/remediation/remediations.go#L105-L113

From the PR, it should be clear that `d.PinnedAt` is assigned the wrong value; it gets assigned `asPointer(asName)` which is the root cause of the bug. `digester.Digest(name)` throws an error like: `parsing reference "python:3.7:build": could not parse reference: python:3.7:build` because of this.

With this change, we create the digest from `python:3.7` in the same line as above. 
 
Fixes https://github.com/ossf/scorecard/issues/2906

NONE

```release-note
Ignore as-name with creating digest.
```


AdamKorcz:
Please reopen

Issue #2318, opened 2022-09-30
Add more options for SAST tools


varunsh-coder:
**Is your feature request related to a problem? Please describe.**
I would like to start a discussion to add more options for SAST tools. As of now, 3 tools are checked in the SAST check - CodeQL, LGTM, Sonar. As per [this issue](https://github.com/ossf/scorecard/issues/2259), LGTM is going away. 

Here are some of the things to consider:
1. **Language specific tools** - as an example, [GoSec](https://github.com/securego/gosec), and [Bandit](https://github.com/PyCQA/bandit) are well known tools for Go and Python respectively. 
2. **Pricing** - if one runs Scorecard in private repo, they might want to pass the SAST check while using free/ open source SAST tools. 
3. **Specialized purpose** - as an example, there are tools that specialize in scanning for secrets. This is an important check which is missing in scorecard today. Also, [CIS benchmark for supply chain security](https://github.com/aquasecurity/chain-bench/blob/main/docs/CIS-Software-Supply-Chain-Security-Guide-v1.0.pdf), has following additional categories: 
    - Scanner for CI pipelines - I think Scorecard already has checks for GitHub Actions. 
    - Scanner for IaC - not sure if this is in-scope for Scorecard
    - Scanner for vulnerabilities in open source packages being used, e.g. Dependency Review Action. 
    - Scanner for open source license issues 

**Describe the solution you'd like**
Would like to have a discussion to come to consensus on what additional SAST tools to add in Scorecard check. Based on the decision, those tools can then be added in the SAST check.

laurentsimon:
Tools for IaC scanning:
- [ ] https://github.com/bridgecrewio/checkov-action (https://github.com/bridgecrewio/checkov)
- [ ] https://github.com/Checkmarx/kics-github-action (https://github.com/Checkmarx/kics)

varunsh-coder:
I am starting the analysis using the table below of popular linting and security-related tools per language and using the [top programming languages ](https://octoverse.github.com/2022/top-programming-languages)blog post as a way to prioritize the analysis. 

I am not an expert in any of these languages or the most used linters/ security tools. This is just a way to organize the info. Please share feedback and info on other popular linters/ security tools.

| Language      | Popular linters   |  Popular tools with emphasis on security|
| ------------- | ----------------|---------------------- |
| JavaScript     | ESLint                 | CodeQL, Semgrep      |
| Python           | Pylint                  | CodeQL, Semgrep, Bandit      |

[...]

PR #4688, opened 2025-06-28
:sparkles: add support for hadolint SAST

AdamKorcz:

feature

Scorecard does not support hadolint SAST.

Scorecard supports hadolint SAST.

Partially fixes https://github.com/ossf/scorecard/issues/2318#issuecomment-1357903694

```release-note
Support Hadolint SAST.
```


checks/raw/sast.go
```diff
@@ -43,6 +43,7 @@ var sastTools = map[string]bool{
 	"lgtm-com":                 true,
 	"sonarcloud":               true,
 	"sonarqubecloud":           true,
+	"hadolint-action":          true,
```
spencerschrock: this section is normally for github apps, is it an app or an action?

checks/sast_test.go
```diff
@@ -294,12 +334,52 @@ func Test_SAST(t *testing.T) {
...
+				NumberOfInfo: 1,
+			},
+		},
 	}
 	for _, tt := range tests {
-		searchRequest := clients.SearchRequest{
-			Query: "github/codeql-action/analyze",
-			Path:  "/.github/workflows",
-		}
 		t.Run(tt.name, func(t *testing.T) {
```
spencerschrock: was there a driver behind redefining this in every test instead of once?

AdamKorcz: yes, I added a new test that needs another `searchRequest`.

checks/raw/sast.go
AdamKorcz: It is an action. Have removed in https://github.com/ossf/scorecard/pull/4688/commits/344a155637030b5e452aa10ad27a2c76dbbca4aa. Thanks!

checks/sast_test.go
spencerschrock: All of the duplication caught my eye initially, but I don't think it's not necessary anymore.
When the test was written 4 years ago, the SAST implementation actually used the search interface.

https://github.com/ossf/scorecard/blob/e5a08af2444ca80704ce2080f073d4c74b6a5249/checks/sast.go#L170-L176

Where as today's implementation uses `getSastUsesWorkflows`
https://github.com/ossf/scorecard/blob/344a155637030b5e452aa10ad27a2c76dbbca4aa/checks/raw/sast.go#L60-L63

which searches the files locally
https://github.com/ossf/scorecard/blob/344a155637030b5e452aa10ad27a2c76dbbca4aa/checks/raw/sast.go#L169-L172

If you uncomment the mock line, nothing panics/breaks, so I say we just get rid of it all the search requests/responses in this test
```
// mockRepoClient.EXPECT().Search(tt.searchRequest).Return(tt.searchresult, nil).AnyTimes()
```

checks/sast_test.go
```diff
@@ -294,12 +334,52 @@ func Test_SAST(t *testing.T) {
...
+						Slug: "hadolint-action",
+					},
+				},
+				{
+					Status:     "notCompletedForTestingOnly",
+					Conclusion: "notSuccessForTestingOnly",
+					App: clients.CheckRunApp{
+						Slug: "hadolint-action",
+					},
+				},
```
spencerschrock: I think this also got leftover?

checks/sast_test.go
AdamKorcz: I am happy to do that, but let me clean up this test in a separate PR to not mix things together? I have reverted my changes in this PR.

spencerschrock: Sure, please submit as a follow-up

Issue #2476, opened 2022-11-21
Feature: Add a Tag protection check to encourage stronger access control across releases


inferno-chromium:
**Is your feature request related to a problem? Please describe.**
Right now, a write/maintain role contributor can delete existing releases and replace with releases pointing to older, vulnerable commits. Read why - https://medium.com/boostsecurity/slsa-dip-source-of-the-problem-a1dac46a976. Have your release tags protected - https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules

**Describe the solution you'd like**
Maybe a check that encourages adding atleast one tag protection rule on existing releases. Ideally a "*" would be nice, so only admins can have those perms to replace/delete releases.

**Describe alternatives you've considered**
None

**Additional context**
None

AdamKorcz:
I am working on this. @spencerschrock would you mind assigning me?

PR #4699, opened 2025-07-10
Fixes false warnings in the "branch protection applies to admins" probe when repositories have tag-only rulesets configured

trask:
## Problem

Scorecard was incorrectly applying repository rulesets that target tags to branch protection analysis. This caused false positive warnings about admin enforcement when a repository had:
- Tag-only rulesets with bypass actors (targeting tags, not branches)
- Proper branch protection rules that do enforce admin compliance

The issue occurred because the `repoRuleSet` struct was missing the `Target` field from the GraphQL query, causing all rulesets to be treated as if they applied to branches.

## Solution

1. **Added missing `Target` field** to the `repoRuleSet` struct to capture the ruleset target from GitHub's GraphQL API
2. **Updated `rulesMatchingBranch` function** to filter out rulesets where `Target != "branch"`
3. **Updated GraphQL query** to include the `target` field in the rulesets query

## Testing

Verified the fix by running Scorecard against the https://github.com/open-telemetry/opentelemetry-java-contrib repository.


spencerschrock:
> incorrectly applying repository rulesets that target tags to branch protection analysis

We do have some feature requests for this kind of analysis though. Both are requested as new checks, so perhaps re-enabling tag analysis is a problem for another day/place

https://github.com/ossf/scorecard/issues/10
https://github.com/ossf/scorecard/issues/2476

trask:
> We do have some feature requests for this kind of analysis though

makes sense, I made some updates to make it more future proof for when those features land

clients/githubrepo/branches.go
```diff
@@ -481,8 +482,9 @@ nextRule:
 			}
 		}
 
+		targetsBranches := rule.Target != nil && *rule.Target == "branch"
 		for _, cond := range rule.Conditions.RefName.Include {
-			if cond == ruleConditionAllBranches {
+			if cond == ruleConditionAll && targetsBranches {
```
spencerschrock: Just to clarify, the issue was only a problem with tag protection rules which apply to all tags?

This is what I see with the graphql explorer for your repository:
```json
{
            "node": {
              "name": "Immutable tags with maintainer bypass",
              "enforcement": "ACTIVE",
              "target": "TAG",
              "conditions": {
                "refName": {
                  "exclude": [],
                  "include": [
                    "~ALL"
                  ]
                }
              },
```

I'm envisioning other cases where the tag include pattern would also match a branch name. So your approach in https://github.com/ossf/scorecard/pull/4699/commits/72db714db12c9d7d8dfa6d77106739ececc63622 may have been better?

clients/githubrepo/branches.go
```diff
@@ -481,8 +482,9 @@ nextRule:
 			}
 		}
 
+		targetsBranches := rule.Target != nil && *rule.Target == "branch"
```
spencerschrock: is this lower case for a reason? When I check one of my test repos I see it in all caps.
```
"target": "BRANCH",
```

If it's inconsistent, or we want to be more defensive, we use `strings.EqualFold` in other places for case insensitive comparisons.

clients/githubrepo/branches.go
trask: I'm good with the original approach, reverted to that

clients/githubrepo/branches.go
trask: oh good point, it should be uppercase, fixed

spencerschrock (approved):
Looks good to me, but you'll need to fix DCO https://github.com/ossf/scorecard/pull/4699/checks?check_run_id=45863683831

PR #4746, opened 2025-08-06
📖 Added a note explaining if you get a runtime error for Branch Protection

### LINKED ISSUE #2946: Feature: Improve error message when branch protection check fails because of the use of GITHUB_TOKEN

**Additional context**

dcaine125:
This PR introduce a doc update.

In the documentation in the CONTRIBUTING.md under Permission for GitHub personal access tokens, it doesn't have a clear warning telling people that if you get a runtime error, it because you didn't put in your personal access token.
Added a note explaining that you will need to put in a personal access token if the user gets that error. Also where to look for the command to implement the access token.

#2946

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -194,6 +194,10 @@ For public repos, classic personal access tokens need the following scopes:
 
 - `public_repo` - Read/write access to public repositories. Needed for branch protection
 
+Note: The `branch-protection` internal error will appear if a Personal Access Token (PAT) is not provided, because the default GitHub token lacks the necessary admin permissions to access certain branch protection settings.
+
+You can find instructions for setting the token in the README under Using Scorecard – Authentication.
+
 ## Where the CI Tests are configured
 
 1.  See the [action files](.github/workflows) to check its tests, and the


spencerschrock:
Can we pivot to documentation in the form of better error messages, like Jakiya asked here https://github.com/ossf/scorecard/issues/2946#issuecomment-3156350162?

The code in question is here:
https://github.com/ossf/scorecard/blob/f6ed084d17c9236477efd66e5b258b9d4cc7b389/clients/githubrepo/branches.go#L302-L304

We would want to check if the error contains the string "resource not accessible by integration", and if so, we can create a new error message which is more clear about what the problem is. 

Pseudo something like this (but with a better phrasing):
```go
	if err := handler.graphClient.Query(handler.ctx, queryData, vars); err != nil {
		if err contains resource not accessible {
			return nil, sce.WithMessage(sce.ErrScorecardInternal, "you probably need to use a PAT")
		}
		return nil, sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("githubv4.Query: %v", err))
	}
```

I actually just submitted a PR to check error substrings in #4747 if you want to take a look at the syntax.

PR #4777, opened 2025-09-04
:bug: add logic for pinned downloads across multiple commands

### LINKED ISSUE #4771: BUG: in-place edit with perl considered to be a `downloadThenRun`
**Describe the bug**

This code:
```shell
wget https://raw.githubusercontent.com/radiantearth/stac-spec/568a04821935cc92de7b4b05ea6fa9f6bf8a0592/item-spec/json-schema/itemcollection.json
perl -pi -e 's#"const": "0.9.0"#"const": "1.1.0"#g' itemcollection.json
```
triggers `Warn: downloadThenRun not pinned by hash: integration_tests/schemas/update.sh:50` where line 50 is the `perl -pi -e ..` line.

There is no running of the JSON file in the script, it downloads the file and performs an in-place edit on it.

I'm aware of issue #3401, but to me, that looks like people are trying to run the downloaded file after verifying they get the expected file, and that is quite different from editing the downloaded file.

**Reproduction steps**
1. Go to https://securityscorecards.dev/viewer/?uri=github.com/opendatacube/datacube-explorer
2. Expand  the pinned dependencies section
3. See the `Warn: downloadThenRun not pinned by hash: integration_tests/schemas/update.sh:50`

**Expected behavior**

No warning from editing a downloaded file in-place.

**Additional context**

spencerschrock:

bug fix

hasUnpinnedURLs was previously introduced in 0c40e14 but only when piping a file to an interpreter in the same command. But we also record downloaded files so we can check if they're used later, and this path wasn't included.

check for pinned URLs in this path too.

Fixes #4771

the `release-note`

```release-note
NONE
```


justaugustus (approved):
Thanks @spencerschrock!

PR #4793, opened 2025-09-16
:sparkles: Implement --org and --repos for scanning organizations or multiple repositories

### LINKED ISSUE #4792: Feature: Support Scanning Github Organizations and Multiple Repositories
When using the CLI, there is no way to scan multiple repositories other than by executing the command once for each repository. For example, I would like to scan all the repositories of my own organization.

The most obvious solution to this problem is to enable a parameter for specifying a list of repositories in the form of a list, and then execute the scan the same way as before. As a further step, we should be able to provide the name of an organization and dynamically fetch all the repositories belonging to that organization.

gabrielsoltz:

Fixes #4792 

the `release-note`

```release-note

```


clients/githubrepo/roundtripper/roundtripper.go
```diff
@@ -68,3 +68,8 @@ func NewTransport(ctx context.Context, logger *log.Logger) http.RoundTripper {
 
 	return MakeCensusTransport(MakeRateLimitedTransport(transport, logger))
 }
+
+// TransportFactory is used to create an http.RoundTripper. It defaults to
+// calling NewTransport but can be overridden in tests to provide a custom
+// RoundTripper that redirects requests to a test server.
+var TransportFactory = NewTransport
```
spencerschrock: I dont think we need this. Most mocking in this repo is done with dependency injection, so whatever uses the round tripper should have one injected in a test.

E.g.
https://github.com/ossf/scorecard/blob/2864c762f36e30ff78b18ae59acd96fcf3b9a9bb/clients/githubrepo/webhook_test.go#L28-L42

https://github.com/ossf/scorecard/blob/2864c762f36e30ff78b18ae59acd96fcf3b9a9bb/clients/gitlabrepo/search_test.go#L172-L176

go.mod
```diff
@@ -41,6 +41,7 @@ require (
 	github.com/caarlos0/env/v6 v6.10.1
 	github.com/gobwas/glob v0.2.3
 	github.com/google/go-github/v53 v53.2.0
+	github.com/google/go-github/v60 v60.0.0
```
spencerschrock: why introduce a second version when the function you need is available in the version we already use?
https://pkg.go.dev/github.com/google/go-github/v53/github#RepositoriesService.ListByOrg

clients/githubrepo/org.go
```diff
@@ -0,0 +1,80 @@
...
+	"net/http"
+
+	"github.com/google/go-github/v60/github"
+
+	"github.com/ossf/scorecard/v5/clients/githubrepo/roundtripper"
+	"github.com/ossf/scorecard/v5/log"
+)
+
+// ListOrgRepos lists all non-archived repositories for a GitHub organization.
+func ListOrgRepos(ctx context.Context, org string) ([]string, error) {
```
spencerschrock: can we move `org.go` and `org_test.go` to `cmd/internal/org/org.go` and `cmd/internal/org/org_test.go`.

Any code we can keep out of the public is good.

cmd/internal/org/org.go
```diff
@@ -0,0 +1,80 @@
...
+		opt.Page = resp.NextPage
+	}
+
+	return urls, nil
+}
+
+// parseOrgName extracts the organization name from a GitHub URL or returns the input if already an org name.
+func parseOrgName(input string) string {
+	// Remove "github.com/" prefix if present
+	const prefix = "github.com/"
```
spencerschrock: other Scorecard arguments accept URLs with schmes. So I would want you to test this works with something like:
```
http://github.com/gabrielsoltz
https://github.com/gabrielsoltz
```

clients/githubrepo/roundtripper/roundtripper.go
spencerschrock: And then when you make a client you can still call this by default, but if the test gives you a round tripper you use it instead.
https://github.com/ossf/scorecard/blob/44f521d287601a8f4d72cb5b107a37d5890fd0c8/clients/githubrepo/client.go#L347-L359

cmd/root.go
```diff
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package cmd implements Scorecard command-line.
```
spencerschrock, line 15: please dont delete comments unrelated to your change.

cmd/root.go
```diff
@@ -76,59 +75,76 @@ func New(o *options.Options) *cobra.Command {
...
-			return fmt.Errorf("making remote repo: %w", err)
-		}
+	// Build the list of repos (only split this logic out)
+	repoURLs, err := buildRepoURLs(ctx, o)
+	if err != nil {
+		return err
+	}
+
+	// Shared setup (unchanged)
+	pol, err := policy.ParseFromFile(o.PolicyFile)
```
spencerschrock: unchanged from what?

cmd/root.go
```diff
@@ -76,59 +75,76 @@ func New(o *options.Options) *cobra.Command {
...
 	}
-	// if commit option set to anything other than HEAD add commit based
 	if !strings.EqualFold(o.Commit, clients.HeadSHA) {
 		requiredRequestTypes = append(requiredRequestTypes, checker.CommitBased)
 	}
-	// this call to policy is different from the one in scorecard.Run
-	// this one is concerned with a policy file, while the scorecard.Run call is
-	// more concerned with the supported request types
+
 	enabledChecks, err := policy.GetEnabled(pol, o.Checks(), requiredRequestTypes)
```
spencerschrock, line 162: please dont delete comments unrelated to your change.

spencerschrock (commented):
Did a partial review of a few things before I ran out of time, I need to dive more into how you build the repo list, scan repos, and present the results. Feel free to address that feedback now or wait until I have more time for full review

spencerschrock:
Also, this repo requires DCO.
https://github.com/ossf/scorecard/blob/2864c762f36e30ff78b18ae59acd96fcf3b9a9bb/CONTRIBUTING.md?plain=1#L10-L14

For instructions on how to fix it:
https://github.com/ossf/scorecard/pull/4793/checks?check_run_id=50569515731

cmd/root.go
gabrielsoltz: Sorry for this, I restored them ✅

cmd/root.go
gabrielsoltz: Based on the previous behavior before my change, I'll remove that comment now because I see it's confusing.

cmd/root.go
gabrielsoltz: restored ✅

clients/githubrepo/org.go
gabrielsoltz: I'm not quite sure I understand this. But if you think that's the correct place for it, I'll move it... My logic was to keep all related github code together....

go.mod
gabrielsoltz: Right. Changed. ✅

clients/githubrepo/org.go
gabrielsoltz: Moved ✅

cmd/root.go
gabrielsoltz: Deleted that part of the comment ✅

cmd/internal/org/org.go
gabrielsoltz: Improved and added test cases ✅

clients/githubrepo/roundtripper/roundtripper.go
gabrielsoltz: This was also changed; still not sure if my implementation is the expected one.

spencerschrock:
Just wanted to update you that this is still on my radar! We are trying to cut a 5.3.0 release this week, so my attention has been on that. But once that's cut this week, I will make time to finish. my review, and happy to cut a 5.4.0 or 5.3.1 shortly after

cmd/internal/org/org.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// Strip optional scheme or leading "//".
+	switch {
+	case strings.HasPrefix(s, "https://"):
+		s = strings.TrimPrefix(s, "https://")
+	case strings.HasPrefix(s, "http://"):
+		s = strings.TrimPrefix(s, "http://")
+	case strings.HasPrefix(s, "//"):
+		s = strings.TrimPrefix(s, "//")
```
spencerschrock: out of curiosity, what case is this intended to solve?

options/flags.go
```diff
@@ -78,6 +80,8 @@ const (
 	FlagCommitDepth = "commit-depth"
 
 	FlagProbes = "probes"
+
+	FlagCombinedOutput = "combined"
```
spencerschrock: is this intended to be only change the behavior if `--format=default`? Currently when I run something like this, I get the "combined" output instead of "json"
```
go run main.go --org=ossf-tests --format json --combined
```

Or should this be it's own format value? e.g. 
```
go run main.go --org=ossf-tests --format combined
```

options/flags.go
```diff
@@ -95,6 +99,21 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
 		"repository to check (valid inputs: \"owner/repo\", \"github.com/owner/repo\", \"https://github.com/repo\")",
 	)
 
+	cmd.Flags().StringSliceVar(
+		&o.Repos,
+		FlagRepos,
+		o.Repos,
+		"repositories to check (<repo1>, <repo2>, ...). Each repo must be one of:"+
+			" \"owner/repo\", \"github.com/owner/repo\", \"https://github.com/repo\"",
```
spencerschrock: nit:

since this is a comma separated list, the space you have in your example would break it: `<repo1>, <repo2>, ...`
This should be `<repo1>,<repo2>, ...`

options/flags.go
```diff
@@ -95,6 +99,21 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
...
+	)
+
+	cmd.Flags().StringVar(
+		&o.Org,
+		FlagOrg,
+		o.Org,
+		"GitHub organization to check (all non-archived repos will be checked), e.g., 'github.com/ossf' or 'ossf'",
+	)
```
spencerschrock:
```
"scans all non-archived repositories in an organization. Currently only supports GitHub, e.g., 'github.com/ossf' or 'ossf'"
```

pkg/scorecard/scorecard_result.go
```diff
@@ -263,6 +266,157 @@ func (r *Result) AsString(writer io.Writer, checkDocs docChecks.Doc, opt *AsStri
...
+	}
+	table.SetHeader(header)
+	table.SetBorders(tablewriter.Border{Left: true, Top: true, Right: true, Bottom: true})
+	table.SetRowSeparator("-")
+	table.SetRowLine(true)
+	table.SetCenterSeparator("|")
+	table.AppendBulk(data)
+	table.SetAlignment(tablewriter.ALIGN_LEFT)
+	table.SetRowLine(true)
+	table.Render()
```
spencerschrock: this all got changed when I had to merge #4796 sorry.

cmd/root.go
```diff
@@ -76,39 +79,65 @@ func New(o *options.Options) *cobra.Command {
 	return cmd
 }
 
-// rootCmd runs scorecard checks given a set of arguments.
-func rootCmd(o *options.Options) error {
-	var err error
-	var repoResult scorecard.Result
+// Build the list of repositories to scan, honoring --repos > --org > --local > --repo/pkg-managers.
+func buildRepoURLs(ctx context.Context, o *options.Options) ([]string, error) {
+	// --repos has highest precedence
```
spencerschrock: FYI: since only one option can be set at a time, precedence doesn't really matter here.

cmd/root.go
```diff
@@ -76,39 +79,65 @@ func New(o *options.Options) *cobra.Command {
...
+		}
+		return urls, nil
+	}
+
+	// --org: expand to all non-archived repos
+	if o.Org != "" {
+		// create a transport to respect auth, rate limiting, etc.
+		logger := sclog.NewLogger(sclog.DefaultLevel)
+		rt := roundtripper.NewTransport(ctx, logger)
+		repos, err := orgpkg.ListOrgRepos(ctx, o.Org, rt)
```
spencerschrock: you wrote `ListOrgRepos` to do all this for you if `rt == nil` so might as well just do:
```
repos, err := orgpkg.ListOrgRepos(ctx, o.Org, nil)
```

cmd/root.go
```diff
@@ -76,39 +79,65 @@ func New(o *options.Options) *cobra.Command {
...
 	if pkgResp.exists {
 		o.Repo = pkgResp.associatedRepo
 	}
 
-	pol, err := policy.ParseFromFile(o.PolicyFile)
-	if err != nil {
-		return fmt.Errorf("readPolicy: %w", err)
-	}
+	return []string{o.Repo}, nil
+}
```
spencerschrock, line 125: just a comment for myself: doing package manager parsing before `--repo` is necessary since as you mention it overwrites `o.Repo`

cmd/root.go
```diff
@@ -158,68 +181,91 @@ func rootCmd(o *options.Options) error {
...
-		return fmt.Errorf("scorecard.Run: %w", err)
+	var allResults []*scorecard.Result
+	// Iterate and scan each repo using a helper to keep rootCmd small.
+	for _, uri := range repoURLs {
+		res, err := processRepo(ctx, uri, o, enabledProbes, enabledChecks, checks, opts, checkDocs, pol)
+		if err != nil {
+			// processRepo already logged details; skip this URI.
+			fmt.Fprintf(os.Stderr, "Skipping %s: %v\n", uri, err)
+			continue
+		}
```
spencerschrock, line 195: this changes the behavior of the `scorecard` binary. Previously if any check had a runtime error, `rootCmd` would return an error, which would lead to a non-zero exit code. But now these runtime errors don't lead to an exit code.

I think continuing to scan all repositories is good, but we should "remember" if we saw an error along the way so we can return it at the end

cmd/root.go
```diff
@@ -253,3 +299,85 @@ func makeRepo(uri string) (clients.Repo, error) {
...
+
+	label := repoLabelFromURI(uri)
+
+	// Start banners with repo label (always show banners even in combined-only mode)
+	if o.Format == options.FormatDefault {
+		if len(enabledProbes) > 0 {
+			printProbeStart(label, enabledProbes)
+		} else {
+			printCheckStart(label, enabledChecks)
+		}
```
spencerschrock: let's get rid of `repoLabelFromURI` and just use `uri` instead of `label`

Otherwise it would be confusing when you analyze repos across forges:
```
--repos=github.com/some/repo,gitlab.com/other/repo
```

Which one is github? Which one is gitlab? What if they both have the exact same repo name?
```
Starting (some/repo) [Maintained]
...
Starting (other/repo) [Maintained]
```

cmd/root.go
```diff
@@ -253,3 +299,85 @@ func makeRepo(uri string) (clients.Repo, error) {
...
+// It returns the Result when successful (or when combined output is requested),
+// or an error describing why the URI should be skipped.
+func processRepo(
+	ctx context.Context,
+	uri string,
+	o *options.Options,
+	enabledProbes []string,
+	enabledChecks checker.CheckNameToFnMap,
+	checksList []string,
+	opts []scorecard.Option,
```
spencerschrock: `checksList` is unused and should likely be removed

spencerschrock (commented):
I like this approach overall, I think building a slice of repos and looping over them is the correct approach. Most of my issue is with the `--combined` output format, is there anyway we can split that into its own PR? I think that would help get `--repos` and `--org` merged first.

cmd/internal/org/org.go
gabrielsoltz: This is not needed, deleted.

pkg/scorecard/scorecard_result.go
gabrielsoltz: Updated ✅

options/flags.go
gabrielsoltz: changed

options/flags.go
gabrielsoltz: changed

cmd/root.go
gabrielsoltz: comment changed

cmd/root.go
gabrielsoltz: Damn 🤦 This was the previous behavior.  Changed.

cmd/root.go
gabrielsoltz: Makes sense. I added a small implementation for this using `sawRuntimeErr`. We can improve it, but I think it already aligns with your point.

cmd/root.go
gabrielsoltz: Good point. And much easier. Changed.

gabrielsoltz: The only cosmetic thing is that running with --org, we will get as uri the full URL, like `https://github.com/ossf-tests/scorecard-check-pinned-dependencies-e2e`...

cmd/root.go
gabrielsoltz: I don't remember why this was here. Deleted.

options/flags.go
gabrielsoltz: This was not intended, but after reviewing, I definitely prefer your approach to treat combined as a new format option.

gabrielsoltz: This is now changed, now we use `--format combined`... I intentionally left the start/finish banner for the format default and combined. Maybe these banners should be optional, using another flag (`--show-progress` or something like that).

gabrielsoltz:
Hey @spencerschrock, thank you so much for reviewing my PR — and sorry it took me a while to address your comments; I couldn’t get to it earlier. I left two small cosmetic notes (one about the banners and one about the URI) for us to discuss. Whenever you have a moment, this is ready for another review. ✌️

spencerschrock:
> Whenever you have a moment, this is ready for another review

I will get to this tomorrow, thanks for your patience.

cmd/root.go
```diff
@@ -253,3 +278,82 @@ func makeRepo(uri string) (clients.Repo, error) {
...
+	sort.Slice(result.Checks, func(i, j int) bool {
+		return result.Checks[i].Name < result.Checks[j].Name
+	})
+
+	// End banners BEFORE RESULTS (always show banners even in combined-only mode)
+	if o.Format == options.FormatDefault {
+		if len(enabledProbes) > 0 {
+			printProbeResults(uri, enabledProbes)
+		} else {
+			printCheckResults(uri, enabledChecks)
```
spencerschrock: as written, this only prints the results for the default format, but the comment makes me think it should also apply to the combined format:

```
if o.Format == options.FormatDefault || o.Format == options.FormatCombined {
```

cmd/root.go
```diff
@@ -253,3 +278,82 @@ func makeRepo(uri string) (clients.Repo, error) {
...
+		if err := scorecard.FormatResults(o, &result, checkDocs, pol); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to format results for %s: %v\n", uri, err)
+		}
+	}
+
+	// Surface per-check runtime errors (non-fatal)
+	for _, r := range result.Checks {
+		if r.Error != nil {
+			fmt.Fprintf(os.Stderr, "Check %s failed for %s: %v\n", r.Name, uri, r.Error)
+		}
```
spencerschrock: let's remove this and handle it in `rootCmd`

cmd/root.go
spencerschrock: Thanks, I may change the implementation in a follow-up as it would just be more effective for me to handle my suggested changes.

options/options.go
```diff
@@ -48,6 +50,7 @@ type Options struct {
 	CommitDepth     int
 	ShowDetails     bool
 	ShowAnnotations bool
+	CombinedOutput  bool
```
spencerschrock: with the switch to using the existing format flag, this `CombinedOutput` field is unused

options/options.go
```diff
@@ -83,7 +86,9 @@ const (
 	FormatDefault = "default"
 	// FormatRaw specifies that results should be output in raw format.
 	FormatRaw = "raw"
-	// FormatInToyo specifies that results should be output in an in-toto statement.
+	// FormatCombined specifies that results should be output as a single combined table for multiple repos.
+	FormatCombined = "combined"
```
spencerschrock: Can we imply this is just for the table output since the other formats naturally combine their output?
```
FormatCombinedTable = "combined-table"
```

pkg/scorecard/scorecard_result.go
```diff
@@ -274,6 +277,179 @@ func (r *Result) AsString(writer io.Writer, checkDocs docChecks.Doc, opt *AsStri
...
+// FormatCombinedResults prints a combined table with extra REPO and AGGREGATED SCORE columns.
+// This expects `results` to contain checks for a single repo; callers scanning
+// multiple repos should aggregate calls or invoke this helper appropriately.
+func FormatCombinedResults(
+	writer io.Writer,
+	results *Result,
+	checkDocs docChecks.Doc,
+	opt *AsStringResultOption,
+	opts *options.Options,
+) error {
```
spencerschrock: one of the reasons I suggested splitting this into a separate PR was to get the other code merged, since I knew I would have trickier feedback on this format that would take another cycle or two.

1. All the other format options are methods on results. Is there a reason you didn't go with that approach instead of `FormatCombinedResults`:
```
func (r *Result) AsCombinedTable(writer io.Writer, checkDocs docChecks.Doc, opt *AsStringResultOption) error
```

2. There's a lot of code overlap with `AsString`, the main difference is the two additional columns. Instead of duplicating the logic, can you make them use a shared helper function with a flag or something?

e.g. 

```
func (r *Result) AsString(writer io.Writer, checkDocs docChecks.Doc, opt *AsStringResultOption) error {
    combined := false
    return common(writer, checkDocs, opt, combined)
}
func (r *Result) AsCombined(writer io.Writer, checkDocs docChecks.Doc, opt *AsStringResultOption) error {
    combined := true
    return common(writer, checkDocs, opt, combined)
}

and then in `common` you can look at that bool to see if you should add the column or not
```

justaugustus (approved):
Approving for a22d201df097e0008196c279419e9f222b274f69.
ref: https://github.com/ossf/scorecard/pull/4793#pullrequestreview-3405877343

spencerschrock:
Thanks Stephen, and of course thanks Gabriel for the multi repo logic!

PR #4835, opened 2025-10-31
🐛 Fix branch-protection ruleset handling when there are no include patterns

trask:

Bug fix – repository rulesets without any include patterns should apply to all branches which aren't explicitly excluded

Scorecard ignores GitHub rulesets that rely on an empty `include` list (apply to all refs unless excluded), so branches covered only by those rulesets are reported as lacking protection. The branch-protection check emits false warnings such as “Warn: branch protection not enabled for branch 'xyz'”.

Rulesets with no explicit include patterns are now treated as applying to every ref except those explicitly excluded, matching GitHub’s semantics. Branches governed by such rulesets are marked protected and no longer generate false warnings.

None

None

```release-note
Fix branch-protection scoring so GitHub rulesets without include patterns are honored, eliminating false warnings for branches covered by those rulesets.
```

### DIFF
--- a/clients/githubrepo/branches.go
+++ b/clients/githubrepo/branches.go
@@ -497,9 +497,13 @@ nextRule:
 		}
 
 		includePatterns := rule.Conditions.RefName.Include
+		excludePatterns := rule.Conditions.RefName.Exclude
 		if len(includePatterns) == 0 {
-			// GitHub treats an empty include list as applying to all refs unless excluded.
-			ret = append(ret, rule)
+			// GitHub treats an empty include list with at least one exclude as applying to all refs unless excluded.
+			// If both include and exclude are empty, the ruleset doesn't apply to any branches.
+			if len(excludePatterns) > 0 {
+				ret = append(ret, rule)
+			}
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


spencerschrock:
> repository rulesets without any include patterns should apply to all branches which aren't explicitly excluded

Are you sure about that? This is what I see on a test repo when making a ruleset without a target:

> This ruleset does not target any resources and will not be applied.

trask:
> Are you sure about that? This is what I see on a test repo when making a ruleset without a target:
> 
> > This ruleset does not target any resources and will not be applied.

ah, good point, fixed: https://github.com/ossf/scorecard/pull/4835/commits/677aec35305d673203af3306d283c30ab630422e

spencerschrock:
Is this intentionally a draft PR?

trask:
yeah, sorry, was going to confirm if we really needed it in OpenTelemetry or not (I think we may need to restructure our rulesets unrelated to this issue). do you prefer that I close until then or would you prefer I just mark it ready for review?

spencerschrock:
Up to you, I mainly wanted to make sure this PR wasn't waiting for a review which wasn't assigned.

But this fixes an edge case I didn't know about, so happy to merge it regardless of if OpenTelemetry ends up needing it.

PR #4848, opened 2025-11-11
:seedling: osv: re-enable transitive scanning without pypi

spencerschrock:

config tweak

All transitive dependency scanning was disabled

Only the Python transitive scanning is disabled.

Plugins are a new osv-scanner/osv-scalibr way of disabling specific features, which give us finer control over what features to toggle.

NONE

the `release-note`

```release-note
Re-enabled transitive scanning in osv-scanner v2 except for PyPI
```

### DIFF
--- a/clients/osv.go
+++ b/clients/osv.go
@@ -61,10 +61,10 @@ func (v osvClient) ListUnfixedVulnerabilities(
 		GitCommits:        gitCommits,
 		CompareOffline:    v.local,
 		DownloadDatabases: v.local,
+		// swap out the transitive requirements scanning for offline extractor
 		ExperimentalScannerActions: osvscanner.ExperimentalScannerActions{
-			TransitiveScanningActions: osvscanner.TransitiveScanningActions{
-				Disabled: true,
-			},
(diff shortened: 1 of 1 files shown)


clients/osv.go
```diff
@@ -62,9 +62,7 @@ func (v osvClient) ListUnfixedVulnerabilities(
 		CompareOffline:    v.local,
 		DownloadDatabases: v.local,
 		ExperimentalScannerActions: osvscanner.ExperimentalScannerActions{
-			TransitiveScanningActions: osvscanner.TransitiveScanningActions{
-				Disabled: true,
-			},
+			PluginsDisabled: []string{"python/requirementsenhanceable"},
```
spencerschrock, line 67: this apparently will disable all requirements.txt scanning, not just transitive

cuixq: to only enable the offline extractor we can add [`python/requirements`](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/python/requirements/requirements.go) to the enable list

Issue #2458, opened 2022-11-15
Feature: New check for average time to update dependencies


olivekl:
Proposal for a new check: project’s mean time to update (MTTU) — what is the average number of days it takes a project to update to a new version of a dependency?

The recent [Sonatype State of the Supply Chain report](https://www.sonatype.com/state-of-the-software-supply-chain/project-quality-metrics) states that 6 out of 7 vulnerabilities are transitive. The report says that “Two critical factors for reducing the risk of transitive vulnerabilities are **minimizing the total number of dependencies** and **maintaining low update times**.”

The MTTU check would give consumers an easy way to judge this risk in a new dependency. 

Potentially it could be combined with a check that uses deps.dev data to also give the total number of dependencies a project uses.

laurentsimon:
Great idea. 

Something simpler (to start with) could be to look at how many versions / how old each dependency is at?

But if we have data available to calculate this average, that'd be really nice.

/cc @oliverchang @another-rex @distractible

laurentsimon:
One limitation I see here is that most open source projects are "libraries" so they declare their dependencies using version ranges. That will make it harder to infer when they update dependencies, because most of the time there is not update required, ie a dependency's new release still be in the range specified by the project.

olivekl:
Bumping this because it's getting a lot of mentions at the Open Source Summit Europe conference!

PR #4868, opened 2025-12-03
🌱 New check: Mean time to update dependencies

AdamKorcz:

New check

This PR adds a new check with 3 underlying probes to scorecard. The check is an implementation of https://github.com/ossf/scorecard/issues/2458. In essence, the check works like this:

1. First, it collects a list of all the project's direct dependencies at the time of checking.
2. Then, it calls deps.dev for a list of releases of each dependency.
3. Then, the check assesses whether each dependency is the latest available version according to deps.dev.
4. With the dependencies that are not the latest available version, the check assesses how much time has passed between the release date of the version that the project uses and the release date of the version right after the project uses. For example, if a project uses 1.2.3, and 1.2.4 was available 100 days ago, the check will record 100 days, even though 1.2.5 is the newest available. This is a design part on my end.
5. Finally, the check creates an average of the dependencies that are not the latest available version and scores accordingly. 

As such, the check considers the negative findings in a project - the dependencies that have not been updated to its latest version - and the check can be considered one that reports how old a project's dependencies are rather than providing a full picture of how quickly a project updates its dependencies. 

A few notes on implementation details:

1. The check relies heavily on osv-scalibr and the deps.dev API which means that any shortcomings in these two libraries will have direct impact on the check. The most significant shortcoming I have found is that the check doesn't always process all dependencies that it finds in a dependency manifest. I haven't debugged this extensively. At the same time, osv-scalibr and deps.dev seem like great sources for this check, and a few edge cases shouldn't in my opinion prevent us from using them. Even if the check doesn't process all dependencies, it is still useful by assessing the majority of a projects dependencies. Initially, I was working with distinct parsers for different lock file types, but this is in my opinion not the way to go. osv-scalibr provides a lot of value out of the box, and given that it is a fairly new library, I think it is fair to expect that it will continue its development.
2. The check only considers direct dependencies. This is a design decision as I don't see a point in evaluating transitive dependencies; The goal of the check is to determine a project's maintenance practices - not to find old dependencies. As such, I have made the decision to give maintainers as much control I could over their score; by only considering direct dependencies, maintainers can simply update their dependencies. If we were to consider transitive dependencies too, maintainers are at the mercy of their direct dependencies to keep _their_ dependencies up to date.

Fixes https://github.com/ossf/scorecard/issues/2458

I would recommend reading the documentation in the `.yml and .yaml` files in this PR and testing the check with `go run main.go --repo=github.com/owner/repo --checks=MTTUDependencies` before reviewing the code.

the `release-note`

```release-note

```

### DIFF
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -40,6 +40,7 @@ type RawResults struct {
 	SBOMResults                 SBOMData
 	MaintainedResults           MaintainedData
 	Metadata                    MetadataData
+	MTTUDependenciesResults     MTTUDependenciesData
 	PackagingResults            PackagingData
 	PinningDependenciesResults  PinningDependenciesData
 	SASTResults                 SASTData
@@ -300,6 +301,36 @@ type SASTWorkflow struct {
 // for the Security-Policy check.
 type SecurityPolicyData struct {
 	PolicyFiles []SecurityPolicyFile
+}
+
+type MTTUDependenciesData struct {
+	Dependencies []LockDependency
+}
+
+// Ecosystem represents the ecosystem of a dependency.
+type Ecosystem string
+
+const (
+	EcosystemCargo    Ecosystem = "CARGO"
+	EcosystemGo       Ecosystem = "GO"
+	EcosystemMaven    Ecosystem = "MAVEN"
+	EcosystemNPM      Ecosystem = "NPM"
+	EcosystemNuget    Ecosystem = "NUGET"
+	EcosystemPypi     Ecosystem = "PYPI"
+	EcosystemRubyGems Ecosystem = "RUBYGEMS"
+)
+
+func (e Ecosystem) String() string {
+	return string(e)
+}
+
+type LockDependency struct {
+	TimeSinceOldestReleast time.Time
+	IsLatest               *bool
+	Name                   string
+	Version                string
+	Comparator             string
+	Ecosystem              Ecosystem
 }
 
 // BinaryArtifactData contains the raw results
--- a/checks/MTTUDependencies.go
+++ b/checks/MTTUDependencies.go
@@ -1,0 +1,64 @@
+// Copyright 2025 OpenSSF Scorecard Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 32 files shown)


spencerschrock:
> osv-scalibr and deps.dev seem like great sources for this check

Agreed. Seems like a good usage of existing libraries to do the heavy lifting for us.

> If we were to consider transitive dependencies too, maintainers are at the mercy of their direct dependencies to keep their dependencies up to date

Also agree generally.

> As such, the check considers the negative findings in a project - the dependencies that have not been updated to its latest version - and the check can be considered one that reports how old a project's dependencies are rather than providing a full picture of how quickly a project updates its dependencies.

Any idea how major versions work with scalibr parsing or version reporting from deps dev? These might be heavier lifts than a minor/patch update.


One reservation with this check is it might encourage upgrade churn for the purpose of churn. But this puts it in a similar boat as the dependency upgrade tool check. For these situational checks, I think we benefit from having the customizability in place and considering what runs by default.

martincostello:
Something to consider for .NET if it isn't accounted for already, is that libraries may match the dependencies for different target frameworks to the specific major version they target to maximise compatibility.

For example, if a library targets `net8.0`, `net9.0` and `net10.0`, it may depend on System.Text.Json `8.0.x`, `9.0.x` and `10.0.x` respectively for those target frameworks.

In that scenario, I wouldn't expect a library to be marked down for using `8.0.x` and `9.0.x` when a `10.0.x` is available.

Microsoft will still support the latest patch version of each major version during their support lifetime ([which overlap](https://dotnet.microsoft.com/platform/support/policy/dotnet-core#cadence)).

For a specific example, if a project was using [System.Text.Json@8.0.6](https://www.nuget.org/packages/System.Text.Json/8.0.6), which is the most up-to-date version for .NET 8, it shouldn't be penalised for not using [System.Text.Json@9.0.0](https://www.nuget.org/packages/System.Text.Json/9.0.0) which was released on 2024-11-12 (394 days ago as of writing).

---

Separately, there's also the question of whether a library that "pins" a version because it wants to maintain compatibility and leave what the specific version used at runtime is to the consuming application should be penalised for doing so. The example of that would be similar to above, except `8.0.0`, `9.0.0` and `10.0.0` would be used indefinitely (unless there was a specific security patch to adopt). That would mean that within 100 days of a major release of .NET (typically ~around the 4th patch, e.g. `10.0.4` assuming a monthly patch release cadence), the dependency would be flagged for the remanining 20 (STS)/32(LTS) months of the major release's support lifetime.

Issue #30, opened 2020-10-21
New check: code is scanning for secrets


kimsterv:

mwarkentin:
A check that something like trufflehog (or other secret scanners) are running would be nice:

* https://github.com/dxa4481/truffleHog
* https://github.com/marketplace/actions/trufflehog-actions-scan

inferno-chromium:
Some examples from Google VRP program, see here - https://docs.google.com/document/d/10GWKW55YD6ZmHcJmQ2ZFu7jv75zyasHvx2O3yyRrfDI/edit

naveensrinivasan:
> Some examples from Google VRP program, see here - https://docs.google.com/document/d/10GWKW55YD6ZmHcJmQ2ZFu7jv75zyasHvx2O3yyRrfDI/edit

Don't have access to the doc.

laurentsimon:
We may check for the presence of this .gitignore and be sure sensitive files like private keys, password, etc are listed.

laurentsimon:
There's also https://github.blog/2022-12-15-leaked-a-secret-check-your-github-alerts-for-free/, which shows a setting we could use.

afmarcum:
This feature does not align with the current project focus. If there is no feedback in the next 7 days to the contrary, then this issue will be closed.

spencerschrock:
Keeping open as there was interest here: #3399

lucasgonze:
Unless this can check for Snyk secret scanning, the output will be noisy.

Danajoyluck:
TAC requested adding secret scanning and push protection to the security baseline,  ossf/tac#333. This check will be a super helpful verification and audit tool

PR #4878, opened 2025-12-12
🐛 New check: Secret scanning

AdamKorcz:

New check

https://github.com/ossf/scorecard/issues/30

This PR adds a new check for secret scanning. A project will score a high score if they have secret scanning and it runs it. The check covers multiple different ways of scanning for secrets:

1. Native GitHub scanning. This can be managed in repositories' "Advanced Security" settings under "Secret Protection".
2. Native Gitlab scanning. The check supports two different ways of protecting Gitlab repositories from leaked secrets: 1) Push protection and 2) The "Prevent Secrets" in "Push Rules".
3. Besides native repository configurations, the check supports a series of CI workflows: 
3.1 [gitleaks](https://github.com/gitleaks/gitleaks)
3.2 [TruffleHog](https://github.com/trufflesecurity/trufflehog)
3.3 [detect-secrets](https://github.com/Yelp/detect-secrets)
3.4 [git-secrets](https://github.com/awslabs/git-secrets)
3.5 [ggshield](https://github.com/GitGuardian/ggshield)
3.6 [sshgit](https://github.com/eth0izzle/shhgit)
3.7 [repo-supervisor](https://github.com/okta-graveyard/repo-supervisor)

Of these tools, the Scorecard Secret-Scanning check will both check if the project has either of these workflows - which will give the project a 1 score - and also whether the tools run. The first 5 tools are supposed to run on commits, and the last 2 run periodically. If they also run, the project will score another 7 points. Currently, the motivation to not score 10 in this case is that projects should enable Native GitHub scanning.

This can be tested out with `go run main.go --checks=SecretScanning --repo=github.com/owner/repo`. Try for example to enable and disable native GitHub secret scanning in your own repository and see the project score 0 when secret scanning is disabled and 10 when it is enabled.

In addition, try to run it on projects with some of the CI tools enabled or use my test repositories - each should score 1 for simply having the workflow and not having native GitHub secret scanning:

1. https://github.com/AdamKorcz/gh-gitleaks
2. https://github.com/AdamKorcz/gh-trufflehog
3. https://github.com/AdamKorcz/gh-detect-secrets
4. https://github.com/AdamKorcz/gh-git-secrets
5. https://github.com/AdamKorcz/gh-ggshield
6. https://github.com/AdamKorcz/gh-shhgit
7. https://github.com/AdamKorcz/gh-repo-supervisor
8. https://github.com/AdamKorcz/repo-with-no-secret-scanning (should score 0 for having no secret scanning)

Currently, details like the scores are merely suggestions as a starting point for getting these formalized. Modifying them won't be a bit challenge, so if a project should score 3 for having a CI workflow without running it, that is fine.

Yes

```release-note
Add new check for secret scanning
```

### DIFF
--- a/checker/raw_result.go
+++ b/checker/raw_result.go
@@ -48,6 +48,7 @@ type RawResults struct {
 	TokenPermissionsResults     TokenPermissionsData
 	VulnerabilitiesResults      VulnerabilitiesData
 	WebhookResults              WebhooksData
+	SecretScanningResults       SecretScanningData
 }
 
 type MetadataData struct {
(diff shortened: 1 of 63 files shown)


spencerschrock (commented):
Thanks for the groundwork Adam, I've tried to organize my thoughts ahead of the community meeting today.

**High level:**
I think this should be added, especially with the focus on baseline measurement. .

If the native platform scanning is the most popular, this might be a case similar to the webhooks check where the lack of an admin token may stop this from being on by default.

**Mid level**

The probes are broken down much too fine-grained in my opinion, to the point where we have N copies of identical stuff, one for each tool/platform. This is similar to how the fuzzing probes were originally before they were combined: https://github.com/ossf/scorecard/pull/3877

At a high level I see 3 probes (names subject to change/tweak):
`hasNativeSecretScanning`
`hasSecretPushProtection`
`hasThirdPartySecretScanning`


**Scoring** 
If we're thinking of this as an admin check, I would envision full points for preventing pushes, less for only reactively scanning, and none for none. Everything in between would need to be discussed further.

**Low level**

I didn't really try to go that low in my skim, I saw a few things, but not going to mention them yet until we align on high/mid level.

PR #4883, opened 2025-12-16
:seedling: Set OSV User-Agent for scorecard cli and cron workers.

### LINKED ISSUE #4029: Specify a user agent for OSV.dev
The Scorecard weekly cron makes many API calls to osv.dev. The OSV team asked if we specify a unique user agent, and we don't.
It would be good to set one. There is a `RequestUserAgent ` variable we can set in the `osv` package. https://pkg.go.dev/github.com/google/osv-scanner@v1.7.1/pkg/osv#pkg-variables.

[osv-scanner](https://github.com/google/osv-scanner) sets user agents like:
`osv-scanner/1.7.1` (where v1.7.1 is the osv-scanner version)

It would be nice if scorecard set a user agent for both our CLI (`scorecard/{version}`) and our weekly cron workers (`scorecard-cron/{version}`). We have version info available to the various binaries via [version.GetVersionInfo](https://pkg.go.dev/sigs.k8s.io/release-utils/version#GetVersionInfo)

kash2104:
This PR introduces improvement to OSV API request by configuring versioned User-Agent for Scorecard.

Currently API requests to `osv.dev` are made without specifying a unique user-agent.

Now a distinct, versioned user agent is set for the OSV API request:
- `scorecard/{version}` for CLI
- `scorecard-cron/{version}` for cron workers
- Uses `GetId()` and `GetAliases()` in `clients/osv.go` as per the latest updates in the osv scanner package.

Fixes #4029 

`osv-scanner/1.9.2` has been installed.

No

the `release-note`

### DIFF
--- a/cron/internal/worker/main.go
+++ b/cron/internal/worker/main.go
@@ -99,7 +99,7 @@ type ScorecardWorker struct {
 	blacklistedChecks []string
 }
 
-func newScorecardWorker() (*ScorecardWorker, error) {
+func newScorecardWorker(vulnsClient clients.VulnerabilitiesClient) (*ScorecardWorker, error) {
 	var err error
 	sw := &ScorecardWorker{}
 	if sw.checkDocs, err = docs.Read(); err != nil {
@@ -134,7 +134,7 @@ func newScorecardWorker() (*ScorecardWorker, error) {
 	if sw.ossFuzzRepoClient, err = ossfuzz.CreateOSSFuzzClientEager(ossfuzz.StatusURL); err != nil {
 		return nil, fmt.Errorf("ossfuzz.CreateOSSFuzzClientEager: %w", err)
 	}
-	sw.vulnsClient = clients.DefaultVulnerabilitiesClient()
+	sw.vulnsClient = vulnsClient
 
 	apiBaseURL, err := config.GetAPIBaseURL()
 	if err != nil {
@@ -373,11 +373,14 @@ func main() {
 	info := version.GetVersionInfo()
 	actions := osvscanner.ExperimentalScannerActions{}
 	actions.RequestUserAgent = fmt.Sprintf("scorecard-cron/%s", info.GitVersion)
+	osvConfig := clients.OSVConfig{}
+	osvConfig.UserAgent = actions.RequestUserAgent
+	vulnsClient := clients.NewOSVClient(&osvConfig)
 	flag.Parse()
 	if err := config.ReadConfig(); err != nil {
 		panic(err)
 	}
-	sw, err := newScorecardWorker()
+	sw, err := newScorecardWorker(vulnsClient)
 	if err != nil {
 		panic(err)
 	}
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 require (
 	github.com/caarlos0/env/v6 v6.10.1
 	github.com/gobwas/glob v0.2.3
-	github.com/google/go-github/v53 v53.2.0
+	github.com/google/go-github/v82 v82.0.0
 	github.com/google/osv-scanner/v2 v2.3.2
 	github.com/hmarr/codeowners v1.2.1
 	github.com/in-toto/attestation v1.1.2
--- a/main.go
+++ b/main.go
@@ -16,20 +16,13 @@
 package main
 
 import (
-	"fmt"
 	"log"
-
-	"github.com/google/osv-scanner/v2/pkg/osvscanner"
-	"sigs.k8s.io/release-utils/version"
 
 	"github.com/ossf/scorecard/v5/cmd"
 	"github.com/ossf/scorecard/v5/options"
 )
 
 func main() {
-	info := version.GetVersionInfo()
-	actions := osvscanner.ExperimentalScannerActions{}
-	actions.RequestUserAgent = fmt.Sprintf("scorecard-cli/%s", info.GitVersion)
 	opts := options.New()
 	if err := cmd.New(opts).Execute(); err != nil {
 		log.Fatalf("error during command execution: %v", err)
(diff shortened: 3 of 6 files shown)


cron/internal/worker/main.go
```diff
@@ -25,7 +25,9 @@ import (
 	_ "net/http/pprof" //nolint:gosec
 	"strings"
 
+	"github.com/google/osv-scanner/pkg/osv"
```
spencerschrock: this should be `"github.com/google/osv-scanner/v2/pkg/osv"`

go.mod
```diff
@@ -41,6 +41,7 @@ require (
 	github.com/caarlos0/env/v6 v6.10.1
 	github.com/gobwas/glob v0.2.3
 	github.com/google/go-github/v53 v53.2.0
+	github.com/google/osv-scanner v1.9.2
 	github.com/google/osv-scanner/v2 v2.2.4
```
spencerschrock: when you fix this import this accidental import of v1 should go away

main.go
```diff
@@ -16,13 +16,19 @@
 package main
 
 import (
+	"fmt"
 	"log"
 
+	"github.com/google/osv-scanner/pkg/osv"
```
spencerschrock: same comment about v2 import

main.go
```diff
@@ -16,13 +16,19 @@
...
+	"sigs.k8s.io/release-utils/version"
+
 	"github.com/ossf/scorecard/v5/cmd"
 	"github.com/ossf/scorecard/v5/options"
 )
 
 func main() {
+	info := version.GetVersionInfo()
+	osv.RequestUserAgent = fmt.Sprintf("scorecard-cron/%s", info.GitVersion)
```
spencerschrock: you have `scorecard-cron/` here too, which is a copy/paste error

cron/internal/worker/main.go
kash2104: @spencerschrock  I went through the v2 version of osv but from v2 onwards variable `RequestUserAgent` is not present. This variable was present till v1.9.2

spencerschrock: I sent a chat message to an osv-scanner maintainer to see if I should create an issue upstream to bring back support for setting user agents. As we use the v2 version of the library, setting the v1 variable wont actually change anything

spencerschrock:
You can track the upstream issue here at https://github.com/google/osv-scanner/issues/2420

We'll need to wait for a new osv-scanner release before we can set it.

another-rex:
@Ly-Joey Can you take a quick look to confirm this is correct?

Ly-Joey:
`ExperimentalScannerActions.RequestUserAgent` is the right variable to set. But it seems `actions` isn't being passed into `osvscanner.DoScan()` or anything equivalent.
I'm not very familiar with the project but I think the user agent values would need to be propagated to [clients/osv.go](https://github.com/ossf/scorecard/blob/85483c21ffbb0f125cf1d16aa53f283d574f4ca5/clients/osv.go#L57) (through `VulnerabilitiesClient`)  where the osvscanner package is actually used to do the scanning.

spencerschrock:
> I'm not very familiar with the project but I think the user agent values would need to be propagated to [clients/osv.go](https://github.com/ossf/scorecard/blob/85483c21ffbb0f125cf1d16aa53f283d574f4ca5/clients/osv.go#L57) (through `VulnerabilitiesClient`) where the osvscanner package is actually used to do the scanning.

+1 to this. The current strategy was based on the old package global. We might need a new flag in our osv-client representing the source, that way when we initialize the client we can pass in a string to use for the user agent.

maybe something like this? And then tweaking DefaultVulnerabilitiesClient to use NewOSVClient for backwards compatibility. (and we can get rid of `ExperimentalLocalOSVClient` now)

```go
type OSVConfig struct {
	ExperimentalLocal bool
	UserAgent string
}

func NewOSVClient(config *OSVConfig) VulnerabilitiesClient {
	if config == nil { // some defaults }
	// store as needed
}
```

kash2104:
Will look into it.

kash2104:
@spencerschrock Have made the appropriate changes as per the reviews.

cron/internal/worker/main.go
```diff
@@ -325,6 +327,12 @@ func startMetricsExporter() (monitoring.Exporter, error) {
 }
 
 func main() {
+	info := version.GetVersionInfo()
+	actions := osvscanner.ExperimentalScannerActions{}
+	actions.RequestUserAgent = fmt.Sprintf("scorecard-cron/%s", info.GitVersion)
+	osvConfig := clients.OSVConfig{}
+	osvConfig.UserAgent = actions.RequestUserAgent
+	clients.NewOSVClient(&osvConfig)
 	flag.Parse()
```
spencerschrock: this client made with the config needs to be used by the code, I think you'll want to store it here:
https://github.com/ossf/scorecard/blob/ade5c77e27e66681a6501c667f3862d8a052db78/cron/internal/worker/main.go#L132

main.go
```diff
@@ -16,13 +16,24 @@
...
 )
 
 func main() {
+	info := version.GetVersionInfo()
+	actions := osvscanner.ExperimentalScannerActions{}
+	config := clients.OSVConfig{}
+	actions.RequestUserAgent = fmt.Sprintf("scorecard-cli/%s", info.GitVersion)
+	config.UserAgent = actions.RequestUserAgent
+	clients.NewOSVClient(&config)
 	opts := options.New()
```
spencerschrock: Lets move this all to `cmd/root.go`, a little above this block:
https://github.com/ossf/scorecard/blob/ade5c77e27e66681a6501c667f3862d8a052db78/cmd/root.go#L173-L179

And then you'll need to use the client you created with `NewOSVClient` by creating a new opt:
```go
scorecard.WithVulnerabilitiesClient(clients.NewOSVClient(&config)),
```

kash2104:
@spencerschrock Have done the appropriate changes.

spencerschrock:
Thank you!

PR #4892, opened 2025-12-24
:sparkles: Specific error message for branch protection permission failure.

### LINKED ISSUE #2946: Feature: Improve error message when branch protection check fails because of the use of GITHUB_TOKEN

kash2104:
This PR introduces a code change to provide specific error message when branch protection rule fails due to the use of `default GITHUB_TOKEN` which doesn't have permission to view branch protection rules.

feature

When the CI fails because of branch protection rules, it shows generic error message which doesn't give the users an idea of what the actual underlying issue is.

When error is shown, string matching is done based on which it shows error message specific to the failure of branch protection CI test.

Fixes #2946 
Here `strings.Contains()` is case-sensitive so need to adjust the substring or the error message string accordingly.
Yes

the `release-note`

```release-note
No change needs to be done from the user side. Users will now see a clear message when instead of generic internal error message when the branch protection check fails due to GITHUB_TOKEN persmission limitations.
```

### DIFF
--- a/clients/githubrepo/branches.go
+++ b/clients/githubrepo/branches.go
@@ -311,6 +311,9 @@ func (handler *branchesHandler) query(branchName string) (*clients.BranchRef, er
 	// Ignore permissions errors if we know the repository is using rulesets, so non-admins can still get a score.
 	queryData := new(branchData)
 	if err := handler.graphClient.Query(handler.ctx, queryData, vars); err != nil {
+		if strings.Contains(err.Error(), "resource not accessible by integration") {
+			return nil, sce.WithMessage(sce.ErrScorecardInternal, "Branch protection failed.Default token insufficient.Use PAT.")
+		}
 		// always report errors which aren't token permission related
 		if !isPermissionsError(err) {
 			return nil, sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("githubv4.Query: %v", err))


clients/githubrepo/branches.go
```diff
@@ -311,6 +311,9 @@ func (handler *branchesHandler) query(branchName string) (*clients.BranchRef, er
 	// Ignore permissions errors if we know the repository is using rulesets, so non-admins can still get a score.
 	queryData := new(branchData)
 	if err := handler.graphClient.Query(handler.ctx, queryData, vars); err != nil {
+		if strings.Contains(err.Error(), "resource not accessible by integration") {
+			return nil, sce.WithMessage(sce.ErrScorecardInternal, "Branch protection check failed. Need to use a PAT.")
```
AdamKorcz: The message doesn't convey that the token doesn't have permissions. Would be great to make that clearer.

kash2104: @AdamKorcz Have made the message more clear.

kash2104:
@AdamKorcz Any more changes required? Would appreciate another review.

kash2104:
@spencerschrock Would appreciate a review.

AdamKorcz (approved):

clients/githubrepo/branches.go
spencerschrock, line 324: I believe this was already fixed in https://github.com/ossf/scorecard/pull/4778. I didn't realize there were multiple issues assigned for this issue, so I didn't know to close #2946,  sorry!

Unless there's an edge case that's still throwing the errors, I'll plan on closing this out in a day or two

PR #4893, opened 2025-12-26
🌱 Maintained check: New probe: Inactive Maintainers

### LINKED ISSUE #2027: Feature: stale maintainers
A new check or as part of an existing check, it would be useful to alert when maintainers have become inactive and could be removed from the maintainers. This is likely to require a PAT with the relevant permissions.

/cc @jeffmendoza

AdamKorcz:

Fixes https://github.com/ossf/scorecard/issues/2027.

New feature

This PR adds a new probe to the Maintained check that assesses whether the project has any inactive maintainers. At a high level, the probe does two things in the following order:

1. First, it fetches the data about users with elevated privileges in the repository. These are the Maintainers of the project.
2. Next, it fetches and analyzes events in the repository for signals about the maintainers' activity going 6 months back. If a maintainer has had any activity in those 6 months, they have been active, otherwise the probe considers them inactive.

The probe scores proportionally based on the proportion between active and inactive maintainers.

The client handlers pull in as many activity signals as the API allows. The Gitlab client that Scorecard uses had some bugs with some signal types, so the Gitlab client handler uses raw requests instead of client methods. The GitHub client does not use graphql; I couldn't get it to be as efficient in getting the same details and batching the calls efficiently. 

At a bit of a lower level, the client handlers will go through all activity data they have fetched and will mark a maintainer active as soon as they see any activity by that maintainer. Once all maintainers are active, the handlers will not process anymore activity records. This is for efficiency.

The clients pull in a fairly comprehensive set of activity data from the repository. They do not include some GitHub enterprise-only activity types, manual job triggers (GitHub) and Direct PR/MR reviews (Gitlab does not make the timestamp for these available in the API response).

Some activity types require higher token permissions. If the check runs without those permissions, it will not consider those activity types.

#### How this probe differs from existing Maintained probes

Currently, the Maintained check has four probes:

1. one that checks whether a project is archived
2. one that checks whether there has been recent commits
3. one that checks for activities on issues by project members
4. one that checks whether the repo was created recently.

The existing probes focus on any activity by the maintainers as a group: Whether _any_ maintainer/project member has recent activity in issues for example. The probe in this PR analyzes whether any of the maintainers have had no activity in a long time. 

the `release-note`

```release-note
Add new check for inactive maintainers.
```


spencerschrock (commented):
> Some activity types require higher token permissions. If the check runs without those permissions, it will not consider those activity types.

Wouldn't you need an admin token to enumerate the maintainers? So none of the check would run without admin?

I could see this working as a probe in Maintained, which would only be enabled when running with an admin token.

In terms of data collection, did you consider using the existing client methods to look for maintainer activity, instead of just putting them in a new `GetMaintainerActivity` method?

AdamKorcz:
> In terms of data collection, did you consider using the existing client methods to look for maintainer activity, instead of just putting them in a new `GetMaintainerActivity` method?

Yes, but I found it unfeasible: The probe in this PR pulls in more data than the existing client handlers allow.

PR #4898, opened 2025-12-31
:bug: detect toJSON(github.event) in Dangerous-Workflow check

### LINKED ISSUE #3554: BUG: Dangerous-Workflow doesn't trigger on `toJSON(github.event)`
**Describe the bug**
The Dangerous-Workflow check currently only triggers when attacker-controlled data is explicitly declared in the workflow.

https://github.com/ossf/scorecard/blob/130a31f/checks/raw/dangerous_workflow.go#L31-L57

However, if the `github.event` object is converted to a JSON object (i.e. for logging) in a `run:` command on a workflow triggered by an external contribution, its contents will contain attacker-controlled data (i.e. the issue/PR title) which can be used for code injection in the workflow.

For example:

```yml
on:
  pull_request_target:

jobs:
  get-pwned:
    runs-on: ubuntu-latest
    steps:
      - name: Log event and get pwned
        run: echo '${{ toJSON(github.event) }}' | jq
```

**Expected behavior**
Dangerous-Workflow should detect this as a dangerous pattern.

heathdutton:

Bug fix

The Dangerous-Workflow check does not detect script injection vulnerabilities when `toJSON(github.event)` or `toJSON(github)` is used in workflow run steps. These serialize the entire event/context object which contains attacker-controlled data.

The check now detects `toJSON(github.event)` and `toJSON(github)` patterns as dangerous, including case variations and whitespace.

Fixes #3554

The regex `(?i)tojson\s*\(\s*github(\s*\)|\.event)` matches:
- `toJSON(github)` - serializes entire context including event
- `toJSON(github.event)` and `toJSON(github.event.*)` - serializes event data
- Case insensitive and handles whitespace

It correctly ignores safe patterns like `toJSON(github.repository)`.

```release-note
Dangerous-Workflow check now detects toJSON(github.event) and toJSON(github) as dangerous patterns
```


checks/raw/dangerous_workflow.go
```diff
@@ -56,6 +56,13 @@ func containsUntrustedContextPattern(variable string) bool {
 	if strings.Contains(variable, "github.head_ref") {
 		return true
 	}
+	// Check for toJSON() serializing github context or event objects.
+	// These contain untrusted data and are dangerous in shell contexts.
+	// See https://github.com/ossf/scorecard/issues/3554
+	dangerousToJSONPattern := regexp.MustCompile(`(?i)tojson\s*\(\s*github(\s*\)|\.event)`)
```
spencerschrock: I think there's a small issue or two with this regex:
Your event alternate doesnt have the closing right parens, so it would have matched `github.events` or similar.

What do you think about:
`(?i)tojson\s*\(\s*github(?:\s*\)|\s*\.event\s*\))`

I made the alternate group non capturing, allowed some spaces before and after event, and look for a closing right paren

checks/raw/dangerous_workflow_test.go
```diff
@@ -105,6 +105,36 @@ func TestUntrustedContextVariables(t *testing.T) {
...
+		},
+		{
+			name:     "toJSON github context",
+			variable: "toJSON(github)",
+			expected: true,
+		},
+		{
+			name:     "toJSON github.event nested",
+			variable: "toJSON(github.event.pull_request)",
+			expected: true,
```
spencerschrock: Hmm, perhaps this is why you didnt close the right paren on event?

There are a lot of tricky cases here. For example:
`toJSON(github.event.pull_request.state)` would not be vulnerable, but detected by your original regex.

spencerschrock (approved):
I think if we just focus on `toJSON(github)` and `to.JSON(github.event)` (and not any other subfields) we avoid false positives for now, while we think of how to get the specificity right.

checks/raw/dangerous_workflow.go
heathdutton: Agreed

PR #4921, opened 2026-01-27
:sparkles: CLI flag to skip unsupported repo checks

### LINKED ISSUE #4795: Feature: Filter-out incompatible repository checks
When Scorecard is run against a GitLab repository, we expect checks incompatible with GitLab to either not execute or result in an inconclusive score. Currently, these incompatible checks run, and the output can be misleading.

My understanding is that Scorecard already [infers](https://github.com/ossf/scorecard/blob/64fa87dc8bca091d2ad11ac3629f56498b151f70/cmd/root.go#L228C6-L228C14) the type of repository it is dealing with. I'm unsure if all tests are currently executed to ensure consistent scores across different providers, I don't have the full context.

We could introduce a flag to filter out incompatible tests. Probably not the best solution, but a flag like `--repo-checks=gitlab` or a subcommand (_e.g._, `scorecard checks --repo=gitlab`) could be introduced to list applicable checks, which could then be used with Scorecard's existing `--checks` flag.

We currently use [checks.yaml](https://github.com/ossf/scorecard/blob/64fa87dc8bca091d2ad11ac3629f56498b151f70/docs/checks/internal/checks.yaml) to programmatically filter out incompatible tests, but we would prefer to avoid maintaining this ourselves.

**Additional context**
We run Scorecard in an air-gapped environment to score private repositories.

dvcorreia:

Introduces a new `--skip-unsupported-checks` CLI flag to skip checks that are not supported by a repo type (e.g. when running against a GitLab repo, it will skip checks that don't have GitLab support, not including them in the final report and score).

For a bit of context, we run Scorecard in an air-gapped environment to score private repositories hosted on a self-hosted GitLab CE instance. Since most of the checks are not supported by GitLab, the score will never be 10/10, being a bit deceptive to our developers in that they cannot do anything to fix that. This change allows the scores to be more representative of what developers can actually achieve for their repository.

When running Scorecard against non-GitHub repositories (e.g., GitLab, Azure DevOps, local), all checks are executed regardless of whether they support the repository type. This results in misleading scores where checks that are incompatible with the repo type fail, producing gaps in the score that developers cannot fix or address since their repository hosting platform doesn't support those checks.

With the new `--skip-unsupported-checks` flag, Scorecard will:
1. Detect the repository type (GitHub, GitLab, Azure DevOps, local)
2. Check each check's supported repository types in `checks.yaml`
3. Skip any checks that don't explicitly support the detected repository type
4. Only run supported checks, providing a score that accurately reflects what's achievable for that repository type

When the flag is enabled with the default format, skipped checks are printed to stderr for visibility: `Skipping (<repo-uri>) [<check-name>]`

Fixes #4795

When using this flag, the number of checks run may be significantly lower than the full suite, particularly for non-GitHub repositories. For GitLab, most checks already have experimental support but are not listed as supported in `checks.yaml`. This can probably be addressed in a follow-up PR (related issue #4700).

the `release-note`

```release-note
A new `--skip-unsupported-checks` CLI flag is now available to skip checks that don't support the repository type. When enabled, Scorecard will only run checks that explicitly support the detected repository type (GitHub, GitLab, Azure DevOps, or local). This provides more accurate scoring for non-GitHub repositories by excluding incompatible checks from the report and final score.
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -42,9 +42,12 @@ import (
 	"github.com/ossf/scorecard/v5/policy"
 )
 
-// errChecksFailed is returned when one or more checks produced a runtime
-// error during execution.
-var errChecksFailed = errors.New("one or more checks failed during execution")
+var (
+	// errChecksFailed is returned when one or more checks produced a runtime
+	// error during execution.
+	errChecksFailed      = errors.New("one or more checks failed during execution")
+	errNoChecksSupported = errors.New("no checks support repository type")
+)
 
 const (
 	scorecardLong = "A program that shows the OpenSSF scorecard for an open source software."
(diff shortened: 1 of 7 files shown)


dvcorreia:
This is a first iteration so if there's a better way to approach this, I'm happy to adjust things! I went with this implementation as it seemed like the least disruptive way to add the flag 😄

spencerschrock:
Let's talk in #4795 first (sorry on the delay, 100% on our end for this not being discussed yet), and then I'll come back to this PR

PR #4968, opened 2026-03-19
fix: detect committer name/email as untrusted input in Dangerous Workflow check

### LINKED ISSUE #3915: Dangerous Workflow: some user input are not being detected as untrusted input.
The current Dangerous Workflow check looks for places where untrusted user inputs could be used for Script Injection, such as:
```
- name: Check title
  run: |
    title="${{ github.event.issue.title }}"
    if [[ ! $title =~ ^.*:\ .*$ ]]; then
      echo "Bad issue title"
      exit 1
    fi
```

The problem is that Scorecard is not considering the whole range of user input that might lead to Script Injection. The [piece of code](https://github.com/ossf/scorecard/blob/70d045b9ef00e7171ce3950aca38eef6ea4d7308/checks/raw/dangerous_workflow.go#L30-L56) that lists the user inputs is the following:

```
func containsUntrustedContextPattern(variable string) bool {
        // GitHub event context details that may be attacker controlled.
        // See https://securitylab.github.com/research/github-actions-untrusted-input/
        untrustedContextPattern := regexp.MustCompile(
                `.*(issue\.title|` +
                        `issue\.body|` +
                        `pull_request\.title|` +
                        `pull_request\.body|` +
                        `comment\.body|` +
                        `review\.body|` +
                        `review_comment\.body|` +
                        `pages.*\.page_name|` +
                        `commits.*\.message|` +
                        `head_commit\.message|` +
                        `head_commit\.author\.email|` +
                        `head_commit\.author\.name|` +
                        `commits.*\.author\.email|` +
                        `commits.*\.author\.name|` +
                        `pull_request\.head\.ref|` +
                        `pull_request\.head\.label|` +
                        `pull_request\.head\.repo\.default_branch).*`)


                return true
        }
        return strings.Contains(variable, "github.event.") && untrustedContextPattern.MatchString(variable)
}
```
And I can point that it isn't able to detect the following inputs, for example:
- `${{ github.event.issue_comment.comment.body }}`, described in [github's doc](https://docs.github.com/en/webhooks/webhook-events-and-payloads#issue_comment)
- `${{ github.event.commit_comment.comment.body }}`, described in [github's doc](https://docs.github.com/en/webhooks/webhook-events-and-payloads#commit_comment)

As Dangerous Workflow is a very important check, Scorecard should be able to identify as many Script Injection risks as possible.

We should study more types of [GitHub Events, their payloads](https://docs.github.com/en/webhooks/webhook-events-and-payloads#about-webhook-events-and-payloads), and make Scorecard as reliable as possible to identify different ways that Script Injections could take place.

**Additional context**
I'd be happy to complete my search and also raise a PR to fix this, but I won't be able to do this in a near future. However, I'm marking this as "Good First issue", as the biggest effort to solve this would be actually a research on GitHub's API. The changes on Scorecard would be simple.

alexchenai:
## Summary

- Adds missing `committer.name` and `committer.email` patterns to the `containsUntrustedContextPattern` function for both `head_commit` and `commits[*]` contexts
- The existing check detects `author.name`/`author.email` but misses the equivalent `committer` fields, which are independently controllable by attackers and equally dangerous for script injection
- Includes 4 new unit test cases covering all added patterns

Fixes #3915 (partially -- addresses the committer gap identified by @pnacht in the issue comments)

## Context

As noted by @pnacht in https://github.com/ossf/scorecard/issues/3915#issuecomment-2179226710, git committer identity is a separate field from author identity. An attacker can set arbitrary committer name/email values (e.g., via `GIT_COMMITTER_NAME`/`GIT_COMMITTER_EMAIL`), making these variables unsafe to use in inline shell scripts without sanitization.

## Test plan

- [x] Added unit tests for `head_commit.committer.name`, `head_commit.committer.email`, `commits[*].committer.name`, `commits[*].committer.email`
- [ ] Existing tests continue to pass

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/checks/raw/dangerous_workflow.go
+++ b/checks/raw/dangerous_workflow.go
@@ -45,8 +45,12 @@ func containsUntrustedContextPattern(variable string) bool {
 			`head_commit\.message|` +
 			`head_commit\.author\.email|` +
 			`head_commit\.author\.name|` +
+			`head_commit\.committer\.email|` +
+			`head_commit\.committer\.name|` +
 			`commits.*\.author\.email|` +
 			`commits.*\.author\.name|` +
+			`commits.*\.committer\.email|` +
+			`commits.*\.committer\.name|` +
 			`blocked_user\.name|` +
 			`blocked_user\.email|` +
 			`pull_request\.head\.ref|` +
--- a/checks/raw/dangerous_workflow_test.go
+++ b/checks/raw/dangerous_workflow_test.go
@@ -86,6 +86,26 @@ func TestUntrustedContextVariables(t *testing.T) {
 			expected: true,
 		},
 		{
+			name:     "head_commit committer name",
+			variable: "github.event.head_commit.committer.name",
+			expected: true,
+		},
+		{
+			name:     "head_commit committer email",
+			variable: "github.event.head_commit.committer.email",
+			expected: true,
+		},
+		{
+			name:     "commits committer name",
+			variable: "github.event.commits[0].committer.name",
+			expected: true,
+		},
+		{
+			name:     "commits committer email",
+			variable: "github.event.commits[0].committer.email",
+			expected: true,
+		},
+		{
 			name:     "blocked_user name",
 			variable: "github.event.pull_request.organization.blocked_user.name",
 			expected: true,


spencerschrock:
Hi can you please fix DCO
https://github.com/ossf/scorecard/pull/4968/checks?check_run_id=67733701532

Otherwise looks good thanks, though

PR #4994, opened 2026-04-01
:book: Consolidate v6 docs and add implementation plan

justaugustus:

Documentation: consolidate Scorecard v6 documents under `docs/v6/` and add a
dependency-ordered implementation plan for Phase 1.

v6 documents are scattered across `openspec/changes/osps-baseline-conformance/`
and `docs/`. There is no implementation plan showing dependency ordering
between v6 work items.

All v6 documents consolidated under `docs/v6/`:
- `proposal.md` — architecture and vision (from merged PR #4952)
- `decisions.md` — reviewer feedback log
- `osps-baseline-coverage.md` — control-by-control coverage analysis
- `plan.md` — **new** dependency-ordered implementation plan

The implementation plan covers Phase 1 (OSPS Baseline Level 1 conformance
evidence) with 6 steps ordered by dependency:
- Step 0: OpenFeature feature flag infrastructure
- Step 1: Framework abstraction (proven with existing checks)
- Step 2: JSON output for conformance results
- Step 3: OSPS Baseline as second framework
- Step 4: Human review of L1 coverage analysis
- Step 5: Complete L1 coverage (gap probes)

The plan also includes:
- Feature promotion table for existing flagged features
- Forge support scope (GitHub in Phase 1; GitLab and Azure DevOps deferred)
- Codebase reuse map documenting existing infrastructure to extend
- Recommendations pending approval (marked as such)

Cross-document links in `ROADMAP.md`, `decisions.md`, and `proposal.md`
updated to reflect new file locations.

N/A — documentation only.

NONE

This is a follow-up to PR #4952 (merged). The proposal and decisions documents
are unchanged except for link fixes. The implementation plan (`plan.md`) is new
content. Several recommendations are marked "pending approval" for Steering
Committee discussion.

```release-note
NONE
```


docs/v6/proposal.md
JamieMagee: I like this principle a lot, and I want to flag something for when non-GitHub platforms come into scope.

Probes operate on `RawResults` and don't know which platform produced the data. When `ListReleases()` returns `ErrUnsupportedFeature` on ADO, the raw check swallows the error, and the probe sees "no releases." It returns `NotApplicable` instead of `UNKNOWN`. Those mean different things: "I looked and there's nothing there" vs. "I couldn't look."

Doesn't need solving in Phase 1, but it'd help to say so in the principle text. Something like: "For non-GitHub platforms, distinguishing UNKNOWN from NOT_APPLICABLE requires platform capability metadata, deferred to a later phase."

docs/v6/plan.md
```diff
@@ -0,0 +1,530 @@
...
+3. Production-ready conformance results in JSON output
+4. Evidence validated against OSPS Baseline v2026.02.19 controls
+5. Existing checks, probes, and scores unchanged (v6 is additive)
+6. Existing flagged features (`SCORECARD_V6`, `SCORECARD_EXPERIMENTAL`) promoted or migrated
+
+**Forge support in Phase 1:**
+- **GitHub:** Primary target (full L1 coverage)
+- **GitLab:** Deferred to a future phase
+- **Azure DevOps:** Deferred to a future phase
+- **Local directory:** Conformance results for file-based probes only
```
JamieMagee, line 31: Deferral makes sense. A few things that'd be easier to account for now while the abstractions are still being designed:

`Framework.Evaluate(findings)` (Step 1) doesn't have a way to say "this probe couldn't run because the platform doesn't support it." Adding that later means changing the interface after other things depend on it.

Also, `ErrUnsupportedFeature` handling in raw varies: `license.go` falls back to file detection, `security_policy.go` silently skips. Picking a canonical pattern before the conformance layer builds on top would save future headaches.

And the enriched JSON schema could include a `reason` field on UNKNOWN statuses from day one. Cheap now, painful later once consumers depend on the shape.

docs/v6/plan.md
```diff
@@ -0,0 +1,530 @@
...
+### Architecture
+
+```go
+// Framework represents an evaluation surface over probe findings
+type Framework interface {
+    // Name returns the framework identifier (e.g., "scorecard-checks", "osps-baseline")
+    Name() string
+
+    // Evaluate takes probe findings and produces framework-specific results
+    Evaluate(findings []finding.Finding) (Result, error)
```
JamieMagee: Thinking about how this plays out on ADO: how does the evaluation layer distinguish "finding is absent because the platform can't see it" from "the project just doesn't do this thing"? Right now branch_protection.go catches `ErrUnsupportedFeature` and the probe downstream sees empty data with no context.

Maybe a `PlatformCapabilities` input alongside findings, or a new outcome like `OutcomeUnobservable` with a reason string. Just so the UNKNOWN-first principle can work beyond GitHub when the time comes.

docs/v6/proposal.md
JamieMagee, line 356: For the Phase 2 release integrity work: six controls depend on "releases" (BR-02.01, BR-04.01, BR-06.01, LE-02.02, LE-03.02, QA-02.02). The concept maps clean to GitHub Releases, but ADO doesn't have a real equivalent. ADO teams ship through Azure Artifacts feeds, classic release pipelines, or pipeline artifacts, which all work differently.

A probe that only understands GitHub Releases would return NotApplicable for an ADO project that ships fine through Azure Artifacts. Probably just worth a note near Phase 2: "release-related probes will need platform-specific implementations."

docs/v6/plan.md
```diff
@@ -0,0 +1,530 @@
...
+### Feature promotion
+
+The following features must be addressed as part of the OpenFeature migration:
+
+| Feature | Current Gate | TODO/Comment | Recommendation |
+|---------|-------------|-------------|----------------|
+| Webhooks check | `SCORECARD_EXPERIMENTAL` | "remove this check when v6 is released" | Promote to always-on in Phase 1 |
+| SBOM check | `SCORECARD_EXPERIMENTAL` | "remove this check when v6 is released" | Promote to always-on in Phase 1 |
+| Raw output format | `SCORECARD_V6` | none | Promote to always-on in Phase 1 |
+| Azure DevOps support | `SCORECARD_EXPERIMENTAL` | none | Keep behind `scorecard.experimental` |
```
JamieMagee, line 98: More of a question: what would graduating ADO from `scorecard.experimental` look like?

justaugustus (commented):
@JamieMagee — I've integrated all of your feedback in the most recent commits.
Thanks for doing the first round!

As for your comment about ADO graduation, I'll let you help decide that, now that you're [a component owner](https://github.com/ossf/scorecard/pull/5024). :)

Copied your [comment](https://github.com/ossf/scorecard/issues/4177#issuecomment-4273228125) to the original ADO feature request, so we don't lose the thread.

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
+# Scorecard v6 Implementation Plan
+
+## Overview
+
+**Phase 1 milestone:** Deliver OSPS Baseline Level 1 conformance evidence using existing infrastructure where possible.
+
+v6 is a clean, backwards-compatible successor to v5. All v6 features land within the v5 module behind feature flags. When all flags graduate to default-on, the module path bumps from `github.com/ossf/scorecard/v5` to `github.com/ossf/scorecard/v6`.
```
spencerschrock, line 7: I know this is AI fluff, but if it's backwards compatible, then it's still v5.

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+The vision and architectural rationale live in [`proposal.md`](proposal.md); this document is the execution plan.
+
+---
+
+## Phase 1: OSPS Baseline Level 1 conformance evidence
+
+**Goal:** Scorecard produces complete OSPS Baseline Level 1 conformance evidence (PASS/FAIL/UNKNOWN/NOT_APPLICABLE per control) via CLI and GitHub Action, using production-ready JSON output.
+
+**Success criteria:**
+1. Complete L1 control coverage (all gap controls closed + existing coverage validated)
```
spencerschrock, line 20: I don't have L1 in front of me, does this include any controls that aren't discoverable via automation?

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+
+The existing `SCORECARD_V6` env var predates the conformance work described in this plan. Several features under flag have an implicit contract to be promoted in v6.
+
+### Feature promotion
+
+The following features must be addressed as part of the OpenFeature migration:
+
+| Feature | Current Gate | TODO/Comment | Recommendation |
+|---------|-------------|-------------|----------------|
+| Webhooks check | `SCORECARD_EXPERIMENTAL` | "remove this check when v6 is released" | Promote to always-on in Phase 1 |
```
spencerschrock, line 95: the problem is the auth token, I don't think we can promote until we have a solution of these ospo style checks

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+| SBOM check | `SCORECARD_EXPERIMENTAL` | "remove this check when v6 is released" | Promote to always-on in Phase 1 |
```
spencerschrock, line 96: Is this tied to a L1 control?

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+| Raw output format | `SCORECARD_V6` | none | Promote to always-on in Phase 1 |
+| Azure DevOps support | `SCORECARD_EXPERIMENTAL` | none | Keep behind `scorecard.experimental` |
```
spencerschrock, line 98: Raw output format hasn't been used and I think probes have taken its place. Do we need it?

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+| SARIF output format | `ENABLE_SARIF` | "UPGRADEv4: remove" | Remove gate entirely (always-on) |
```
spencerschrock, line 99: Full support for this one

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+| Raw results in CheckRequest | none (commented) | "UPGRADEv6: return raw results instead of scores" | Implement in Phase 1 |
```
spencerschrock, line 100: Again, I think probes have taken this place. I'm not sure this work is directly related to any of the baseline work

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+Per-feature granular flags (e.g., `v6.conformance-evaluation`, `v6.probe.<name>`) are deferred until actual need arises (e.g., cron rollout requiring partial enablement). Two flags is sufficient for Phase 1.
+
+### Testing strategy
+
+E2E tests currently do not set feature flag env vars, meaning experimental and v6 features have zero e2e coverage. The OpenFeature migration should include:
+- E2E test suite runs twice: once with default flags, once with `SCORECARD_V6=1`
+- Test helper that initializes OpenFeature `TestProvider` with specific flag states
+- Each new v6 feature gets at least one e2e test with the flag enabled
+
+**Deliverable:** Single PR introducing `internal/featureflags/`, migrating existing flags, promoting graduated features, adding `FeatureGate` to `checker.Check`. Behavioral changes limited to promoting features that were already committed to ship in v6.
```
spencerschrock, line 140: I guess I don't see what this layer of abstraction affords us

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+2. Probe grouping (`probes/entries.go`)
+3. Evaluation function's `expectedProbes` list (`checks/evaluation/*.go`)
+
+A mismatch between any of these causes "invalid probe results" runtime failures. Fixing this early de-risks the entire plan — Steps 2, 6, and 8 all involve probe registration or evaluation.
+
+### Solution
+
+Derive `expectedProbes` from the probe registry rather than hardcoding them in evaluation functions. The probe registration system (`internal/probes/probes.go`) already knows which probes exist and which checks they belong to. Evaluation functions should query this registry instead of maintaining independent lists.
+
+**Deliverable:** PR that eliminates hardcoded `expectedProbes` lists in evaluation functions, replacing them with registry-derived probe sets.
```
spencerschrock, line 164: It's a little more nuanced than this.

The grouping and expected probes lists are separate based on what check's raw data the probe consumes, and whether or not the probe runs by default

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+Phase 1 delivers L1. Phase 2 adds L2 controls to the same framework. Phase 3 adds L3. One framework, expanding control set.
+
+### Architecture
+
+```go
+// PlatformCapabilities describes what the underlying forge supports.
+// On GitHub, all capabilities are true. On other forges, unsupported
+// features are false, allowing the evaluation layer to distinguish
+// "platform can't observe this" (UNKNOWN) from "project doesn't do
+// this" (NOT_APPLICABLE or FAIL).
```
spencerschrock, line 208: what is the granularity of a capability? Is this a probe? a check? something else?

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+}
+
+// Result represents a framework's evaluation outcome
+type Result interface {
+    // Type returns the result type (score, conformance label, etc.)
+    Type() string
+
+    // Value returns the type-specific value
+    Value() any
+}
```
spencerschrock, line 232: I'm curious what this looks like. `any` is obviously broad, and this would be annoying as a library user, e.g. allstar

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+
+**What this proves:**
+1. Existing checks work through the abstraction (no behavior change)
+2. The interface supports different evaluation semantics (scores vs. labels)
+3. Probe findings are framework-agnostic (same input, different outputs)
+4. Evaluation layer can distinguish platform limitations from project state
+
+### Prerequisite for non-GitHub conformance: `ErrUnsupportedFeature` standardization
+
+`ErrUnsupportedFeature` handling in raw checks is currently inconsistent: `license.go` falls back to file detection, `security_policy.go` silently skips, `branch_protection.go` produces an error result. Before non-GitHub conformance support is introduced, this must be standardized so that `PlatformCapabilities` can be correctly populated from the client's actual behavior.
```
spencerschrock, line 245: This isn't accurate. Security policy also falls back to file detection.

The goal semantic is to have as much info as possible for all the forges that implement it.

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+
+This standardization is not a Phase 1 deliverable (Phase 1 targets GitHub only, where `ErrUnsupportedFeature` never fires), but it is a prerequisite for any phase that adds GitLab or Azure DevOps conformance support.
+
```
spencerschrock, line 248: I guess the above comment isn't immediately relevant

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+// Framework represents an evaluation surface over probe findings
+type Framework interface {
+    // Name returns the framework identifier (e.g., "scorecard-checks", "osps-baseline")
+    Name() string
+
+    // Evaluate takes probe findings and platform capabilities, producing
+    // framework-specific results. Capabilities allow the evaluator to
+    // return UNKNOWN (with reason) when the platform cannot observe a
+    // control, rather than misinterpreting absent findings.
+    Evaluate(findings []finding.Finding, caps PlatformCapabilities) (Result, error)
```
spencerschrock, line 222: Findings would already take into account supported features via the finding outcomes `finding.OutcomeNotSupported`, `finding.OutcomeNotAvailable`, etc.

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+}
+```
+
+In Phase 1 (GitHub-only), `PlatformCapabilities` is fully-capable (all features `true`). The parameter has no behavioral effect but is present in the interface from day one, avoiding a breaking change when non-GitHub conformance support is introduced.
+
+**What this proves:**
+1. Existing checks work through the abstraction (no behavior change)
+2. The interface supports different evaluation semantics (scores vs. labels)
+3. Probe findings are framework-agnostic (same input, different outputs)
+4. Evaluation layer can distinguish platform limitations from project state
```
spencerschrock, line 241: I see the desire for a Framework interface, but I see the output format itself as being the manifestation of the probe -> output mapping.

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+        {"id": "AC-01.01", "status": "PASS", ...},
+        {"id": "AC-02.01", "status": "UNKNOWN", "reason": "Requires org-level permissions", ...}
+      ]
+    }
+  }
+}
+```
+
+Old schema (`"checks": [...]`) stays available as the default for backward compatibility. New schema available via `--format=json-v6` or similar. When v6 becomes default, new schema becomes the default JSON output.
+
```
spencerschrock, line 319: i was under the impression there would just be a new `--output=baseline` option which would spit out its own JSON. Do we really need one JSON schema with all evaluation formats?

docs/v6/plan.md
```diff
@@ -0,0 +1,640 @@
...
+## Step 4: CLI integration
+
+**Requires:** Step 2 (framework abstraction exists to wire into the run)
+**Enables:** Phase 1 completion (users can invoke conformance evaluation)
+**Can proceed in parallel with:** Steps 3, 5, and 6
+**Estimated scope:** 1 PR (~200 lines)
+
+### Problem
+
+Users need to know how to invoke conformance evaluation. The proposal's architectural target state #5 says: "A single Scorecard run produces both check scores and conformance results — users MUST NOT need to run Scorecard twice."
```
spencerschrock, line 341: for the library this is certainly true/obtainable. For the CLI, I question if this is necessary.

PR #5030, opened 2026-04-20
:bug: checks/pinned: strip inline comments from uses value before evaluatingchecks/pinned: strip inline comments from uses value before evaluating

miguel-baptista07:
Fixes #5004

## Problem
GitHub Actions workflow steps can include inline YAML comments after the `uses` value:
```yaml
- uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # main
```
The comment was being included in the ref evaluation, causing valid pinned SHAs to be flagged as unpinned dependencies.

## Fix
Strip everything from ` #` onwards inside `isActionDependencyPinned` before evaluating the ref. This is the single evaluation point so fixing it here handles all callers.

## Tests
Added two regression tests to `TestGithubWorkflowPinningPattern`:
- `pinned SHA with inline comment should be pinned` → `true`
- `mutable branch ref remains unpinned` → `false`

### DIFF
--- a/checks/raw/pinned_dependencies.go
+++ b/checks/raw/pinned_dependencies.go
@@ -831,6 +831,9 @@ func newGHActionDependency(uses, pathfn string, line int) checker.Dependency {
 }
 
 func isActionDependencyPinned(actionUses string) bool {
+	// Strip inline YAML comment (e.g. "# pinned to main") before evaluating.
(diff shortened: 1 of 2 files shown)


miguel-baptista07:
Hi, just wanted to check in, is there anything else needed from my side before this can move forward? Happy to make any additional changes.

PR #5118, opened 2026-07-08
:bug: Fix Prow revision ID guard for lgtm-labeled commits

arpitjain099:

bug fix

In `getProwRevisionID` (checks/raw/code_review.go) the intent is to treat a commit as Prow-reviewed only when its associated merge request carries an `lgtm` or `approved` label and has a real number. Because `&&` binds tighter than `||` in Go, the `mr.Number != 0` guard only applies to the `approved` branch. So a commit whose merge request has the `lgtm` label but `Number == 0` returns `strconv.Itoa(0)`, i.e. "0". Since that revision ID isn't empty, it never falls back to the commit SHA, and every such commit ends up grouped into a single bogus Prow changeset keyed on "0", which skews the Code-Review results.

The label check is parenthesized so the `mr.Number != 0` guard covers both labels, matching the existing `getGithubRevisionID` behavior. Commits without a real merge request number now correctly fall back to their own SHA instead of collapsing into a "0" changeset.

NONE

Added `Test_getProwRevisionID` covering the lgtm/no-number, approved/no-number, and lgtm/valid-number cases. It fails on the current code (returns "0" for lgtm without a number) and passes with the fix. Verified with `go test ./checks/raw/` plus `go vet` and `gofmt`.

NONE

```release-note

```


arpitjain099:
This one's still relevant, just got caught by the stale bot. It's a small operator-precedence fix in getProwRevisionID: because && binds tighter than ||, the mr.Number != 0 guard only applied to the approved branch, so an lgtm-labeled commit with Number 0 returns "0" instead of falling back to the SHA, and those commits all collapse into one bogus "0" changeset. Parenthesizing the label check makes the guard cover both labels, matching getGithubRevisionID. The added Test_getProwRevisionID fails on the current code and passes with the fix, and CI is green. Happy to rebase whenever someone has a chance to take a look.

PR #5141, opened 2026-07-25
:bug: Branch-Protection: don't fail the whole check on one bad release branch

### LINKED ISSUE #2161: Improve Score Reporting: Branch-Protection check fails with -1
Branch-Protection fails with `-1` for multiple reasons:

- lookup of a branch name that no more exists on the repository
- unable to locate the branch even though it existed (maybe similar to #1320)
- the main branch had a different name than the “main”

None of these seem reasonable cases to be returning `-1`. Need to fix this.

pujitha24:
Motivation:
Issue #2161 reports the Branch-Protection check scoring -1 (internal
runtime error) when a release's target branch can't be resolved, e.g.
a release pointing at a branch name that no longer exists, a branch
that briefly can't be located, or a legacy master->main redirect that
doesn't resolve. In checks/raw/branch_protection.go, any single such
release aborted the entire raw data collection with an error, which
checks/branch_protection.go turned into a runtime error result (-1)
for the whole check, discarding every other branch that resolved
successfully.

Approach:
In raw.BranchProtection, the three per-release lookup failures (empty
TargetCommitish, GetBranch failing for the release's target branch,
and GetBranch failing for the master->main redirect fallback) now log
a Debug message via cr.Dlogger (matching the existing non-fatal
degradation pattern already used in checks/raw/signed_releases.go and
checks/raw/sast.go) and skip that release instead of returning an
error. The check now completes and scores based on whichever branches
it could resolve. This does not change behavior for errors from
GetDefaultBranch, ListReleases, or CODEOWNERS collection, which still
fail the check as before, nor does it touch any other check. The
dead errInternalCommitishNil sentinel is removed along with its only
use.

Validation:
go build ./...
go test ./checks/... ./checks/raw/...
All packages pass, including updated/added cases in
checks/raw/branch_protection_test.go and checks/branch_protection_test.go
covering the empty-target-commitish, release-branch-lookup-error, and
redirect-lookup-error paths.

Fixes #2161

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>


pujitha24:
This has been rebased and green for a while now — happy to make any changes that would make it easier to review.

PR #5164, opened 2026-08-08
:bug: contain git-mode file reads within the checkout root

Sahana2524:

Bug fix (security): keeps file reads in git mode inside the repository checkout.

In git mode (`--file-mode git`) the target repo is cloned to a temp dir with a
non-bare `git.PlainClone`, which materializes symlinks on disk for symlink
entries in the tree. `internal/gitfile`'s `GetFile` checks the requested name for
`../` traversal but then reads with `os.Open`, which follows symlinks. A scanned
repo can commit a symlink whose name matches a check's file predicate (a workflow
file, a `Dockerfile`, and so on) and make the read resolve outside the checkout:

    innocent.txt -> /etc/passwd   # committed in the scanned repo
    GetFile("innocent.txt")       # os.Open follows the link, reads the host file

`clients/git`'s `GetFileReader` has the same non-bare-clone plus `os.Open` shape
with no containment check at all.

Both readers now use `os.OpenInRoot(tempDir, filename)`, which refuses names that
escape the root through `../` or a symlink, so a within-repo path reads normally
while an escaping one returns an error. This is the same approach
`clients/localdir` already uses. A regression test commits an escaping symlink
and checks the read is refused; the existing `../` traversal test still passes.

NONE

`os.OpenInRoot` landed in Go 1.24 and the module is already on 1.25, so no
version bump is needed. I left the archive/zip handlers alone since they skip
symlink entries during extraction and never write them to disk, so the same read
path is not reachable there.

```release-note
In git file mode, file reads are now contained to the repository checkout, so a symlink committed in a scanned repo can no longer redirect a read to a file outside it.
```

### DIFF
--- a/clients/git/client.go
+++ b/clients/git/client.go
@@ -239,13 +239,12 @@ func (c *Client) ListFiles(predicate func(string) (bool, error)) ([]string, erro
 }
 
 func (c *Client) GetFileReader(filename string) (io.ReadCloser, error) {
-	// Create the full path of the file
-	fullPath := filepath.Join(c.tempDir, filename)
-
-	// Read the file
-	f, err := os.Open(fullPath)
+	// Read the file while staying inside the checkout root. os.OpenInRoot
+	// refuses names that escape via "../" or a symlink materialized by the
+	// clone that points outside the worktree.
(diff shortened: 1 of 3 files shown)


Sahana2524:
gentle ping

Issue #5100, opened 2026-06-21
Feature: Add winget package manager support


andy778:
**Is your feature request related to a problem? Please describe.**
When assessing the security posture of Windows software, there is no way to look up a package by its winget identifier. Users must manually find the source repository before running Scorecard, which is inconsistent with the experience for npm, PyPI, RubyGems, and NuGet.

**Describe the solution you'd like**
Add a --winget flag that accepts a winget package identifier and automatically resolves the source repository from the [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) manifest, then runs Scorecard against it.

`scorecard --winget="Notepad++.Notepad++"`

**Describe alternatives you've considered**
Using the --repo flag directly after manually looking up the repository URL. This works but requires an extra step and breaks the consistency with other package manager flags.

**Additional context**
Winget is the official Windows package manager with a large and growing catalog. The manifest repository (microsoft/winget-pkgs) is publicly accessible on GitHub and contains the source URL for each package, making resolution straightforward.

PR #5169, opened 2026-08-09
:sparkles: feat: detect winget publishing workflows in Packaging check

### LINKED ISSUE #5168: Feature: Packaging check should detect winget publishing workflows

The `Packaging` check does not recognize any winget (Windows Package Manager) publishing patterns. `IsPackagingWorkflow` in `checks/fileparser/github_workflow.go` has matchers for npm, PyPI, RubyGems, NuGet, cargo, Docker, ko, Maven/Gradle, sbt, Elixir and semantic-release, but none for winget.

As a result, a project that automates publishing to winget — via the [`winget-releaser`](https://github.com/vedantmgoyal9/winget-releaser) action or Microsoft's official [`wingetcreate`](https://github.com/microsoft/winget-create) CLI — scores `Packaging` as N/A ("packaging workflow not detected"), even though it does publish packages through an automated workflow.

Add winget matchers to `IsPackagingWorkflow`:

- `vedantmgoyal9/winget-releaser` (the de-facto community action)
- `vedantmgoyal2009/winget-releaser` — the same action under its previous owner name; the account was renamed and GitHub redirects it, so unmodified workflows still reference the old path
- a `Run` matcher for `wingetcreate ... submit`, which matches Microsoft's official CLI by command name and is therefore owner-agnostic

Matching only `wingetcreate` would avoid hardcoding a personally-owned action, but would miss the most common real-world publishing path. Supporting a pattern instead of an exact owner would require `JobMatcherStep.Uses` to accept a regex — `stepsMatch` currently does exact prefix matching — which touches code shared by all checks and seems better suited to its own change. The configurable allowlist discussed in #1642 would be the general fix.

**Additional context**

Related to but independent of #5100 / #5101, which propose a `--winget` *input* flag so Scorecard can resolve a winget package identifier to its source repo. That is the input side; this issue is the detection side (noticing that a scanned repo publishes to winget). Neither depends on the other.

This is the same category as #5145 (cargo-dist), #5131 (step-security/goreleaser) and #5096 (changesets).

### Side effect on Token-Permissions worth deciding on

Anyone implementing this should know `IsPackagingWorkflow` has two call sites, not one. Besides the `Packaging` check (`checks/raw/github/packaging.go`), it is also called by `requiresPackagesPermissions` (`checks/raw/permissions.go:422`), whose result feeds `createIgnoredPermissions` (`:341`):

```go
if requiresPackagesPermissions(workflow, fp, pdata) {
    ignoredPermissions[permissionPackages] = true
}
```

So adding these matchers also causes `Token-Permissions` to stop checking the `packages` permission in winget publishing workflows. Winget publishing does not actually need `packages: write`: `winget-releaser` requires a classic PAT with `public_repo` scope to open a PR against `microsoft/winget-pkgs`, and `wingetcreate submit` works the same way — neither publishes to GitHub Packages.

In practice the exemption only bites when a workflow actually declares a `packages` permission, which a typical winget workflow does not, so the effect is narrow. It is also pre-existing and broad rather than specific to winget: `gem push` (RubyGems.org), `cargo publish` (crates.io), `pypa/gh-action-pypi-publish` (PyPI), `sbt ci-release` (Maven Central), `hex.publish` (hex.pm) and `setup-node` + `registry.npmjs.org` don't require `packages: write` either.

For context, #1642 is where this coupling originated: the matcher list was justified there as "actions we know of" (classification rather than trust), even though the bug being fixed was a `Token-Permissions` false positive. Narrowing it now would change `Token-Permissions` results for many existing repos, so it likely deserves its own issue rather than riding along with a new matcher.

mangokingTW:

New feature — the `Packaging` check now recognizes winget (Windows Package Manager) publishing workflows.

`IsPackagingWorkflow` (in `checks/fileparser/github_workflow.go`) has no matchers for winget publishing patterns, so repositories that publish to winget via `winget-releaser` or `wingetcreate` don't get credit for having an automated packaging workflow.

Adds three matchers to `IsPackagingWorkflow`:
- `vedantmgoyal9/winget-releaser` (current owner)
- `vedantmgoyal2009/winget-releaser` (older owner name, still used by some projects)
- a `Run` matcher for `wingetcreate.*submit`

Fixes #5168

This only teaches the `Packaging` check to recognize winget publishing workflows. It's independent of #5100 / #5101, which are about adding a `--winget` input flag so Scorecard can score a package by its winget identifier.

Yes — repositories using `winget-releaser` or `wingetcreate` to publish packages will now pass the `Packaging` check.

```release-note
The `Packaging` check now detects winget publishing workflows using `winget-releaser` or `wingetcreate`.
```

### DIFF
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -624,10 +624,12 @@ func IsPackagingWorkflow(workflow *actionlint.Workflow, fp string) (JobMatchResu
 			LogText: "candidate winget publishing workflow using winget-releaser",
 		},
 		{
-			// Winget packages using wingetcreate.
+			// Winget packages using wingetcreate. These workflows run on Windows and
+			// commonly split the command with PowerShell's backtick continuation,
+			// which is not stripped like a backslash is, so match across newlines.
 			Steps: []*JobMatcherStep{
 				{
-					Run: "wingetcreate.*submit",
+					Run: "(?s)wingetcreate.*submit",
 				},
 			},
 			LogText: "candidate winget publishing workflow using wingetcreate",
--- a/checks/fileparser/github_workflow.go
+++ b/checks/fileparser/github_workflow.go
@@ -604,6 +604,34 @@ func IsPackagingWorkflow(workflow *actionlint.Workflow, fp string) (JobMatchResu
 			},
 			LogText: "candidate publishing workflow using elixir",
 		},
+		{
+			// Winget packages using winget-releaser.
+			Steps: []*JobMatcherStep{
+				{
+					Uses: "vedantmgoyal9/winget-releaser",
+				},
+			},
+			LogText: "candidate winget publishing workflow using winget-releaser",
+		},
+		{
+			// Winget packages using winget-releaser under its old owner name,
+			// which is still used by some projects.
+			Steps: []*JobMatcherStep{
+				{
+					Uses: "vedantmgoyal2009/winget-releaser",
+				},
+			},
+			LogText: "candidate winget publishing workflow using winget-releaser",
+		},
+		{
+			// Winget packages using wingetcreate.
+			Steps: []*JobMatcherStep{
+				{
+					Run: "wingetcreate.*submit",
+				},
+			},
+			LogText: "candidate winget publishing workflow using wingetcreate",
+		},
 	}
 
 	return AnyJobsMatch(workflow, jobMatchers, fp, "not a publishing workflow")
--- a/checks/fileparser/github_workflow_test.go
+++ b/checks/fileparser/github_workflow_test.go
@@ -1004,6 +1004,16 @@ func TestIsPackagingWorkflow(t *testing.T) {
 			filename: "../testdata/.github/workflows/github-workflow-packaging-elixir.yaml",
 			expected: true,
 		},
+		{
+			name:     "winget publish using winget-releaser",
+			filename: "../testdata/.github/workflows/github-workflow-packaging-winget.yaml",
+			expected: true,
+		},
+		{
+			name:     "winget publish using wingetcreate",
+			filename: "../testdata/.github/workflows/github-workflow-packaging-wingetcreate.yaml",
+			expected: true,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/checks/fileparser/github_workflow_test.go
+++ b/checks/fileparser/github_workflow_test.go
@@ -1010,8 +1010,18 @@ func TestIsPackagingWorkflow(t *testing.T) {
 			expected: true,
 		},
 		{
+			name:     "winget publish using winget-releaser under its previous owner name",
+			filename: "../testdata/.github/workflows/github-workflow-packaging-winget-legacy-owner.yaml",
+			expected: true,
+		},
+		{
 			name:     "winget publish using wingetcreate",
 			filename: "../testdata/.github/workflows/github-workflow-packaging-wingetcreate.yaml",
+			expected: true,
+		},
+		{
+			name:     "winget publish using wingetcreate with PowerShell backtick continuation",
+			filename: "../testdata/.github/workflows/github-workflow-packaging-wingetcreate-powershell.yaml",
 			expected: true,
 		},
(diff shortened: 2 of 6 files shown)


mangokingTW:
Flagging a secondary effect for reviewers, since `IsPackagingWorkflow` has two call sites rather than one.

Besides the `Packaging` check (`checks/raw/github/packaging.go`), it's also called by `requiresPackagesPermissions` (`checks/raw/permissions.go:422`), whose result feeds `createIgnoredPermissions` (`:341`):

```go
if requiresPackagesPermissions(workflow, fp, pdata) {
    ignoredPermissions[permissionPackages] = true
}
```

So these matchers also cause `Token-Permissions` to stop checking the `packages` permission in winget publishing workflows.

Worth being explicit that winget publishing doesn't actually need `packages: write`: `winget-releaser` requires a classic PAT with `public_repo` scope in order to open a PR against `microsoft/winget-pkgs`, and `wingetcreate submit` works the same way. Neither publishes to GitHub Packages.

That said, this is consistent with how the existing matchers already behave — `gem push` (RubyGems.org), `cargo publish` (crates.io), `pypa/gh-action-pypi-publish` (PyPI), `sbt ci-release` (Maven Central), `hex.publish` (hex.pm), and `setup-node` + `registry.npmjs.org` don't require `packages: write` either. I kept this PR consistent with that existing behavior rather than changing it, since the docstring on `requiresPackagesPermissions` ("A packaging workflow using GitHub's supported packages", plus the `TODO: add support for GitHub registries`) hints at a narrower original intent than the current matcher list implements. Narrowing it would change `Token-Permissions` results for many existing repos, which seems like it deserves its own issue rather than riding along here.

Happy to restructure so the winget matchers only feed the `Packaging` check if you'd prefer that instead.

mangokingTW:
Pushed two follow-ups after testing this against real repositories, plus some evidence of impact.

### Real-world impact

[`microsoft/PowerToys`](https://github.com/microsoft/PowerToys) is in `cron/internal/data/projects.csv` and publishes to winget from [`package-submissions.yml`](https://github.com/microsoft/PowerToys/blob/main/.github/workflows/package-submissions.yml) using Microsoft's own `wingetcreate`, with 19 successful runs of that workflow. Its current public result scores `Packaging` as `-1` / "packaging workflow not detected":

```
$ curl -s https://api.scorecard.dev/projects/github.com/microsoft/PowerToys | jq '.checks[] | select(.name=="Packaging")'
  score: -1
  reason: "packaging workflow not detected"
```

With this PR:

```
$ go run . --repo=github.com/microsoft/PowerToys --checks=Packaging --show-details
| 10 / 10 | Packaging | packaging workflow detected | Info: Project packages its releases by way of
GitHub Actions.: .github/workflows/package-submissions.yml:9 |
```

### `:bug:` wingetcreate needed to match across PowerShell line continuations

The first version of the `wingetcreate.*submit` matcher did **not** detect PowerToys. These workflows run on Windows and commonly split the command with PowerShell's backtick continuation:

```yaml
.\wingetcreate.exe update Microsoft.PowerToys `
  --version $env:VERSION `
  --submit
```

`stepsMatch` strips backslash continuations only, and Go's `.` does not match a newline, so the pattern never spanned the break. Since wingetcreate is inherently a Windows/PowerShell tool, this is the common case rather than an edge case, so the matcher now uses `(?s)`.

Worth flagging for reviewers: the underlying gap is that the line-continuation stripping in `stepsMatch` is POSIX-shell-only. Other `Run` matchers (`docker.*push`, `gem.*push`, `cargo.*publish`, ...) have the same blind spot on PowerShell workflows. I've kept this PR scoped to the winget matcher rather than changing shared behaviour for every check, but happy to open a separate issue for that if it's useful.

### Missing test

The `vedantmgoyal2009` (previous owner name) matcher had no test coverage. Added one, with testdata that documents why the entry exists.

### One note on scope

Detection alone isn't sufficient for a score change — `packaging.go` also requires `ListSuccessfulWorkflowRuns` to be non-empty. For example `gerardog/gsudo` uses `winget-releaser` but its workflow is `workflow_dispatch`-only with 0 successful runs, so it correctly still reports as not detected. This PR only widens what counts as a publishing workflow; it doesn't relax that requirement.

PR #5173, opened 2026-08-10
:bug: reject oversized zip entries in azuredevopsrepo extractZip

Sahana2524:

Bug fix. The per-entry size limit in the Azure DevOps zip extractor never rejects anything today, so an oversized entry is silently truncated instead of being refused.

`extractZip` copies each entry with `io.CopyN(outFile, rc, maxSize)` and then checks `written > maxSize` to return `errFileTooLarge`. `io.CopyN` copies at most its third argument, so `written` tops out at `maxSize` and the comparison is never true. The effect is that `errFileTooLarge` is unreachable, and an entry larger than `maxSize` is written out truncated to `maxSize` bytes and appended to `z.files` as if it extracted cleanly. Later checks then read a partial file believing it is the whole thing. With `resolveLfs=true` on the download, a small LFS pointer in the repo can expand server-side into an entry past the cap, so this is reachable without pushing a large blob through git.

The copy now reads `maxSize+1`, so an entry that exceeds the cap makes `written > maxSize` true and is rejected, while an entry of exactly `maxSize` still extracts. The new test builds a zip whose single entry is one byte over the cap and checks `extractZip` returns `errFileTooLarge`; on the current code that entry extracts and the test fails.

NONE

I left the GitHub and GitLab tarball handlers alone. They intentionally waive the size cap with a `//nolint:gosec` and a comment, so this stays scoped to the Azure path where the check already exists but does not fire.

```release-note
In Azure DevOps mode, a repository file larger than the zip extraction size limit is now rejected instead of being silently truncated.
```

### DIFF
--- a/clients/azuredevopsrepo/zip.go
+++ b/clients/azuredevopsrepo/zip.go
@@ -189,31 +189,42 @@ func (z *zipHandler) extractZip() error {
 			continue
 		}
 
-		outFile, err := os.OpenFile(filenamepath, os.O_CREATE|os.O_WRONLY, 0o644)
-		if err != nil {
-			return fmt.Errorf("os.OpenFile: %w", err)
+		if err := extractEntry(file, filenamepath); err != nil {
+			return err
 		}
-
-		rc, err := file.Open()
-		if err != nil {
-			return fmt.Errorf("file.Open: %w", err)
-		}
-
-		// Read one byte past the limit so an entry that is exactly maxSize still
-		// succeeds while anything larger trips the check below. io.CopyN caps the
-		// copy at its third argument, so passing maxSize here would make written
-		// top out at maxSize and the comparison could never be true.
-		written, err := io.CopyN(outFile, rc, maxSize+1)
-		if err != nil && !errors.Is(err, io.EOF) {
-			return fmt.Errorf("%w io.Copy: %w", errZipNotFound, err)
-		}
-		if written > maxSize {
-			return errFileTooLarge
-		}
-		outFile.Close()
 
 		filename := strings.TrimPrefix(filenamepath, destinationPrefix)
 		z.files = append(z.files, filepath.ToSlash(filename))
+	}
+	return nil
+}
+
+// extractEntry writes a single zip entry to filenamepath, enforcing the
+// per-entry size cap. The defers guarantee both the destination file and the
+// entry reader are closed on every return path, including the error ones.
+func extractEntry(file *zip.File, filenamepath string) error {
+	outFile, err := os.OpenFile(filenamepath, os.O_CREATE|os.O_WRONLY, 0o644)
+	if err != nil {
+		return fmt.Errorf("os.OpenFile: %w", err)
+	}
+	defer outFile.Close()
+
+	rc, err := file.Open()
+	if err != nil {
+		return fmt.Errorf("file.Open: %w", err)
+	}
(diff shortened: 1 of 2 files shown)


bilaldeveloper4312 (commented):
I found one resource-lifecycle issue in the new oversized-entry path. After `io.CopyN`, the `written > maxSize` branch returns before either `outFile` or the ZIP entry reader `rc` is closed. That can retain descriptors until garbage collection; on Windows, the new `t.TempDir()` cleanup may also fail because the extracted file remains open.

Could the per-entry processing guarantee both closes on every return path (for example, through a small helper with defers), then keep the size check after the copy? The existing success path also never closes `rc`, so handling both resources together would make the new error path reliable.

This is based on inspecting commit `1ea04b0`. I could not execute the Go test locally because this environment does not currently have the Go toolchain.

bilaldeveloper4312:
Verification clarification: after submitting the review I located a portable Go 1.25.6 toolchain in the workspace. I attempted `go test ./clients/azuredevopsrepo -run '^TestExtractZipTooLarge$' -count=1`, but dependency compilation did not complete within the bounded validation window, so I stopped it. The test result remains unverified; the resource-close finding above is still based on code-path inspection.

Sahana2524:
Good catch. The oversized-entry return did leak both handles, and the success path never closed rc either. I moved the per-entry work into an extractEntry helper so outFile and rc are closed via defer on every path, with the size check still after the copy. Both zip tests pass locally, including the new oversized-entry one.

bilaldeveloper4312 (commented):
Verified the follow-up at `b4f7e56`: moving the per-entry work into `extractEntry` ensures both `outFile` and `rc` are closed on the oversized-entry and success paths, addressing the resource-lifecycle issue I raised.

Focused verification now passes locally with Go 1.25.6:

```
go test ./clients/azuredevopsrepo -run '^TestExtractZipTooLarge$' -count=1
```

The broader existing `TestExtractZip/Basic` still hits Windows path-separator and file-lock failures, so I am not treating that unrelated package behavior as verification of this change. The specific resource-close fix looks good.

clients/azuredevopsrepo/zip.go
```diff
@@ -189,31 +189,46 @@ func (z *zipHandler) extractZip() error {
...
+	// succeeds while anything larger trips the check below. io.CopyN caps the
+	// copy at its third argument, so passing maxSize here would make written
+	// top out at maxSize and the comparison could never be true.
+	written, err := io.CopyN(outFile, rc, maxSize+1)
+	if err != nil && !errors.Is(err, io.EOF) {
+		return fmt.Errorf("%w io.Copy: %w", errZipNotFound, err)
+	}
+	if written > maxSize {
+		return errFileTooLarge
+	}
```
JamieMagee:
```suggestion
	written, err := io.Copy(outFile, io.LimitReader(rc, maxSize+1))
	if err != nil {
		return fmt.Errorf("%w io.Copy: %w", errZipNotFound, err)
	}
	if written > maxSize {
		return errFileTooLarge
	}
```

Sahana2524: Applied. io.Copy with the LimitReader drops the EOF special case CopyN needed, which reads cleaner for the same bound. Both zip tests still pass locally.

JamieMagee (approved):

PR #5174, opened 2026-08-11
fix: improve error messages for branch-protection and pip-install checks

### LINKED ISSUE #2444: Feature: specify the cause of failure if it is unpinned pip install
When the user sees a report on code scanning about a "pip install" line in his script, it is not clear what he would have to do to solve it. To make it easier to the user to identify what to do, the description of the "pipCommand not pinned by hash" could also mention something like: "--require-hashes  expected". Or if it fails due to additional arguments, would be interesting to also specify.

A possible solution would be to adapt the error message to make the logic bellow clear to the user.
https://github.com/ossf/scorecard/blob/a6983edf6e320a0878dc062cb60aba4f2f215668/checks/raw/shell_download_validate.go#L469-L527

None

**Additional context**
An example of a report on code scanning about pipCommand

image

### LINKED ISSUE #2946: Feature: Improve error message when branch protection check fails because of the use of GITHUB_TOKEN

rickgcv:
## Summary

- Makes `isPermissionsError` case-insensitive so the helpful error message is always shown when `GITHUB_TOKEN` lacks branch protection read permission, regardless of API capitalisation
- Rewrites `classicBranchErrMsg` to name the missing permission (`administration:read`) and link directly to the fix
- Adds a pip-specific message in `generateTextUnpinned` explaining how to pin pip dependencies with `--require-hashes` or a requirements file

Fixes #2946
Fixes #2444

## Notes

The previous PR #5043 addressed these same issues but went stale and was closed without merging. This picks up that work.

## Test plan

- [ ] `make unit-test` passes with no failures
- [ ] `go test ./clients/githubrepo/... ./probes/pinsDependencies/...` passes

### DIFF
--- a/clients/githubrepo/branches.go
+++ b/clients/githubrepo/branches.go
@@ -32,7 +32,8 @@ import (
 const (
 	refPrefix = "refs/heads/"
 	//nolint:lll
-	classicBranchErrMsg = "some github tokens can't read classic branch protection rules: https://github.com/ossf/scorecard-action/blob/main/docs/authentication/fine-grained-auth-token.md"
+	classicBranchErrMsg = "branch protection rules are not accessible with the current token (GITHUB_TOKEN lacks administration:read permission); " +
+		"use a fine-grained personal access token with 'administration: read' scope instead: https://github.com/ossf/scorecard-action/blob/main/docs/authentication/fine-grained-auth-token.md"
 )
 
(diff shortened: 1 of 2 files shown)


rickgcv:
Friendly bump — just rebased onto current main. Happy to address any feedback. Thanks for taking a look!

PR #5189, opened 2026-08-19
:bug: don't forward gitlab PRIVATE-TOKEN across cross-host redirects

Sahana2524:

Bug fix.

`apiFunction` in `clients/gitlabrepo/tarball.go` sets the `PRIVATE-TOKEN` header to `GITLAB_AUTH_TOKEN` and sends the request with `http.DefaultClient`, which follows redirects. Go's `net/http` only strips its built-in auth headers (`Authorization`, `Cookie`, and friends) when a redirect crosses to a different host; a custom header like `PRIVATE-TOKEN` is copied to every hop. So when the `archive.tar.gz` or `ci/lint` endpoint responds with a 302, the operator's GitLab token rides along to whatever host the redirect points at. That target can be object storage or a CDN for a legitimate instance, or an arbitrary host for a self-managed URL, and the same request path is reachable from the `?repo=` input and the cron project list.

The request now goes through a client whose `CheckRedirect` removes `PRIVATE-TOKEN` once a redirect leaves the original host, while keeping the standard 10-redirect limit. Same-host redirects are unaffected, so normal downloads keep working.

NONE

The relevant stdlib logic is `net/http/client.go` (the sensitive-header list is fixed and does not include custom headers). I compared on `req.URL.Host` rather than hostname alone so a change in host or port drops the credential; that is the conservative choice, and GitLab's archive redirects go to a pre-signed URL that does not need the token anyway. The added test drives `apiFunction` through a redirect to a second server and asserts the header is not forwarded; it fails on the current code and passes with the fix.

```release-note
Avoid forwarding the GitLab `PRIVATE-TOKEN` to redirect targets when downloading a project archive.
```

### DIFF
--- a/clients/gitlabrepo/tarball.go
+++ b/clients/gitlabrepo/tarball.go
@@ -43,6 +43,7 @@ var (
 	errTarballNotFound  = errors.New("tarball not found")
 	errTarballCorrupted = errors.New("corrupted tarball")
 	errZipSlip          = errors.New("ZipSlip path detected")
+	errTooManyRedirects = errors.New("stopped after 10 redirects")
 )
 
 func extractAndValidateArchivePath(path, dest string) (string, error) {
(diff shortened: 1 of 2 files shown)


Sahana2524:
any update?

PR #5191, opened 2026-08-21
🐛 Recognize GitHub self-repository action references

trask:
Treat GitHub's `$/` syntax as a same-repository reference in the Pinned-Dependencies check, matching the existing handling for `./` references.

```yaml
uses: $/.github/actions/example
```

See [the shared-workflows usage that exposed this false positive](https://github.com/open-telemetry/shared-workflows/blob/52987636e38f35a2bdb712f5a2b5989f6a5f6f8a/.github/workflows/pull-request-dashboard-repo.yml#L48-L50).

[GitHub documents](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/find-and-customize-actions#adding-an-action-from-the-same-repository) that this syntax resolves from the workflow repository at the running commit, so Scorecard no longer reports these actions or reusable workflows as unpinned third-party dependencies.

Note: the CI failure appears to be unrelated to this PR: [golangci-lint / check-linter (pull_request)](https://github.com/ossf/scorecard/actions/runs/32495932588/job/96814189007?pr=5191)

### DIFF
--- a/checks/raw/pinned_dependencies.go
+++ b/checks/raw/pinned_dependencies.go
@@ -768,10 +768,7 @@ var validateGitHubActionWorkflow fileparser.DoWhileTrueOnFileContent = func(
 		}
 
 		if job.WorkflowCall != nil && job.WorkflowCall.Uses != nil {
-			//nolint:lll
-			// Check whether this is an action defined in the same repo,
-			// https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#referencing-an-action-in-the-same-repository-where-a-workflow-file-uses-the-action.
-			if !strings.HasPrefix(job.WorkflowCall.Uses.Value, "./") {
+			if !isSameRepositoryReference(job.WorkflowCall.Uses.Value) {
 				dep := newGHActionDependency(job.WorkflowCall.Uses.Value, pathfn, job.WorkflowCall.Uses.Pos.Line)
 				pdata.Dependencies = append(pdata.Dependencies, dep)
 			}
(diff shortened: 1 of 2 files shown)


trask:
> This pull request has been marked stale because it has been open for 10 days with no activity

adding some activity to hopefully keep this alive 🙏

jkreileder:
@spencerschrock friendly ping - this is about to affect a lot of repos.

[zizmor 1.30.0](https://github.com/zizmorcore/zizmor/releases/tag/v1.30.0) (2026-08-30) added a [`self-repository` audit](https://docs.zizmor.sh/audits/#self-repository) that flags every `uses: ./...` in the default persona, recommends `$/` instead, and ships an auto-fix. Repos running zizmor in pre-commit/CI will switch to `$/` to clear it - and each one then loses a point on Pinned-Dependencies, even though GitHub itself treats `$/` as pinned under its "require actions pinned to a SHA" policy.

It already hit two of my repos after adopting the syntax (both 10/10 → 9/10, the `$/` references being the only warnings). With this PR's diff applied locally, the new tests pass and both are back to 10/10 with no other change in the results.

The `check-linter` failure is unrelated: `panic: file requires newer Go version go1.27 (application built with go1.26)`.

bilaldeveloper4312 (approved):
Independent code/documentation review:\n\n- GitHub's current Actions documentation explicitly defines $/path/to/dir as a same-repository reference resolved at the running commit and recommends it alongside the existing ./ form: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/find-and-customize-actions#adding-an-action-from-the-same-repository\n- isSameRepositoryReference is applied to both step actions and job-level reusable workflow calls, and isActionDependencyPinned now uses the same rule.\n- The tests cover the helper, the pinning predicate, and parsing a workflow containing both $/.github/actions/... and $/.github/workflows/....\n- The PR's upstream unit, integration, CodeQL, dependency, generation, validation, DCO, Kusari, and patch-coverage checks passed.\n\nThe remaining check-linter failure reports a Go toolchain-version mismatch and is not caused by these two changed files. Since this branch is based on an older main commit, it should still be rebased or have CI rerun by the author/maintainer before merge. I was unable to capture a reliable independent local test result, so I am not claiming one here.\n\nApproved on the implementation and coverage; this is an independent review, not maintainer acceptance authority.

PR #5199, opened 2026-08-26
🐛 fix: match selected security policies by exact path

### LINKED ISSUE #5198: BUG: Security-Policy scores the contents of an unrelated security.md found elsewhere in the tree
### What happened

`Security-Policy` scored my repository 3/10 with:

```
score is 3: security policy file detected:
Warn: no linked content found
Warn: One or no descriptive hints of disclosure, vulnerability, and/or timelines in security policy
```

The repository's `SECURITY.md` contains, at the analysed commit, **2 URLs matching the check's own `reURL`, 1 address matching `reEML`, and 93 substrings matching `reDIG`**. All three security-policy probes should evaluate true.

The check appears to detect the correct file but scan a *different* one.

### Repro

- Repo: `tufantunc/ssh-mcp` (public)
- Analysed commit: `8c008907e420461042f84de761501bbe6927c382`
- Scorecard `v5.5.0`, via `ossf/scorecard-action@v2.4.4`
- Policy file: `SECURITY.md` at the repo root
- Also present: `.review-pro/node/security.md` and `.review-pro/ssh-mcp/security.md` — unrelated code-review rubric files that happen to be named `security.md`

I edited `SECURITY.md` to add a plain email address and an explicit disclosure-timeline table, re-ran, and the score and both warnings were **byte-identical**. The run was fresh — `Code-Review` moved 1/19 → 1/20 and `CI-Tests` 22/29 → 23/29 in the same run.

### Root cause (from reading the source)

Discovery and content-scanning use **different matchers**.

Discovery, `checks/raw/security_policy.go` — full-path, exact:

```go
func isSecurityPolicyFilename(name string) bool {
	return strings.EqualFold(name, "security.md") ||
		strings.EqualFold(name, ".github/security.md") || ...
```

`.review-pro/node/security.md` correctly does **not** match, so the discovered policy file is the root `SECURITY.md`. Its path is then handed to the content pass:

```go
err := fileparser.OnMatchingFileContentDo(c.RepoClient, fileparser.PathMatcher{
	Pattern:       data.files[idx].File.Path,   // "SECURITY.md"
	CaseSensitive: false,
}, checkSecurityPolicyFileContent, ...)
```

But `isMatchingPath` in `checks/fileparser/listing.go:38-48` falls back to the **basename**:

```go
filename := path.Base(fullpath)
match, err := path.Match(pattern, fullpath)
// No match on the fullpath, let's try on the filename only.
if !match {
	if match, err = path.Match(pattern, filename); err != nil { ... }
}
```

So the pattern `security.md` matches `.review-pro/node/security.md` via its basename. And `checkSecurityPolicyFileContent` stops at the first file with non-empty content:

```go
// stop here found something, no need to look further (false)
return false, nil
```

Whichever `security.md` the file listing yields first wins. A dot-directory sorts before `SECURITY.md`.

### Why this reconciles exactly

`.review-pro/node/security.md` is 23 lines with **0 URL matches, 0 email matches and 0 `reDIG` matches**. Feeding it through the three probes:

| Probe | Condition | Result | Points |
|---|---|---|---|
| `securityPolicyContainsLinks` | `(urls + emails) > 0` | false | 0 of 6 |
| `securityPolicyContainsVulnerabilityDisclosure` | `CountSecInfo(Text, false) > 1` | false | 0 of 1 |
| `securityPolicyContainsText` | `FileSize > 1 && FileSize > linkedContentLen + (urls+emails)*2` | **true** | 3 of 3 |

Total: **3** — the observed score, including the counter-intuitive part where `ContainsText` passes while the other two fail, because it tests `FileSize` rather than the collected hits.

### Impact

A project gets scored on the contents of an arbitrary file that merely shares a basename with a security policy. It is silent: the check reports "security policy file detected", so the maintainer sees a low score on a policy that plainly satisfies the documented criteria, and editing the real `SECURITY.md` changes nothing. `security.md` is not an unusual filename for docs, rubrics or templates in a subdirectory.

### Suggested fix

Have the content pass address the exact path that discovery selected, rather than a pattern that can match elsewhere — e.g. fetch the file directly by path, or add an `ExactPath` option to `PathMatcher` that skips the basename fallback. The basename fallback is presumably load-bearing for other callers, so a per-call opt-out seems safer than changing `isMatchingPath` for everyone.

### Note

I have not attached a debugger — the evidence is the published results, the source at `v5.5.0`, and the arithmetic above matching the observed score exactly. Happy to test a patch against the repo, which reproduces this on every run.

bilaldeveloper4312:
Fixes #5198\n\nSecurity-Policy first discovers an approved policy path, but its content pass previously allowed a basename fallback. An unrelated nested security.md could therefore be read instead of the selected policy.\n\nThis adds an ExactPath option to the shared matcher and uses it only for Security-Policy content reads, preserving existing basename fallback behavior for other callers.\n\nValidation:\n- go test ./checks/fileparser ./checks/raw (Go 1.26.1, GitHub Codespaces)

### DIFF
--- a/checks/fileparser/listing.go
+++ b/checks/fileparser/listing.go
@@ -41,8 +41,9 @@ func isMatchingPath(fullpath string, matchPathTo PathMatcher) (bool, error) {
 		return false, sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("%v: %v", errInternalFilenameMatch, err))
 	}
 
-	// No match on the fullpath, let's try on the filename only.
-	if !match {
+	// No match on the fullpath, optionally try on the filename only.
+	// ExactPath callers have already selected a specific repository path.
+	if !match && !matchPathTo.ExactPath {
 		if match, err = path.Match(pattern, filename); err != nil {
 			return false, sce.WithMessage(sce.ErrScorecardInternal, fmt.Sprintf("%v: %v", errInternalFilenameMatch, err))
 		}
@@ -63,6 +64,9 @@ func isTestdataFile(fullpath string) bool {
 type PathMatcher struct {
 	Pattern       string
 	CaseSensitive bool
+	// ExactPath disables the filename fallback and matches Pattern only against
+	// the complete repository path.
+	ExactPath bool
 }
 
 // DoWhileTrueOnFileReader takes a filepath, its reader and
--- a/checks/fileparser/listing_test.go
+++ b/checks/fileparser/listing_test.go
@@ -185,6 +185,7 @@ func Test_isMatchingPath(t *testing.T) {
 		pattern       string
 		fullpath      string
 		caseSensitive bool
+		exactPath     bool
 	}
 	tests := []struct {
 		name    string
@@ -192,6 +193,26 @@ func Test_isMatchingPath(t *testing.T) {
 		want    bool
 		wantErr bool
 	}{
+		{
+			name: "exact path does not fall back to matching a filename",
+			args: args{
+				pattern:       "SECURITY.md",
+				fullpath:      ".review-pro/node/security.md",
+				caseSensitive: false,
+				exactPath:     true,
+			},
+			want: false,
+		},
+		{
+			name: "exact path still matches the selected file",
+			args: args{
+				pattern:       "SECURITY.md",
+				fullpath:      "SECURITY.md",
+				caseSensitive: false,
+				exactPath:     true,
+			},
+			want: true,
+		},
 		{
 			name: "matching path",
 			args: args{
@@ -317,6 +338,7 @@ func Test_isMatchingPath(t *testing.T) {
 			got, err := isMatchingPath(tt.args.fullpath, PathMatcher{
 				Pattern:       tt.args.pattern,
 				CaseSensitive: tt.args.caseSensitive,
+				ExactPath:     tt.args.exactPath,
 			})
 			if (err != nil) != tt.wantErr {
 				t.Errorf("isMatchingPath() error = %v, wantErr %v", err, tt.wantErr)
--- a/checks/raw/security_policy.go
+++ b/checks/raw/security_policy.go
@@ -50,6 +50,7 @@ func SecurityPolicy(c *checker.CheckRequest) (checker.SecurityPolicyData, error)
 			err := fileparser.OnMatchingFileContentDo(c.RepoClient, fileparser.PathMatcher{
 				Pattern:       data.files[idx].File.Path,
 				CaseSensitive: false,
+				ExactPath:     true,
 			}, checkSecurityPolicyFileContent, &data.files[idx].File, &data.files[idx].Information)
 			if err != nil {
 				return checker.SecurityPolicyData{}, err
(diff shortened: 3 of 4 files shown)


noqt (approved):
NOQT independently reviewed current head `f2848a6`.

The change is narrowly scoped: `ExactPath` preserves the existing basename fallback by default and is enabled only when Security-Policy rereads a path it already selected. The two focused exact-path cases and `TestSecurityPolicy` pass with Go 1.25.6; `go vet ./checks/fileparser ./checks/raw` and `git diff --check` also pass. The wider affected-package run still hits existing Windows workflow-fixture failures, which I reproduced on unchanged base `d1fab88` as well.

This fixes the reported misread: a selected root `SECURITY.md` no longer matches `.review-pro/node/security.md` by basename.

Approved. One non-blocking hardening suggestion: add a `SecurityPolicy`-level regression containing both the selected policy and a same-basename nested file, so the caller's `ExactPath` wiring is protected as well as the matcher itself.

tufantunc:
I filed #5198, so a note from the reporting side in case it helps a reviewer.

**The shape here is the right one.** Gating the fallback per call rather than changing `isMatchingPath` for everyone was what the issue suggested, for the reason the diff's own comment gives: the basename fallback is presumably load-bearing for other callers, and Security-Policy is the one caller that has already resolved an exact path before it re-reads the file. `ExactPath` defaulting to false keeps that asymmetry explicit.

**One detail worth knowing when you check the fix.** The affected repo scores **3**, not 0, and the difference is diagnostic rather than cosmetic. With the wrong file read, the three probes land like this:

| probe | outcome | points |
|---|---|---|
| `securityPolicyContainsLinks` | false — the nested file has no URLs | 0 of 6 |
| `securityPolicyContainsVulnerabilityDisclosure` | false — no `reDIG` hits either | 0 of 1 |
| `securityPolicyContainsText` | **true** | 3 of 3 |

The third one passes because it tests `policy.File.FileSize > linkedContentLen + (urls+emails)*2` rather than the collected hits, so a non-empty wrong file still satisfies it. That combination — text passing while links and disclosure fail — is the signature of reading the wrong file, and it is why the symptom is a stubborn `3` rather than a `0`. Expect `3 → 10` on a fixed run, and if you see `0` something else is wrong.

**On the fixture.** I see the tests use `.review-pro/node/security.md`, which is the real path from the reporting repo rather than a synthetic one — good, because it is also the reason the bug is reachable at all: a dot-directory sorts before `SECURITY.md`, so it wins the first-match. Worth keeping that property in the fixture if it ever gets renamed.

`tufantunc/ssh-mcp` is public and reproduces this on every run, if a live case is more useful than a fixture at any point: root `SECURITY.md` (12 KB, links + email + disclosure timeline) and two unrelated `.review-pro/*/security.md` rubric files with no policy signals in them.

I could not build scorecard in my own environment to confirm the fix end-to-end against that repo — the module download times out here — so I am not claiming to have verified it, only that the analysis and the approach line up.

bilaldeveloper4312:
Thanks @noqt for the review and @tufantunc for confirming the real-world failure shape. I added the suggested Security-Policy-level regression in 558e7026: it covers a selected root SECURITY.md alongside .review-pro/node/security.md and ensures the nested same-basename file is not read. The focused checks/raw test passes, and DCO and Kusari are green. @spencerschrock @AdamKorcz, this is ready for maintainer review when you have capacity.

